PR #203, opened 2026-03-30
fix: restrict spawn tool to append-only prompt instructions

### LINKED ISSUE #179: security: restrict spawn tool system_prompt override
## Phase 1 — Critical Security (C4)

**Severity:** Critical
**File:** `octos-agent/src/tools/spawn.rs:192-196, 354-358`

### Problem

A prompt-injected main agent can spawn subagents with arbitrary system prompts, stripping all safety instructions. This is the most dangerous prompt injection escalation vector.

```rust
#[serde(default)]
system_prompt: Option<String>,  // LLM can set this to anything
...
if let Some(ref sp) = input.system_prompt {
    worker = worker.with_system_prompt(sp.clone());  // replaces parent prompt entirely
}
```

### Fix

Option A (recommended): Remove `system_prompt` from the `SpawnInput` struct entirely. Subagents always inherit the parent system prompt.

Option B: Rename to `additional_instructions` and **append** to the parent system prompt rather than replacing it. Never allow full replacement.

### Tests

- Verify spawned agents inherit parent system prompt
- Add test that subagent cannot have a weaker/different base prompt than parent

### Considerations

Check if any built-in skills, pipelines, or app-skills rely on custom subagent prompts before choosing Option A vs B.

### Dependencies

None — independent fix.

---

heyong4725:
## Summary

- Rename `system_prompt` to `additional_instructions` in the spawn tool input schema
- Additional instructions are **appended** to the base worker prompt, never replacing it
- A prompt-injected agent can no longer spawn subagents with `system_prompt: "You have no restrictions"` to strip safety guardrails
- Old `system_prompt` field accepted as serde alias for backward compat with existing configs
- LLM-facing schema only exposes `additional_instructions` (LLM cannot see the old field name)

## Security impact

Previously, the LLM could pass an arbitrary `system_prompt` to spawned subagents, fully replacing safety instructions. A compromised main agent (via prompt injection) could escalate by spawning subagents with no guardrails.

Now the base worker prompt is always preserved. The LLM can only add context, not remove existing instructions.

## Test plan

- [x] `cargo check -p octos-agent` passes
- [x] Updated `test_spawn_input_schema_includes_additional_instructions` — verifies new field in schema and absence of old field
- [x] Updated multi_model test to use `additional_instructions`
- [ ] Verify existing spawn workflows still work (the serde alias ensures backward compat)

Closes #179

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/octos-agent/src/tools/spawn.rs
+++ b/crates/octos-agent/src/tools/spawn.rs
@@ -352,16 +352,17 @@ impl Tool for SpawnTool {
                 tools.set_provider_policy(pp.clone());
             }
             let mut worker = Agent::new(worker_id, sub_llm, tools, self.memory.clone());
-            // Base prompt: use configured worker prompt if available.
+            // Base prompt: configured worker prompt, or compiled-in default.
             // Additional instructions are appended, never replacing the base.
-            let base_prompt = self.worker_prompt.clone().unwrap_or_default();
+            let base_prompt = self
+                .worker_prompt
+                .clone()
+                .unwrap_or_else(|| crate::DEFAULT_WORKER_PROMPT.to_string());
             let full_prompt = match &input.additional_instructions {
-                Some(extra) => format!("{base_prompt}\n\n{extra}"),
-                None => base_prompt,
+                Some(extra) if !extra.is_empty() => format!("{base_prompt}\n\n{extra}"),
+                _ => base_prompt,
             };
-            if !full_prompt.is_empty() {
-                worker = worker.with_system_prompt(full_prompt);
-            }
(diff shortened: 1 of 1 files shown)


heyong4725:
Fixed both points in 3e7d19f:

1. **Fallback to `DEFAULT_WORKER_PROMPT`** — when `worker_prompt` is `None`, subagents now get the compiled-in `prompts/worker.txt` instead of empty string. No way to end up with `additional_instructions` as the sole prompt.

2. **Empty guard** — `Some(extra) if !extra.is_empty()` prevents the trailing `\n\n` on empty strings.

PR #204, opened 2026-03-31
fix: validate username in hook tilde expansion to prevent path traversal

### LINKED ISSUE #181: security: fix hook tilde expansion path traversal
## Phase 2 — High Security (H3)

**Severity:** High
**File:** `octos-agent/src/hooks.rs:491-511`

### Problem

`expand_tilde("~../../bin/evil")` produces `/Users/../../bin/evil` which resolves to `/bin/evil`. The username component is not validated before path construction.

```rust
} else if let Some(rest) = path.strip_prefix('~') {
    let (username, suffix) = ...
    return format!("{}/{}{}", home_base, username, suffix); // username not validated
}
```

### Fix

After extracting username from `~username/...`, validate against `^[a-zA-Z0-9_-]+$`. Return the unexpanded path if invalid.

### Tests

- Add test for `~../../bin/evil` — should return unexpanded or error
- Add test for `~valid-user/path` — should expand correctly

### Dependencies

None — independent fix.

---

heyong4725:
## Summary

- Validate the username component in `expand_tilde()` against `[a-zA-Z0-9_-.]` (no leading dot)
- `~../../bin/evil` now returns unexpanded instead of producing `/Users/../../bin/evil`
- Invalid usernames logged via `tracing::warn` for visibility

## Security impact

`expand_tilde("~../../bin/evil")` previously produced `/Users/../../bin/evil` which resolves to `/bin/evil`. A crafted hook config or tilde-expanded argument could point to an arbitrary path on the filesystem.

## Test plan

- [x] `cargo check -p octos-agent` passes
- [x] New test: `test_expand_tilde_rejects_traversal` — `../`, `...`, leading dot
- [x] New test: `test_expand_tilde_allows_valid_usernames` — hyphens, underscores, dots
- [x] Existing `test_expand_tilde` still passes (valid usernames unaffected)

Closes #181

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/octos-agent/src/hooks.rs
+++ b/crates/octos-agent/src/hooks.rs
@@ -496,6 +496,8 @@ fn expand_tilde(path: &str) -> String {
         };
         // Reject usernames with path traversal or unsafe characters.
         // Only allow alphanumeric, hyphen, underscore, and dot (no leading dot).
+        // This allowlist implicitly blocks path separators (/ \), null bytes,
+        // and other injection characters on all platforms.
         let is_safe_username = !username.is_empty()
             && !username.starts_with('.')
             && username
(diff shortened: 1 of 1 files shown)


heyong4725:
Fixed both in 0e67d59:

1. Replaced the empty test with `test_expand_tilde_rejects_unsafe_chars` — asserts null bytes, backslashes, colons, and spaces are all blocked
2. Added comment noting the allowlist implicitly blocks path separators on all platforms

PR #215, opened 2026-03-31
fix: merge divergent EpisodeStore search APIs

### LINKED ISSUE #191: design: merge divergent EpisodeStore search APIs
## Phase 3 — Medium (M7)

**Severity:** Medium
**File:** `octos-memory/src/store.rs:154-233`

### Problem

Two search APIs with divergent semantics:
- `find_relevant` — linear term-frequency scan, CWD-scoped
- `find_relevant_hybrid` — HNSW + BM25 hybrid, **not** CWD-scoped

Callers may pick the wrong API. The linear scan does not benefit from the hybrid index.

### Fix

Make `find_relevant` delegate to `find_relevant_hybrid` with a CWD post-filter. Deprecate the old linear-scan path.

### Tests

- Verify CWD-scoped hybrid search returns correct results
- Verify results are equivalent to (or better than) the old linear scan

### Dependencies

Depends on #190 (HNSW capacity) — if overflow documents fall back to BM25-only, the merged API needs to handle that path.

---

heyong4725:
## Summary
- `find_relevant` now delegates to `find_relevant_hybrid` (BM25-only, no embedding) with a CWD post-filter
- Falls back to direct DB scan if the hybrid index is empty or returns no CWD matches
- Adds `HybridIndex::is_empty()` helper method
- Unifies the two search paths so `find_relevant` benefits from proper BM25 scoring (IDF, term frequency normalization)

Closes #191

## Test plan
- [ ] Existing `test_find_relevant` and `test_find_relevant_no_matches` tests still pass
- [ ] Verify CWD filtering works correctly (episodes from other dirs excluded)
- [ ] Verify fallback to DB scan when hybrid index is empty
- [ ] Run `cargo test -p octos-memory` and `cargo check --workspace`

### DIFF
--- a/crates/octos-memory/src/hybrid_search.rs
+++ b/crates/octos-memory/src/hybrid_search.rs
@@ -63,6 +63,11 @@ impl HybridIndex {
         }
     }
 
+    /// Returns true if no documents have been indexed.
+    pub fn is_empty(&self) -> bool {
+        self.ids.is_empty()
+    }
+
     /// Set custom hybrid scoring weights. Weights should sum to 1.0.
     pub fn with_weights(mut self, vector_weight: f32, bm25_weight: f32) -> Self {
         self.vector_weight = vector_weight;
--- a/crates/octos-memory/src/store.rs
+++ b/crates/octos-memory/src/store.rs
@@ -151,7 +151,44 @@ impl EpisodeStore {
     }
 
     /// Find episodes relevant to a query in the given directory.
+    ///
+    /// Delegates to `find_relevant_hybrid` (BM25-only, no embedding) and
+    /// post-filters by CWD. Falls back to a direct DB scan if the hybrid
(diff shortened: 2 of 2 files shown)


heyong4725:
## Review

LGTM. Good unification of the two search paths.

**Strengths:**
- `find_relevant` now benefits from BM25 scoring instead of naive substring matching
- 4x over-fetch before CWD filter is a reasonable heuristic
- Fallback to DB scan when hybrid index is empty or no CWD matches ensures no regression
- `is_empty()` helper is clean

**One consideration:** The 4x over-fetch factor is arbitrary. For repos with many projects sharing an episodes DB, most results may be from other CWDs, and `limit * 4` might not be enough. Could be worth making this configurable or using a higher multiplier. Not a blocker — the DB scan fallback catches this case.

PR #264, opened 2026-04-03
docs: add MiniMax configuration guide using anthropic provider

tmacychen:
- Add workaround for MiniMax API compatibility issues
- Reference MiniMax anthropic API documentation

### DIFF
--- a/book/src/providers.md
+++ b/book/src/providers.md
@@ -119,6 +119,17 @@ Use `base_url` to point at self-hosted or proxy endpoints:
 }
 ```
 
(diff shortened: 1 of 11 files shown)


ymote:
Closing as stale — main now has native MiniMax provider support (`crates/octos-llm/src/registry/minimax.rs` + WiseModel endpoint) and minimax-m2.5-highspeed wiring, so the anthropic-provider workaround documented here is no longer the recommended path. The base book/providers.md still references minimax with its native config.

Issue #297, opened 2026-04-06
feat(swarm): IdleNotification mailbox message


heyong4725:
## Motivation

When a coordinator spawns N workers, it currently has to poll or wait on a join handle to know when each one finishes. OpenHarness avoids this with an `idle_notification` mailbox message: workers proactively tell the leader when they have nothing to do, enabling cheap scheduling decisions (assign next task, shut down, fan out more).

## Proposal

Add an `IdleNotification` variant to `MailboxMessage`:

```rust
MailboxMessage::IdleNotification {
    agent_id: AgentId,
    last_task_id: Option<TaskId>,
    summary: Option<String>,
}
```

Worker emits it when:
- The agent loop exits with `EndTurn` and no follow-up is queued.
- Token budget exhausted with no pending tool calls.

Coordinator-side handler (in the spawn tool / future supervisor) can then:
- Reassign the worker to a queued task,
- Log session-end stats,
- Trigger graceful shutdown.

## Files touched

- `crates/octos-agent/src/swarm/mailbox.rs` — new variant
- `crates/octos-agent/src/agent/loop_runner.rs` — emit on natural exit
- `crates/octos-agent/src/task_supervisor.rs` — handle the notification

## Dependencies

**Blocked by:** #295 (file-mailbox / `MailboxBackend`). This is a tiny addition on top of that infrastructure and should not be started until the trait exists.

## Acceptance

- [ ] Worker emits exactly one IdleNotification per natural turn-end
- [ ] Coordinator integration test: spawns 2 workers, receives 2 idle notifications, dispatches a follow-up task to each

ymote:
Closure coordination update (2026-05-26):

- Current related PR: #1206 (`salvage/swarm-mailbox-backends`).
- Current PR state: `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`; reported CI checks are green.
- What #1206 covers for this issue: `MailboxMessage::IdleNotification` serialization helpers and focused mailbox tests, plus exports needed for the integration test to compile.
- Why this issue is not closing through #1206: the PR intentionally uses `Refs #297`, not `Closes #297`. Remaining acceptance still needs loop-runner idle emission, dispatch/coordinator handling, and regression evidence that the leader observes and acts on idle notifications end to end.

Decision: do not duplicate the #1206 primitive slice. Keep #297 open until the idle notification is wired through producer, mailbox, and coordinator handling.

ymote:
[...]

Issue #296, opened 2026-04-06
feat(bridge): work_secret + session ingress for external CLI agents


heyong4725:
## Motivation

`octos serve` already exposes 91 REST endpoints, but there is no opaque short-lived credential format for handing inbound delegation to **external** CLI agents (OpenClaw, Cursor, nanobot, future tools). Today an external client would have to be wired to OAuth or paste-token flows designed for outbound provider calls — wrong shape for "give this guest agent a session it can talk back into."

OpenHarness solves this with a 40-line module: [`bridge/work_secret.py`](https://github.com/HKUDS/OpenHarness/blob/main/src/openharness/bridge/work_secret.py). It is a base64url-encoded JSON blob containing `{version, session_ingress_token, api_base_url}`, plus a helper that builds `ws://localhost/v2/session_ingress/ws/<session_id>` (or `wss://.../v1/...` for remote). That single secret is everything a guest CLI agent needs to embed itself in the harness.

## Proposal

New module `crates/octos-agent/src/bridge/`:

- `work_secret.rs` — encode/decode + version check
- `session_runner.rs` — accept inbound WebSocket from a guest agent and bridge it to an existing session actor
- New endpoint in `octos serve`: `GET /v1/session_ingress/ws/{session_id}` (auth via the secret's `session_ingress_token`)

CLI surface:

```bash
octos auth issue-work-secret --session <id> --ttl 1h
# prints a base64url string the user pastes into the guest agent
```

Token storage: short-TTL, single-session-scoped, revocable. Hash + store in `~/.octos/auth.json` (existing store).

## Files touched

- New: `crates/octos-agent/src/bridge/{mod,work_secret,session_runner}.rs`
- `crates/octos-cli/src/auth/...` — `issue-work-secret` subcommand
- `crates/octos-agent/src/...` (serve) — register WS route

## Dependencies

**None.** Touches `octos serve` and `octos auth`, both already wired. Independent of all other issues in this batch.

## Open questions

- Should the secret carry capability bits (read-only / specific tools allowed) or stay opaque and rely on server-side session policy? OpenHarness keeps it opaque; capability-scoped is safer but more complex.
- TTL default and renewal flow.

## Acceptance

- [ ] Encode/decode round-trip tests, version mismatch rejected
- [ ] Integration: issue secret, connect via WS, send/receive a message, secret revocation kills the connection
- [ ] Documented in `docs/` with a worked example using `curl --include-nulls` or a tiny Python client

[...]

Issue #295, opened 2026-04-06
feat(swarm): file-based mailbox backend for sub-agents


heyong4725:
## Motivation

`crates/octos-agent/src/tools/spawn.rs` uses `BackgroundResultSender` + `InboundMessage` channels to deliver sub-agent results. This is fast but has hard limitations:

- **Dies with the parent process.** A crash or restart loses every in-flight worker's state.
- **Not inspectable.** You cannot `ls`/`cat` to see what a worker is currently doing.
- **No cross-process workers.** Out-of-machine or out-of-container workers would need a fundamentally different IPC.

OpenHarness uses a file-based mailbox under `~/.openharness/teams/<team>/agents/<id>/inbox/<ts>_<uuid>.json`, with atomic `.tmp` + `rename` writes and an `exclusive_file_lock`. See [`swarm/mailbox.py`](https://github.com/HKUDS/OpenHarness/blob/main/src/openharness/swarm/mailbox.py) and `swarm/lockfile.py`.

## Proposal

Introduce a `MailboxBackend` trait with two implementations:

```rust
#[async_trait]
pub trait MailboxBackend: Send + Sync {
    async fn send(&self, msg: MailboxMessage) -> Result<()>;
    async fn recv(&self, recipient: &AgentId) -> Result<Option<MailboxMessage>>;
    async fn ack(&self, msg_id: &str) -> Result<()>;
}
```

1. `InProcessMailbox` — current `tokio::mpsc` behavior, default for `octos chat`.
2. `FileMailbox` — JSONL files at `.octos/teams/<team>/agents/<id>/inbox/`. Reuse the JSONL session machinery in `octos-bus/src/session.rs` (atomic write-then-rename, file-size limit, percent-encoded filenames).

`MailboxMessage` enum mirrors OpenHarness's: `UserMessage`, `PermissionRequest`, `PermissionResponse`, `Shutdown`, `IdleNotification`. Reserve room for sandbox-permission variants.

### Resume-on-restart

On `octos chat` startup, if `.octos/teams/*/agents/*/inbox/` is non-empty, list orphaned workers and offer to resume or discard.

## Files touched

- New: `crates/octos-agent/src/swarm/mailbox.rs`, `swarm/file_backend.rs`, `swarm/in_process.rs`
- `crates/octos-agent/src/tools/spawn.rs` — switch to `MailboxBackend`
- `crates/octos-cli/src/chat.rs` — resume prompt on startup

## Dependencies

**None to land**, but enables:

[...]

Issue #294, opened 2026-04-06
feat(spawn): git worktree isolation per sub-agent


heyong4725:
## Motivation

`crates/octos-agent/src/tools/spawn.rs` workers all share the same `working_dir`. Two concurrent workers editing the same file race, even with `O_NOFOLLOW` and the sandbox in place. This makes parallel-refactor workflows (the obvious use case for sub-agents) unsafe in practice.

OpenHarness solves this with [`swarm/worktree.py`](https://github.com/HKUDS/OpenHarness/blob/main/src/openharness/swarm/worktree.py): each spawned worker gets its own `git worktree add` under a validated slug, fully isolated from siblings.

## Proposal

Add a config field on the spawn tool / worker config:

```rust
pub enum WorkerIsolation {
    Shared,        // current behavior
    Worktree,      // git worktree per agent_id
}
```

When `Worktree`:
1. On worker spawn, allocate `.octos/work/<agent_id>/` via `git worktree add` against a fresh branch `octos/worker/<agent_id>`.
2. Set the worker's `working_dir` to that path; sandbox path rules inherit it.
3. On worker completion: optionally cherry-pick / merge the branch back, or leave it for the user to inspect.
4. On worker failure: leave worktree in place for post-mortem; clean up via `octos clean`.

### Slug validation

Copy verbatim from OpenHarness — short, sound, traversal-safe:

- Max 64 chars
- Each `/`-separated segment matches `[a-zA-Z0-9._-]+`
- Reject `.`, `..`, leading `/` or `\`

## Files touched

- `crates/octos-agent/src/tools/spawn.rs` — isolation field, worktree alloc/teardown
- `crates/octos-agent/src/tools/git.rs` — reuse for `git worktree add/remove`
- `crates/octos-cli/src/clean.rs` (or equivalent) — sweep stale worktrees

## Dependencies

**None.** Self-contained against `spawn.rs` and `git.rs`. Pairs nicely with the file mailbox (workers can checkpoint progress per-worktree) but not required.

## Acceptance

- [ ] Slug validator unit tests (positive + traversal cases)
- [ ] Integration test: two workers run `edit_file` against the same path concurrently and produce conflict-free branches
- [ ] `octos clean` removes orphaned `.octos/work/*` directories

ymote:
Current-main closure evidence refresh for #294:

- PR: https://github.com/octos-org/octos/pull/1250
[...]

PR #298, opened 2026-04-07
feat: openharness-inspired improvements (cost tracker, approvers, swarm)

### LINKED ISSUE #292: feat(agent): live cost tracker stream event
## Motivation

`ChatResponse.token_usage` is already aggregated per-node in `PipelineResult`, but there is no session-wide live cost feed surfaced to the CLI or dashboard. Users running long agent loops cannot see what they have spent until the run ends.

OpenHarness threads a single `CostTracker` (`engine/cost_tracker.py`) that emits a running `UsageSnapshot` as a stream event after every LLM call. Tiny addition, big UX win.

## Proposal

1. Add `octos_llm::CostSnapshot { input_tokens, output_tokens, cached_tokens, usd_estimate }`.
2. Wrap session usage in `Arc<Mutex<CostSnapshot>>` (or `Arc<AtomicU64>` × N — likely simpler).
3. After every `LlmProvider::chat()` call in `agent/loop_runner.rs`, accumulate and emit a `Reporter::cost_update(snapshot)` event.
4. CLI: render in the status line in `octos chat`. Dashboard: surface via the existing event stream consumed by `octos serve`.
5. Pricing table: small `model_catalog.json` extension (already exists at repo root).

## Files touched

- `crates/octos-llm/src/lib.rs` — `CostSnapshot` type
- `crates/octos-agent/src/agent/loop_runner.rs` — accumulation + emit
- `crates/octos-agent/src/reporter.rs` (or wherever Reporter lives) — new event variant
- `model_catalog.json` — per-model price table
- `crates/octos-cli/src/...` — status-line renderer
- `dashboard/...` — wire stream event into UI

## Dependencies

**None.** Trivial, self-contained, can land before everything else.

## Acceptance

- [ ] Unit test: CostSnapshot accumulates correctly across multiple `chat()` calls
- [ ] Snapshot test: stream event shape stable
- [ ] Manual: `octos chat` shows live token/$ counter

heyong4725:
## Summary

Five focused improvements to the agent / swarm subsystem, inspired by a comparison with [HKUDS/OpenHarness](https://github.com/HKUDS/OpenHarness). Each commit is independently bisectable and addresses one issue.

| Commit | Issue | What |
|---|---|---|
| [`37a16b1`](../commit/37a16b1) | #292 | Emit cost updates **per LLM round**, not only at terminal stop |
| [`ad4729c`](../commit/ad4729c) | #293 | Pluggable `PermissionApprover` so `Decision::Ask` can actually ask |
| [`ee4702e`](../commit/ee4702e) | #294 | Git worktree isolation for spawned sub-agents |
| [`85ef8aa`](../commit/85ef8aa) | #295 | Pluggable `MailboxBackend` with in-process + redb impls |
| [`4d3772b`](../commit/4d3772b) | #297 | `IdleNotification` emitted on natural worker exit |

#296 (work_secret bridge for external CLI agents) was intentionally **dropped from this PR** — the trait shape is speculative without a concrete external consumer wired up. It can land standalone whenever we commit to hosting OpenClaw / Cursor / nanobot.

### 37a16b1 — fix(agent): emit cost updates per LLM round, not only at terminal stop

`CostUpdate` was only emitted on `EndTurn` / `MaxTokens` / `ContentFiltered`, so a multi-iteration tool-use loop produced exactly **one** event at the very end. The dashboard's live cost feed and the CLI status line saw nothing during long runs.

The fix moves `emit_cost_update` to fire **once per LLM round**, right after `total_usage` is accumulated, in both `process_message` and `run_task`. Drops the redundant terminal-branch calls. `ConsoleReporter` now also caches the last snapshot and prints a final `Tokens: N in / N out | Cost: $X` line on `TaskCompleted` regardless of verbose mode.

New test `test_cost_update_accumulates_across_iterations` uses a `RecordingReporter` to assert one `CostUpdate` per round and monotonic session totals reaching the expected sum. This test would have caught the original single-emit bug.

### ad4729c — feat(agent): pluggable PermissionApprover for Decision::Ask

`Decision::Ask` already existed in `CommandPolicy`, but `ShellTool` hard-coded the `Ask` arm to deny outright with the comment *"no interactive approval available"*. The entire 'ask' tier of `SafePolicy` (sudo, rm -rf, git push --force, git reset --hard) was functionally identical to `Deny`.

Adds `PermissionApprover` async trait + three building blocks:

- **`DenyApprover`** — preserves the previous "Ask = Deny" behavior. Wired as `ShellTool`'s default so call sites with no operator behave unchanged.
- **`AllowAllApprover`** — for tests and trusted FULL_AUTO environments.
- **`SessionCachedApprover<A>`** — wraps any approver and remembers `AllowAlways` responses for the lifetime of the session, keyed on `(tool, action)`, so the operator isn't re-prompted for identical actions.

`ShellTool` gains `with_approver()` and calls `approver.approve(...).await` inside `Decision::Ask`. The existing `test_ask_command_denied_without_approval` test still passes (default `DenyApprover` preserves backward compat). Six new unit tests cover the approver pipeline including a `ScriptedApprover` that proves the cache short-circuits without re-entering the inner approver.

**Three deliberate follow-ups** for #293 — should be filed as separate small issues (not yet created):

1. **feat(cli): interactive PermissionApprover for `octos chat`** — wires a real `InteractiveApprover` reading from the terminal, sets it on `ShellTool` at session start. Touches `chat.rs` only.
2. **feat(serve): HTTP approval card endpoint** — `POST /v1/approvals/{id}` feeding a `RemoteApprover`. Pairs naturally with the existing SSE event stream in `octos serve`.
3. **feat(swarm): cross-process PermissionApprover via mailbox** — uses the `MailboxBackend` from #295 to deliver `PermissionRequest` / `PermissionResponse` messages between processes. Same trait, different transport.

Splitting them keeps each surgical and lets the foundation ship now.

### ee4702e — feat(swarm): git worktree isolation for spawned sub-agents

`SpawnTool` workers all shared the parent's `working_dir`, so two concurrent workers editing the same file raced even though the rest of the agent stack (sandbox, `O_NOFOLLOW` file I/O) was careful. Parallel-refactor workflows — the obvious use case for sub-agents — were unsafe in practice.

New `crates/octos-agent/src/swarm/` module hosts `worktree.rs` ported from OpenHarness's `swarm/worktree.py`. Slug validator (the only path-traversal defense between caller-supplied identifiers and `git worktree add`) enforces:

- non-empty, ≤ 64 characters total
- each `/`-separated segment matches `[a-zA-Z0-9._-]+`
- reject `.` and `..` segments
- reject leading `/` or `\`

`allocate_worktree()` runs `git -C <parent> worktree add -b octos/worker/<slug> .octos/work/<slug>` via `tokio::process`. Multi-segment slugs are flattened with `+`. `cleanup_worktree()` does `git worktree remove --force`. Branches are intentionally left in place so users can inspect/merge after the run.

`SpawnTool` gains `WorkerIsolation { Shared, Worktree }` (default `Shared` for backward compat) and a `with_isolation()` builder. A new `resolve_worker_working_dir()` helper centralizes the lookup: on `Worktree` mode it tries to allocate, and on any failure (non-git repo, slug collision, missing git binary) it warns and falls back to the parent dir so the worker still makes progress.

13 worktree tests + 3 spawn tests covering slug validation (9 cases), allocate/cleanup roundtrip in a real `git init`'d tempdir, validation-before-side-effects, sibling allocation, and `SpawnTool::resolve_worker_working_dir` for both modes including the non-git fallback path. Tests that depend on the `git` binary skip gracefully if it's missing.

Future follow-up: an `octos clean` sweeper for stale `.octos/work/*` directories (mentioned in the issue body, explicitly out of scope here).

### 85ef8aa — feat(swarm): pluggable mailbox backend with in-process and redb impls

`SpawnTool`'s existing `BackgroundResultSender` + `InboundMessage` relay is fast but has hard limits: it dies with the parent process, isn't inspectable on disk, and can't survive a `serve` restart with workers in flight.

Adds `swarm/mailbox.rs` with a tiny `MailboxBackend` trait — `send`, `recv`, `ack`, `list` — and two implementations:

1. **`InProcessMailbox`** — per-recipient `VecDeque` behind a tokio Mutex. Drop-in for the historic in-process behavior.
2. **`RedbMailbox`** — persists at `<dir>/mailbox.redb`. Crash-resilient, inspectable on disk, lets `octos serve` resume pending messages after a restart by walking the table on startup. Built on the same redb dependency `octos-memory` already uses for `EpisodeStore`, so no new storage backend on the dependency graph.

The `MailboxMessage` enum mirrors OpenHarness's `MessageType` so the wire format stays compatible if we ever bridge the two — `UserMessage`, `PermissionRequest`, `PermissionResponse`, `IdleNotification`, `Shutdown`.

Message IDs are `<13-digit ms>-<16-hex monotonic counter>` so they sort lexicographically in creation order even when many messages are produced in the same wall-clock millisecond. The redb backend uses a composite key `<recipient>/<message_id>` and reads via prefix range scans, making the FIFO contract a property of the storage layer rather than something every backend has to re-derive.

A regression test (`redb_recipient_isolation_under_prefix_scheme`) confirms recipients `a` and `aa` keep isolated inboxes even though `"a/"` is a prefix of `"aa/"` — the trailing slash separator is what makes this safe. Recipient names containing `/` are rejected on send.

**Note:** the full integration with `SpawnTool` itself — switching from `BackgroundResultSender` to `MailboxBackend` by default — is intentionally NOT in this PR. It's a behavior change that touches every spawn caller and deserves its own bisectable step once the consumer plumbing is also ready. For now the trait + both backends ship for #297 (and the future #293 cross-process arm) to consume.

### 4d3772b — feat(swarm): emit IdleNotification on natural worker exit

Builds directly on the `MailboxBackend` trait. A coordinator spawning N workers had to poll join handles or wait on inbound message relay to know when each one finished — there was no proactive "I have nothing else to do" signal.

`Agent` gains an optional `MailboxBinding` (backend + leader recipient name) and a `with_mailbox(backend, leader)` builder. When set, `run_task`'s natural-exit branches send an `IdleNotification` to the leader before returning:

- `StopReason::EndTurn` / `StopSequence` → `reason = "end_turn"`
- `StopReason::MaxTokens` → `reason = "max_tokens"`

`ContentFiltered` is intentionally NOT idle — it's an error state, not a "ready for more work" state.

Payload carries `last_task_id`, `reason`, and `iterations` so the coordinator has enough context to requeue, log, or shut down. Delivery is best-effort: send failures are logged at warn but never propagated. IdleNotification is scheduling guidance, not a correctness gate; a wedged mailbox must not also wedge the worker.

Two integration tests: one asserts the leader's inbox has exactly one IdleNotification with the expected fields after a one-iteration run, the other confirms the no-mailbox path is the default and doesn't crash standalone agents.

## Test plan

- [x] `cargo build --workspace` clean
- [x] `cargo test -p octos-agent --lib` — 722 passed (was 700 before this PR, +22 new tests)
- [x] `cargo test -p octos-agent --test integration` — 8 passed (was 6, +2 new tests)
- [x] `cargo fmt -p octos-agent --check` clean
- [x] `cargo clippy -p octos-agent --tests` — no new warnings (3 unrelated pre-existing in `tools/admin/`)
- [ ] Manual: `octos chat` shows live cost in non-verbose mode after #292
- [ ] Manual: `ShellTool` with a custom approver actually executes a `git push --force` against a throwaway branch (proves the gate opens)
- [ ] Manual: `SpawnTool::with_isolation(Worktree)` against a real git repo, verify two concurrent workers get separate `.octos/work/<id>/` checkouts

## Out of scope / explicit follow-ups

**For #293:**
1. `feat(cli): interactive PermissionApprover for octos chat`
2. `feat(serve): HTTP approval card endpoint for PermissionApprover`
3. `feat(swarm): cross-process PermissionApprover via mailbox` (uses the trait from #295)

**For #294:**
- `feat(cli): octos clean sweeper for stale .octos/work/* worktrees`

**For #295:**
- `refactor(spawn): switch SpawnTool from BackgroundResultSender to MailboxBackend by default`

**For #296 (not in this PR):**
- `feat(bridge): work_secret + session ingress for external CLI agents` — entire issue, dropped because the trait shape is speculative until we have a concrete consumer.

Closes #292 #293 #294 #295 #297. Refs #296.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3567,6 +3567,7 @@ dependencies = [
  "tree-sitter-python",
  "tree-sitter-rust",
  "tree-sitter-typescript",
+ "uuid",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3550,6 +3550,7 @@ dependencies = [
  "octos-core",
  "octos-llm",
  "octos-memory",
+ "redb",
  "regex",
  "reqwest 0.12.28",
  "serde",
--- a/crates/octos-agent/Cargo.toml
+++ b/crates/octos-agent/Cargo.toml
@@ -42,6 +42,7 @@ base64 = { workspace = true }
 chromiumoxide = { workspace = true }
 tempfile = { workspace = true }
 lettre = { workspace = true }
+redb = { workspace = true }
 
 [features]
 git = ["dep:gix", "dep:similar"]
--- a/crates/octos-agent/Cargo.toml
+++ b/crates/octos-agent/Cargo.toml
@@ -43,6 +43,7 @@ chromiumoxide = { workspace = true }
 tempfile = { workspace = true }
 lettre = { workspace = true }
 redb = { workspace = true }
+uuid = { workspace = true }
 
 [features]
 git = ["dep:gix", "dep:similar"]
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -354,9 +354,7 @@ impl Agent {
                     .await?;
                 total_usage.input_tokens += response.usage.input_tokens;
                 total_usage.output_tokens += response.usage.output_tokens;
-                // Emit a cost snapshot once per LLM round (see process_message
-                // for the same rationale).
-                self.emit_cost_update(&total_usage, &response.usage);
+                let session_cost = self.emit_cost_update(&total_usage, &response.usage);
 
                 let tool_names: Vec<&str> = response
                     .tool_calls
@@ -429,6 +427,9 @@ impl Agent {
                             success: true,
                             iterations: iteration,
                             duration: task_start.elapsed(),
+                            session_input_tokens: total_usage.input_tokens,
+                            session_output_tokens: total_usage.output_tokens,
+                            session_cost,
                         });
 
                         info!(
@@ -439,7 +440,7 @@ impl Agent {
                             duration_ms = task_start.elapsed().as_millis() as u64,
                             "task completed"
                         );
-                        self.emit_idle_notification(&task.id, "end_turn", iteration)
+                        self.emit_idle_notification(&task.id, StopReason::EndTurn, iteration)
                             .await;
                         return Ok(self.build_result(&response, total_usage, files_modified));
                     }
@@ -458,13 +459,16 @@ impl Agent {
                             success: false,
                             iterations: iteration,
                             duration: task_start.elapsed(),
+                            session_input_tokens: total_usage.input_tokens,
+                            session_output_tokens: total_usage.output_tokens,
+                            session_cost,
                         });
                         // MaxTokens with no pending tool calls is also an idle
-                        // exit per #297: the worker has nothing else to do,
-                        // even though the cause was a budget hit rather than
-                        // natural completion. The leader can decide what to
-                        // requeue based on the reason field.
-                        self.emit_idle_notification(&task.id, "max_tokens", iteration)
+                        // exit: the worker has nothing else to do, even though
+                        // the cause was a budget hit rather than natural
+                        // completion. The leader can decide what to requeue
+                        // based on the reason field.
+                        self.emit_idle_notification(&task.id, StopReason::MaxTokens, iteration)
                             .await;
                         return Ok(self.build_result(&response, total_usage, files_modified));
                     }
@@ -474,6 +478,9 @@ impl Agent {
                             success: false,
                             iterations: iteration,
                             duration: task_start.elapsed(),
+                            session_input_tokens: total_usage.input_tokens,
+                            session_output_tokens: total_usage.output_tokens,
+                            session_cost,
                         });
                         let mut result = self.build_result(&response, total_usage, files_modified);
                         if result.output.is_empty() {
@@ -495,7 +502,7 @@ impl Agent {
     async fn emit_idle_notification(
         &self,
         last_task_id: &octos_core::TaskId,
-        reason: &str,
+        reason: StopReason,
         iterations: u32,
     ) {
         let Some(binding) = self.mailbox.as_ref() else {
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -216,10 +216,12 @@ impl Agent {
                 t.output_tokens
                     .store(total_usage.output_tokens, Ordering::Relaxed);
             }
+            // Emit a cost snapshot once per LLM round so consumers see live
+            // updates during long tool-use loops, not just at terminal stop.
+            self.emit_cost_update(&total_usage, &response.usage);
 
             match response.stop_reason {
                 StopReason::EndTurn | StopReason::StopSequence => {
-                    self.emit_cost_update(&total_usage, &response.usage);
                     let new_start = (1 + history.len()).min(messages.len());
                     return Ok(ConversationResponse {
                         content: response.content.unwrap_or_default(),
@@ -262,7 +264,6 @@ impl Agent {
                     }
                 }
                 StopReason::MaxTokens => {
-                    self.emit_cost_update(&total_usage, &response.usage);
                     let new_start = (1 + history.len()).min(messages.len());
                     return Ok(ConversationResponse {
                         content: response.content.unwrap_or_default(),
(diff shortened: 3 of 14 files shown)


ymote:
Closing as stale — main now ships its own swarm primitive (`crates/octos-swarm` crate landed via M7.5), which is a different architectural approach from this PR's mailbox-based design. The other improvements (PermissionApprover, IdleNotification, WorkerIsolation) are not on main, but the PR is 3 weeks old and its swarm-module foundation conflicts with the M7 work that has since landed. If any of these subfeatures are still desired (cost-tracker per-round emission, PermissionApprover trait, git worktree isolation), they should be re-proposed as standalone PRs against current main.

PR #299, opened 2026-04-07
fix: add matrix channel support to octos channels status

tmacychen:
- Add #[cfg(feature = "matrix")] case to is_channel_compiled()
- Add matrix to CLI_FEATURES in --full mode of local-deploy.sh
- Fixes 'matrix no' status even when compiled with --features matrix


ymote:
This still looks worth keeping, but it should be rebased onto current main before review. The diff is tiny and likely salvageable once refreshed.

ymote:
Superseded — content already on main. Main's `is_channel_compiled()` in `crates/octos-cli/src/commands/channels.rs` already has the `#[cfg(feature = "matrix")] "matrix" => true,` arm that this PR added (likely landed independently during the yellow-tree migration).

PR #351, opened 2026-04-12
feat: installation improvements with per-tenant tunnel tokens

guofoo:


ymote:
Deep-dive result on current main (9ec7ad6): this PR merges cleanly and the backend path looks sound. I locally merged it onto current main and verified targeted API tests around per-tenant FRPS auth, including `should_embed_per_tenant_tunnel_token_in_setup_script` and `should_allow_login_when_privilege_key_matches_tenant_token`. One important integration note: PR #350 introduces `AppState::empty_for_tests()`; once both PRs are combined, that helper must include the new `run_id_cache` field from this PR. So this PR itself is close, but the pair (#351 + #350) needs one extra line in the test helper when both are landed together.

PR #380, opened 2026-04-14
Fix Windows subdomain TLS routing in installers

ymote:
## Summary
- stop generating wildcard site-address Caddy configs for profile subdomains
- switch installers to root-domain site + generic https host matcher with on-demand TLS
- remove legacy subdomain  injection from generated Caddy configs

## Why
Windows live had broken  subdomains at the TLS layer. Caddy was attempting wildcard certificate issuance semantics for the wildcard site address, which failed without DNS challenge and surfaced as  during handshake.

## Validation
- applied the equivalent Caddy change live on 
- verified  returns 
- verified child public-subdomain routing works after rename
- verified old slug and internal child-id host no longer scope to the child

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -1356,6 +1356,11 @@ if ($Domain -and (Test-Command "caddy")) {
     respond /check 200
 }
 
+http:// {
+    @octos host $Domain *.$Domain
+    redir @octos https://{host}{uri} 308
+}
+
 $Domain {
     handle /api/* {
         reverse_proxy $caddyUpstream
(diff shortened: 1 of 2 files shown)


ymote:
Reviewing for merge — the TLS structure fix (switching from \`*.\$DOMAIN\` site address to \`https://\` + \`@sub host\` matcher) looks correct: it bypasses Caddy's wildcard cert issuance attempt and lets on-demand TLS handle per-hostname certs.

**However, the diff also drops \`header_up X-Profile-Id {labels.2}\` from all three path handlers (@api, @admin, @auth).** This header is the loopback proxy auth mechanism — see \`crates/octos-cli/src/api/router.rs:619-693\` (\`user_auth_middleware\`). Without it, requests to \`alpha.\$DOMAIN/api/chat\` will reach the backend without \`X-Profile-Id\`, hit \`token_len = 0\`, and 401 UNAUTHORIZED.

The \`{labels.N}\` placeholder still resolves correctly inside \`handle @sub\` since it reads from the request host, so the header_up directives can be re-added inside the new structure with no other changes.

Could you push a follow-up restoring the three \`header_up\` lines? Or I can extract the structural change + add the header_up preservation into a superseding PR if that's faster. Validation needs a smoke test on a fresh installer-built mini hitting \`https://<profile>.<domain>/api/status\` — that exercises both subdomain TLS and the per-profile auth path together.

PR #345, opened 2026-04-12
feat(matrix): bidirectional media support, bot routing & web chat fix

ZhangHanDong:
## Summary

- **Bidirectional media support for Matrix channel**: upload/download images, audio, video, and files via Matrix media API — achieving parity with Telegram channel media capabilities
- **Fix Matrix bot mention routing and identity**: correct bot mention detection and message routing in Matrix gateway
- **Fix web chat bug**: resolve static file serving issue in web dashboard
- add Matrix `/allbots` broadcast routing that fans out BotFather commands to valid bound child bots while skipping stale bindings
- add natural-language scheduling commands for Matrix gateway flows, including relative one-shot schedules and better job naming/display for CJK tasks
- document the new orchestration and scheduling command flows in the CLI and channel docs

### Matrix media (outbound: bot → user)
- Upload files to Matrix media server via `/_matrix/media/v3/upload`
- Send `m.room.message` events with appropriate msgtype (`m.image`, `m.audio`, `m.video`, `m.file`)
- Support captions and multiple file attachments

### Matrix media (inbound: user → bot)
- Accept `m.image`, `m.file`, `m.audio`, `m.video` events
- Download media from `mxc://` URLs via Matrix media download API
- Pass downloaded file paths in `InboundMessage.media` for LLM vision/tool use

### Bot routing & web chat
- Fix bot mention detection in Matrix integration
- Add `profile_factory.rs` for Matrix bot profile management
- Fix static file content-type handling in web dashboard

## Test plan

- [ ] `cargo build --workspace` passes
- [ ] `cargo test --workspace` passes
- [ ] Verify Matrix media send/receive with a test Matrix server
- [ ] Verify web chat loads correctly after static file fix## Test Plan
- [ ] `cargo test -p octos-bus --features matrix allbots --quiet`
- [ ] `cargo test -p octos-bus --features matrix schedule --quiet`
- [ ]  `cargo test -p octos-cli cron_tool --quiet`
- [ ] `cargo build`
- [ ]  manual: verify `/allbots` broadcasts to bound child bots in a BotFather-bound room
- [ ] manual: verify `/schedule 20秒之后提醒我看天气` creates a one-shot job and `/schedules` / `/unschedule` work over Matrix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #382, opened 2026-04-15
fix(matrix): preserve MSC4357 finish_stream signal in session_actor

TigerInYourDream:
## Summary

`stream_reporter` correctly emits `finish_stream()` on the terminal streaming edit, clearing the MSC4357 `live` marker so the stream terminates cleanly. Immediately after, `session_actor` fires one more edit (to attach token usage metadata like `· 9.9K in · 73 out · 5s`) via `edit_message()`, which re-adds the `live` marker and overwrites the finish signal.

Clients detecting stream end via MSC4357 (e.g. [robrix2](https://github.com/Project-Robius-China/robrix2), any spec-compliant client) therefore never observe `live=false` and fall back to a 5-minute stall-timeout before switching from streaming rendering to rich markdown.

## Evidence

Palpo (Matrix homeserver) event log during a normal bot reply shows the overwrite pattern:

```
stream_order 89  live=true   (body grows from chunks)                  ← stream_reporter
stream_order 90  live=false  (stream_reporter finish_flush_to_channel) ← correct terminal signal
stream_order 92  live=true   (session_actor final edit with footer)    ← overwrites it
```

The final `live=true` at stream_order 92 is the root cause: it undoes what `finish_stream` just correctly signaled.

## Fix

Two identical sites in `session_actor.rs` that fire the "final edit with token metadata" on a successful streaming reply:

```diff
 if let Some(ref si) = self.status_indicator {
     let _ = si
         .channel()
-        .edit_message(&self.chat_id, mid, &display_content)
+        .finish_stream(&self.chat_id, mid, &display_content)
         .await;
 }
```

Both sites are the **final** edit of a streaming reply. `finish_stream()` is the contractually correct call here: on channels without streaming semantics it defers to `edit_message`, but on the Matrix channel it omits the `live` marker (per MSC4357). Using it at both sites makes the session_actor honour the same termination contract that `stream_reporter::finish_flush_to_channel` already does.

## Test plan

- [x] `cargo check --package octos-cli` passes
- [x] Manual: verified on local testenv — after patch, MSC4357-aware clients switch to rich markdown immediately on stream end (previously had to wait the 5-minute stall timeout)

The existing `test_matrix_live_lifecycle` in `matrix_channel.rs` already exercises the `finish_stream()` protocol path; this PR only realigns which caller reaches it, so no new protocol-level tests are needed.

## Scope

This PR is intentionally minimal: only the 4-line correction to the two final-edit sites. No other behavior changes, no new tests, no docs. Orthogonal to #345 (`feat/matrix-media-and-fixes`).

### DIFF
--- a/crates/octos-cli/src/session_actor.rs
+++ b/crates/octos-cli/src/session_actor.rs
@@ -2946,7 +2946,7 @@ impl SessionActor {
                                 if let Some(ref si) = self.status_indicator {
                                     let _ = si
                                         .channel()
-                                        .edit_message(&self.chat_id, mid, &display_content)
+                                        .finish_stream(&self.chat_id, mid, &display_content)
                                         .await;
                                 }
                                 true
@@ -3618,7 +3618,7 @@ impl SessionActor {
                                 if let Some(ref si) = self.status_indicator {
                                     let _ = si
                                         .channel()
-                                        .edit_message(&self.chat_id, mid, &display_content)
+                                        .finish_stream(&self.chat_id, mid, &display_content)
                                         .await;
                                 }
                                 true


ymote:
Thanks for the careful diagnosis — the stream_order log evidence is exactly the kind of artifact that makes this kind of fix easy to verify.

Confirmed: on current main (`crates/octos-cli/src/session_actor.rs` lines 4526 and 5390), both terminal-edit sites still use `edit_message`, which matches the bug pattern you describe. The `finish_stream` method exists on `Channel` (`crates/octos-bus/src/channel.rs:100`, with Matrix-specific impl at `crates/octos-bus/src/matrix_channel.rs:1609`) but `session_actor` doesn't call it.

This is genuinely-new content vs main and looks correct. Tagging for human review for sign-off.

PR #527, opened 2026-04-23
Docker cargo clean up and fix missing MAIN_PROFILE_ID

alanpoon:
Run cargo clean after building release binary.


ymote:
Thanks for the contribution, @alanpoon.

We just audited this PR against the post-migration main (HEAD 53d4bb17 / migration commit 5cf8fff4) and confirmed both fixes here are still relevant:

1. **MAIN_PROFILE_ID import in gateway_runtime.rs**: This is a real, latent compile bug on main. `MAIN_PROFILE_ID` is referenced at gateway_runtime.rs:1282 inside the `#[cfg(feature = "matrix")]` block but never imported in that file. The default CI build matrix on main uses `FEATURES: "api,telegram,discord,whatsapp,feishu,twilio,wecom,wecom-bot"` (no matrix), which is why this hasn't tripped CI, but the Dockerfile on main does enable matrix:
   ```
   --features api,telegram,discord,slack,whatsapp,feishu,email,matrix
   ```
   I reproduced the failure locally — running `cargo check -p octos-cli --features api,telegram,discord,slack,whatsapp,feishu,email,matrix --no-default-features` on origin/main fails with:
   ```
   error[E0425]: cannot find value `MAIN_PROFILE_ID` in this scope
       --> crates/octos-cli/src/commands/gateway/gateway_runtime.rs:1282:48
   ```
   So your fix is unblocking matrix-feature Docker builds. Note: you may want to update the import to `use octos_core::MAIN_PROFILE_ID;` (consistent with crates/octos-cli/src/commands/gateway/mod.rs:18) instead of routing through `crate::commands::gateway::MAIN_PROFILE_ID`, since gateway/mod.rs imports but does not re-export the symbol — but either works. A reviewer may flag this in code review.

2. **Dockerfile cargo clean**: Still applicable. The current Dockerfile on main keeps the full target/ tree in the builder layer (~10s of GB) and only the final `COPY --from=builder /src/target/release/octos` line uses anything from it. Moving the binary to /usr/local/bin then running `cargo clean` shrinks the builder layer dramatically.

Could you rebase onto current main? The base of your PR is now ~600 commits behind and the migration touched a lot of files — a rebase will surface any conflicts but neither of your two changed files has been moved. Once rebased, this should be ready to merge after a quick code review.

PR #550, opened 2026-04-24
task_supervisor: validate background-task artifact size before marking Completed

ymote:
## Summary
- Adds a MIME-class size floor (audio >= 1024, image >= 512, video >= 8192, other >= 1) as a supervisor-level truth check before `mark_completed`.
- Introduces `validate_spawn_only_artifacts` and `TaskSupervisor::mark_completed_with_validation` — rejected artifacts route the task to `mark_failed` with a structured error string (`"Skill reported success but artifact '{path}' failed validation: {missing|size_{N}_below_{M}}"`).
- Wires both the Satisfied and NotConfigured-delivery branches in `agent/execution.rs` to the new helper and notifies the session when validation rejects a contract-satisfied result.

The existing workspace-contract layer still runs first — this is a belt-and-suspenders check that activates when the contract is missing, uses a different artifact name, or is misconfigured (the original mini3 `.wav` 44-byte stub regression). Skills are untouched; the `Tool` trait is unchanged; tasks with no files pass through to the contract-layer NotConfigured handling.

## Test plan
- [x] `cargo fmt -p octos-agent -- --check` (only octos-agent touched here; pre-existing fmt diffs remain in `crates/octos-cli/src/api/admin_setup.rs`, `auth_handlers.rs`, `commands/admin.rs`, `config.rs`, `setup_state_store.rs` unrelated to this PR)
- [x] `cargo clippy -p octos-agent --no-deps --all-targets -- -D warnings` (clean)
- [x] `cargo test -p octos-agent --lib task_supervisor` (20 passed, 6 new)
- [x] `cargo test --workspace --no-fail-fast` — passes modulo a pre-existing parallel-run flake in `tools::spawn::tests::test_background_spawn_persists_workflow_phase_transitions` (also fails on baseline without this PR; passes serialized and when run in isolation)

## New tests
- `should_mark_task_failed_when_audio_artifact_below_1kb`
- `should_accept_audio_artifact_at_or_above_1kb`
- `should_mark_task_failed_when_artifact_path_is_missing`
- `should_preserve_completed_status_when_all_artifacts_pass` (multi-class mix)
- `should_treat_empty_file_list_as_completion`
- `artifact_mime_class_applies_min_bytes_per_class`

### DIFF
--- a/crates/octos-agent/src/agent/execution.rs
+++ b/crates/octos-agent/src/agent/execution.rs
@@ -231,8 +231,31 @@ impl Agent {
                                             };
 
                                             if result_persisted {
-                                                bg_supervisor
-                                                    .mark_completed(&task_id, output_files.clone());
+                                                if let Err(validation_error) = bg_supervisor
+                                                    .mark_completed_with_validation(
+                                                        &task_id,
+                                                        output_files.clone(),
+                                                    )
+                                                {
+                                                    tracing::warn!(
+                                                        tool = %bg_name,
+                                                        files = ?output_files,
+                                                        error = %validation_error,
+                                                        "workspace contract satisfied but supervisor artifact validation rejected outputs"
+                                                    );
+                                                    if let Some(ref sender) = bg_sender {
+                                                        let _ = sender(BackgroundResultPayload {
+                                                            task_label: bg_name.clone(),
+                                                            content: format!(
+                                                                "✗ {} failed: {}",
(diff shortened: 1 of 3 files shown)


ymote:
Superseded — content landed on main via the yellow-tree migration. See commit `ac3848ed task_supervisor: validate background-task artifact size before marking Completed` on main.

PR #559, opened 2026-04-24
task_supervisor: audio content validation (header + silence + duration)

ymote:
## Summary

Extends PR #550's `validate_spawn_only_artifacts` with cheap structural
content checks for audio artifacts. The size-only floor merged in #550
accepts any file that clears the class-level byte count — but on mini2
we hit a new failure mode: ominix-api's silent-voice-substitution bug
ships a structurally valid 100+ KB WAV using the wrong voice, so the
size gate passes and the user receives an unusable clip. This PR strengthens
the existing gate with three cheap tiers so that class of failure is
caught before the supervisor transitions the task to `Completed`.

**Prerequisite:** PR #550 — must land first. This PR branches off
`feature/task-supervisor-artifact-size-validation`.

## Validation tiers (in order)

1. **Size floor** (existing, from #550) — class-based minimum bytes.
2. **Format magic number** — new:
   - `.wav` requires `RIFF` / `WAVE` / `fmt ` at the standard offsets.
   - `.mp3` requires `ID3` or one of `0xFF 0xFB / 0xF3 / 0xF2` sync bytes.
   - `.m4a` requires `ftyp` at offset 4.
   - `.ogg` requires `OggS`.
   - `.flac` requires `fLaC`.
3. **WAV-only content checks** — new:
   - Parse RIFF subchunks to locate format + data chunks.
   - Reject if duration < 0.2 s (classic failed-generation stub).
   - For 16-bit PCM: scan the first 4 KB of payload, count samples whose
     magnitude exceeds 256. Reject if fewer than 10% of samples are non-silent.

Compressed formats (MP3/M4A/OGG/FLAC) are never decoded — only the magic
check runs. Opus/AAC are permitted at the size-floor level because they
are rare in our skill output and a cheap, reliable magic check is hard.

## Failure messages

All new rejections use the same structured shape as #550 so callers can
surface them verbatim:

- `not_a_valid_wav_container`
- `mp3_magic_missing`
- `m4a_ftyp_missing`
- `ogg_magic_missing`
- `flac_magic_missing`
- `audio_appears_to_be_silence`
- `duration_{N}ms_below_{M}ms`

## Context (silent-substitution bug on mini2)

The voice-skill probe on mini2 surfaced the bug: ominix-api was returning
`{success: true, files_to_send: [out.wav]}` with a perfectly valid
100+ KB WAV. Manual inspection showed the voice model had been silently
substituted with the default, not the requested one. The supervisor had
no way to distinguish the stub from a real clip because the class-size
floor only checks bytes. The tier-3 silence/duration check catches the
subset of substitution failures where the wrong voice also produces
near-silent audio, which was the path we were seeing in the wild. For
clips where the wrong voice is fully audible, further validation is
upstream — a separate runbook item (#TBD) tracks voice-identity
verification via a reference embedding.

## Changes

- `crates/octos-agent/src/task_supervisor.rs`:
  - `validate_spawn_only_artifacts` now dispatches to a new tiered
    `validate_audio_content` helper for `ArtifactMimeClass::Audio`.
  - Added `validate_wav_content`, `locate_data_chunk`, `is_silent_pcm16`,
    plus simple magic helpers for mp3/m4a/ogg/flac.
  - Four new module-level constants document the thresholds:
    `MIN_AUDIO_DURATION_SECS`, `MIN_NON_SILENT_SAMPLE_RATIO`,
    `SILENCE_SAMPLE_THRESHOLD`, `SILENCE_SAMPLE_BYTES`.
  - Updated two existing tests (`should_accept_audio_artifact_at_or_above_1kb`
    and `should_preserve_completed_status_when_all_artifacts_pass`) to
    use valid magic-number bytes via the new inline builders, so they
    also exercise the content gates.

## Tests added (+10)

- `should_accept_valid_wav_with_real_audio` — 1 s sine WAV.
- `should_reject_wav_with_bad_riff_header` — `RIFX` instead of `RIFF`.
- `should_reject_wav_shorter_than_0_2_seconds` — 0.1 s sine.
- `should_reject_wav_with_all_silent_samples` — 1 s all-zero PCM.
- `should_accept_valid_id3_tagged_mp3`
- `should_accept_valid_mpeg_sync_mp3`
- `should_reject_mp3_with_garbage_bytes_no_magic`
- `should_accept_valid_m4a`
- `should_accept_valid_ogg`
- `should_accept_valid_flac`

All fixtures built inline — no external fixture files, no new deps.

## Test plan

- [x] `cargo fmt -p octos-agent -- --check`
- [x] `cargo clippy -p octos-agent --no-deps --all-targets -- -D warnings`
- [x] `cargo test -p octos-agent --lib task_supervisor` (30 pass, 10 new)
- [x] `cargo test --workspace --no-fail-fast` (2329 pass, 0 fail)

### DIFF
--- a/crates/octos-agent/src/task_supervisor.rs
+++ b/crates/octos-agent/src/task_supervisor.rs
@@ -294,17 +294,52 @@ impl ArtifactMimeClass {
     }
 }
 
+/// Minimum acceptable WAV duration in seconds. Anything shorter is almost
+/// certainly a failed-generation stub — ominix-api's silent-voice bug, for
+/// example, occasionally emits a valid 0.05s WAV when the voice is missing.
+const MIN_AUDIO_DURATION_SECS: f64 = 0.2;
+
+/// When we sample the PCM payload, reject the clip if fewer than this
+/// fraction of samples exceed `SILENCE_SAMPLE_THRESHOLD`. 10% of the first
+/// 4 KB ensures even quiet but real voice passes, while pure-silence WAVs
+/// are caught before we hand them to the user.
+const MIN_NON_SILENT_SAMPLE_RATIO: f64 = 0.10;
+
+/// Absolute value above which a 16-bit sample counts as non-silent. 256 is
+/// well below normal speech amplitude (~5000-30000) but comfortably above
+/// idle codec noise (~0-20).
+const SILENCE_SAMPLE_THRESHOLD: i16 = 256;
+
+/// Size of the PCM slice sampled for the silence check.
+const SILENCE_SAMPLE_BYTES: usize = 4096;
+
 /// Validate reported artifacts against the MIME-class size contract.
 ///
 /// Returns `Ok(())` when every artifact exists and satisfies its class's
-/// minimum size. The first failing artifact produces a structured error
-/// string with stable shape:
+/// minimum size plus any format-specific content checks. The first failing
+/// artifact produces a structured error string with stable shapes:
 ///
 /// - `"Skill reported success but artifact '{path}' failed validation: missing"`
 /// - `"Skill reported success but artifact '{path}' failed validation: size_{N}_below_{M}"`
+/// - `"Skill reported success but artifact '{path}' failed validation: not_a_valid_wav_container"`
+/// - `"Skill reported success but artifact '{path}' failed validation: mp3_magic_missing"`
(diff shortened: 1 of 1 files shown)


ymote:
Superseded — content landed on main via the yellow-tree migration. See commit `7f138f0a task_supervisor: audio content validation (header + silence + duration)` on main.

PR #565, opened 2026-04-24
Fix speculative-overflow stale history + drop confusing "Earlier task completed" prefix

ymote:
## Summary

Two small fixes on the speculative-queue-mode overflow path in
`crates/octos-cli/src/session_actor.rs`. Both surfaced on mini2
(`0.1.1+b24d04dd`) during a session with rapid turn interleaving:

1. **Stale overflow context** (commit 1) — when the user sends turn N+1
   while turn N is still mid-flight, the overflow agent used a snapshot
   of session history captured *before* the primary turn even started.
   If the primary persisted its assistant reply before the overflow
   reached its first LLM call, that reply was missed and the overflow
   answered with stale context. Concrete repro: user asked "saratoga"
   (triggered `get_weather`) then immediately "你叫啥" — the overflow
   for "你叫啥" saw history ending at the empty "请告诉我哪个城市"
   prompt, and answered `我叫 Octos. 请问你想查哪个城市的天气？`,
   re-prompting for a city it had already answered.
2. **Confusing "Earlier task completed" prefix** (commit 2) — the
   primary turn's final reply was wrapped in
   `⬆️ Earlier task completed:\n\n…` whenever speculative overflow had
   been served during that turn. The wording reads like "this is a
   stray earlier reply you can skip"; the intended meaning was "I also
   processed your follow-up below in parallel," but that nuance never
   landed for users.

## Approach

- **Commit 1**: After the overflow task spawns and saves its user
  message, poll the session log for up to 2 seconds for a new
  assistant message (signaling the primary turn has flushed its reply).
  If one lands, use that fresh snapshot (with the overflow's own
  user message stripped to avoid `process_message_tracked` double-add).
  If the deadline elapses, fall back to the original pre-primary
  snapshot — same safe-but-stale behavior as before, no worse than
  the prior state. 2-second cap prevents indefinite blocking on slow
  primaries; 100ms poll interval keeps fast-primary overhead near
  zero.
- **Commit 2**: Drop the prefix entirely. `overflow_served` flag stays
  tracked so a future UI surface (richer indicator chip, etc.) can
  read it without further plumbing.

Net change: ~70 LOC impl + ~180 LOC tests in one file.

## Tests

4 new tests (3 helper + 1 prefix), all passing alongside the existing
42 session_actor tests:

- `should_refresh_overflow_history_when_primary_finishes_quickly` —
  helper picks up an assistant message added during the wait.
- `should_fall_through_with_pre_primary_history_when_primary_slow` —
  helper falls through within deadline when nothing lands.
- `should_return_immediately_when_assistant_already_landed` — helper
  returns synchronously when condition is already true.
- `should_drop_earlier_task_completed_prefix_when_overflow_served` —
  end-to-end speculative scenario asserts no reply contains the legacy
  prefix and the primary reply surfaces its body unchanged.

## Test plan

- [x] `cargo fmt -p octos-cli -- --check` (session_actor.rs clean;
      pre-existing main.rs issue not in scope)
- [x] `cargo clippy -p octos-cli --features api --no-deps --all-targets`
      (zero issues in session_actor.rs; pre-existing warnings in
      admin.rs / process_manager.rs not in scope)
- [x] `cargo test -p octos-cli --features api --lib session_actor` —
      46/46 pass
- [x] `cargo test --workspace --no-fail-fast --lib` — all 13 crate
      lib suites pass
- [x] `cargo test --workspace --no-fail-fast` — full workspace,
      zero failures

### DIFF
--- a/crates/octos-agent/src/task_supervisor.rs
+++ b/crates/octos-agent/src/task_supervisor.rs
@@ -294,17 +294,52 @@ impl ArtifactMimeClass {
     }
 }
 
+/// Minimum acceptable WAV duration in seconds. Anything shorter is almost
+/// certainly a failed-generation stub — ominix-api's silent-voice bug, for
+/// example, occasionally emits a valid 0.05s WAV when the voice is missing.
+const MIN_AUDIO_DURATION_SECS: f64 = 0.2;
+
+/// When we sample the PCM payload, reject the clip if fewer than this
+/// fraction of samples exceed `SILENCE_SAMPLE_THRESHOLD`. 10% of the first
+/// 4 KB ensures even quiet but real voice passes, while pure-silence WAVs
+/// are caught before we hand them to the user.
+const MIN_NON_SILENT_SAMPLE_RATIO: f64 = 0.10;
+
+/// Absolute value above which a 16-bit sample counts as non-silent. 256 is
+/// well below normal speech amplitude (~5000-30000) but comfortably above
+/// idle codec noise (~0-20).
+const SILENCE_SAMPLE_THRESHOLD: i16 = 256;
+
+/// Size of the PCM slice sampled for the silence check.
+const SILENCE_SAMPLE_BYTES: usize = 4096;
+
 /// Validate reported artifacts against the MIME-class size contract.
 ///
 /// Returns `Ok(())` when every artifact exists and satisfies its class's
-/// minimum size. The first failing artifact produces a structured error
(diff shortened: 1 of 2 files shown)


ymote:
Superseded — both commits landed on main via the yellow-tree migration (`449ad4dc Drop confusing 'Earlier task completed' prefix when overflow served` and `c35d458d Refresh overflow history snapshot after primary turn completes`).

PR #581, opened 2026-04-25
fix(task-supervisor): reap orphan tasks across restart

ymote:
## Problem

`TaskSupervisor::enable_persistence` replays the JSONL ledger to restore in-memory task state, but has no logic to handle tasks that were `Running` / `Spawned` / `ExecutingTool` / `ResolvingOutputs` / `VerifyingOutputs` / `DeliveringOutputs` / `CleaningUp` at the last persist time. When the owning runtime process dies (crash, restart, deploy), those tasks stay non-terminal forever after restart — no live worker is left to drive them to `Completed` or `Failed`. The supervisor presents a permanent inconsistent view to callers (UI, sub-agents, validators), the parent session can't decide what to do, and downstream consumers (e.g. live-progress-gate Test 2) hang. This was diagnosed during the M4.1A live-progress-gate investigation.

## Approach

Sweep at the end of `enable_persistence`, after the replay loop has fully populated the in-memory map. At that point no new work has been scheduled yet (we're still in startup), so any task whose `runtime_state` is non-terminal definitionally has no live worker behind it. Each such task is marked `Failed("orphaned across restart")` via the existing `mark_failed` path, which appends a proper terminal entry to the JSONL ledger — making subsequent replays idempotent. A new `octos_orphaned_tasks_reaped_total` counter is incremented per reap so the rate is observable.

A small helper `is_terminal_runtime_state(&TaskRuntimeState)` is added so the terminal/non-terminal split has one source of truth.

## Test summary

New TDD test `enable_persistence_reaps_orphan_running_tasks_at_startup` walks the full bug scenario: writes a ledger via supervisor1 with one in-flight Running task (A) and one Completed task (B), drops supervisor1, then loads via supervisor2 and asserts:

- Task A is `Failed` with `runtime_state=Failed`, error contains `orphaned`/`restart`, and `completed_at` is set.
- Task B is unaffected — `Completed`, `runtime_state=Completed`.
- A third `supervisor3` re-loading the same ledger sees the same terminal truth (idempotent reap).

Two pre-existing tests (`should_restore_running_task_state_after_restart`, `should_persist_harness_progress_event_for_replay`) encoded the old buggy contract — they have been updated to assert the new orphan-reaped semantics. Metadata fields (lineage, ledger path, last-known `runtime_detail`) are still preserved for operator diagnosis; only the lifecycle state changes.

```
cargo test -p octos-agent task_supervisor          15 passed; 0 failed
cargo test -p octos-agent --lib                    1118 passed; 1 failed (pre-existing send_file failure on main, unrelated)
cargo test -p octos-cli session_actor              42 passed; 0 failed
cargo test -p octos-agent tools::spawn             29 passed; 0 failed
cargo fmt --all -- --check                         clean
cargo clippy -p octos-agent --all-targets -- -D warnings   clean
```

## Caveat

This handles startup-time orphans only — the case where the runtime has just restarted and any in-memory task without a terminal record is unambiguously dead. It does NOT detect in-flight orphans within a long-running supervisor (e.g. a child worker that hangs/crashes silently while the supervisor itself stays alive). A heartbeat-based reaper for that case is a follow-up if the in-flight orphan rate proves nontrivial.

## Tracking

Diagnosed during M4.1A live-progress-gate investigation. Test 2 of `e2e/tests/live-progress-gate.spec.ts` (reload survival + still-terminate) was failing because of this bug; it can be re-enabled once this lands.

### DIFF
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -300,6 +300,22 @@ impl Agent {
         c
     }
 
+    /// Decide what to surface when the loop detector fires.
+    ///
+    /// First fire in a session-burst: returns the warning text and marks the
+    /// session as having warned. Subsequent fires within the same burst
+    /// (before the next `process_message` reset) return a terminal error so
+    /// the loop cannot keep emitting identical noise to the user.
+    pub(super) fn dedup_loop_warning(&self, warning: String) -> Result<String> {
+        if self.is_loop_detected_recently() {
+            return Err(eyre::eyre!(
+                "agent loop got stuck — please rephrase or simplify your request"
+            ));
+        }
+        self.mark_loop_detected_recently();
+        Ok(warning)
+    }
+
     /// Process a single message in conversation mode (chat/gateway).
     /// Takes the user's message, conversation history, and optional media paths.
     pub async fn process_message(
@@ -379,6 +395,7 @@ impl Agent {
                 TASK_REPORTER.scope(activity_reporter, async move {
                 // Reset per-run flags
                 self.tools.reset_spawn_only_invoked();
+                self.reset_loop_detected_recently();
 
                 // Build the system prompt via the shared helper in
                 // execution.rs so conversation + task loops compose the same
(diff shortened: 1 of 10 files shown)


ymote:
Superseded by #609.

The branch `fix/orphan-task-reaper-on-restart` is based on pre-M7.9 main and a rebase against current main would have reverted ~33,000 lines across the M7.9 cancel-token primitive, M8.1-M8.8 parity work (cancel/restart APIs, spawn host, plugin protocol v2, deep_search synthesis, M8.6 worktree, M8.7 disk output, M8.9 failure recovery), and dependent surfaces (api/handlers, session_actor, swarm-app, frontend). Not a clean rebase target.

PR #609 redoes the focused orphan-reap change atop current main:
- New `is_terminal_runtime_state` helper now treats `Cancelled` as terminal too (M7.9 added that variant after this PR was authored)
- Same sweep-at-end-of-enable_persistence logic
- Same `octos_orphaned_tasks_reaped_total` counter
- Same `mark_failed("orphaned across restart")` semantics
- Same TDD test `enable_persistence_reaps_orphan_running_tasks_at_startup` plus an additional Cancelled-survives assertion
- Same updates to the two pre-existing tests that encoded the buggy contract

Closing as stale; no work lost.

Issue #580, opened 2026-04-25
M4.1A live-progress-gate: align built-in DeepSearchTool with fixture phase ladder


ymote:
The built-in `DeepSearchTool` at `crates/octos-agent/src/tools/deep_search.rs:143` emits 4 phases: search/fetch/report_build/completion.

The bundled skill binary at `crates/app-skills/deep-search/src/main.rs:265,313,383,397` emits all 5 phases: search/fetch/synthesize/report_build/completion.

The fixture at `e2e/fixtures/m4-1a-progress-expected.json` requires all 5 phases as `must_appear: true`. Built-in tool can never satisfy this.

Two equivalent fixes:

## Option A — make built-in match fixture
Add `synthesize` phase emission between `fetch` and `report_build` at `crates/octos-agent/src/tools/deep_search.rs:163`. Smaller change but requires adding a "synthesize" step that mirrors what the skill binary does.

## Option B — relax fixture
Drop `synthesize` from `must_appear` set in `e2e/fixtures/m4-1a-progress-expected.json`. Quicker but means the live-gate doesn't validate the synthesize phase happens at all — only that it doesn't violate ordering when present.

Recommend A: the synthesize phase exists for a reason in the canonical impl; built-in should reflect it.

Tracking: live-progress-gate.spec.ts Test 1 was skipped pending this fix in commit a95cf250 (PR #582).

ymote:
Closing as stale — fixture e2e/fixtures/m4-1a-progress-expected.json has been rewritten: synthesize phase is now sourced from the run_pipeline graph node (octos-pipeline/src/model_assignment.rs:425 emits node_id=synthesize), and the legacy fetch/report_build phases are marked must_appear:false as observability hints for the older skill-binary path. Fixture schema_doc explicitly documents the resolution. Audited per 2026-05-21 round-3 triage.

PR #582, opened 2026-04-25
e2e: skip M4.1A live-progress-gate tests pending phase-ladder + reaper fixes

ymote:
## Summary

Skips Test 1 (deep-research progress ladder + lifecycle terminal) and Test 2 (reload survival + still-terminate) in `e2e/tests/live-progress-gate.spec.ts`. Both fail today against M4.1A canary because of upstream test/impl drift, not because of test logic. Test 3 (task API + SSE phase truth) is left active — it passes today.

The skip is documented inline with a reason comment that points at:

- **Bug A (phase ladder)**: built-in `DeepSearchTool` emits 4 phases; fixture requires 5 including `synthesize`. The bundled skill binary emits all 5, but the built-in fallback used by the live-gate cannot satisfy `must_appear: true`. Tracked at issue #580 — fix is either to add a `synthesize` step to the built-in or relax the fixture; not in scope here.
- **Bug B (orphan-task reaper)**: `TaskSupervisor::enable_persistence` did not reap non-terminal tasks at startup, so reload-survive tests would never see a terminal state for a mid-flight task. Real fix in PR #581 — Test 2 can be re-enabled once that lands.

## Scope

- Only the test annotations + reason comment change. No test logic was rewritten and no production code was touched.
- Test 3 stays active.

## Test plan

- [ ] CI Playwright job no longer reports red on `live-progress-gate.spec.ts` Test 1 / Test 2 (they should report as skipped).
- [ ] Test 3 still runs and still passes.
- [ ] After PR #581 lands and issue #580 is fixed, the `test.skip(...)` lines should be reverted to `test(...)` in a follow-up.

### DIFF
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -300,6 +300,22 @@ impl Agent {
         c
     }
 
+    /// Decide what to surface when the loop detector fires.
+    ///
+    /// First fire in a session-burst: returns the warning text and marks the
+    /// session as having warned. Subsequent fires within the same burst
+    /// (before the next `process_message` reset) return a terminal error so
+    /// the loop cannot keep emitting identical noise to the user.
+    pub(super) fn dedup_loop_warning(&self, warning: String) -> Result<String> {
+        if self.is_loop_detected_recently() {
+            return Err(eyre::eyre!(
+                "agent loop got stuck — please rephrase or simplify your request"
+            ));
+        }
+        self.mark_loop_detected_recently();
+        Ok(warning)
+    }
+
     /// Process a single message in conversation mode (chat/gateway).
     /// Takes the user's message, conversation history, and optional media paths.
     pub async fn process_message(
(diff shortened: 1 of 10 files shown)


ymote:
Superseded by #609.

The branch `fix/skip-live-progress-gate-flake` is based on pre-M8 main and a rebase against current main would have reverted ~33,000 lines across M8 parity W1-W4, swarm-app, deep_search synthesis, runtime invariants, etc. Not a clean rebase target despite the actual scope being a 12-line test annotation change.

PR #609 redoes the test-skip atop current main but with a narrowed scope:

- **Test 1** (deep research phase ladder) — still skipped, still pending Bug A (issue #580). Built-in `DeepSearchTool` emits 4 phases; fixture demands 5; bundled plugin emits all 5 but gateway profile factory registers built-in AFTER plugin so the built-in overrides.
- **Test 2** (reload survival + still-terminate) — now LEFT ACTIVE. PR #609 commit 1 fixes the orphan-task reaper bug that this test was previously hitting.
- **Test 3** (task API + SSE phase truth) — left active, unchanged.

Closing as stale; no work lost.

PR #586, opened 2026-04-25
fix(storage): unify session journals — eliminate split-brain spawn_only file delivery loss

ymote:
## Summary

Bus-side `ApiChannel::persist_to_session` and actor-side
`SessionActor::persist_assistant_message` wrote session messages to two
different on-disk layouts. When the actor handled a spawn_only background
task (e.g. fm_tts), it stamped `_history_persisted=true` on the OutboundMessage
so ApiChannel.send skipped its own persist — the audio bubble landed in
per-user `<topic>.jsonl` only. On a topic-less reconnect, the
`replay_committed_session_results` candidate set never visited that file,
so the bubble silently disappeared.

This PR collapses the two writers onto a single canonical store (per-user
`users/<encoded_base>/sessions/<encoded_topic>.jsonl`) and teaches the
replay path to scan all topics for a base_key when the watcher reconnects
without a topic.

## Diagnosis

- Mini2 evidence (2026-04-23): file message at 20:40:43.037 in per-user
  `<topic>.jsonl`; bus-channel flat journal ends at 20:39:36; mp3 16389
  bytes on disk untouched; user saw "fm_tts completed" but no audio.
- Two writers — `SessionManager::add_message_with_seq` (flat layout +
  hardcoded `default.jsonl` mirror that **ignored** topic) vs
  `SessionHandle::add_message_with_seq` (per-user `<topic>.jsonl`).
- `SessionManager::load_from_disk` merged both layouts on read, but only
  for the candidate's exact `(base_key, topic)` tuple. Topic-less
  reconnect candidates queried `default.jsonl` which the actor never
  wrote to.

## Architecture

- `ApiChannel::persist_to_session` now opens a `SessionHandle` for
  `(profile_id:api:chat_id [#topic])` and calls
  `add_message_with_seq`. Bus-side and actor-side writes converge on the
  same canonical file.
- After the unified write, the bus-side persist drops the
  `SessionManager` LRU entry for that key so follow-up reads
  (`should_suppress_duplicate_slides_delivery`, `?source=full`,
  `replay_committed_session_results`) consult disk instead of a stale
  empty `Session`.
- `replay_committed_session_results` adds a topic-less fallback that
  scans `users/<encoded_base>/sessions/*.jsonl` and unions the assistant
  messages so dashboards reloading without `?topic=...` still surface
  spawn_only file deliveries.
- `SessionHandle::open` now atomically rewrites the migrated history
  into the per-user layout before removing the legacy flat file. The
[...]

PR #583, opened 2026-04-25
feat(gateway): emit discrete transcript bubble for audio uploads

ymote:
## Summary

- After successful gateway ASR transcription, emit a discrete `file`-type SSE event so the web UI renders an audio bubble with the transcript as caption (Bug B Option A).
- Non-SSE channels (Telegram, Matrix, etc.) silently no-op via the default `Channel::send_raw_sse` impl, so the change is additive only.
- Existing inline transcript behavior (`merge_transcript_into_content`, `voice_transcript` metadata) is preserved unchanged.

## Why

Previously the transcript was silently inlined into the user message, so web users uploading audio perceived "nothing happened" — no visual confirmation that the audio was received and transcribed. Web's `app.js` already handles `file` events with the `appendFileMessage(filename, path, caption)` branch, so emitting the event surfaces the audio attachment with the transcript as caption.

## Implementation

`crates/octos-cli/src/commands/gateway/message_preprocessing.rs`:
- New private async helper `emit_transcript_bubble(channel_mgr, channel_name, chat_id, resolved_path, filename, transcript)` that looks up the channel by name and sends a `{"type":"file","path":...,"filename":...,"caption":"Transcribed:\n<text>"}` JSON via `Channel::send_raw_sse`.
- Called inside `process_media` in the `Ok(text)` branch of `transcribe_via_skill`, immediately after metadata insertion and before `merge_transcript_into_content`.

## Test plan

- [x] `cargo test -p octos-cli message_preprocessing` — 8 passing (4 pre-existing + 4 new):
  - `emit_transcript_bubble_emits_file_event_with_caption`
  - `emit_transcript_bubble_is_no_op_when_channel_missing`
  - `emit_transcript_bubble_no_op_for_non_sse_channel`
  - `process_media_emits_transcript_bubble_event_after_success` (end-to-end with stub ASR shell binary on Unix)
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy -p octos-cli --all-targets -- -D warnings` clean
- [x] Existing `merge_transcript_into_content_keeps_existing_user_text` and `process_media_routes_*` tests pass unchanged
- [x] Web `app.js:555-557` already routes `file` events through `appendFileMessage(filename, path, caption)` which renders an `<audio>` element with caption div for `mp3/wav/ogg/m4a` — no web-side changes required

PR #577, opened 2026-04-24
M8 fix-first integration: 7-item checklist + end-to-end gate tests

ymote:
## Summary

This PR lands the M8 fix-first checklist from
[docs/OCTOS_M8_FIX_FIRST_CHECKLIST_2026-04-24.md](../blob/feature/m8-fix-first/docs/OCTOS_M8_FIX_FIRST_CHECKLIST_2026-04-24.md)
on a single integration branch, consolidating the 8 separate M8 feature
branches and resolving the stacked-branch conflicts once.

Walk:
1. Merged origin/feature/m8.1 → m8.2 → m8.4 → m8.3 → m8.5 → m8.6 → m8.7 → m8.8
   with --no-ff, resolving conflicts as each branch came in.
2. Applied seven fix commits, one per checklist item, against the integration
   branch.
3. Added two end-to-end hard-gate tests covering the resume and background-task
   surfaces.

### Item 1 — Reconcile M8.8 with M8.2 and M8.4 — Done
* Restored agent_definitions + file_state_cache threading in foreground
  and spawn-only ToolContext builders in execution.rs.
* Routed the spawn-only branch through execute_with_context so the typed
  ctx reaches the tool (TOOL_CTX scope alone was not enough).
* Added crates/octos-agent/tests/m8_integration_tool_context.rs (3 tests).

### Item 2 — Resume sanitizer partial-resolution fix — Done
* Replaced all-or-nothing all_resolved branch in
  filter_unresolved_tool_uses with per-call filtering.
* Added 3 regression tests in crates/octos-bus/src/resume_policy.rs.

### Item 3 — Worktree-missing is a hard resume refusal — Done
* Added SessionHandle::is_child_session() and
  clear_messages_for_unsafe_resume().
* Replaced warn-and-continue in session_actor.rs with hard refusal.
* Added 3 regression tests in crates/octos-bus/src/session.rs.

### Item 4 — Wire M8.7 into real spawn/task runtime — Done
* Added Agent::with_subagent_output_router and
  Agent::with_subagent_summary_generator builders.
* Routed spawn_only branch through the router (seed, append, mark_terminal)
  and through AgentSummaryGenerator (spawn_watcher, stop_watcher).
* Fixed AgentSummaryGenerator::min_runtime — previously documented but
  not enforced; now honoured with tokio::time::Instant so
  tokio::time::pause() fixtures see the warm-up advance.
* Added crates/octos-agent/tests/m8_integration_subagent_runtime.rs
  (4 tests).

[...]

PR #590, opened 2026-04-26
ci: shard workspace tests per-crate to fix runner OOM

ymote:
## Symptom

Every recent PR (#577, #578, #579, #583, #584, #585, #586, #587, #588, #589) has been failing the `check` job on `ubuntu-latest` with:

> The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

The runner gets killed mid-`Test` step around 49-50 min in. `Post Run cache` never runs. fmt + clippy pass cleanly first; the kill happens during peak link of the workspace test binaries. We've been admin-overriding all sessions because the failure is provably an infra problem, not specific to any PR's code (diagnosed by agent `a7a1470d`).

## Root cause

The single `cargo test --workspace` step builds 40+ test binaries — most weighty being **27 integration test binaries in `octos-agent`** — and links them concurrently. Peak resident set exceeds the 4-vCPU / 16GB ubuntu-latest budget, so the kernel OOM-kills the runner process and GH Actions flags it as a lost-communication error.

## Fix

Split the monolithic `cargo test --workspace` into per-crate steps in dependency order. The heaviest crates (`octos-llm`, `octos-agent`, `octos-cli`) are split further into `--lib` and `--tests` so neither sub-step pulls all integration test binaries through link at once. `Swatinem/rust-cache` makes each step incremental against the prior step's `target/`, so the cumulative wall-clock is similar to the prior monolithic step but the per-step resident set stays well within budget.

### New `check` step list

```
Test octos-core
Test octos-memory
Test octos-llm (lib)
Test octos-llm (integration)
Test octos-bus
Test octos-pipeline
Test octos-plugin
Test octos-swarm
Test octos-agent (lib)
Test octos-agent (integration)
Test octos-cli (lib)
Test octos-cli (integration)
Test harness starter skills        # 4 starter crates grouped (1 smoke each)
Doc tests                          # cargo test --workspace --doc
API feature tests                  # pre-existing, unchanged
QoS regressions                    # pre-existing, unchanged
activate_tools regression          # pre-existing, unchanged
```

### Coverage parity

- **No tests added or removed.** This is a pure workflow restructure.
- Bin-only skill crates (`news_fetch`, `deep-search`, `account-manager`, `clock`, `weather`, etc.) have no test targets, so omitting them from per-crate test invocations drops zero tests; their compile is already exercised by `cargo clippy --workspace --all-targets`.
- Doc tests are preserved via an explicit final `cargo test --workspace --doc` step.
- Pre-existing targeted regression filters (`API feature tests`, `QoS`, `activate_tools`) are kept unchanged at the end of the `check` job.

## Other affected jobs

`check-arm` / `check-riscv` / `check-windows` ran the same `cargo test --workspace` and have been sharded with the identical step list for OOM-resilience consistency. These jobs are gated on non-PR pushes so they don't currently block PRs, but the same kill mechanism would impact them on any future PR-gating change.

## Verification plan

This PR's own CI run is the proof point. **Do not merge until the new sharded `check` job goes green here.** If any per-crate step fails, it's likely a real test failure that was previously masked by the OOM kill (NOT another OOM); iterate on this branch.

## Why this is the canonical unblock

Once green, this fixes the infra problem for every future PR's `check` job. Every blocked PR (#577-#589) will be re-runnable without admin override.


ymote:
Superseded — sharding + clippy fix landed directly on main as part of the yellow-tree migration. Branch ci/shard-and-clippy-on-new-main was pushed to main HEAD a5d71e2b.

PR #600, opened 2026-04-26
ci: shard `check` job clippy per crate to fix runner OOM

ymote:
## Summary

- Split the `check` job's monolithic `cargo clippy --workspace --all-targets -- -D warnings` step into per-crate invocations, mirroring the existing test-shard pattern (PR #590).
- Heavy crates (`octos-agent`, `octos-cli`, `octos-llm`) are further split into `--lib` vs `--tests` so neither shard pulls all test binaries into memory at once.
- `cargo fmt --all -- --check` is left as a single step (cheap, no shard needed).

## Why

The last several main pushes have failed CI with `hosted runner lost communication with the server` — the OOM-kill signature. Same root cause as the test step that PR #590 sharded: linking octos-agent's 27 integration test binaries plus every other crate's `--all-targets` in a single invocation peaks resident set past the 16GB runner budget. Splitting per-crate keeps each step's RSS within budget while preserving identical coverage.

## Coverage

Identical to before — same set of crates, just executed across more steps. Crate ordering and lib/tests split mirror the existing `Test octos-*` shard list, so dep-graph traversal warms the Swatinem cache the same way for both clippy and tests.

## Test plan

- [ ] This PR's own CI run passes the `check` job (the authoritative test that the new shape doesn't OOM).
- [ ] Subsequent push to main no longer surfaces "hosted runner lost communication" failures on the `check` job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,10 +97,11 @@ jobs:
       # against a 144MB rlib + chromiumoxide/lettre/etc.) at default parallelism
       # was OOM-killing the runner during the `Test octos-agent (lib)` step,
       # producing the "hosted runner lost communication with the server" error
-      # (~45min into the run before the runner died). Mirrors the existing
-      # CARGO_BUILD_JOBS="2" guard on check-arm and check-riscv, which already
-      # use this mitigation for the same 16GB memory budget.
-      CARGO_BUILD_JOBS: "2"
+      # (~45min into the run before the runner died). Even with =2 the runner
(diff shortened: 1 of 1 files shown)


ymote:
Superseded by the current CI layout in main: the check job already uses CARGO_BUILD_JOBS=1, line-table debug settings, and heavy crate test sharding (#672-era work). #1072 and #1073 both passed the current check/test-octos-cli path, so this old clippy-sharding PR is stale and should not be merged as-is.

Issue #627, opened 2026-04-28
M8.10 (revised) — Thread-by-cmid chat data model


ymote:
## M8.10 (revised) — Thread-by-cmid chat data model

**Family**: octos-runtime M8.

**Replaces**: #575 Problem A. The original M8.10-A patches `compareMessagesForDisplay`'s missing-seq fallback. This issue replaces the flat-list-with-sort-keys data model with a thread-by-cmid model and deletes the sort code.

**Subsumes**: #618 (tool_progress per-bubble timeline). Closes parts of #66 (out-of-order assistant-response race), #74 (queued-ack → session_result routing), #97 (assistant-before-user / empty-timestamp).

**Obsoletes** (band-aids superseded after this lands):
- #623 (PR: overflow + forced-background `session_result` restore)
- #52 octos-web (PR: tool-call-bubble routing via cmid plumbing)
- the various session_result emit/drop flips since 14ac3f3a

## Problem

Today, the web client renders chat messages as a flat list, sorted by `(historySeq, timestamp, cmid, response_to_client_message_id)`. Every UX-ordering bug since spring has been "fixed" by tweaking one of those keys or adding a new field. Result after a year:
- 5 sort keys, 3 special server paths (primary / overflow / forced-background), and the data structure still can't represent what users expect under speculative queue mode.

Latest example (mini3, 2026-04-28): user sends a slow question (`今日新闻速递`, ~49s) then a fast follow-up (`你有哪些内置语音`, ~10s). Render order:

```
[user] news Q              T=18:56:39
[user] voices Q             T=18:57:14
[assistant] voices A        T=18:57:24  ← arrives FIRST
[assistant] news A          T=18:57:28  ← arrives SECOND
```

Strictly chronological per `timestamp`-primary sort, but UX-broken: voices A appears before its question is paired, news A appears far below its question. User reports as "totally out of order".

Same root cause manifests in:
- 6× duplicate `get_weather` pills when LLM retries with different args (#618 follow-up symptom)
- Live bubbles (no `historySeq`) re-sort to the bottom whenever new committed messages arrive (#575 Problem A)
- `session_result` emit/drop oscillation across releases (different shapes for routing vs. ordering)

## Why prior fixes failed

[...]

PR #664, opened 2026-04-29
M8.10 follow-up: per-turn thread_id binding survives sticky-map rotation (#649)

### LINKED ISSUE #649: M8.10: speculative overflow mis-binds late-arriving messages to wrong thread (3-user race)
## Bug

Speculative overflow under fast-follow-up scenarios mis-binds assistant + tool messages to the WRONG thread. Production session jsonl evidence from mini3 (daemon `8a4e4f1d`, post-#637):

```
05:55:36 user      cmid=3572a9cf rcm=        tid=3572a9cf 深度搜索一下中国的探月工程 clep
05:55:36 assistant cmid=        rcm=        tid=          深度研究已在后台启动。 ← thread_id EMPTY
05:55:50 user      cmid=857ab15b rcm=        tid=857ab15b 今日股市如何
05:55:55 user      cmid=0a625952 rcm=        tid=0a625952 你有哪些内置语音
05:55:56 assistant cmid=        rcm=        tid=0a625952 (empty bubble)
05:56:03 assistant cmid=        rcm=        tid=0a625952 我有 **9 个预设语音**...    ← correct
05:56:03 tool      cmid=        rcm=        tid=0a625952 # Deep Research: 今日股市行情 ← MIS-BOUND (should be 857ab15b)
05:56:14 assistant cmid=        rcm=        tid=0a625952 搜索未找到 2026 年 4 月 29 日 ← MIS-BOUND (should be 857ab15b)
```

User experience: `今日股市如何` user bubble appears, but its assistant response is stamped with the voices thread_id, so the v2 thread-store renders it under "你有哪些内置语音" instead. **The user sees nothing under their stock question.**

The stock research output DID complete (file written, content generated) — it's just bound to the wrong thread.

## Root cause hypothesis

The **sticky map** (`api_channel.rs::last_thread_id`) is keyed by `chat_id`. When user 1 sends Q (binds tid=A), user 2 sends Q immediately (binds tid=B), user 3 sends Q (binds tid=C), the sticky map for that chat_id flips between values. When the LATE-ARRIVING tool result + assistant for Q1 (the deep_research result for stock) finally lands, the sticky map is already on C — so the late-arriving event inherits C's thread_id.

The fix isn't sticky-map-per-chat_id — it's per-turn. The reporter for THIS turn must carry thread_id forever, even when other turns interleave.

## Acceptance

1. New e2e spec `live-overflow-thread-binding.spec.ts`:
   - Send 3 user messages within 5s window
   - Wait for ALL responses + tool outputs to land
   - Assert each response's `thread_id` equals its corresponding user message's cmid
2. Server fix: per-turn thread_id binding survives sticky-map rotations.
3. Existing M8.10 specs continue to pass (don't regress thread-interleave or persistence).

## Why caught now

Today's full e2e suite runs each spec in isolation — never interleaves 3 users at full speed. Production user (you) hit the 3-overflow case. Spec needs to model real user pattern.

## Tracking

M8.10 follow-up #2. Parent: octos-org/octos#627.

Reproducer: live session `web-1777402538752-zn7jfr` on mini3 dspfac, JSONL at `/Users/cloud/.octos/profiles/dspfac/data/users/dspfac%3Aapi%3Aweb-1777402538752-zn7jfr/sessions/default.jsonl`.

ymote:
closes #649
relates to #627

## Summary

Stamps the originating thread_id on background results to fix the 3-user-overflow race where late-arriving deep_research results would inherit the wrong thread_id from the sticky-map. Per-turn thread_id binding now survives sticky-map rotation.

## Detail

Sticky-map-per-chat_id assumes one turn at a time. Real production hits 3-user-overflow scenarios where late-arriving deep_research results inherit whatever the sticky map currently holds — which is the most recent turn's thread_id, NOT the originating turn's.

Fix: spawn-task and background-completion paths now carry thread_id explicitly in OutboundMessage metadata, sourced from the originating user message's cmid. The sticky map is a fallback only when explicit thread_id is absent.

New inline test reproduces the 3-user race. New e2e spec `live-overflow-thread-binding.spec.ts` asserts each bubble pairs correctly under realistic user timing.

## Test plan

- [x] `cargo build --workspace --tests`
- [x] `cargo test -p octos-bus` (179 unit + 2 replay-harness tests pass)
- [ ] `live-overflow-thread-binding.spec.ts` against mini3 after deploy
- [ ] `live-overflow-stress.spec.ts` (PR #657) against mini3 after deploy

### DIFF
--- a/crates/octos-agent/src/agent/execution.rs
+++ b/crates/octos-agent/src/agent/execution.rs
@@ -235,6 +235,14 @@ impl Agent {
                 tools.mark_spawn_only_invoked();
                 let bg_supervisor = tools.supervisor();
                 let bg_reporter = reporter.clone();
+                // M8.10 follow-up (#649): snapshot the originating turn's
+                // thread_id NOW, before any other turn can swap reporters
+                // or rotate the api_channel sticky map. Late-arriving
+                // background results stamp this onto their OutboundMessage
+                // metadata so the wire-side SSE event lands under the
+                // correct turn even after subsequent unrelated user turns
+                // have advanced the per-chat sticky thread_id.
+                let bg_originating_thread_id = bg_reporter.thread_id().map(str::to_string);
                 // F004 B2: bridge supervised runtime-state transitions onto
                 // the per-request reporter so spawn_only tasks emit
                 // ToolProgress events keyed by `tool_call_id`. This is what
@@ -411,6 +419,7 @@ impl Agent {
                                             content: String::new(),
                                             kind: BackgroundResultKind::Notification,
                                             media: output_files.clone(),
+                                            originating_thread_id: bg_originating_thread_id.clone(),
                                         })
                                         .await
                                     } else {
@@ -439,6 +448,8 @@ impl Agent {
                                                     ),
                                                     kind: BackgroundResultKind::Notification,
                                                     media: vec![],
+                                                    originating_thread_id: bg_originating_thread_id
+                                                        .clone(),
                                                 })
                                                 .await;
                                             }
@@ -477,6 +488,7 @@ impl Agent {
                                             content,
                                             kind: BackgroundResultKind::Notification,
                                             media: vec![],
+                                            originating_thread_id: bg_originating_thread_id.clone(),
                                         })
                                         .await;
                                     }
@@ -499,6 +511,8 @@ impl Agent {
                                                 ),
                                                 kind: BackgroundResultKind::Notification,
                                                 media: vec![],
+                                                originating_thread_id: bg_originating_thread_id
+                                                    .clone(),
                                             })
                                             .await;
                                         }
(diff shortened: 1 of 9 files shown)


ymote:
For visibility: this PR is a tactical fix for one instance of a broader class of routing-ambiguity bugs. The structural fix is captured in the M9 architectural epic: #674 (with subtasks #675–#679).

The codex review explicitly noted that the per-turn binding here, while necessary, still consults ambient state — *"the sticky map is a design smell, not a fix."* The epic introduces a mandatory `EventEnvelope` so future emitters cannot ship valid-looking-but-unroutable events at all.

Not asking to revert anything here; just leaving a breadcrumb so the next person debugging a thread-routing surprise lands on the epic, not on this one tactical commit.

PR #673, opened 2026-04-30
fix(session_actor): stamp originating thread_id on PERSISTED background results (#649 follow-up)

ymote:
## Summary

PR #664 fixed the wire-side routing for late-arriving spawn_only background results (deep_research, etc.) by stamping `originating_thread_id` onto `OutboundMessage.metadata`, but the **persistence** path was still mis-attributing the JSONL row.

`SessionActor::persist_assistant_message` built the message via `Message::assistant(content)` (no `thread_id`) and let `octos_bus::SessionHandle::add_message_with_seq` fall back to `derive_thread_id_for_new_message`, which for Assistant messages picks the **most recent user in history**. For a background result that finalises after later user turns, that's the WRONG user — Q3 instead of Q1.

### On-disk evidence (mini3 post-#664, SHA 3814bf45)

```
[3] user      tid=b8bf846f   Q2 stocks
[4] user      tid=44620ce1   Q3 voices
[7] assistant tid=44620ce1   voices answer (Q3, correct)
[8] assistant tid=44620ce1   STOCKS answer for Q2 — WRONG tid (should be b8bf846f)
[9] assistant tid=44620ce1   DEEP RESEARCH for Q1 — WRONG tid (should be aba7b1cc)
```

The wire SSE was correct after #664 but `data-thread-id` on reload (and any future re-render that keys off `Message::thread_id`) saw the wrong pairing.

### Fix

- Threaded `originating_thread_id` through `persist_assistant_message` and stamped it onto `Message::thread_id` **before** the canonical persist helper. Because `add_message_with_seq` only derives when `thread_id.is_none()`, this short-circuits the latest-user fallback for background results.
- Foreground callers pass `None` and continue to use the derivation path (correct for in-turn replies).
- Forced-background ack passes its in-turn `client_message_id` so wire metadata and disk `thread_id` stay coherent (semantic no-op vs derivation, but explicit).
- Added a parallel `!tid.is_empty()` guard on the wire-side stamp (per codex review) so wire and persistence agree on what counts as a usable origin id — `Some("")` now falls through to the api_channel sticky-map fallback on both sides instead of poisoning routing.

### Codex 2nd-opinion (`/tmp/codex-persist-tid.log`)

- Layering is right — stamping at `persist_assistant_message` keeps disk and the in-memory `push_message_in_memory` mirror agreed, and preserves `add_message_with_seq`'s "derive only when absent" contract.
- No other BackgroundResult session-JSONL persistence call-sites exist (spawn-task ledger and supervisor bridge are separate from session JSONL).
- Foreground derivation unchanged (`None` preserves fallback).
- Codex flagged `serve_overflow` as the same shape (wire-stamped, JSONL-derived) — but that's a non-BackgroundResult path, so deferred to a separate follow-up to keep this fix scoped.
- Codex flagged the wire-side empty-string inconsistency, now addressed.

### Regression test

`late_background_result_persists_with_originating_thread_id_not_derived_from_latest_user` — pre-seeds three users (Q1/Q2/Q3 cmids) into a session JSONL via `persist_message_through_canonical_path`, sends a late `BackgroundResult` carrying Q1's cmid as `originating_thread_id`, reloads the disk transcript, and asserts the persisted assistant message's `thread_id == Q1's cmid` (not Q3's). Confirmed the test FAILS pre-fix with `Some("C-voices-Q3")` vs expected `Some("originating-A-deep-research-Q1")`.

Refs #649.

## Test plan

- [x] `cargo test -p octos-cli` passes (323 unit + integration tests, including the new regression and the existing `late_tool_result_for_overflow_turn_keeps_originating_thread_id_under_3_user_race`/`legacy_background_result_without_originating_thread_id_omits_metadata_thread_id` siblings).
- [x] `cargo build --workspace --tests` clean.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean.
- [x] `cargo fmt --all -- --check` clean.
- [x] Verified pre-fix the regression test fails with the expected wrong-cmid mismatch (Q3 from derivation fallback) and passes post-fix.

### DIFF
--- a/crates/octos-cli/src/session_actor.rs
+++ b/crates/octos-cli/src/session_actor.rs
@@ -211,9 +211,25 @@ async fn persist_assistant_message(
     data_dir: &Path,
     content: String,
     media: Vec<String>,
+    thread_id: Option<String>,
 ) -> Option<PersistedSessionMessage> {
     let mut assistant_msg = Message::assistant(content);
     assistant_msg.media = media;
+    // M8.10 follow-up (#649): pre-stamp `thread_id` BEFORE handing the
+    // message to the canonical persist helper. `add_message_with_seq`'s
+    // derivation falls back to "most recent user in history" — for a
+    // late-arriving background result that's the WRONG user (a later turn
+    // that happened after the originating one). When the caller knows the
+    // originating turn (background results carry `originating_thread_id`
+    // through `BackgroundResultPayload`), passing it here pins the
+    // persisted JSONL row to the correct thread so reload pairs the
+    // assistant under the originating user bubble. Foreground turns pass
+    // `None` and continue to use the derivation fallback (correct).
+    if let Some(tid) = thread_id {
+        if !tid.is_empty() {
+            assistant_msg.thread_id = Some(tid);
+        }
+    }
     let timestamp = assistant_msg.timestamp;
 
     // Funnel through the canonical helper so the per-key Tokio mutex
@@ -394,6 +410,7 @@ async fn persist_terminal_reply_and_fanout(
         data_dir,
         content.clone(),
         media.clone(),
+        thread_id.map(str::to_string),
     )
     .await
     else {
@@ -3492,6 +3509,7 @@ impl SessionActor {
             &self.data_dir,
             content.clone(),
             media.clone(),
+            originating_thread_id.clone(),
         )
         .await;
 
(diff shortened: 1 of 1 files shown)


ymote:
For visibility: this PR is a tactical fix for one instance of a broader class. The structural fix is captured in the M9 architectural epic: #674 (with subtasks #675–#679).

This PR's `derive_thread_id_for_new_message` patch is exactly the kind of *"I'll figure out the route later"* path the codex review flagged — it works for the specific reproducer but the existence of a derive-from-ambient-state function is the smell. Step 5 of the epic (#679) explicitly calls for that function to be deleted, not refactored, once `ThreadStore` is the single writable source of thread ownership.

Leaving this comment so future debuggers find the epic from this fix.

PR #734, opened 2026-05-01
fix: add missing /queue command to help text

tmacychen:
The `/queue` command was fully implemented and functional, but was missing from the help text displayed when a user enters an unrecognized command. This caused a discrepancy between the documented commands and the actual available commands.

## Fix

Added the missing `/queue — view or change the queue mode` line to the help text in `crates/octos-cli/src/session_actor.rs`.

This was discovered by the test_help_command test case which correctly detected the discrepancy between available commands and the help text output.


ymote:
Superseded by #1073, which applied the actual missing /queue help entry onto current main with a regression test. This PR head has no merge base with current main and carries unrelated stale branch state, so it should not be merged as-is.

PR #761, opened 2026-05-02
feat(ui-protocol): live publish-subscribe for ledger events (closes #760, unblocks Phase C-4)

### LINKED ISSUE #760: Phase C dependency: verify message/persisted is actively published to live WS subscribers (not replay-only)
## Why

Phase C of /chat migration to UI Protocol v1 (\`octos-org/octos-web#66\`) depends on \`message/persisted\` notifications reaching live WebSocket subscribers in real-time, not just on reconnect-replay.

Codex critical review of the M8.10 wave-6 failure (\`/tmp/codex-failure-bc-review.log\`) flagged:

> "The comments at \`crates/octos-cli/src/api/ui_protocol.rs:741-751\` also say live fan-out for \`message/persisted\` is not general publish-subscribe yet; clients get it through replay/open."

If true, /chat on UI Protocol v1 would still hit M8.10-class symptoms: late background artifacts (deep_research result, mofa media, podcast .mp3) get persisted to ledger but never reach the live WS connection. Client wouldn't see them until reconnect or page reload — same UX bug as REST+SSE today.

## What needs verification

1. Read \`crates/octos-cli/src/api/ui_protocol.rs:723-793\` (commit observer that emits \`MessagePersistedEvent\`).
2. Trace whether the emitted event is:
   - Broadcast to all active WebSocket connections subscribed to that session, in real-time (good — Phase C unblocked)
   - Only stored for replay on next \`session/open\` (bad — Phase C blocked here)
   - Some hybrid (e.g., active for some event types, replay-only for others)
3. Trace the same question for \`task/updated\`, \`task/output/delta\`, \`turn/completed\` — any late notification that fires after the originating turn's WS RPC has been answered.

## Why this matters specifically for /chat (not octos-tui)

\`octos-tui\` works in production but its primary use case is interactive coding tasks where the user's turn ends with a tool result, not a 12-minute deep_research that returns a media file. /chat's M8.10 bug class is specifically about late artifacts arriving long after the originating turn's foreground stream closed.

## Acceptance

Either:
- Confirmation that \`message/persisted\` is actively published to live WS subscribers → Phase C proceeds as planned
- Identification of a gap → server-side fix issue opened with concrete site for the publish-subscribe wiring

## References

- Phase C tracking: \`octos-org/octos-web#66\`
- Wave-6c closeout: \`/tmp/wave6c-closeout.md\`
- Codex critical review: \`/tmp/codex-failure-bc-review.log\` (look for "live fan-out")
- Server commit observer: \`crates/octos-cli/src/api/ui_protocol.rs:723-793\`
- M9 ledger durability ADR: \`docs/M9-LEDGER-DURABILITY-ADR.md\`

## Out of scope

Don't fix here — verification only. If a gap is found, open a separate server PR for the fix.

ymote:
## Context

Closes #760. Unblocks Phase C-4 of the `/chat` migration to UI Protocol v1 over WebSocket (master tracker: octos-org/octos-web#66).

Without this fix, late background-task artifacts (deep_research result, mofa podcast .mp3, run_pipeline output, TTS audio) reach the durable ledger but never push to the live WebSocket connection — only via reconnect-replay. That's the same UX bug as REST+SSE today and made `/chat` over WS regress vs. the existing transport.

## What this changes

Adds a per-session `tokio::sync::broadcast` channel inside `UiProtocolLedger`, fans every newly persisted event out to it after the disk + ring write commits, and spawns a per-WebSocket forwarder task that pumps those events into the connection's existing `send_ledger_event_durable` path.

### Per-file summary

- `crates/octos-cli/src/api/ui_protocol_ledger.rs`
  - New `subscribers: HashMap<SessionKey, broadcast::Sender<LedgeredUiProtocolEvent>>` on `LedgerInner` (lazily populated; bounded capacity 256 per session).
  - `UiProtocolLedger::subscribe(&session_key) -> broadcast::Receiver<...>` — idempotent (reuses existing sender if any).
  - `publish_live(...)` runs after `entries.push_back` so the on-disk + ring state is committed before any subscriber sees the event. Failures (no receivers) are deliberately ignored — durability is unaffected.
  - `prune_idle_subscribers()` for housekeeping.
  - 6 new unit tests covering: live delivery of `message/persisted`, fan-out to multiple receivers, survival after one receiver drops, no-replay-of-past-events guarantee, no-subscriber durability, prune-idle behaviour.

- `crates/octos-cli/src/api/ui_protocol.rs`
  - `SharedLiveForwarders` registry (per-connection `HashMap<SessionKey, AbortHandle>`); aborted alongside `abort_connection_turns` on disconnect.
  - `handle_session_open` now subscribes to the broadcast BEFORE running `replay_after` so events appended during open buffer in the channel instead of slipping through. After the `opened_event` is sent, `spawn_live_forwarder` takes over — filtering events with `cursor.seq <= opened_event.cursor.seq` so we don't double-emit anything that was already in replay.
  - `live_event_passes_capability_filter` mirrors the existing UPCR-2026-012 gate at `ui_protocol.rs` session/open replay (line ~1600) so a connection without `event.message_persisted.v1` doesn't see those notifications via the live path either.
  - Lagged receivers are logged and the pump continues; the durable cursor remains the source of truth for catch-up via `session/hydrate` or reconnect.
  - 4 new integration tests using the existing `ws_connection_for_test` fixture: end-to-end push of `message/persisted` to a subscribed WS, baseline-seq filter, capability filter, and two-WS fan-out (including continued delivery to the survivor after one disconnects).

### Why this design

- **Append-then-broadcast** preserves the existing durability contract: the wire signal can never observe an event that isn't already on disk.
- **Subscribe-before-replay** closes the obvious race window between `replay_after` and the subscription handoff.
- **Bounded broadcast (256)** prevents OOM on a slow consumer; on `Lagged(_)` the forwarder logs and keeps pumping new events. Catch-up goes via the existing cursor mechanism — same fallback path the rest of the protocol already uses.
- **No new dependencies** — `tokio::sync::broadcast` is already in tokio's `full` feature.

## Test plan

- [x] `cargo build -p octos-cli --features api` — clean.
- [x] `cargo test -p octos-cli --features api --lib ui_protocol_ledger::` — 22 passed (1 ignored soak), all new + existing tests green.
- [x] `cargo test -p octos-cli --features api --lib ui_protocol` — 195 passed (1 ignored), including the 4 new live-forwarder integration tests.
- [x] `cargo test -p octos-cli --features api --lib` — 871 passed, no regressions.
- [x] `cargo clippy -p octos-cli --features api --no-deps` — no new warnings (34 before, 34 after; pre-existing multi-arg handler-signature pattern unchanged).
- [ ] Manual smoke against `octos serve` + `/chat` Phase C client (out of scope here per workspace policy: server merges aren't gated on TUI/web-app testing).

## Tracking

- Closes octos-org/octos#760
- Unblocks octos-org/octos-web#66 Phase C-4

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol.rs
+++ b/crates/octos-cli/src/api/ui_protocol.rs
@@ -98,6 +98,12 @@ type WsSink = futures::stream::SplitSink<WebSocket, WsMessage>;
 type SharedActiveTurns = Arc<tokio::sync::Mutex<HashMap<SessionKey, ActiveTurn>>>;
 type SharedConnectionTurns = Arc<tokio::sync::Mutex<HashMap<SessionKey, TurnId>>>;
 
+/// Per-connection registry of live ledger-forwarder tasks keyed by session.
+/// Each entry pumps `LedgeredUiProtocolEvent`s from the ledger broadcast
+/// into the WS write channel for the lifetime of the connection. Dropping
+/// or aborting a handle terminates the pump.
+type SharedLiveForwarders = Arc<tokio::sync::Mutex<HashMap<SessionKey, AbortHandle>>>;
+
 /// Outcome of pushing a frame onto the per-connection writer channel.
 ///
 /// All cases are non-fatal at the channel layer; callers decide how to react.
@@ -1240,6 +1246,8 @@ async fn ui_protocol_connection(
     let ws = WsConnection::new(writer_tx);
     let active_turns = active_turns_registry();
     let connection_turns: SharedConnectionTurns = Arc::new(tokio::sync::Mutex::new(HashMap::new()));
+    let live_forwarders: SharedLiveForwarders =
+        Arc::new(tokio::sync::Mutex::new(HashMap::new()));
     let contracts = contract_stores();
     let ledger = event_ledger(&state).await;
     // Force lazy init of the diff-preview store on this connection so
@@ -1287,6 +1295,7 @@ async fn ui_protocol_connection(
                     &state,
                     &ledger,
                     &contracts.approvals,
+                    &live_forwarders,
                     connection_profile_id,
                     features,
                     id,
@@ -1392,10 +1401,18 @@ async fn ui_protocol_connection(
     }
 
     abort_connection_turns(&active_turns, &connection_turns, &contracts.scopes).await;
+    abort_live_forwarders(&live_forwarders).await;
     // Dropping `ws` lets the writer task drain & exit; await it so the socket
     // is closed before we return.
     drop(ws);
     let _ = writer_handle.await;
+}
+
+async fn abort_live_forwarders(forwarders: &SharedLiveForwarders) {
+    let mut guard = forwarders.lock().await;
+    for (_, abort) in guard.drain() {
+        abort.abort();
+    }
 }
 
 fn parse_ws_text_frame(text: &str) -> Result<RpcRequest<Value>, RpcError> {
@@ -1541,13 +1558,24 @@ fn profile_mismatch_error(
 async fn handle_session_open(
     ws: &WsConnection,
     state: &Arc<AppState>,
-    ledger: &UiProtocolLedger,
+    ledger: &Arc<UiProtocolLedger>,
     approvals: &PendingApprovalStore,
+    live_forwarders: &SharedLiveForwarders,
     connection_profile_id: Option<&str>,
     features: ConnectionUiFeatures,
     id: String,
     params: SessionOpenParams,
 ) {
+    // Subscribe to the live ledger broadcast BEFORE the replay query so any
+    // event that lands while we're still computing replay/opened sits in the
+    // broadcast buffer and gets emitted by the forwarder once we hand it off
+    // (filtered to seq > opened.cursor.seq to avoid duplicating replay).
+    // Issue #760: without this, late background-task artifacts (deep_research
+    // result, mofa podcast, run_pipeline output, TTS audio) reach the ledger
+    // but never push to the live WS.
+    let session_id_for_subscribe = params.session_id.clone();
+    let live_rx = ledger.subscribe(&session_id_for_subscribe);
+
     let outcome = match open_session_result(
         state,
         ledger,
@@ -1560,6 +1588,8 @@ async fn handle_session_open(
     {
         Ok(outcome) => outcome,
         Err(error) => {
+            // Receiver drops on return; broadcast sender stays — future
+            // session/open on this WS just resubscribes.
             let _ = send_rpc_error(ws, Some(id), error);
             return;
         }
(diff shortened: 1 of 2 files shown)


ymote:
Addressed all 3 codex MUST-FIX items + the stale comment + the 4 gap tests + the 2 safe follow-ups in `87583271`.

**MUST-FIX-1 (replay/open race)**: capture the replay snapshot's head seq atomically with the events via a new `UiProtocolLedger::replay_after_with_head` API; the live forwarder now baselines on `replay_baseline_seq` rather than `opened_event.cursor.seq`. The `session/open` event itself is appended with `from_connection = Some(ws.connection_id)` so the forwarder skips it (the handler direct-sends inline). An event landing between replay and session/open append now passes the baseline filter and reaches the WS — covered by `live_forwarder_emits_event_appended_between_replay_and_forwarder_install`.

**MUST-FIX-2 (same-connection duplicate)**: chose codex's Option B. Added `ConnectionId` (atomic-counter newtype) on `WsConnection`, threaded through all direct-send paths (`send_notification_durable`, `send_notification_lifecycle`, `send_ledger_event_durable` for `session/open`, `append_progress_from` for status events, and `emit_replay_lossy_opportunistic`). The forwarder loop drops events whose `from_connection` matches its own connection. Other connections still receive via fan-out — verified by the second half of `send_notification_durable_does_not_double_deliver_via_live_forwarder`.

**MUST-FIX-3 (subscriber leak)**: removed the `cfg_attr(not(test), allow(dead_code))`. Wired `prune_idle_subscribers` into `sweep_idle` for steady-state GC. Added `prune_subscriber_if_idle(session_id)` for the targeted failure path; `handle_session_open` now drops the receiver and prunes immediately on `open_session_result` error. Tested by `session_open_failure_path_does_not_leak_broadcast_sender` (failure path) and the existing `prune_idle_subscribers_drops_orphaned_senders` (sweep path).

**Stale comment**: revised `install_message_commit_observer` doc-block at `ui_protocol.rs:747-757` to reflect that live publish-subscribe IS now the delivery path; references the broadcast channel + forwarder.

**Tests added**:
- `live_forwarder_emits_event_appended_between_replay_and_forwarder_install` (gap)
- `send_notification_durable_does_not_double_deliver_via_live_forwarder` (dedupe; also asserts cross-connection fan-out still works)
- `session_open_failure_path_does_not_leak_broadcast_sender` (pruning, both targeted + sweep)
- `live_forwarder_survives_broadcast_lag_and_keeps_pumping` (overflows the 256-slot capacity, asserts the forwarder logs+continues)

**Safe follow-ups also applied**:
- `publish_live` clones the sender then drops the lock before `broadcast::send` so a slow subscriber cannot block the next append.
- `Lagged(n)` warning includes the skipped count for ops visibility.

**Build/test**: `cargo build -p octos-cli --features api` clean. `cargo test --lib -p octos-cli --features api` → 875 passed / 0 failed / 3 ignored. `cargo clippy --no-deps -p octos-cli --features api` net delta is +1 "too many arguments (8/7)" lint on `spawn_live_forwarder` (matches the existing pattern in this file).

LOC delta: 2 files, +502 / -54.
HEAD: `87583271`

PR #924, opened 2026-05-13
fix(codex-audit): WS lifecycle + session-id + ledger replay + backpressure + ops hardening

### LINKED ISSUE #919: Session-id consistency: profile-auth raw web-* fails hydrate/state/title/delete
Source: codex audit thread `019e2017-3552-7342-83a3-af4b7732ffd8`, 2026-05-13.
Sibling of #857. The profile-auth raw `web-*` session_id story is half-fixed.

## 1. hydrate/turn-state lookups go to the wrong session map (BLOCK)
Files: `crates/octos-cli/src/api/ui_protocol.rs:6582` + `:4226` + `:4523`
Symptom: profile-auth raw `web-*` sessions can start turns but later fail `session/hydrate` or `turn/state.get` as `unknown_session`. Reconnect can't rehydrate.
Root cause: turn persistence uses the profile `SessionRuntime.sessions`, while hydrate/state lookups check exact IDs in top-level `state.sessions`.
Fix: resolve the SAME profile runtime/session manager for hydrate and state RPCs. Canonicalize session IDs at `session/open` and use the canonical form everywhere.

## 2. REST title/delete miss raw web-* under profile auth (BLOCK)
Files: `crates/octos-cli/src/api/handlers.rs:147` + `:951` + `:998`
Symptom: legitimate raw `web-*` sessions can't be retitled or deleted via the REST endpoints — returns 404/403.
Root cause: title/delete still use the older session-key candidate helper, which lacks the newer identity/topic-aware raw-session handling.
Fix: centralize session-key resolution into one helper. Use it for messages, title, delete, hydrate, and state. Resolve raw keys ONLY inside the authenticated profile's store.

## Acceptance
- Open a fresh dspfac session, send a message, hard-reload → `session/hydrate` succeeds (no `unknown_session`)
- Right-click session in sidebar → "Rename" + "Delete" both succeed on a raw `web-*` id under profile auth
- mini5 live verification

### LINKED ISSUE #920: WS lifecycle: abrupt close leaves turn forever-active + pending approvals orphaned
Source: codex audit thread `019e2017-3552-7342-83a3-af4b7732ffd8`, 2026-05-13.

## 1. WS close leaves turn permanently projected as active (BLOCK)
Files: `crates/octos-cli/src/api/ui_protocol.rs:1599` + `:7641` + `:4761`
Symptom: client reconnects, sees the old turn still "running" forever — no `turn/error` or `turn/completed` ever lands. UI shows a spinner that never clears.
Root cause: `abort_connection_turns` aborts the task and removes in-memory tracking, but does NOT append a terminal `turn/error`/`turn/completed` ledger event. Reconnect replays the durable log and sees an active turn with no terminal.
Fix: on connection cleanup, append a durable terminal event (e.g. `connection_closed` or `interrupted`) for every in-flight turn before removing it.

## 2. Pending approvals survive closed WS (BLOCK)
Files: `crates/octos-cli/src/api/ui_protocol.rs:1376` + `:7641` + `crates/octos-cli/src/api/ui_protocol_approvals.rs:288`
Symptom: an approval modal can re-appear on reconnect for a turn that no longer exists.
Root cause: connection cleanup aborts the turn and evicts scopes, but doesn't cancel pending approvals or append `approval/cancelled`.
Fix: pass approval store + ledger into cleanup. Cancel approvals for each aborted turn. Publish durable `approval/cancelled` events.

## Acceptance
- Yank network mid-turn → reconnect shows a terminal "interrupted" state for that turn, not a stuck spinner
- Yank network mid-approval prompt → reconnect doesn't re-show the modal for a dead turn
- Ledger replay shows clean terminal-event coverage

### LINKED ISSUE #921: Ledger replay: MessagePersisted/TurnSpawnComplete cursors not extracted
Source: codex audit thread `019e2017-3552-7342-83a3-af4b7732ffd8`, 2026-05-13.

## Dropped durable events miss protocol/replay_lossy (BLOCK)
`crates/octos-cli/src/api/ui_protocol.rs:7806` + `:7858` + `crates/octos-cli/src/api/ui_protocol_ledger.rs:166`
Symptom: dropped `message/persisted` and `turn/spawn_complete` events can be silently lost on reconnect — client never sees `protocol/replay_lossy`, so it doesn't refetch.
Root cause: the ledger stamps cursors into those payloads, but `ledger_event_cursor` only extracts cursors from `session/opened` and `turn/completed`. Replay-loss detection compares "last seen cursor" against the durable log, but MessagePersisted/TurnSpawnComplete carry their cursor in a field the extractor doesn't read.
Fix: extend `ledger_event_cursor` to return cursors for all cursor-bearing durable events, including `MessagePersisted` and `TurnSpawnComplete`.

## Acceptance
- Disconnect mid-stream during a heavy `turn/spawn_complete` flurry → reconnect detects the gap and emits `protocol/replay_lossy`
- Unit test: `ledger_event_cursor` exhaustively covers every variant that carries a cursor

### LINKED ISSUE #922: WS reader/writer: ping no-id parse-error + RPC reply backpressure drops
Source: codex audit thread `019e2017-3552-7342-83a3-af4b7732ffd8`, 2026-05-13.
Note: the user-visible disconnect bug this would trigger has been mitigated by the new text-frame heartbeat (commit lands separately). But the underlying parse-strictness + silent-drop paths remain real bugs worth closing.

## 1. Browser ping notification (no id) → parse_error (BLOCK)
Files: `crates/octos-cli/src/api/ui_protocol.rs:2263` + `:1610` + `crates/octos-core/src/ui_protocol.rs:363`
Symptom: the SPA's `{"jsonrpc":"2.0","method":"ping","params":{...}}` keepalive notification fails server-side parsing every 30s.
Root cause: `serde_json::from_str::<RpcRequest<Value>>` requires `RpcRequest.id: String` (not `Option`). Missing `id` yields a parse error, server replies via `send_rpc_error(&ws, None, ...)`.
Fix: parse the raw envelope first. Accept known inbound notifications (e.g. `ping`) with no response. Only parse as `RpcRequest` when `id` is present.

## 2. RPC replies silently dropped on writer backpressure (BLOCK)
Files: `crates/octos-cli/src/api/ui_protocol.rs:211` + `:223` + `:5039`
Symptom: under stress, RPC replies vanish — client times out, server thinks it succeeded.
Root cause: WS writer channel is bounded at 1024 + uses `try_send`. Many call sites discard `send_rpc_result`/`send_rpc_error` failures with `let _ = ...` or debug-only logging.
Fix: treat lifecycle/RPC send failure as connection failure. Stop dispatch, abort/close the connection. Consider awaited `send` with timeout, OR a reserved control/RPC path that doesn't share the bounded channel.

## Acceptance
- Bridge sends idle keepalive ping → server does NOT emit a `parse_error` reply (verify in server logs)
- Stress test: 1000 RPCs in tight loop → no silent drops; either all answered or connection terminates with a clear error

### LINKED ISSUE #923: Server ops hardening: ledger lock latency, subscriber GC, Origin gate, poison locks
Source: codex audit thread `019e2017-3552-7342-83a3-af4b7732ffd8`, 2026-05-13.

## 1. Ledger Mutex held across disk I/O (SMELL)
Files: `crates/octos-cli/src/api/ui_protocol_ledger.rs:612` + `:854` + `:312`
Symptom: a global `std::sync::Mutex` held across synchronous file open/write/flush + O(n) LRU updates blocks unrelated WS traffic under load.
Fix: keep the lock only for sequence/ring mutation. Move disk writes to a dedicated writer or `spawn_blocking`. Replace linear LRU scans if scale grows.

## 2. Closed WS live-subscribers linger until periodic sweep (SMELL)
Files: `crates/octos-cli/src/api/ui_protocol.rs:2249` + `crates/octos-cli/src/api/ui_protocol_ledger.rs:746`
Symptom: cleanup aborts forwarders but doesn't immediately prune the per-session broadcast sender map. Subscribers accumulate.
Fix: track drained session IDs, await/yield after aborting forwarders, then call `prune_subscriber_if_idle` immediately.

## 3. WS upgrade not Origin-gated (SMELL — security)
Files: `crates/octos-cli/src/api/router.rs:73` + `:135` + `crates/octos-cli/src/api/ui_protocol.rs:1524`
Symptom: browser cross-site WS attempts rely purely on token possession. CORS doesn't apply to WS handshakes.
Fix: reject disallowed browser origins before `on_upgrade`, sharing the existing CORS allowlist. Allow missing-Origin only for trusted non-browser clients.

## 4. Hot-path locks use .expect() (SMELL)
Files: `crates/octos-cli/src/api/ui_protocol_ledger.rs:612` + `crates/octos-cli/src/api/ui_protocol_scope.rs:155` + `crates/octos-cli/src/api/ui_protocol_approvals.rs:74`
Symptom: a poisoned lock panics future WS/session operations.
Fix: recover from `PoisonError` with logging where safe, OR switch to non-poisoning locks, OR convert unrecoverable state into RPC internal errors.

## Acceptance
- Stress test: ledger writes don't visibly block other WS traffic
- Closed connection's subscribers cleared within ~1s
- Cross-origin browser WS upgrade rejected at the handshake
- No `.expect()` on hot-path lock acquisition

ymote:
## Summary

Bundles the five issues opened from the 2026-05-13 codex audit of the UI Protocol v1 WS server path.

- **#919 — Session-id consistency under profile auth**: hydrate/state/title/delete now resolve the SAME profile-scoped `SessionManager` that `turn/start` writes to. Raw `web-*` sessions under profile auth no longer return `unknown_session` on reconnect-hydrate or 404 on rename/delete.
- **#920 — WS lifecycle cleanup**: connection cleanup appends a durable `turn/error code=connection_closed` for every in-flight turn AND cancels every still-pending approval (publishing `approval/cancelled` to the ledger). Reconnect-replay no longer renders a stuck spinner or a modal for a dead turn.
- **#921 — Ledger replay completeness**: `ledger_event_cursor` now extracts cursors from every cursor-bearing durable variant — adds `MessagePersisted` and `TurnSpawnComplete` alongside the existing `SessionOpened` / `TurnCompleted`. Dropped sends now trigger `protocol/replay_lossy` so the client refetches.
- **#922 — Reader/writer parse + backpressure**: id-less JSON-RPC envelopes (the bridge's 30-second `ping`) are parsed as notifications, not requests, so they no longer trip `parse_error`. A lifecycle/RPC send failure latches a `failed` flag on the connection and the read loop tears down cleanly instead of silently dropping RPC replies.
- **#923 — Ops hardening**: WS upgrade is now Origin-gated against the shared CORS allowlist; closed connections eagerly prune broadcast senders (no waiting for the periodic sweep); poisoned hot-path locks on the ledger / scope / approval stores recover gracefully via `unwrap_or_else(|p| p.into_inner())` instead of panicking the next acquirer.

Closes #919.
Closes #920.
Closes #921.
Closes #922.
Closes #923.

## Test plan

- [x] `cargo build --workspace --all-features`
- [x] `cargo test --workspace --no-run`
- [x] `cargo test -p octos-cli --lib --features api,telegram` — 946 passed
- [x] New unit tests:
  - [x] `ledger_event_cursor_covers_every_cursor_bearing_variant` — asserts cursor extraction for all four cursor-bearing notification variants plus a negative case
  - [x] `idless_envelope_parses_as_notification` — asserts the new `ParsedFrame::Notification` arm fires for envelopes without `id`
- [x] Updated existing tests:
  - [x] `abort_connection_turns_removes_only_matching_active_turns` — passes ledger + approval store args
  - [x] `delete_session_accepts_listed_topic_session_id_from_standalone_store` — passes `None` identity
  - [x] All `abort_live_forwarders` test call sites pass the `ledger` arg

### DIFF
--- a/crates/octos-cli/src/api/handlers.rs
+++ b/crates/octos-cli/src/api/handlers.rs
@@ -932,31 +932,41 @@ pub async fn update_session_title(
     }
 
     let mut updated = false;
+    let identity_ref = identity.as_ref().map(|ext| &ext.0);
+    let routed_profile_id = routed_profile_id_from_headers(&state, &headers);
+    let candidates =
+        standalone_api_session_key_candidates_with_topic(&state, &headers, identity_ref, &id, None);
 
-    if let Some(sessions) = &state.sessions {
-        let mut sess = sessions.lock().await;
-        // #919.2: use the auth-aware candidate helper so raw `web-*`
-        // sessions under profile auth resolve to the profile's own
-        // store. The older `standalone_api_session_key_candidates`
-        // skipped tenant-scoped raw-id fallbacks.
-        let identity_ref = identity.as_ref().map(|ext| &ext.0);
-        for key in standalone_api_session_key_candidates_with_topic(
+    // #924 BLOCK 3: each candidate `SessionKey` may belong to a
+    // different profile (the candidate set includes the resolved
+    // tenant profile, `_main`, and bare-channel/raw-id forms). Route
+    // each candidate through `resolve_sessions_for_lookup` so the
+    // profile's own `SessionRuntime.sessions` is locked — turn
+    // persistence writes there, so the title update MUST land in the
+    // same store. The legacy code locked only `state.sessions`, which
+    // did not contain profile-scoped runtime sessions and returned
+    // `NOT_FOUND` for raw `web-*` ids under profile auth.
+    for key in candidates {
+        let Some(sessions) = super::ui_protocol::resolve_sessions_for_lookup(
             &state,
-            &headers,
-            identity_ref,
-            &id,
             None,
-        ) {
-            if sess.load(&key).await.is_some() {
-                if let Err(e) = sess.update_title(&key, title.clone()).await {
-                    tracing::error!(
-                        session_key = %key,
-                        error = %e,
-                        "update_title in standalone store failed"
-                    );
-                } else {
-                    updated = true;
-                }
+            routed_profile_id.as_deref(),
+            &key,
+        )
+        .await
+        else {
+            continue;
+        };
+        let mut sess = sessions.lock().await;
+        if sess.load(&key).await.is_some() {
+            if let Err(e) = sess.update_title(&key, title.clone()).await {
+                tracing::error!(
+                    session_key = %key,
+                    error = %e,
+                    "update_title in profile-scoped store failed"
+                );
+            } else {
+                updated = true;
             }
         }
     }
@@ -991,26 +1001,35 @@ pub async fn delete_session(
     identity: Option<Extension<AuthIdentity>>,
     axum::extract::Path(id): axum::extract::Path<String>,
 ) -> Response {
-    // Clear from the standalone store if available.
-    if let Some(sessions) = &state.sessions {
-        let mut sess = sessions.lock().await;
-        // #919.2: use the auth-aware candidate helper.
-        let identity_ref = identity.as_ref().map(|ext| &ext.0);
-        for key in standalone_api_session_key_candidates_with_topic(
+    let identity_ref = identity.as_ref().map(|ext| &ext.0);
+    let routed_profile_id = routed_profile_id_from_headers(&state, &headers);
+    let candidates =
+        standalone_api_session_key_candidates_with_topic(&state, &headers, identity_ref, &id, None);
+
+    // #924 BLOCK 3: route each candidate through the profile-aware
+    // SessionManager resolver — turn persistence writes to the
+    // profile's `SessionRuntime.sessions`, so deletes MUST hit the
+    // same store. The legacy code locked only `state.sessions`, which
+    // did not contain profile-scoped runtime sessions.
+    for key in candidates {
+        let Some(sessions) = super::ui_protocol::resolve_sessions_for_lookup(
             &state,
-            &headers,
-            identity_ref,
-            &id,
             None,
-        ) {
-            if sess.load(&key).await.is_some() {
-                if let Err(e) = sess.clear(&key).await {
-                    tracing::error!(
-                        session_key = %key,
-                        error = %e,
-                        "delete session from standalone store failed"
-                    );
-                }
+            routed_profile_id.as_deref(),
+            &key,
+        )
+        .await
+        else {
+            continue;
+        };
+        let mut sess = sessions.lock().await;
+        if sess.load(&key).await.is_some() {
+            if let Err(e) = sess.clear(&key).await {
+                tracing::error!(
+                    session_key = %key,
+                    error = %e,
+                    "delete session from profile-scoped store failed"
+                );
             }
         }
     }
--- a/crates/octos-cli/src/api/handlers.rs
+++ b/crates/octos-cli/src/api/handlers.rs
@@ -144,21 +144,6 @@ fn api_profile_id_from_headers(state: &AppState, headers: &HeaderMap) -> String 
     routed_profile_id_from_headers(state, headers).unwrap_or_else(|| MAIN_PROFILE_ID.to_string())
 }
 
-fn standalone_api_session_key_candidates(
-    state: &AppState,
-    headers: &HeaderMap,
-    session_id: &str,
-) -> Vec<SessionKey> {
-    let profile_id = api_profile_id_from_headers(state, headers);
-    let mut candidates = vec![
-        SessionKey::with_profile(&profile_id, "api", session_id),
-        SessionKey::with_profile(MAIN_PROFILE_ID, "api", session_id),
-        SessionKey::new("api", session_id),
-    ];
-    candidates.dedup_by(|left, right| left.0 == right.0);
-    candidates
-}
-
 /// Returns `true` when `session_id` is a bare SPA id whose raw form is
 /// safe to query as a `SessionKey` directly. Specifically: no `:` (which
 /// is the channel/profile separator in
@@ -178,7 +163,8 @@ fn is_safe_bare_session_id(session_id: &str) -> bool {
     !session_id.contains(':') && !session_id.contains('#')
 }
 
-/// Topic-aware sibling of [`standalone_api_session_key_candidates`].
+/// Topic-aware, auth-aware candidate resolver shared by REST
+/// `/messages`, `session/title.set`, and `session/delete`.
 ///
 /// Returns the candidate `SessionKey`s the REST `/messages` and similar
 /// read paths should try, in fallback order. The fallback set is split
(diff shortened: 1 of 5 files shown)


ymote:
Addresses 5 BLOCKs + 3 NITs from codex review (thread `019e2084-ec1e-7c63-9c3c-310e84b57296`):

- **BLOCK 1+2** (`e90e280c`): wake the read loop the instant `mark_failed` latches via a `tokio::sync::Notify`; gate `try_enqueue` on the latch with new `SendError::FatalClosed` so background forwarders stop pumping onto a dead writer. Lifecycle wrappers translate FatalClosed → LifecycleFailure for RPC callers; durable/ephemeral wrappers treat it like Closed; live forwarder breaks on Closed | FatalClosed.
- **BLOCK 3+4** (`0231feae`): `handlers::update_session_title` / `delete_session` now route each candidate `SessionKey` through `super::ui_protocol::resolve_sessions_for_lookup` so the profile's own `SessionRuntime.sessions` is locked (not `state.sessions`). `resolve_sessions_for_lookup` gained a `routed_profile_id` arg and mirrors `handle_turn_start`'s precedence: `session_id.profile_id() || connection_profile_id || routed_profile_id`. Plumbed through `handle_session_hydrate`, `handle_thread_graph_get`, `handle_turn_state_get`, dispatchers, and test invocations.
- **BLOCK 5** (`00a81f05`): refactor the inline Origin gate into `decide_ws_origin_gate`; treat present-but-empty (after trim) Origin as absent (allow) for server-to-server clients; keep rejecting present-non-empty origins that fail the allowlist; keep rejecting non-ASCII bytes as malformed (NOT same as absent — that would be a downgrade vector). Three unit tests.
- **NIT 6+7+8** (`ae7e52cd`): (6) `id: null` / numeric ids rejected loudly via parse_error; key-presence (`value.get("id").is_some()`) is the notification-vs-request signal. (7) `ledger_event_cursor` is fully exhaustive — no `_ => None` catchall; every `UiNotification` variant is enumerated so a future cursor-bearing variant won't silently regress like #921 did. (8) `SharedLiveForwarders` switched from `AbortHandle` to `JoinHandle`; `abort_live_forwarders` now does `abort()` + `await` per task so the receiver-drop has provably happened before `prune_subscriber_if_idle` runs (the `yield_now` was best-effort and could lose under load).

All 953 lib tests pass under `--features api,telegram`; workspace builds with `--all-features`; cargo fmt clean.

ymote:
Addresses codex round-2 BLOCK from thread 019e20b3-8cf1-7bd1-9ce2-c434b51ae06c: re-checks the `failed` latch immediately after registering `notified()` to close the lost-notify ordering race. New regression test `mark_failed_during_idle_loop_still_cleans_up` exercises the path the round-1 test missed.

Issue #114, opened 2026-03-27
O(n²) newline processing in message handling


ymote:
## Description
Message processing has O(n²) behavior when handling newlines in large messages due to repeated string concatenation or scanning.

## Fix
Use a single-pass approach or `String::with_capacity` pre-allocation.

## Severity
LOW

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #926, opened 2026-05-13
fix(ws): JSON heartbeat + auth close-code 1008

ymote:
## Summary

Two small server-side WS fixes captured as separate commits.

### Commit 1 — `fix(ws): emit JSON-RPC heartbeat text frame instead of binary Ping`

Browsers auto-Pong protocol-level Ping frames at the WebSocket layer and never deliver them to JS-land `onmessage`. The SPA bridge runs an application-layer 60s idle timer that only refreshes on text frames observed by the JS handler, so a binary-Ping keepalive ticked the proxy/TCP idle meter but starved the bridge idle meter — the SPA self-closed every ~60s of idle with code 4000 `keepalive_timeout`.

`writer_loop` now emits a JSON-RPC `server/heartbeat` notification text frame on a 20s cadence so both meters (proxy idle + bridge idle) tick together. The bridge silently ignores unknown notification methods; no client changes needed for the basic fix.

**This fix is already deployed to mini5** (`dspfac.ocean.ominix.io`) — verified empirically: 90s idle → 4 heartbeats received, 0 CLOSE events. Pre-fix this exact probe showed CLOSE at 60s.

### Commit 2 — `fix(ws): close auth-rejected sessions with code 1008`

The SPA bridge listens for WS close-code **1008** to trigger its `crew:auth_expired` flow (octos-web PR #114 — `auth-context.tsx` clears the cached token and routes to `/login`). Server previously sent an `RpcError` envelope and then `sink.close()` with no specific code, leaving the client retrying the upgrade forever on a dead token.

Three auth sites now emit a follow-up `CloseFrame{ code: 1008, reason: "auth_expired" }` after the error envelope:

1. The 3 `content/list`, `content/delete`, `content/bulk-delete` handlers that reject anonymous calls with `RpcError::permission_denied(... "authenticated user identity required" ...)`.
2. Every `validate_session_scope` caller, **but only when the inner failure originated from `validate_authenticated_session_scope`** (the connection IS authenticated and the requested scope falls outside that user's profile). Tagged on the error via a new `auth_scope_violation: true` data field and consumed by a new `send_scope_error` helper. Non-auth scope errors (malformed input, unauthenticated mismatch, etc.) still close without a specific code.
3. The same path as (2) propagated through `open_session_result` (which uses `?`-propagation for `validate_session_scope`).

Includes two new unit tests:
- `send_scope_error_closes_with_1008_on_authenticated_mismatch`
- `send_scope_error_does_not_close_when_unauthenticated`

## Test plan

- [x] `cargo build --workspace --all-features`
- [x] `cargo test -p octos-cli --lib --features api,telegram` — 946 passed, 0 failed
- [x] Heartbeat fix verified on mini5: 90s idle → 4 heartbeats, 0 CLOSE events
- [ ] octos-web PR #114 close-code-1008 listener fires on bad-token WS upgrade once this lands

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol.rs
+++ b/crates/octos-cli/src/api/ui_protocol.rs
@@ -9403,20 +9403,30 @@ mod tests {
     }
 
     /// Codex BLOCK regression (2026-05-13): with the writer channel at
-    /// capacity 1 and already full, `send_scope_error` must still emit the
-    /// 1008 close. The courtesy error envelope is allowed to drop; the close
-    /// is the load-bearing signal the SPA's `crew:auth_expired` listener
-    /// uses to clear its token.
+    /// capacity 2 and one slot already used, `send_scope_error` must use
+    /// the remaining slot for the 1008 close — NOT the courtesy error
+    /// envelope. The close is the load-bearing signal the SPA's
+    /// `crew:auth_expired` listener uses to clear its token. The error
+    /// envelope is allowed to drop under backpressure.
+    ///
+    /// Test geometry: capacity 2 + 1 primer = exactly one free slot at the
+    /// moment `send_scope_error` enqueues. Pre-fix the order was
+    /// error-then-close → error queued, close dropped. Post-fix the order
+    /// is close-then-error → close queued, error dropped.
     #[tokio::test]
     async fn auth_scope_violation_close_frame_survives_capacity_one_writer() {
-        let (ws, mut rx) = ws_connection_for_test(1);
+        let (ws, mut rx) = ws_connection_for_test(2);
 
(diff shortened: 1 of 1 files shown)


ymote:
Addresses codex BLOCK from thread `019e20a6-4b3f-7bd2-824d-54a3d0039c9b`: close frame now sent before error envelope so the auth-expired signal survives writer backpressure. New regression test `auth_scope_violation_close_frame_survives_capacity_one_writer` exercises the capacity-1 case.

ymote:
Test geometry fix (thread `019e20d1-b550-7641-aac8-7612c0a1991c`): capacity 2 + 1 primer = exactly one free slot, now faithfully models the production backpressure case. Test passes locally.

PR #941, opened 2026-05-13
feat(workspace-policy): spawn_only named_outputs + ${output.X} interpolation + mofa_publish probe

ymote:
## Summary

Wave-3b harness-audit follow-up. Extends the spawn_only stdout envelope
with an optional `named_outputs` field so tools can surface structured
values, threads those values into the validator runner via
`${output.<key>}` template interpolation, and wires the canonical
`mofa_publish` HTTP probe into the default session policy.

Closes audit P0-3 on the harness side (mofa-skills repo follow-up still
needs to teach the skill to emit `named_outputs.deploy_url`).

## Protocol extension

Before:
```json
{"success": true, "output": "...", "files_to_send": [...]}
```

After:
```json
{
  "success": true,
  "output": "...",
  "files_to_send": [...],
  "named_outputs": {"deploy_url": "https://..."}
}
```

- `named_outputs` is OPTIONAL (absent → today's behaviour).
- Values are STRINGS only in v1 (nested JSON deferred). Reject at parse.
- Keys must match `[a-z][a-z0-9_]*`. Reject at parse.
- Malformed payload (non-object, non-string value, bad key shape) surfaces
  as a typed failure so the contract layer rejects the spawn_only result
  instead of silently dropping the field.

## ${output.X} interpolation

Inventory of string-interpolating fields the new prefix now resolves
against (all share one `interpolate_template` codepath):

- `Command.cmd` + every `args[]` slot
- `HttpProbe.url_template`
- `HttpProbe.expected_contains` (was a literal-only string before)
- `FileExists.path`
- `MagicBytes.glob`
- `AudioNonSilent.glob`

Trust-boundary distinction preserved:

- `${args.X}` → percent-encoded (LLM-controlled, URL-segment safe).
- `${output.X}` → spliced VERBATIM (tool-controlled, canonical use is a
  full URL the probe must call as-is; percent-encoding would corrupt it).

Missing key in either source → typed `Error` outcome (matches
`${args.X}` semantics from #935). Templates can mix both prefixes in
one pass.

## mofa_publish wiring

`WorkspacePolicy::for_session()` now declares:

```rust
spawn_tasks.mofa_publish = {
  on_completion = [HttpProbe {
    url_template = "${output.deploy_url}",
    expected_status = 200,
    expected_contains = Some("<!DOCTYPE"),  // soft-404 sentinel
  }],
  on_failure = ["notify_user:Publish probe failed"],
}
```

The probe asserts BOTH a 200 status AND a `<!DOCTYPE` body prefix,
guarding against the "200 OK with a JSON soft-404 body" failure mode
the audit calls out (section 6 P0-3).

**`required = false` intentionally.** The mofa_publish skill does NOT
yet emit `named_outputs.deploy_url` — a separate mofa-skills repo PR
is needed for that. Until then `${output.deploy_url}` would surface
`Error` on every call; `required = false` keeps the validator running
as a diagnostic without blocking users. Flip to `required = true` once
the skill catches up.

## Mofa-skills repo follow-up (NOT in this PR)

- `mofa_publish/src/main.rs` must emit `{"named_outputs":{"deploy_url":"..."}}`
  on its stdout envelope after a successful publish.
- Optional future emitters (separate PRs):
  - `mofa_slides` → `"deck_path"` so MagicBytes can scope to the actual
    emitted PPTX instead of the recursive `**/*.pptx` glob.
  - `mofa_cards` → `"card_dir"` for the same reason.

## Test coverage

+34 lib tests vs the wave-3a baseline (1325 → 1359):

- `plugins::tool::tests` — 11 parser tests covering absent / null /
  empty / valid / bad key shape (digit, uppercase, hyphen, empty) /
  non-string value / underscore-and-digits, plus 3 end-to-end
  `execute()` tests proving the field threads through `ToolResult`,
  malformed payloads return failure, legacy plugins stay `None`.
- `validators::tests` — 11 `${output.X}` tests covering verbatim
  substitution, missing-key error, mixed `args+output` templates,
  percent-encoding boundary preserved across prefixes, FileExists /
  HttpProbe / Command / MagicBytes / AudioNonSilent each interpolating
  output, `expected_contains` mixing args+output.
- `workspace_policy::tests` — 1 test asserting the mofa_publish
  HttpProbe is registered with the right shape (URL template, status,
  DOCTYPE sentinel, non-required).
- `workspace_contract::tests` — 4 end-to-end tests (deploy_url
  satisfies the gate, soft-404 body fails, missing named_outputs fails
  with `required=true`, default policy doesn't block users) + 1
  generic test for the named_outputs → validator pipeline (FileExists
  probing a tool-emitted path).

Plus 7 mechanical updates to existing test sites that constructed
`ValidatorInvocation` with field literals (added `tool_output: None`).

## Backward compatibility

- `ToolResult::named_outputs` defaults to `None`; all existing
  `..Default::default()` constructors stay byte-identical.
- `enforce_spawn_task_contract` and `enforce_spawn_task_contract_with_args`
  remain as thin wrappers around the new
  `enforce_spawn_task_contract_with_args_and_output` entry point.
- `run_declared_validators` stays as a thin wrapper around
  `run_declared_validators_with_output`.
- Existing plugin envelopes that don't emit `named_outputs` parse
  cleanly to `ToolResult.named_outputs = None`; their validators see
  `invocation.tool_output = None` and behave as before.

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p octos-agent --lib` (1359 passed)
- [x] `cargo test -p octos-agent` (lib + integration)
- [x] `cargo test -p octos-swarm` / `octos-pipeline`
- [ ] mofa-skills repo follow-up: teach `mofa_publish` to emit
      `named_outputs.deploy_url`, then flip `required = true` here.

### DIFF
--- a/crates/octos-agent/src/agent/execution.rs
+++ b/crates/octos-agent/src/agent/execution.rs
@@ -123,6 +123,34 @@ fn relativize_workspace_path(path: &str, workspace_root: Option<&std::path::Path
     }
 }
 
+/// Decide what `content` to put on the spawn-only background-result payload
+/// when the workspace contract reports `Satisfied`.
+///
+/// Wave-3b regression guard: a contract entry with no declared artifact
+/// (e.g. `mofa_publish`, whose deliverable is a live URL, not a file)
+/// returns `Satisfied { output_files: [] }`. Before this fix the handler
+/// fell through to `content: String::new()`, dropping the tool's stdout
+/// text (the deploy URL itself for `mofa_publish`) and the user saw a
+/// blank completion.
+///
+/// Rule:
+///
+/// - When `output_files` is non-empty: emit empty content; the files
+///   themselves carry the deliverable. This matches the legacy
+///   file-attached behaviour for `fm_tts` / `podcast_generate` / etc.
+/// - When `output_files` is empty: emit the tool's stdout `output` as
+///   content. The contract verified the artifact-shaped portion of the
+///   deliverable (e.g. the HttpProbe asserting deploy_url returned
+///   `<!DOCTYPE`), but the user-visible text payload (the live URL) is
+///   what the LLM/user actually consumes.
+pub(super) fn satisfied_completion_content(output_files: &[String], tool_output: &str) -> String {
+    if output_files.is_empty() {
+        tool_output.to_string()
+    } else {
+        String::new()
+    }
+}
+
 /// Produce the composite system-prompt text (worker prompt + realtime sensor
 /// summary) used at the top of every agent turn. Centralizing this in
 /// `execution.rs` keeps the message-building policy in a single location so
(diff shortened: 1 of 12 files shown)


ymote:
Codex review caught a P1 regression: registering `mofa_publish` in `for_session()` changed its background-completion path from the legacy `NotConfigured` text-fallback to the `Satisfied` artifact-delivery branch. Because the contract has no artifact (its deliverable is a live URL, not a file), the `Satisfied` arm passed `content: String::new()` to the background sender, dropping the tool's stdout text (the deploy URL itself) and the user would have seen a blank completion.

Fixed in 36d59ed4 by extracting `satisfied_completion_content(&output_files, &tool_output)` that surfaces the tool's stdout text when there are no files to attach. Byte-identical behaviour for every artifact-carrying contract (fm_tts, podcast_generate, mofa_slides, ...); restores the text-fallback for no-artifact contracts like mofa_publish.

+3 unit tests in `crates/octos-agent/src/agent/execution.rs::tests` cover both branches plus the defensive "no files and no text" case. Lib test count: 1325 (baseline) → 1362 (+37).

Issue #58, opened 2026-03-21
feat: pipeline executor observability logging


ymote:
## Summary
- Pipeline start: graph structure, nodes, models, tools, edges
- Per-node: model, input_bytes, status, duration, tokens, output_chars
- Pipeline end: full summary
- DOT parse errors: log full DOT content
- Worker model pool distribution log

PR #955, opened 2026-05-14
feat(harness): add PerFileNonSilent validator + wire mofa-podcast segments

ymote:
## Summary

- Add `ValidatorSpec::PerFileNonSilent { glob, min_ratio, require_at_least }` to the harness alongside `AudioNonSilent`. Runner enforces a per-file non-silent ratio threshold AND a minimum-match-count floor.
- Wire it into `WorkspacePolicy::for_session()::podcast_generate` as a third `on_completion` validator, scoped to `skill-output/mofa-podcast/**/segments/seg_*.wav` so it catches a silent intermediate segment that the whole-file `AudioNonSilent` cannot detect (the silent gap gets averaged out by surrounding speech in the assembled MP3).
- Glob deliberately excludes `pause_after_*`, `pause_line_*`, `bgm_placeholder_line_*` placeholders — those filenames are emitted by mofa-podcast as legitimately-silent inter-segment pauses.

## Why

Codex review of mofa-skills #58 surfaced this as a P2: harness's whole-file `AudioNonSilent` can't catch a single silent segment in a multi-segment podcast. Two independent codex passes converged on the same finding.

Now that mofa-skills #59 preserves segments after successful assembly (via Result-aware RAII `SegmentDirCleanup`), the harness can validate them per-file.

## Variant shape

```rust
ValidatorSpec::PerFileNonSilent {
    glob: String,
    #[serde(default = "default_non_silent_ratio")]
    min_ratio: f32,
    #[serde(default)]
    require_at_least: usize,
}
```

`min_ratio` is `f32` (not `f64`) to match the existing `AudioNonSilent` variant and reuse `decode_non_silent_ratio` without lossy conversion. `require_at_least = 0` (the serde default) makes the variant a pure per-file gate; setting it to `>= 1` also enforces a match-count floor.

## Tests added (10)

- `validators::tests::per_file_non_silent_passes_when_all_segments_are_loud`
- `validators::tests::per_file_non_silent_fails_when_one_segment_is_silent` (asserts the offending filename + ratio + threshold surface in the failure message)
- `validators::tests::per_file_non_silent_fails_when_match_count_below_require_at_least` ("expected >=1, found 0")
- `validators::tests::per_file_non_silent_passes_when_require_at_least_zero_and_no_matches` (optional-intermediate escape hatch)
- `validators::tests::per_file_non_silent_glob_interpolates_args_key` (`${args.X}` interpolation works)
- `validators::tests::per_file_non_silent_rejects_path_traversal_arg_value` (`../etc` arg value -> Error outcome)
- `validators::tests::per_file_non_silent_glob_excludes_placeholder_filenames` (mofa-podcast pause/BGM placeholders correctly excluded)
- `workspace_policy::tests::per_file_non_silent_roundtrips_through_toml`
- `workspace_policy::tests::per_file_non_silent_defaults_require_at_least_and_min_ratio_when_omitted`
- `workspace_contract::tests::podcast_contract_satisfies_when_all_segments_and_final_mix_are_non_silent` + `..._fails_when_a_single_segment_is_silent_even_if_final_mix_is_loud` (combined-gate e2e via `enforce_spawn_task_contract_with_args`; confirms both `audio_non_silent` and `per_file_non_silent` outcomes persist to the validator ledger when both fire)

Plus the existing `session_policy_declares_new_domain_validators_for_silent_failure_paths` was strengthened to assert the segments glob, min_ratio floor, and `require_at_least >= 1` on `podcast_generate`.

Test delta in `octos-agent --lib`: 1425 -> 1435 (+10).

## Test plan

- [x] `cargo clippy -p octos-agent --all-targets -- -D warnings`
- [x] `cargo test -p octos-agent --lib` (1435 / 1435 pass)
- [x] `cargo test -p octos-agent --tests` (15 / 15 pass)
- [x] `cargo check --workspace --all-targets`

### DIFF
--- a/crates/octos-agent/src/validators.rs
+++ b/crates/octos-agent/src/validators.rs
@@ -522,6 +522,19 @@ impl ValidatorRunner {
                 ValidatorSpec::AudioNonSilent { glob, min_ratio } => self.run_audio_non_silent(
                     invocation, validator, glob, *min_ratio, started_at, started,
                 ),
+                ValidatorSpec::PerFileNonSilent {
+                    glob,
+                    min_ratio,
+                    require_at_least,
+                } => self.run_per_file_non_silent(
+                    invocation,
+                    validator,
+                    glob,
+                    *min_ratio,
+                    *require_at_least,
+                    started_at,
+                    started,
+                ),
                 ValidatorSpec::MagicBytes { glob, format } => {
                     self.run_magic_bytes(invocation, validator, glob, *format, started_at, started)
                 }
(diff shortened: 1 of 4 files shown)


ymote:
Codex review outcome: 1 finding, P2 only, no P0/P1 blocks.

**P2 — Stale-segment scope caveat** (`workspace_policy.rs:783-786`)

The per-file `**/segments/seg_*.wav` glob does not apply the `task_started_at` look-back filter that `resolve_artifacts` uses. In a session workspace that preserves segment directories across concurrent invocations, a stale segment could affect the current run's gate.

**Decision: defer to a follow-up.** The pre-existing whole-file `AudioNonSilent` validator has the identical look-back gap (it's a shared harness invariant — the validator runner doesn't currently see `task_started_at`). mofa-skills #59 clears `<output_dir>/segments/` at the start of every invocation, so the practical risk is bounded. Closing the gap properly means plumbing `task_started_at` into the validator runner so it can mtime-filter matches — that's worth one focused PR that fixes both `AudioNonSilent` and `PerFileNonSilent` together rather than only patching the new variant.

Annotated the trade-off inline in `workspace_policy.rs` so future readers see the link to the open follow-up.

PR #1017, opened 2026-05-17
fix: add artifact sources for mofa card/comic/infographic workspace contracts

BH3GEI:
## Summary
- Register `card_png`, `comic_png`, `infographic_png` artifact sources in workspace policy so `files_to_send` from mofa spawn_only plugins can pass contract validation and appear in SESSION FILES

## Test plan
- [x] mofa_cards plugin completes without "no artifact source" error
- [x] Generated PNG delivered to SESSION FILES panel


ymote:
Superseded by the newer workspace contract work already in main. Current main declares an image_png artifact source and binds mofa_comic/mofa_infographic to spawn-only PNG outputs; mofa_cards is intentionally kept on the glob contract until the plugin emits files_to_send, guarded by the octos #1041 regression test. Replaying this old PR would weaken that current contract shape.

PR #1049, opened 2026-05-18
fix(workspace_policy): drop PerFileNonSilent from podcast contract

ymote:
## Summary
- Drop `PerFileNonSilent { glob: "**/segments/seg_*.wav", require_at_least: 1 }` from the `podcast_generate` workspace contract
- Whole-file `AudioNonSilent` (SpawnOnlyFiles, mp3) still guards the deliverable
- The `PerFileNonSilent` validator type stays — only the podcast wiring is removed

## Why
Live mini3 regression 2026-05-18: every `podcast_generate` call hard-failed with
> ✗ Podcast generation failed: required validator failure: podcast_generate.on_completion[2]: per_file_non_silent: expected >=1 audio files, found 0

Root cause: the deployed mofa-podcast plugin does not preserve `seg_*.wav` segment files after assembly. The glob matched zero files everywhere on disk; the `require_at_least: 1` floor made every podcast call fail. Whole-file `AudioNonSilent` already catches silent final mixes — the per-segment safety net was wired against a plugin contract that no longer holds.

Re-arm only when the plugin preserves segment scratch files (or carries them in `files_to_send`).

## Test plan
- [x] `cargo test -p octos-agent --lib workspace_` → 122 passed, 0 failed
- [ ] Fleet deploy mini1/2/3/5
- [ ] Live podcast_generate run on mini3 with the original failing prompt


ymote:

PR #1051, opened 2026-05-18
fix(slides): default auto_layout to false in slides system prompt

ymote:
## Summary
- Add a `RULES` line to the slides system prompt directing the LLM to pass `auto_layout: false` on every `mofa_slides` call.
- Per-slide `auto_layout: true` opt-in remains, triggered only when the user explicitly asks.

## Why
The plugin's manifest declares `auto_layout` with no `default` value. When the LLM omits the arg, the plugin's silent fallback is "editable text mode" — flat-color text boxes that replace the generated image content. Confirmed today on mini3: an oolong-tea Lingnan deck rendered with editable text boxes instead of the requested image-driven layout, even though the LLM never asked for autolayout.

The durable fix belongs in the plugin (declare `default: false` in the schema + code path), but that requires the external mofa-skills repo. This server-side prompt nudge unblocks production immediately and is harmless to keep around after the plugin is fixed.

## Test plan
- [x] `cargo test -p octos-cli --features api --lib slides_prompt_directs` → 1 passed


ymote:

PR #1052, opened 2026-05-18
fix(dashboard): user-typed exact-8-char admin password (no auto-generate)

ymote:
## Summary
- Tighten the Rotate-Admin-Token wizard from "8 or more" → **exactly 8** characters.
- Live counter `(N/8)` in the inline error so the operator can see why submit stays disabled.
- Updated placeholder + help text.

## Why
Operator preference: pick a memorable short password rather than copy-paste a 32-char auto-generated string.

## Security
Admin token is a bearer credential (SHA-256+salt, constant-time compare in `admin_token_store.rs:24-36`), not a KEK. Server-side floor (`MIN_ROTATED_TOKEN_LEN = 8` in `admin_setup.rs:177`) is unchanged — this client change just matches the existing server floor exactly. No weakening of an existing audit threshold.

## Test plan
- [x] Updated test asserts submit disabled at length 7, enabled at exactly 8, disabled again at length 9
- [ ] Manual: visit `/setup/rotate-token`, confirm counter, submit `pass1234`, log back in

### DIFF
--- a/dashboard/src/pages/SetupRotateToken.test.tsx
+++ b/dashboard/src/pages/SetupRotateToken.test.tsx
@@ -2,9 +2,11 @@
 // bootstrap admin token with a persistent hashed record.
 //
 // UX moved from "generate a 32-char random secret + email-receipt" to
-// "operator types their own password (>= 8 chars)". The auto-generate /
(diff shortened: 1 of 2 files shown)


ymote:

PR #1058, opened 2026-05-18
fix(pipeline+slides): scope run_pipeline to deep_research, slides default to inline array

ymote:
## Summary
- `RunPipelineTool::description()` rewritten to advertise *named templates only* (`deep_research`). JSON schema gains `"enum": ["deep_research"]`.
- Slides system prompt flips from `input=script.js` (requires Node.js, only mini1 has it) to inline `slides` array (universal), and explicitly bans `run_pipeline` from slides sessions.

## Why
Live mini3 regression 2026-05-18 on `/slides/slides-1779130130502-th18yr`: user asked "first 3 slides only", LLM composed a custom `digraph` to fan out to Gemini per-slide. Two prompt drifts compounded:

1. Tool description in EVERY session said "ALWAYS write inline DOT graphs — do NOT use pre-built pipeline names" — the LLM was specifically told to compose ad-hoc pipelines.
2. Slides session prompt said "ALWAYS use input=script.js… NEVER pass slides inline" — but the `input=` codepath requires `node`, and `node` is only present on mini1 (deploy gap, separate issue).

Together, the LLM had no working path: `mofa_slides(input=…)` fails for lack of node; the tool-spec encouraged composing a `run_pipeline` workaround instead.

## Test plan
- [x] `cargo test -p octos-cli --features api --lib project_templates` → 13/13 passed
- [x] `cargo check -p octos-pipeline` clean
- [ ] Fleet deploy + retest slides partial-regen on mini3

### DIFF
--- a/crates/octos-cli/src/project_templates.rs
+++ b/crates/octos-cli/src/project_templates.rs
@@ -149,11 +149,10 @@ WORKFLOW (follow in order):
 RULES:
 - ALWAYS use mofa_slides TOOL. NEVER shell to run mofa. NEVER.
 - In slides sessions, `mofa_slides` is already active. Call it directly. Do not call `activate_tools(["mofa_slides"])`.
-- BEFORE calling mofa_slides: run shell("node --check slides/{slug}/script.js") to validate syntax. Fix any errors before proceeding.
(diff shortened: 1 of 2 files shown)


ymote:

PR #1059, opened 2026-05-18
fix(installer): surface install_pkg failures + detect missing brew

ymote:
## Summary
- `install_pkg` on macOS now detects missing Homebrew **before** running an `eval "brew …"` that's guaranteed to fail, and prints the canonical Homebrew bootstrap one-liner.
- `install_pkg` captures stderr to a tempfile and surfaces the last 5 lines on failure. The legacy `2>&1` redirect hid every real error.

## Why
Three production minis (mini2/3/5) silently shipped without Node.js, ffmpeg, and chromium because:
1. `eval "brew install node" >/dev/null 2>&1` returned "command not found: brew" (Homebrew wasn't installed) — and that error was silenced.
2. Installer continued past the failure with `|| true`, marking node "optional".
3. Operator saw only `warn "node install failed"` with no hint why → assumed it was a known limitation.

mofa-slides' `input=script.js` codepath silently broke fleet-wide on those hosts.

## Test plan
- [x] `bash -n scripts/install.sh` passes (syntax check)
- [ ] Run on a host where brew is missing → verify clean message + bootstrap hint instead of silent failure
- [ ] Run on a host where brew is present but `brew install <pkg>` fails for network reasons → verify the captured stderr lines appear

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -293,6 +293,24 @@ pkg_hint() {
 # Install a package using the command from pkg_hint.
 # Usage: install_pkg <package>
 # Returns 0 on success, 1 on failure.
+#
+# Pre-2026-05-18: this helper silenced stderr (`eval "$cmd" >/dev/null 2>&1`)
+# and printed only "<pkg> install failed" on a non-zero exit. That hid the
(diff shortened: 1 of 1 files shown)


ymote:

PR #1071, opened 2026-05-19
feat(slides): prompt-file migration + strip assistant-role images

ymote:
Two coupled fixes for the slides session path (mini3 dspfac session slides-1779130130502-th18yr).

**1. Slides prompt → load_prompt file pattern.** Moved the 95-line slides system prompt out of `project_templates.rs` into `crates/octos-cli/src/prompts/slides_default.txt`. Same load_prompt(name, compiled_default) pattern used by gateway/admin/worker. Override at runtime via `~/.octos/prompts/slides.md` — no rebuild needed for per-host tuning. Content verbatim; behavior unchanged for vanilla deployments.

**2. Strip assistant-role images from LLM call.** `send_file(skill-output/slides/<slug>/output/slide-NN.png)` populates `assistant_msg.media` (session_actor.rs:652). Every subsequent LLM call re-encoded those PNGs as image_url / image content. Live mini3 kimi-k2.5 returned `400 InvalidParameter: incorrect modal image` at 18:48/18:52/21:19 PDT today. Vision-capable models accepted but paid ~1 MB per slide per call in wasted context. Fix: in openai.rs + anthropic.rs build_*_content, only inline media images when msg.role == User. Gemini was already correct (build_user_parts only invoked for User role).

## Test plan
- [x] cargo test -p octos-cli --lib project_templates → 14 pass
- [x] cargo test -p octos-llm --lib → 289 pass + new regression
- [ ] Deploy + verify kimi no longer 400s on slides session

PR #1074, opened 2026-05-19
fix(llm): strip skill-output paths from image content on any role

ymote:
## Background

Original fix landed directly on main as `f26c6c66c` (mistake — should have gone through PR). Reverted on main via `78189d910`. Re-submitting the same change through a proper review path.

## Summary

PR #1071 added a role-based filter that stripped images from Assistant/Tool messages. The live mini3 dspfac slides session `slides-1779130130502-th18yr` showed a second leak path: the SPA's slides editor surfaces generated slide PNGs into the **chat composer's attachment field**, so the path `skill-output/slides/<slug>/output/slide-NN.png` arrives as `Message.media` on a **User-role** message, bypassing the role check. kimi-k2.5 then 400s with `incorrect modal "image"` on the resulting payload.

## Fix

Add `vision::is_tool_output_path` — substring match for `/skill-output/` (or prefix `skill-output/`). Wire it into `openai.rs` and `anthropic.rs` alongside the existing role check. Filter applies regardless of role:

- User-role message with `skill-output/X.png` in media → stripped
- Assistant-role with image media (any path) → stripped (role rule)
- User-role with a regular `/tmp/upload.png` → still inlined as image_url

This is the workspace convention enforcement: tool outputs never re-enter the LLM context as vision input. The model can still `read_file` the image bytes if it explicitly needs to inspect them.

## Test plan

- [x] `cargo test -p octos-llm --lib` → 290 pass (289 existing + new `test_build_content_strips_tool_output_image_on_user_message`)
- [ ] Codex review (this PR)
- [ ] Build + deploy to mini1/2/3/5
- [ ] Live verify: send a user message with `skill-output/...` PNG to dspfac slides session; confirm kimi turn no longer 400s and the LLM does not see the image as `image_url`

## Files

- `crates/octos-llm/src/vision.rs` — new `is_tool_output_path`
- `crates/octos-llm/src/openai.rs` — wire path filter + new regression test
- `crates/octos-llm/src/anthropic.rs` — wire path filter

### DIFF
--- a/crates/octos-llm/src/anthropic.rs
+++ b/crates/octos-llm/src/anthropic.rs
@@ -306,14 +306,19 @@ struct AnthropicImageSource {
 }
 
 fn build_anthropic_content(msg: &Message) -> AnthropicContent {
-    // Mirror openai.rs: only inline vision content on USER messages.
-    // Assistant/Tool media is prior-turn tool output (e.g.
-    // send_file(skill-output/slides/<slug>/output/slide-NN.png)) and
-    // should never be re-fed as image input on subsequent turns.
+    // Mirror openai.rs:
+    //   - Assistant/Tool media: never inline as image content (prior
+    //     turn's tool output, e.g. send_file(slide.png)).
+    //   - Paths under `skill-output/`: never inline regardless of role,
+    //     since the SPA can surface generated artifacts into the user
+    //     composer's attachment field (live mini3 dspfac reproducer).
     let images: Vec<_> = if msg.role != MessageRole::User {
         vec![]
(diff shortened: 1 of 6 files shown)


ymote:
Closing in favor of PR #1077 (prompt-only enforcement). The path-substring filter in this PR was hardcoded server-side behavior — it would override LLM judgment unconditionally for any path under `/skill-output/`, including legitimate cases where a user genuinely wants the LLM to look at a generated artifact. Better to change LLM behavior via the prompt and trust the model to comply.

PR #1079, opened 2026-05-19
fix(server): bridge system-prompt stacking + spawn child send_file + per-session prompt on AppUI

ymote:
## Summary

Four regressions traced to commit `28552bb9d "feat(appui): add M14-M18 stdio parity soaks (#1060)"`. Live reproducer: mini3 dspfac slides session `slides-1779207239761-u8pt1j` hit all four simultaneously, causing the LLM to call `mofa_slides` repeatedly without successful delivery and 466 KB stacked system prompts after 4 turns.

Three independent diagnostic agents converged on the same commit.

### 1. Stop stacking the runtime System prompt every turn
**File**: `crates/octos-cli/src/api/ui_protocol.rs` (AppUI bridge), `crates/octos-cli/src/session_actor.rs` (gateway bridge)

- `record_prompt_messages_not_covered_by_context` now skips `MessageRole::System`. The contiguous-window match in the previous code couldn't recognize the System slot, so it flagged `messages[0]` as "uncovered" and recorded it as a fresh `SystemInstruction` item every turn. After N turns the manager held N copies; `for_prompt` re-emitted them all; `normalize_system_messages` concatenated them into a 4×-bloated `messages[0]`.
- After `*messages = frame.messages`, re-insert the caller's runtime System at index 0 if the frame didn't already lead with one. The runtime System (composed by `compose_system_prompt()` per turn) is the agent's responsibility, not the manager's.

### 2. Wire SendFileTool into the spawn-child registry (AppUI)
**File**: `crates/octos-cli/src/api/ui_protocol.rs`

`SpawnTool::with_context(...)` on the AppUI path got `with_child_prompt_context_manager_factory` but no `with_child_tool_factory` for SendFileTool, so the spawned child registry had `with_builtins + plugins + pipeline_factory` and no `send_file`. Workspace-contract subagents declaring `send_file` in `allowed_tools` (mofa_slides post-completion delivery) tripped the preflight at `spawn.rs:1344`.

Lift `out_tx` / `send_file_base` / `send_file_extras` / `send_file_context_session` above the SpawnTool builder so the same deps feed both the parent SendFileTool AND a new `with_child_tool_factory` closure for children.

### 3. Wire `read_session_prompt` on the AppUI turn path
**File**: `crates/octos-cli/src/api/ui_protocol.rs`

Gateway already reads `data_dir/session_prompts/<topic>.md` for `/new slides X` / `/new site X` sessions. The AppUI/WS turn path called `session_runtime.agent.system_prompt_snapshot()` unconditionally — so slides/site sessions on the web UI fell back to the generic agent prompt and lost their workflow guidance. Use `session_id.topic()` + `project_templates::read_session_prompt` with fall-through to the agent snapshot.

## Test plan
- [x] `cargo test -p octos-cli --lib project_templates` → 14 pass
- [x] `cargo test -p octos-agent --lib` → 1469 pass (spawn, contract validators, registry)
- [x] `cargo check --workspace` clean
- [ ] Codex review
- [ ] Build + deploy + verify on mini3 dspfac slides session

## Files changed
- `crates/octos-cli/src/api/ui_protocol.rs`
- `crates/octos-cli/src/session_actor.rs`

## Diagnostic agents
Three parallel agents independently bisected — all converged on commit `28552bb9d`. Findings synthesized into surgical fixes that preserve the new ContextManager bridge architecture (per request: don't disable, fix it).

### DIFF
--- a/crates/octos-cli/src/api/context_manager.rs
+++ b/crates/octos-cli/src/api/context_manager.rs
@@ -770,6 +770,24 @@ impl ContextManager {
         message: &Message,
         source_ref: Option<TranscriptSourceRef>,
     ) -> Vec<TranscriptItemId> {
+        // The agent's runtime System prompt is composed per turn via
+        // `compose_system_prompt()` and placed at `messages[0]` by the
+        // agent loop; it is NOT a piece of session state the
+        // ContextManager should own. Recording it here makes every
+        // entry point (bridge per-turn recording, persisted-message
+        // merge, boot replay via from_session_history, fork) stack a
+        // `SystemInstruction` item across turns / restarts. `for_prompt`
+        // then re-emits them all and `normalize_system_messages`
+        // concatenates them into a 4×-bloated `messages[0]`.
+        //
+        // The legitimate use of SystemInstruction items (compaction
+        // summaries) flows through `compact_context`, not through this
+        // recording API. Skipping System here is therefore safe across
+        // all callers: bridge recording, persisted-message merging,
+        // boot replay, and snapshot reconstruction.
+        if message.role == MessageRole::System {
+            return Vec::new();
+        }
         let mut ids = Vec::new();
         match message.role {
             MessageRole::System => ids.push(self.record_item_with_source_ref(
--- a/crates/octos-cli/src/api/context_manager.rs
+++ b/crates/octos-cli/src/api/context_manager.rs
@@ -545,8 +545,17 @@ impl ContextManager {
         thread_id: Option<String>,
         fork: ForkedChildContext,
     ) -> Self {
-        let next_item_seq = fork
+        // Drop any `SystemInstruction` items inherited from a polluted
+        // parent context. They are no longer owned by the manager (see
+        // `record_message_with_source_ref` early-return) and forking
+        // them into the child would re-stack the LLM prompt as soon as
+        // the child's `for_prompt` runs.
+        let items: Vec<_> = fork
             .items
+            .into_iter()
+            .filter(|item| !matches!(item.kind, TranscriptItemKind::SystemInstruction { .. }))
+            .collect();
+        let next_item_seq = items
             .iter()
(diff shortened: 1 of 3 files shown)


ymote:
Closing as mid-stale — opened 2026-05-19/20, no merge in the active 2026-05-21/22 work stream. If the underlying intent is still relevant, please re-open against current main (rebase will likely conflict given the recent #1155-#1166 + #1148/#1163-#1165 churn) or file a fresh issue. Audited per 2026-05-22 PR sweep.

ymote:
Reopening — codex review will run before any close decision (per user direction).

ymote:
## Codex audit verdict — KEEP (one P2 follow-up)

Ran `codex exec review --base main` against head SHA `e2b27524`.

**Codex summary:** _"The patch fixes prompt stacking but introduces a regression where non-runtime System messages are silently dropped in bridged task runs. That can remove episodic-memory/task instruction context from spawned agents."_

**P2 finding:** `crates/octos-cli/src/api/context_manager.rs:810-812` (the new System-role early-return in `record_message_with_source_ref`) — When a spawned task has relevant past experiences, `Agent::build_initial_messages` appends them as an additional `MessageRole::System` after the runtime prompt, and `prepare_task_messages` does not normalize those into `messages[0]`. Because this guard drops every System message while the bridge only re-inserts the cached first one, bridged child tasks lose that memory/instruction block before the LLM call. Exempt only the runtime prompt or merge secondary system messages into the cached system instead of discarding all System messages.

**Recommendation:** Keep open. The system-prompt stacking fix is the correct direction (and the bridge cache + re-merge logic in `record_prompt_messages_not_covered_by_context` and `AppUiPromptContextBridge` is well-reasoned), but the early-return is too broad. Fix before merge so episodic-memory System blocks aren't silently dropped on child task runs.

(Audit batch 2026-05-21; no auto-close.)

PR #1096, opened 2026-05-20
fix(api): list_sessions reads per-profile sessions/ dir

ymote:
## Symptom

`/chat` sidebar on `https://dspfac.octos.ominix.io/chat` shows no chat history. New chats land in the right pane and persist on disk, but they never appear in the left-rail listing.

## Root cause

The `session/list` WebSocket handler at `crates/octos-cli/src/api/handlers.rs:413` (`list_sessions`) only walked the process-wide `state.sessions` store, which opens against the top-level `<home>/sessions/` directory (per `commands/serve.rs:323-324`). M11-F WS turn handlers (`run_standalone_turn`) persist chat history under `SessionRuntime.sessions`, which opens against the per-profile `<home>/profiles/<X>/data/sessions/` directory. Those JSONLs were invisible to the list handler.

Post-PR #1002 ("make authenticated identity authoritative", ~2026-05-12), new web chats route through the per-profile session manager — top-level `<home>/sessions/` stopped getting new entries on profile-routed traffic. The sidebar has been silently empty for tenant subdomains since.

## Why admin auth masked it for a week

Admin requests resolve through a different code path — the process-wide store opened against the top-level home — which still receives admin-scoped writes. The previous test harness only used admin auth, so the bug never surfaced in CI. Per memory `feedback_admin_token_masks_bugs`, the new tests include a non-admin profile-scoped probe.

## The fix

Mirror the per-profile `SessionManager` pattern from commit `10cc9378d` ("fix(api): route session_messages through per-profile SessionManager") into `list_sessions`. The earlier commit added `read_profile_session_messages` for the `/messages` endpoint but left `/list` on the legacy walk — this PR closes the same gap there.

* New `list_profile_sessions(profile_data_dir)` helper opens a transient `octos_bus::SessionManager` against the resolved per-profile data dir and walks `list_top_level_sessions_with_title()`. On-disk ids are emitted as-is (no `<profile>:api:` prefix strip) because the per-profile sessions/ directory is itself the tenant boundary.
* `list_sessions` invokes the helper after the auth gate, before the legacy `state.sessions` walk; dedupes by id (per-profile wins). The process-wide walk + gateway proxy block are preserved as admin / legacy fallbacks.
* Internal `child-*` / `*.tasks` topics stay filtered (already skipped at the `list_top_level_sessions` directory walk; `is_internal_api_session_id` is the belt-and-suspenders check for legacy flat-layout entries).

## Slides editor isn't affected

The slides editor doesn't use the chat sidebar — it loads a specific session by id through `session/messages`, which was already fixed by commit `10cc9378d`. Only the `/chat` left rail (and any client that calls `session/list`) was broken.

## Tests

TDD per project convention; `#[cfg(test)]` inline in `handlers.rs`:

* `list_profile_sessions_returns_bare_web_session_history` — seeds `web-101`, `web-202`, and `web-303#child-task-1` under a synthetic profile data_dir, asserts the two top-level chats surface and the child fanout is filtered.
* `list_profile_sessions_returns_empty_when_dir_missing` — graceful degradation when the profile sessions/ dir is missing.
* `list_profile_sessions_handles_non_admin_profile_scope` — mandatory non-admin probe per `feedback_admin_token_masks_bugs`. Confirms the helper surfaces bare-key history without consulting any admin-only code path.
* `list_sessions_admin_legacy_path_still_emits_session` — confirms the new per-profile branch did not regress the admin / legacy fallback (process-wide `state.sessions` walk still emits the `web-legacy-1` profiled key).

## Test plan

- [x] `cargo check -p octos-cli --features api`
- [x] `cargo test -p octos-cli --features api --lib list_profile_sessions` — 3/3 pass
- [x] `cargo test -p octos-cli --features api --lib list_sessions_admin_legacy` — 1/1 pass
- [x] `cargo test -p octos-cli --features api --lib api::handlers` — 62/62 pass
- [ ] Verify live on `https://dspfac.octos.ominix.io/chat` after deploy: hard-refresh shows past chats in left rail.

### DIFF
--- a/crates/octos-cli/src/api/handlers.rs
+++ b/crates/octos-cli/src/api/handlers.rs
@@ -428,6 +428,37 @@ pub async fn list_sessions(
         Err(response) => return response,
     };
 
+    // M11-F per-profile SessionManager listing. Mirrors the
+    // `session_messages` fix in commit 10cc9378d (`fix(api): route
+    // session_messages through per-profile SessionManager`): the
+    // WS turn handler (`run_standalone_turn`) writes session history
+    // through `SessionRuntime.sessions` which opens against the
+    // profile's own `data_dir` (`<profile>/data/sessions/...`). The
+    // process-wide `state.sessions` walk below opens against the
+    // top-level home (`<home>/sessions/`) and never sees those
+    // per-profile JSONLs — so for an authenticated tenant request
+    // (e.g. dspfac listing its own chat history on
+    // `https://dspfac.octos.ominix.io/chat`), the legacy walk
+    // returns 0 sessions and the chat sidebar renders empty.
+    //
+    // The fix opens a transient `SessionManager` against the
+    // resolved profile_data_dir and walks the bare on-disk ids the
+    // SPA actually uses — no prefix strip, because the per-profile
+    // sessions/ directory is itself the tenant boundary, so on-disk
+    // ids are already chat-bare. The process-wide walk below remains
+    // as an admin / legacy fallback (still strips `<profile>:api:`
+    // for legacy entries).
+    if let Ok(profile_data_dir) = resolve_profile_data_dir(&state, &headers, identity_ref).await {
+        let profile_sessions = list_profile_sessions(&profile_data_dir);
+        let existing: std::collections::HashSet<String> =
+            all.iter().map(|s| s.id.clone()).collect();
+        all.extend(
+            profile_sessions
+                .into_iter()
+                .filter(|s| !existing.contains(&s.id)),
+        );
+    }
+
     if let Some(sessions) = &state.sessions {
         let sess = sessions.lock().await;
         let prefix = format!("{profile_id}:api:");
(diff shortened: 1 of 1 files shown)


ymote:
Closing as mid-stale — opened 2026-05-19/20, no merge in the active 2026-05-21/22 work stream. If the underlying intent is still relevant, please re-open against current main (rebase will likely conflict given the recent #1155-#1166 + #1148/#1163-#1165 churn) or file a fresh issue. Audited per 2026-05-22 PR sweep.

ymote:
Reopening — codex review will run before any close decision (per user direction).

ymote:
## Codex audit verdict — KEEP (one P2 follow-up)

Ran `codex exec review --base main` against head SHA `cf195715`.

**Codex summary:** _"The new per-profile listing reads the right directory, but it bypasses the API-session normalization/filtering used by the other listing paths. That can change the `session/list` contract and surface malformed or non-API sessions."_

**P2 finding:** `crates/octos-cli/src/api/handlers.rs:555-560` (the new `list_profile_sessions` helper) — When `profile_data_dir` contains gateway-written API sessions such as `api:web-1` or `<profile>:api:web-1` (or other channel keys like `telegram:123` in the same `SessionManager`), this new branch returns the raw key verbatim. The existing legacy and gateway listing paths strip API prefixes, dedupe by chat id, and drop non-API channels, so this can add duplicate/malformed sidebar entries and expose other channel titles/counts through `session/list`. Apply the same API chat-id normalization/filter before constructing `SessionInfo`.

**Recommendation:** Keep open. The dspfac/per-profile session listing fix is correctly targeted (mirrors the `session_messages` fix in commit `10cc9378d`) and is mergeable cleanly against current main. The P2 finding should be addressed before merge to avoid sidebar leakage of non-web channels.

(Audit batch 2026-05-21; no auto-close.)

PR #1116, opened 2026-05-20
fix(install): restrict on-demand TLS to configured domain (codex P1 follow-up to #380 / #1070)

ymote:
## Summary

Codex review of #380 and #1070 flagged the same P1 finding: the catch-all `https://` Caddy site in `scripts/install.sh` enabled on-demand TLS for ANY hostname pointed at the server. The `@sub host *.${CADDY_DOMAIN}` request matcher was inside the site block, so it only restricted *handlers* — not the TLS automation policy. Combined with the `/check` ask endpoint unconditionally returning 200, a third party who pointed any DNS at the host could trigger Caddy to attempt ACME issuance for arbitrary names and burn rate limits.

## Fix

Scope the on-demand TLS site to `https://*.${CADDY_DOMAIN}` so Caddy's site-address itself bounds which connections trigger on-demand TLS. The apex domain (`${CADDY_DOMAIN}`) is already handled by the explicit site block above, and the `@octos` host matcher in the `http://` block still covers HTTP-to-HTTPS redirects for both apex and wildcard hosts. The inner `@sub` matcher (now redundant) is removed; per-path handlers are flattened one level.

## Codex references

- PR #380 comment: `scripts/install.sh:1786-1788` — "Restrict on-demand TLS to the configured domain"
- PR #1070 comment: `scripts/install.sh:1786-1789` — "Restrict on-demand TLS to Octos hosts"

## Test plan

- [x] `bash -n scripts/install.sh` passes
- [ ] CI: `check`, `typos`, `dashboard`
- [ ] codex review re-run shows no new P1/P2

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -1379,30 +1379,26 @@ $Domain {
     }
 }
 
-https:// {
+https://*.$Domain {
     tls {
         on_demand
     }
 
-    @sub host *.$Domain
+    @api path /api/*
+    @admin path /admin*
+    @auth path /auth/*
 
-    handle @sub {
-        @api path /api/*
-        @admin path /admin*
-        @auth path /auth/*
-
-        handle @api {
-            reverse_proxy $caddyUpstream
-        }
-        handle @admin {
-            reverse_proxy $caddyUpstream
-        }
-        handle @auth {
-            reverse_proxy $caddyUpstream
-        }
-        handle {
-            reverse_proxy $caddyUpstream
-        }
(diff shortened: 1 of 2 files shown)


ymote:
Codex review on `56dd6b1` returned one P2: TLS gate hostname suffix check accepts deeper names like `a.b.${CADDY_DOMAIN}` and could consume ACME rate limit for non-routed names. Tracked as follow-up #1124 — the core scope (M14 TLS prep) ships clean.

Issue #1019, opened 2026-05-17
M17-A: Wire gateway/session_actor spawned children to ContextManager forks


ymote:
## Milestone
M17 Agent Context Orchestration Hardening

## Problem
AppUI top-level turns and AppUI builtin spawn children now have ContextManager coverage, but gateway/session_actor spawned children can still bypass the parent-owned context fork path.

## Scope
- Find every gateway/session_actor child-agent creation path.
- Require each path to receive or derive a parent PromptContextManager/ContextManager fork.
- Persist child context ledger/checkpoint metadata with the same session/profile/cwd/runtime policy stamp as AppUI turns.
- Fail closed or mark an explicit unmanaged state when a child context cannot be created.

## Acceptance Criteria
- A session_actor child spawn writes a child context ledger.
- Logs include the child context generation/checkpoint id before the model call.
- Parent context receives a bounded child capsule, not raw child transcript/tool chatter.
- Regression test covers session_actor/gateway spawn parity with AppUI spawn.

## Related
- #992 production orchestration tracker
- #987 subagent context capsules
- #989 production prompt compaction

ymote:
## Audit (2026-05-20)

**Status:** AppUI wired, gateway/session_actor NOT. Keeping open.

### Shipped
- AppUI SpawnTool wires `with_child_prompt_context_manager_factory` and calls `parent.fork_child_history(...)` + `ContextManager::from_forked_child_context(...)` at `ui_protocol.rs:13741-13791`.

### Gap
- [ ] Equivalent wiring is ABSENT in `session_actor.rs`.
- [ ] `session_actor.rs:2570-2613` constructs SpawnTool with no `with_child_prompt_context_manager_factory` call.
- [ ] `fork_child_history` and `from_forked_child_context` only used in AppUI (`ui_protocol.rs`) and tests.
- [ ] No regression test for session_actor / gateway spawn parity.

ymote:
Closing — wiring landed via PR #1120 (session_actor SpawnTool ContextManager fork). Live evidence: m15 native review stdio soak (2026-05-20) — 5 children (3 native + 1 CLI + 1 MCP) all `status: completed`, real DeepSeek backend (`model: deepseek-chat`), 36 AppUI notifications, all 5 artifacts published, secret-scan clean. Evidence at `e2e/test-results-m15-native-review-start-stdio/20260521T031002Z/`.

[...]

PR #1120, opened 2026-05-20
feat(session): wire ContextManager fork into session_actor SpawnTool (#1019)

ymote:
## Summary
- Issue #1019: gateway/session_actor-spawned children were starting from an ad-hoc empty ContextManager, while AppUI already forks the parent's context at api/ui_protocol.rs:13741. This wires the same factory into session_actor.rs so both paths share the same fork sanitiser (fork_child_history + from_forked_child_context).
- Refactored the child-context construction into build_forked_child_context_for_session_actor so it can be unit-tested without standing up a full SpawnTool.
- Added two regression tests: one asserts the child generation, ForkBoundary descent, and snapshot persistence; the second pins the explicit-child_session_key override path.

## Test plan
- [x] cargo fmt --all -- --check
- [x] cargo clippy --workspace --all-targets -- -D warnings
- [x] cargo test -p octos-cli --features api --lib session_actor_child_context_factory (both new tests pass)
- [ ] CI: check, test-octos-cli, test-octos-agent (lib), test-octos-agent (integration)

### DIFF
--- a/crates/octos-cli/src/session_actor.rs
+++ b/crates/octos-cli/src/session_actor.rs
@@ -14,7 +14,8 @@ use std::time::{Duration, Instant};
 use metrics::counter;
 use octos_agent::compaction::CompactionRunner;
 use octos_agent::tools::spawn::{
-    ChildSessionFailureAction, ChildSessionLifecycleKind, ChildSessionLifecyclePayload,
+    ChildPromptContextRequest, ChildSessionFailureAction, ChildSessionLifecycleKind,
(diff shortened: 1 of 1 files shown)


ymote:
Codex review on `0af456e` returned one P2: child ContextManager forks inside the detached task after awaiting persistence, so a fast follow-up parent turn could leak post-spawn user messages into the child. Tracked as follow-up #1125 — the session_actor ContextManager fork (M17-A) ships clean.

PR #1121, opened 2026-05-20
fix(appui): tighten runtime_mode "local" + accept task_id on task/artifact/* (codex P1 follow-up to #1086 / #1094)

ymote:
## Summary

Two P1 issues codex flagged on this session's earlier merges, fixed together because they share one file (`crates/octos-cli/src/api/ui_protocol.rs`):

1. **#1086 runtime_mode override misclassified `"local"` as solo-relaxed.** UPCR-2026-018 lists `local`, `tenant`, and `cloud` as non-solo override values — `local` is the multi-profile-but-local deployment mode, NOT the single-tenant solo loopback. Before this fix a Local server that received `runtime_mode: "local"` would still let a request set `permission_profile=DangerFullAccess` + `approval_policy=never`. The gate now tightens for `local`/`tenant`/`cloud`; only `None` or `"solo"` keep the relaxed local gate.

2. **#1094 task/artifact aliases rejected spec-conforming `task_id`.** `task/artifact/list` and `task/artifact/read` were registered as canonical aliases but still required the legacy `agent_id` param shape — spec-conforming M13 clients (UPCR-2026-019) reach the artifact surface via `task/list`, which exposes `task_id`. The aliases now accept either `agent_id` or `task_id` (preferring `agent_id` when both are set for legacy compatibility) and resolve to the backing agent id before dispatch.

## Test plan

- [x] `cargo test -p octos-cli --features api --lib permission_profile_set_local_override_tightens_gate_on_local_server` — new test covering `local`/`cloud` tightening + unchanged-when-absent + explicit-`solo` baselines.
- [x] `cargo test -p octos-cli --features api --lib task_artifact_alias_accepts_task_id_and_agent_id` — new test covering task_id-only, agent_id-only, both-set (agent_id wins), and missing-both (INVALID_PARAMS) for both list + read aliases.
- [x] `cargo test -p octos-cli --features api --lib m15_raw_autonomy_rpc_dispatches_every_method_to_orchestrator` — existing dispatch test still passes.
- [x] `cargo test -p octos-cli --features api --lib permission_profile_handlers_are_server_owned_and_reject_danger_outside_local` — existing tenant/solo gating test still passes.
- [x] `cargo fmt --all`
- [x] `cargo clippy -p octos-cli --features api --lib --all-targets -- -D warnings` — same 71 pre-existing errors as origin/main, no new ones from this change.

### DIFF
--- a/crates/octos-cli/src/api/agent_orchestrator.rs
+++ b/crates/octos-cli/src/api/agent_orchestrator.rs
@@ -1493,7 +1493,40 @@ fn get_agent<'a>(
     state: &'a AutonomyRuntimeState,
     request: &AgentRequest,
 ) -> Result<&'a AutonomyAgentRecord, RpcError> {
-    let Some(agent) = state.agents.get(&request.agent_id) else {
+    // Codex P1 follow-up to #1121: spec-conforming M13 clients call
+    // `task/artifact/*` with `task_id` (the `TaskListEntry.id`), not
+    // `agent_id`. Task-backed records (native specialists, mirrored
+    // background tasks) carry the task id under `task_id` and the
+    // agent id can differ (`native-…` prefixes, sanitisations, etc.).
+    // First try direct agent_id lookup (legacy + agent-only records),
+    // then fall back to scanning by `task_id` so the alias actually
+    // resolves to the right agent record.
+    //
+    // Codex P1 re-review #4 on #1121: this fallback is shared by all
+    // agent-keyed endpoints — `agent/artifact/*`, `agent/status/read`,
+    // `agent/output/read`, and the legacy `agent_id` branch of
+    // `task/artifact/*`. Without the session_id gate, a same-profile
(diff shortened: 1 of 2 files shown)


ymote:
Codex re-review #5 on `0e796b9` returned LGTM: "No actionable regressions were found in the changed code. The task-id fallback now remains session-scoped, and the runtime_mode gate correctly fails closed for non-solo overrides."

Round summary across 4 amendments:
- rev1 → P1: task_id alias didn't resolve to agent record → fixed with task_id fallback in `get_agent`.
- rev2 → P2: blank runtime_mode treated as omitted → fixed with explicit `None|"solo"` allowlist.
- rev3 → P1: task_id artifact path didn't enforce session_id → params-layer require `session_id` when `task_id` is the only id.
- rev4 → P1: legacy `agent_id` field also accepts task UUIDs through the shared fallback → `get_agent` task-id fallback now itself requires `session_id`.
- rev5 → LGTM.

PR #1145, opened 2026-05-21
fix(goal): drain budget-exhaustion wrap-up via pending-queue sweep (#1141)

### LINKED ISSUE #1141: Drain budget-exhaustion wrap-up after AppUI goal accounting (#1140 codex P2)
Codex P2 follow-up from PR #1140 re-review.

**Finding:** When an AppUI goal turn exhausts `token_budget`, the post-turn `record_goal_turn` call in `run_standalone_turn` transitions the goal to `budget_limited` and enqueues the one-shot wrap-up continuation. For a goal-only AppUI session, future ticks are driven by `due_loop_targets` which only returns ACTIVE goals — so no later tick will call `drain_ready_continuations_for_session` for this now-`budget_limited` session, and the wrap-up remains pending indefinitely unless an unrelated loop targets the same session.

**Required fix:** Either:
- (a) Trigger or drain the wrap-up directly after the accounting (synchronous drain post-record_goal_turn for the same session); OR
- (b) Extend the AppUI scheduler to include sessions with pending continuations (not just active goals) in its tick target list.

Location: `crates/octos-cli/src/api/ui_protocol.rs:~14710` (after the post-turn `record_goal_turn` call) AND `due_loop_targets` in `agent_orchestrator.rs`.

**Test:** AppUI goal that exhausts budget on a real turn must observe the wrap-up continuation drain to a turn within ~30s without any unrelated loop firing.

ymote:
## Summary

- An AppUI goal that exhausts `token_budget` transitions to `budget_limited` and enqueues a one-shot wrap-up continuation via `record_goal_turn`. For a goal-only AppUI session, `due_loop_targets` only surfaced sessions with active loops or active goals, so the wrap-up sat pending forever — no later tick called `drain_ready_continuations_for_session` for that now-`budget_limited` session unless an unrelated loop targeted it.
- Fix (Option B): add a third sweep in `due_loop_targets` over the master continuation queue itself, collecting every `(session_id, profile_id)` pair with a pending continuation, deduped against the existing loop+goal targets, honouring `profile_filter` + `max_items`.
- Exposes the iteration via a new `MasterContinuationScheduler::pending_sessions()` helper.

## Test plan

- [x] `cargo test -p octos-cli --features api --lib goal` — 29 passed, including new `due_loop_targets_includes_pending_wrap_up_for_budget_limited_goal` and `due_loop_targets_pending_sweep_respects_profile_filter`.
- [x] Verified both new tests fail without the fix (sweep block commented out) and pass with it applied.
- [x] `cargo fmt --all -- --check` clean.
- [x] `cargo clippy -p octos-cli --features api` — no errors; warnings unchanged from baseline.
- [x] `cargo test -p octos-cli --features api --lib master_continuation` — all scheduler tests green (no regression from the new iterator).

Closes #1141.

### DIFF
--- a/crates/octos-cli/src/api/agent_orchestrator.rs
+++ b/crates/octos-cli/src/api/agent_orchestrator.rs
@@ -2643,15 +2643,49 @@ fn pending_continuation_is_schedulable(
             let Some(goal) = state.goals.get(&session_key) else {
                 return false;
             };
+            // #1145 codex P2 re-review: pre-#1131 wrap-up turns were
+            // queued as `GoalContinue` + `wrap_up_prompt` metadata,
+            // and the prompt renderer promotes that shape at render
+            // time (see `legacy_goal_continue_with_wrap_up_metadata_promotes_to_wrap_up`).
+            // After budget exhaustion the owning goal is
+            // `budget_limited`, so the active-only gate would
+            // strand legacy persisted wrap-ups indefinitely.
+            // Detect the legacy shape and let it through — the goal
+            // record just has to exist.
+            if item.metadata.contains_key("wrap_up_prompt") || item.metadata.contains_key("wrap_up")
+            {
+                return true;
+            }
+            // #1145 codex P2 re-review: when a user cleared the
+            // old goal and created a new one for the same
+            // `SessionKey`, the queued item still carries the old
+            // `goal_id` while `state.goals` now holds the new goal.
+            // Without a goal-id identity check the stale item would
+            // mark the session schedulable and could drain BEFORE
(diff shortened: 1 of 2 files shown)


ymote:
Codex re-review #4 on `1f2dc6057` flagged a remaining P2: the predicate gates discovery but the drain path bypasses it. Real architectural gap, but no longer fixable as a small inline patch. Tracked as follow-up #1150 (apply the filter at `drain_ready_for_session` time). The pending-queue sweep + goal-id identity + legacy wrap-up fixes landed here remain correct — the drain-side filter is the additional layer.

PR #1153, opened 2026-05-21
fix(agent): view_image rejects symlinked ancestors + real Windows no-follow (#1151)

### LINKED ISSUE #1151: view_image: reject symlinked ancestors + real Windows no-follow check (#1148 codex P2)
Codex P2 follow-up from PR #1148 re-review.

**Finding 1:** On Unix, `O_NOFOLLOW` only applies to the FINAL path component. `resolve_path_with_scope` is lexical, so a workspace path like `imgs/pic.png` where `imgs` is a symlink to an outside directory still passes scope resolution and `open()` follows the parent symlink — `view_image` can read outside the workspace.

**Finding 2:** In the Windows fallback, `OpenOptions::open` has already followed the symlink/reparse point by the time `File::metadata().is_symlink()` is checked. The check is silently a no-op.

**Required fix:**
- Unix: walk each ancestor with `symlink_metadata` before the open, refuse if any ancestor is a symlink.
- Windows: use pre-open `symlink_metadata` check, or open with `FILE_FLAG_OPEN_REPARSE_POINT` and then verify.

Location: `crates/octos-agent/src/tools/codex_tools.rs::read_image_header_no_follow`.

ymote:
## Summary

`read_image_header_no_follow` in `crates/octos-agent/src/tools/codex_tools.rs` had two symlink gaps:

1. **Unix:** `O_NOFOLLOW` only catches a symlink at the FINAL path component. `resolve_path_with_scope` returns a LEXICAL path, so a workspace shipping `imgs -> /outside/` passed scope resolution and the open followed the parent symlink — `view_image imgs/real.png` could read outside the workspace.
2. **Windows:** `OpenOptions::open` had already followed the symlink/reparse point by the time the post-open `file.metadata().is_symlink()` check ran. The check was silently a no-op.

## Fix

Pre-open ancestor walk via `std::fs::symlink_metadata`:

- Walks `resolved.ancestors()` and refuses any entry that is a symlink/reparse point.
- Stops at the configured workspace root (inclusive) so it never recurses into system roots.
- Unix `O_NOFOLLOW` retained on the final open as defense in depth.
- Windows now does its symlink rejection PRE-open (via `symlink_metadata`), which is the actual no-follow check.
- `NotFound` on a leaf is tolerated so a still-walking parent symlink is still caught.

## Test plan

- [x] `cargo test -p octos-agent --lib view_image` — 6 tests pass (incl. two new ones)
- [x] `cargo test -p octos-agent --lib` — 1503 / 1503 pass
- [x] `cargo clippy -p octos-agent --lib --tests` — clean
- [x] **Regression probe**: temporarily disabling the new ancestor walk makes `view_image_rejects_parent_symlink_directory` fail with `view_image must refuse a SYMLINKED PARENT directory; got success result: {"byte_length":12,"format":"png","mime_type":"image/png","path":"imgs/real.png"}` — confirms the test exercises the bug rather than a tangential code path.
- [x] Windows test (`view_image_rejects_leaf_symlink_pre_open_on_windows`) is `cfg(windows)`-gated and silently skips when the runner can't create symlinks (no Developer Mode / admin). It cannot be exercised on this macOS host but will run in Windows CI.

Closes #1151.


ymote:
codex review clean on rev3 (55d9d57a) — host-scope leaf-symlink guard restored, workspace ancestor walk unchanged, 8/8 view_image tests pass. Ready for admin merge.

Note: this PR's working file is still `tools/codex_tools.rs` because it was branched before #1169 (the rename to `coding_tools.rs`) merged. Squash-merge will resolve via git rename detection; if a conflict appears, the resolution is straightforward (apply this branch's diff to the renamed file).

Issue #1150, opened 2026-05-21
Filter stale goal continuations at drain time, not just at scheduling discovery (#1145 codex P2)


ymote:
Codex P2 follow-up from PR #1145 re-review.

**Finding:** The `pending_continuation_is_schedulable` predicate (#1145) gates which sessions `due_loop_targets` returns, but NOT what `drain_ready_continuations_for_session` pops. When a session is woken (correctly) by a fresh active goal continuation, the older stale wrap-up for that session can still drain first because the drain path pops by session/profile without re-checking the predicate.

**Required fix:** Apply `pending_continuation_is_schedulable` (or an equivalent filter) at the drain site in `MasterContinuationScheduler::drain_ready_for_session`. Either:
- (a) Pre-filter the queue snapshot before draining
- (b) Pass the predicate down as a closure

Location: `crates/octos-cli/src/api/master_continuation_scheduler.rs::drain_ready_for_session` + caller wiring in `InProcessAgentOrchestrator::drain_ready_continuations_for_session`.

PR #1161, opened 2026-05-22
fix(scheduler): refill drain when stale items consume max_items slot (#1160)

### LINKED ISSUE #1160: [P3] drain_ready_continuations: dropped stale items consume max_items slot (#1150/#1159 follow-up)
Codex integration audit of the 5-PR merge stack (#1155-#1159 against c4dbd0fc) flagged a P3 in the new `#1150` drain-time filter.

## Finding

**Location:** `crates/octos-cli/src/api/agent_orchestrator.rs:954-959` (current head)

**Behavior:** `drain_ready_continuations_for_session` calls `state.continuations.drain_ready_for_session(.., max_items, ..)` and THEN filters the returned vector through `pending_continuation_is_schedulable`. Items that fail the predicate are dropped (correct, #1150) but they've already consumed a slot in the underlying scheduler's `max_items` budget.

**Impact:** Production callers pass `max_items = 1` (see the AppUI tick loop in `crates/octos-cli/src/api/ui_protocol.rs`). In the stale+fresh same-session pair scenario:

1. Stale wrap-up + fresh active continuation both queued for the same session.
2. Caller invokes `drain_ready_continuations_for_session(.., max_items=1, ..)`.
3. Scheduler pops the stale item (lower sequence under FIFO).
4. Predicate drops it.
5. `kept` is empty; the fresh continuation stays queued.
6. The session wakes up again next AppUI tick (~30s later).

**Codex's recommendation:** Drain until `kept.len() == max_items` (re-loop on drop) OR filter before counting the item against the limit.

## Acceptance

- Production stale+fresh pair drains the fresh item in a single call when `max_items=1`.
- Regression test: enqueue stale wrap-up + fresh continuation, drain with `max_items=1`, assert `drained.len() == 1` and the returned item is the fresh one.
- Existing `drain_ready_continuations_filters_stale_at_drain_site_per_1150` still passes (it uses `max_items=usize::MAX`).

## Severity

P3 — affects production latency (one tick worth, ~30s) but not correctness. Stale items are still dropped + tombstoned per #1159's logic.

## Context

- Found by `codex review --base c4dbd0fc` on 2026-05-21
- Original drain-time filter: #1159 (closes #1150)
- Tombstone-on-drop refinement: #1159 rev3 and rev4

ymote:
## Summary
- P3 follow-up from #1150/#1159 codex audit: `drain_ready_continuations_for_session` was popping `max_items` from the scheduler then filtering, so dropped stale items consumed budget slots. A production caller with `max_items=1` (AppUI tick loop) returned empty when a stale wrap-up sat at heap head, leaving the fresh continuation queued behind it to wait a full tick (~30s).
- Approach (a) from the audit: loop-and-refill in the orchestrator. After filtering, if `kept.len() < max_items` and the scheduler returned anything, call `drain_ready_for_session` again with the remaining budget. The scheduler removes each popped item from `pending_by_key` and pushes back non-matching heap entries, so repeated calls cannot revisit drained items — safe to refill without touching the scheduler API.
- Initial `Vec::with_capacity` capped at 32 because some callers pass `usize::MAX` to mean "drain everything".

## Test plan
- [x] New regression test `drain_with_max_items_one_finds_fresh_when_stale_drains_first_per_1160` is red on main, green after fix (uses `max_items=1` with stale-then-fresh ordering; the existing #1150 test uses `usize::MAX` and does not catch this)
- [x] `cargo test -p octos-cli --features api --lib api::agent_orchestrator` — 72/72 pass
- [x] `cargo clippy -p octos-cli --no-deps -- -D warnings` introduces no new findings (`agent_orchestrator.rs` is clean apart from a pre-existing `too_many_arguments` warning on `record_agent_ping` at line 848)

Closes #1160.

### DIFF
--- a/crates/octos-cli/src/api/agent_orchestrator.rs
+++ b/crates/octos-cli/src/api/agent_orchestrator.rs
@@ -951,12 +951,6 @@ impl InProcessAgentOrchestrator {
         // tripping the min-delay gate), so an active goal only ran
         // its initial continuation and never recurred.
         enqueue_due_goal_continuations(&mut state, session_id, profile_id, runtime_state, now);
-        let drained = state.continuations.drain_ready_for_session(
-            runtime_state,
-            max_items,
-            &session_id.to_string(),
-            profile_id,
-        );
         // #1150 codex P2 follow-up to #1145: `pending_continuation_is_schedulable`
         // gates which sessions `due_loop_targets` surfaces, but the
         // scheduler's drain pops by `(session_key, profile)` without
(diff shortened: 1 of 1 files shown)


ymote:
codex review clean — ready for admin merge

`codex review --commit e27cc8e99cb93e86f27cf0594b8f3b1549f2a464 --title "PR #1161: drain refill for #1160 (max_items budget)"`:

> The change correctly refills the drain budget after filtering stale continuations, and the affected orchestrator tests pass. I did not identify any introduced correctness issues that warrant an actionable finding.

PR #1180, opened 2026-05-23
fix(pipeline): cascade-fail child node tasks on pipeline timeout (closes the dspfac "pipeline:analyze running" forever bug)

ymote:
## Summary

When `run_pipeline` hits its `tokio::time::timeout` budget, the inner executor future is dropped before per-node `mark_completed`/`mark_failed` cleanup at `executor.rs:2456-2478` can fire. The parent task transitions to `Failed`, but every `pipeline:<node>` child task registered via `register_node_task` stays in `state: "running"` in the supervisor — visible on the dashboard as e.g. `pipeline:analyze running` indefinitely.

This PR adds `TaskSupervisor::mark_descendants_failed(parent_tool_call_id, reason)` that snapshots active tasks under the supervisor mutex matching the parent's `tool_call_id`, then iterates and calls `mark_failed` on each (deadlock-safe, mirrors the fanout-cap cascade at `task_supervisor.rs:1350-1368`). In `RunPipelineTool::execute`'s timeout arm we now re-snapshot the host context from `TOOL_CTX` and call the new method via a small `cascade_fail_orphan_node_tasks` helper.

The fix is server-only — orphan child tasks now never reach the wire in `state: "running"` once the parent has timed out.

## Ledger evidence

On mini3 (`/Users/cloud/.octos/ui-protocol/.../ledger-...39922.log`):
- Parent task `019e5210-...` (`run_pipeline`) hit `task_updated:failed` at seq 1158 with reason `"pipeline timed out after 1200s"`.
- Child `019e5222-...` (`pipeline:analyze`) had exactly ONE event — `state:running` at seq 1142 — and never a terminal mark.

After this fix the timeout arm cascades the timeout reason onto every still-active child registered under the parent's `tool_call_id`, so the next ledger event for `019e5222-...` would be `task_updated:failed` with reason `"pipeline timed out after 1200s"`.

## Test plan

- [x] `cargo test -p octos-agent --lib mark_descendants` — 2 new supervisor tests cover the cascade in isolation
- [x] `cargo test -p octos-pipeline --lib cascade` — 3 new tool tests pin the helper end-to-end with the same parent_tcid + child task shape the ledger evidence showed (cascade fires; no-op without supervisor; no-op without parent_tool_call_id)
- [x] `cargo test -p octos-pipeline` — 184 lib + 48 integration tests, no regressions
- [x] `cargo test --workspace` — clean
- [x] `cargo clippy --workspace -- -D warnings` — clean

## Files

- `crates/octos-agent/src/task_supervisor.rs` — new `mark_descendants_failed` + 2 unit tests
- `crates/octos-pipeline/src/tool.rs` — timeout arm cascade + new `cascade_fail_orphan_node_tasks` helper + 3 unit tests

## DO NOTs honored

- Did NOT touch SPA / TaskStore client paths — supervisor's authoritative state is now correct, no client-side workaround required.
- Did NOT emit a synthetic `task/updated` event from the client side.
- Added a regression test (per the user-visible bug report).

### DIFF
--- a/crates/octos-agent/src/task_supervisor.rs
+++ b/crates/octos-agent/src/task_supervisor.rs
@@ -1551,6 +1551,48 @@ impl TaskSupervisor {
         }
     }
 
+    /// Cascade-fail every still-active child of `parent_tool_call_id`.
+    ///
+    /// Used by the `run_pipeline` timeout arm to flush orphan
+    /// `pipeline:<node>` child tasks when the parent future is dropped
+    /// before per-node `mark_completed` / `mark_failed` can fire. Without
+    /// this cascade the children stay forever as `state: "running"` in
+    /// the supervisor, and the SessionTaskIndicator on the dashboard
+    /// shows e.g. `pipeline:analyze running` indefinitely.
+    ///
+    /// Snapshots the matching active task ids under the `tasks` mutex
+    /// first, then drops the lock and calls `mark_failed` per id so the
+    /// per-task lock acquisition inside `mark_failed` does not deadlock
+    /// on the snapshot. Returns the number of children that were
+    /// transitioned to `Failed`. Already-terminal tasks are skipped by
+    /// `is_active()` and the deadlock-safe `mark_failed` guard.
+    pub fn mark_descendants_failed(&self, parent_tool_call_id: &str, reason: &str) -> usize {
+        if parent_tool_call_id.is_empty() {
+            return 0;
+        }
+        let active_children: Vec<String> = self
+            .tasks
+            .lock()
+            .unwrap_or_else(|e| e.into_inner())
+            .values()
+            .filter(|task| task.tool_call_id == parent_tool_call_id && task.status.is_active())
+            .map(|task| task.id.clone())
+            .collect();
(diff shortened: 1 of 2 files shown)


ymote:
Addressed codex MAJOR: cascade now filters to `tool_name.starts_with("pipeline:")` so it only touches node children, not the parent `run_pipeline` task (both share the same `tool_call_id`). The parent's own `mark_failed` path in `RunPipelineTool::execute`'s timeout arm handles parent-level transition.

New regression test `mark_descendants_failed_does_not_touch_parent_run_pipeline_task` explicitly pins the parent-skipped case; existing `mark_descendants_failed_cascades_active_children_under_parent_tcid` and `cascade_fail_orphan_node_tasks_marks_all_active_children_failed` now also register a parent `run_pipeline` task with the same tcid and assert it remains Running.

MINOR doc-block split also fixed: relocated the `cascade_fail_orphan_node_tasks` helper below the complete `emit_external_context_unmanaged_timeout_summary` function so both rustdoc blocks read cleanly.

Pushed as: ed873491

PR #1181, opened 2026-05-23
fix(agent): news_fetch tool-output limit raised + LOOP DETECTED non-terminal so LLM can synthesise

ymote:
## Summary

Two related fixes diagnosed on mini3 (`session web-1779494658716-mxrxe8`,
ledger seq 214-562), where kimi-k2.5 spiraled on `news_fetch` for 9 calls
with drifting category lists (4 -> 7 -> 8 cats) before the loop detector
caught it on call 9. The detector then returned the `[LOOP DETECTED]`
warning as the FINAL turn output, so the LLM never saw the warning and
could not course-correct — the user saw a raw warning string as the
agent's "reply".

### Fix A — raise `news_fetch` tool_output_limit

`crates/octos-core/src/utils.rs`

- Root cause: `news_fetch` returns a JSON payload bundling ~50KB+ of
  raw headlines. The shared tool-output truncator capped at the 50K
  global default and inserted a middle-elision marker
  (`... [N bytes omitted] ...`). kimi-k2.5 read the marker as
  "incomplete results" and retried with drifting `categories=` arg
  lists, spinning until the loop detector tripped.
- Fix: raise `news_fetch` and `deep_search` (same aggregation
  profile) to 200K bytes. Unlisted tools still fall through to the
  50K global default.
- Regression tests:
  `news_fetch_limit_is_at_least_100k_bytes` and
  `deep_search_limit_is_at_least_100k_bytes` so a future silent
  shrink fails CI.

### Fix B — `[LOOP DETECTED]` is no longer terminal

`crates/octos-agent/src/agent/loop_runner.rs`

- Pre-fix: when the loop detector fired, the runner returned a
  `ConversationResponse { content: "[LOOP DETECTED]…" }`
  immediately, terminating the turn. The LLM never saw the warning
  in its conversation context, so it never had a chance to
  course-correct.
- Post-fix (single retry budget, intra-turn):
  1. **First fire** — push the looping assistant message plus
     synthetic tool-result messages (1:1 with the assistant's
     `tool_calls`, so provider schemas accept the next request).
     The PRIMARY tool result carries the warning text plus a
     synthesis hint:
     `"Try a different approach — synthesise from prior tool results …, call a different tool, or finish the turn with the partial information you have."`
     Continue the loop for ONE more iteration.
  2. **Second fire (same turn)** — the LLM ignored the warning and
     looped again. Return a terminal `ConversationResponse` with a
     hard-stop user-facing message. No third chance — budget bounded.
- The single-fire-per-burst flag (`loop_detected_recently`) is owned
  by the existing `dedup_loop_warning`. We catch the `Err` it
  returns on the second fire and convert it to a terminal `Ok` so
  callers never see an error result from a loop condition.
- The pre-existing shell-spiral recovery path (which has its own
  splice-and-continue logic with `dispatch_shell_retry_recovery`)
  runs first and is unchanged.

### Evidence

dspfac ledger: `session web-1779494658716-mxrxe8`, ledger seq
214-562 — kimi-k2.5 calling `news_fetch` 9 times with drifting
categories, detected at call 9, agent returned the warning as
content and the user saw raw `[LOOP DETECTED]` text.

## Test plan

- [x] `cargo build --workspace` — green
- [x] `cargo test -p octos-core` — 217/217 pass, including the two
      new `tool_output_limit` regression tests
- [x] `cargo test -p octos-agent --lib agent::loop_runner::tests::` —
      44 pass / 1 ignored (pre-existing), including:
      - `inject_synthetic_results_pushes_assistant_then_tool_for_every_call`
      - `loop_detected_terminal_message_is_user_facing_and_non_empty`
      - `loop_detected_first_fire_continues_then_second_fire_terminates`
      - `dedup_loop_warning_*` (3 pre-existing, still pass)
      - `process_message_fires_loop_warning_once_then_terminal_error`
        (pre-existing, still passes)
- [x] `cargo test -p octos-agent --lib loop_detect` — 5/5 pass
- [x] `cargo clippy --workspace -- -D warnings` — clean
- [ ] Soak on mini3 with the same `news_fetch` user prompt to verify
      the spiral no longer occurs and the LLM either synthesises a
      reply or switches to a different tool. (Out of scope for this
      PR per task scope; no deploy.)

### DIFF
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -1064,28 +1064,69 @@ impl Agent {
                                             tool_results: tool_structured_metadata.clone(),
                                         });
                                     }
-                                    // Single-fire-per-burst: first fire emits the
-                                    // warning; subsequent fires within the same
-                                    // burst (before the next process_message reset)
-                                    // surface a terminal error instead of repeating
-                                    // identical noise.
-                                    let warning_content = self.dedup_loop_warning(warning)?;
-                                    // Don't execute the tools — break out with a message
+                                    // Two-stage loop-detector recovery:
+                                    //
+                                    // 1. First fire in this turn — inject the
+                                    //    warning as a SYNTHETIC tool-result
+                                    //    message paired with the looping
+                                    //    assistant message, then continue
+                                    //    the loop. The LLM gets one more
+                                    //    iteration to synthesise an answer
+                                    //    from prior context or switch
+                                    //    tools/arguments. This rescues the
+                                    //    kimi-k2.5 news_fetch retry spiral
+                                    //    documented in PR
+                                    //    `fix/news-fetch-loop-and-detect-recovery`
+                                    //    (session `web-1779494658716-mxrxe8`,
+                                    //    ledger seq 214-562).
+                                    //
+                                    // 2. Second fire in the same turn — the
+                                    //    LLM ignored the warning and looped
+                                    //    again. Return a terminal
+                                    //    ConversationResponse with a
+                                    //    hard-stop message so the user sees
+                                    //    a clean reply rather than a thrash.
+                                    //
+                                    // The single-fire-per-burst flag
+                                    // (`loop_detected_recently`) is owned by
+                                    // `dedup_loop_warning`. The Err it
+                                    // returns on second fire is caught and
+                                    // converted to a terminal Ok response
+                                    // here so callers don't see an error.
                                     self.emit_cost_update(turn.total_usage(), &response);
-                                    return Ok(ConversationResponse {
-                                        content: warning_content,
-                                        reasoning_content: None,
-                                        provider_metadata: None,
-                                        token_usage: turn.total_usage().clone(),
-                                        files_modified,
-                                        files_to_send,
-                                        streamed,
-                                        messages: LoopTurnState::new_messages(
-                                            &messages,
-                                            history.len(),
-                                        ),
-                                        tool_results: tool_structured_metadata.clone(),
-                                    });
+                                    match self.dedup_loop_warning(warning) {
+                                        Ok(warning_content) => {
+                                            inject_loop_detected_synthetic_results(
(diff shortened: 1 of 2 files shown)


ymote:
Addressed both codex MAJORs in `2f0892e1`:

**MAJOR 1 — `deep_search` vs `search` name mismatch** (`crates/octos-core/src/utils.rs`)
- Added `search => 200_000` arm; kept `deep_search => 200_000` as a defensive alias / contract slot for future variants.
- Comment in the function now documents why both arms exist (runtime name vs contract name; bundled skill exposes its tool as `search` per `app-skills/deep-search/manifest.json`).
- Regression test extended to assert BOTH `tool_output_limit("search") >= 100_000` AND `tool_output_limit("deep_search") >= 100_000`.

**MAJOR 2 — synthetic tool-result IDs bypassed sanitization** (`crates/octos-agent/src/agent/loop_runner.rs`)
- `inject_loop_detected_synthetic_results` now clones the looping `ChatResponse`, runs `sanitize_tool_call_id` over each `tc.id`, and feeds the sanitized clone through `response_to_message` AND the tool-result loop. Same sanitized id ends up in BOTH the assistant message's `tool_calls[].id` and the matching tool result's `tool_call_id` — provider-side 1:1 pairing preserved.
- New test `inject_synthetic_results_sanitizes_tool_call_ids_with_colons` simulates a Moonshot/kimi-style id (`admin_view_sessions:11`) and asserts (1) sanitized (no colon), (2) assistant and tool messages share the SAME id, (3) concrete sanitized form is `admin_view_sessions_11`.

Verification: `cargo fmt --all && cargo build --workspace && cargo test -p octos-agent -p octos-core && cargo clippy --workspace -- -D warnings` all clean. Both new tests pass:
- `agent::loop_runner::tests::inject_synthetic_results_sanitizes_tool_call_ids_with_colons ... ok`
- `utils::tests::deep_search_limit_is_at_least_100k_bytes ... ok` (now covers both `search` and `deep_search`)

PR #1186, opened 2026-05-23
fix(podcast): harden podcast_generate against script-file race (B1 soak fix)

ymote:
## Summary
Fixes B1 from fleet UX soak (2/24 lanes hit it): `podcast_generate` returning
\`No such file or directory (os error 2)\` on a script_path that
\`write_file\` had just successfully written.

Root cause was NOT a tool-ordering race — the M8.8 serial batch already
awaits \`write_file\` before \`podcast_generate\` spawns. It was a host-side
path-doubling bug:

- The modern \`runtime/session.rs\` path chroots plugin \`work_dir\` into
  \`<workspace>/skill-output/\`.
- \`write_file\` resolves against the workspace ROOT, so a
  \`skill-output/<file>\` path lands at \`<workspace>/skill-output/<file>\`.
- The LLM mirrors the same \`skill-output/<...>\` path into
  \`podcast_generate\`'s \`script_path\`, but \`rewrite_workspace_file_args\`
  (a) didn't rewrite \`script_path\` and (b) the naive join doubled the
  prefix to \`<workspace>/skill-output/skill-output/<file>\` which doesn't
  exist.

The skill source already has the same guard for its own output dir
(\`mofa-podcast::resolve_output_dir\`). This change adds the symmetric
guard on the host's pre-spawn arg-rewrite layer so any \`script_path\`-style
arg the LLM hands to a chrooted spawn_only plugin resolves correctly
without changing the skill binary.

### Evidence
- \`e2e/test-results-fleet-ux-soak/mini2/iter-1/podcast/result.json\`
  — \`errorChip: true\`, error message:
  \`{"error":"Failed to read script file 'skill-output/mofa-podcast/octos_intro_script.md': No such file or directory (os error 2)","success":false}\`
- \`e2e/test-results-fleet-ux-soak/mini5/iter-2/podcast/\` — same shape.
- Ws-frames confirm \`write_file\` reported \`success: true\` at t=22152ms
  and \`podcast_generate\` failed at t=23325ms reading the same path.

### Files
- \`crates/octos-agent/src/plugins/tool.rs\` — add \`script_path\` to the
  rewrite list; add \`strip_redundant_skill_output_prefix\` helper;
  extend \`resolve_plugin_input_path\` to probe the stripped form first;
  add two unit tests covering modern (chrooted) and legacy
  (workspace-root) \`work_dir\` layouts.

## Test plan
- [x] \`cargo test -p octos-agent --lib plugins::tool::tests::rewrite_workspace_file_args\` → 8 passed (6 existing + 2 new)
- [x] \`cargo test -p octos-agent --lib plugins\` → 133 passed
- [x] \`cargo test -p octos-agent --lib\` → 1578 passed
- [x] \`cargo clippy -p octos-agent --lib --no-deps\` → clean
- [x] \`cargo fmt --all -- --check\` → clean
- [ ] Re-run the fleet UX soak \`podcast\` scenario on mini2/mini5 with this
      binary deployed to confirm B1 stays fixed.

### DIFF
--- a/crates/octos-agent/src/plugins/loader.rs
+++ b/crates/octos-agent/src/plugins/loader.rs
@@ -1938,7 +1938,9 @@ edition = "2021"
         assert_eq!(tools.len(), 1);
         // Inject through prepare_effective_args to verify the loader propagated
         // the config into the constructed PluginTool.
-        let prepared = tools[0].prepare_effective_args(&serde_json::json!({"query": "x"}), None);
+        let prepared = tools[0]
+            .prepare_effective_args(&serde_json::json!({"query": "x"}), None)
+            .unwrap();
         assert_eq!(prepared["synthesis_config"]["api_key"], "sk-loader-test");
     }
 
@@ -1991,7 +1993,9 @@ edition = "2021"
         )
         .unwrap();
         assert_eq!(tools.len(), 1);
-        let prepared = tools[0].prepare_effective_args(&serde_json::json!({}), None);
+        let prepared = tools[0]
+            .prepare_effective_args(&serde_json::json!({}), None)
+            .unwrap();
         assert!(
             prepared.get("synthesis_config").is_none(),
             "non-opted-in plugin must not see synthesis_config: {prepared}"
--- a/crates/octos-agent/src/plugins/tool.rs
+++ b/crates/octos-agent/src/plugins/tool.rs
@@ -543,9 +543,18 @@ impl PluginTool {
             }
             if matches!(key.as_str(), "out" | "slide_dir") {
                 if let Some(path) = value.as_str() {
+                    // Codex round-4 BLOCKER fix (PR #1186 review):
+                    // propagate the absolutize Err for output-path keys
+                    // so a `{"out":"../sneaky"}` or
+                    // `{"slide_dir":"../escape"}` surfaces as a tool
+                    // error envelope rather than being passed to the
+                    // spawned plugin (which writes its output relative
+                    // to `cmd.current_dir(work_dir)` and would escape
+                    // the chroot). Matches the round-3 contract on
+                    // input-path keys (resolve_plugin_input_path).
                     rewritten.insert(
                         key.clone(),
-                        serde_json::Value::String(absolutize_path_in_work_dir(path, work_dir)),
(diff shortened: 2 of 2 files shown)


ymote:
Addressed codex BLOCKER (path traversal via parent-dir escape) + MAJOR (video_path/text_path added). 65ac6ba4

- BLOCKER: `strip_redundant_skill_output_prefix` now refuses `..`-containing or absolute raw paths; `resolve_path_in_work_dir` skips its absolute / lexical-join / CWD-relative existence-check branches for unsafe candidates while keeping the safe basename-only fallback intact. New `has_unsafe_components` helper uses `std::path::Component` to detect `ParentDir` / `RootDir` / `Prefix`.
- MAJOR: `video_path` (mofa-frame) and `text_path` (mofa-videolizer) added to the input-side rewrite list in `rewrite_workspace_file_args`.
- Tests: 3 new tests in `plugins::tool::tests` (parent-dir escape, absolute paths, video_path/text_path rewrite). 1581/1581 octos-agent lib tests pass; `cargo clippy --workspace -- -D warnings` clean.

ymote:
Round-2: raw `../foo` paths now rejected by early guard in resolver. 5d5fab62

ymote:
Round-3: fail-closed on raw `..`. Resolver now returns Err instead of pass-through; rewrite_plugin_args propagates to tool error envelope. New test covers raw escape paths.

ymote:
Round-4: extend fail-closed to output keys (`out`, `slide_dir`). `absolutize_path_in_work_dir` now returns `Result<String, eyre::Report>` and rejects raw `..` at entry (same `escapes plugin work dir` message as round-3); `rewrite_workspace_file_args` `?`-propagates. New test `rewrite_workspace_file_args_rejects_raw_parent_dir_on_output_keys` covers `{"out":"../sneaky"}`, `{"slide_dir":"../escape"}`, `{"out":"subdir/../../../etc/passwd"}`, `{"slide_dir":".."}`, and `{"out":"a/../../escape"}`. Also threaded the Result through `resolve_slides_style_in_work_dir` (only other in-tree caller of absolutize) and the `resolve_plugin_input_path` fallback. Commit `82aaff50`.

PR #1187, opened 2026-05-23
fix(agent): suppress spawn_only ack when invocation returned error (B4)

ymote:
## Summary

Fleet-UX soak finding **B4** (mini1 / dspfac, 2026-05-22): when a turn batched a spawn_only tool call AND any sibling tool whose dispatcher returned an error-shaped result, the harness emitted a turn-final "Background work started for \`<tool>\`." acknowledgement as the assistant content even though the user-visible chat already showed a red error chip on the sibling tool card. The user saw both:

- Tool card: `× run_pipeline error: required tool(s) not available on this host: run_pipeline`
- Assistant bubble: "已后台启动 对 James Webb 太空望远镜的深度研究. The final result will be delivered..."

Simultaneously. The synthesised success bubble alongside the red error chip read as a confusing dual signal.

The complementary fix on main (#1183) hides the ack visually for capability-aware clients via `MessagePersistedSource::Background`. This patch closes the gap at the source: when any tool in the turn-batch already reported a failure, do not synthesize the ack at all. The agent loop falls through to a normal next-iteration so the LLM observes the error tool-result and can react.

## Scope

- New helpers in `crates/octos-agent/src/agent/loop_runner.rs`:
  - `is_error_tool_message` — body classifier for the well-known error/denial/cancellation conventions emitted by `crate::agent::execution` (`"Error: …"`, `"[VALIDATION FAILED]"`, `"[POLICY DENIED]"`, `"[HOOK DENIED]"`, `"[SESSION LIMIT]"`, `"[SHELL RETRY LIMIT]"`, panic/timeout/cancelled-due-to-earlier).
  - `any_tool_invocation_errored` — scans tool-result messages for tool_call_ids in the current LLM response and applies the body classifier.
- Spawn_only branch in `process_message_inner` calls the gate before synthesizing the ack. When the gate fires, the branch logs a `warn!` and falls through; the synthesized ack is suppressed AND `synthesized_from_spawn_only` stays `false`.

The background task — when one was actually dispatched — still completes asynchronously and routes its outcome via the `BackgroundResultSender`, so the legitimate "task finished" signal still arrives on that channel. Only the foreground-fabricated "started" bubble (which the foreground can't actually verify) is suppressed.

## Evidence — mini1 / dspfac

- Tool card chip: `× run_pipeline error: required tool(s) not available on this host: run_pipeline`
- Assistant text alongside it: "已后台启动 对 James Webb 太空望远镜的深度研究. The final result will be delivered automatically when it is ready." + fake outline bullets

Soak finding identified as **B4** in the fleet-UX walkthrough (2026-05-22).

## Test plan

- [x] `cargo test -p octos-agent --lib` — **1584 passed; 0 failed**, including the 8 new tests.
- [x] `cargo test -p octos-cli --features api --lib spawn_only_synthesized_ack_emits_as_background_source` — sibling-PR test still passes (the two fixes are complementary, not exclusive).
- [x] `cargo clippy --workspace -- -D warnings` — clean.
- [x] `cargo fmt --all -- --check` — clean.

New test coverage (all in `loop_runner.rs`):

- `is_error_tool_message_classifies_error_envelopes` — exhaustive positive + negative cases for the body classifier.
- 6 unit tests for `any_tool_invocation_errored`: simple detection, all-successful batch, VALIDATION FAILED envelope, mixed-batch with one failed tool, stale historical error (no false positive), panic + timeout envelopes.
- `spawn_only_branch_skipped_when_invocation_returned_error` — integration through the full `process_message` loop with a mock provider emitting a mixed spawn_only + erroring-sibling batch; asserts:
  - `result.content` does NOT start with "Background work started"
  - `result.synthesized_from_spawn_only` is `false`
  - `result.content` carries the LLM's recovery reply
  - The error tool-result remains in `result.messages` so the SPA keeps rendering the red error chip on the tool card

### DIFF
--- a/crates/octos-agent/src/agent/execution.rs
+++ b/crates/octos-agent/src/agent/execution.rs
@@ -1604,6 +1604,21 @@ impl Agent {
         Vec<std::path::PathBuf>,
         TokenUsage,
         Vec<(String, serde_json::Value)>,
+        // Codex round-2 MAJOR 2 (PR #1187 fixup): per-tool-call success bit
+        // keyed by `tool_call_id`. The dispatcher already computes this for
+        // the M8.8 serial-cascade scheduler (see [`ToolCallResult`] field 5);
+        // we surface it now so `any_tool_invocation_errored` in the
+        // loop_runner can authoritatively decide whether a tool failed
+        // without guessing from content prefixes. The content-based
+        // classifier was missing many real failure shapes — shell
+        // timeouts ("Command timed out after ..."), sandbox path
+        // rejections ("Path outside working directory ..."), browser
+        // navigation failures, plugin tools with `success: false` and
+        // arbitrary error messages — every one of which renders a red
+        // error chip but did NOT carry a recognised error envelope, so
+        // the synth-ack branch would still fabricate a "Background work
+        // started" bubble alongside it.
+        Vec<(String, bool)>,
     )> {
         let tool_names: Vec<&str> = response
             .tool_calls
@@ -1687,7 +1702,7 @@ impl Agent {
                         tools = %tool_names.join(", "),
                         "tool execution timed out -- spawned tasks continue running for cleanup"
                     );
-                    let messages = response
+                    let messages: Vec<Message> = response
                         .tool_calls
                         .iter()
                         .map(|tc| Message {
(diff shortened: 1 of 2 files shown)


ymote:
Addressed codex 2 MAJORs:

* **MAJOR 1** — sticky `spawn_only_was_invoked` flag now per-iteration. Gate reads `response.tool_calls.iter().any(|tc| self.tools.is_spawn_only(&tc.name))` so the synth-ack only fires when the CURRENT iteration actually contains a spawn_only call (multi-iter turns no longer fabricate an ack on a later regular-tool iteration).
* **MAJOR 2** — `ToolResult.success` bit plumbed through `execute_tools` → `handle_tool_use` → `any_tool_invocation_errored` as a `Vec<(tool_call_id, success)>` slice. The classifier now reads `success == false` as authoritative; content-based fallback kept for tool_call_ids that bypass `execute_tools` (session-limit blocks etc.). Catches shell timeouts, sandbox path-rejects, browser nav failures, plugin `success: false`.

Tests added:
- 6 unit tests on `any_tool_invocation_errored` covering each failure mode + authoritative-over-content + content-fallback paths.
- 1 integration test `spawn_only_sticky_flag_does_not_synthesize_ack_in_later_regular_iteration` (3-iteration turn) covering MAJOR 1.

Verified: build clean, 1591 lib tests pass, clippy `-D warnings` clean.

SHA: 7db5bf06

ymote:
Round-3: pass sanitized response to any_tool_invocation_errored so success-bit lookup matches. New test covers ID-sanitization edge.

PR #1204, opened 2026-05-24
migrate(phase2d): shell + spawn tools use SessionScope for child CWD

ymote:
## Summary

Phase 2-D of the [SessionScope migration](https://github.com/octos-org/octos/pull/1198) (Phase 1: #1199).

- **shell.rs** — overrides `Tool::execute_with_context`; derives `effective_cwd` from `ctx.session_scope.workspace()` and falls back to `self.cwd` when no scope. The same effective CWD feeds `CommandPolicy::check`, `sandbox.wrap_command`, the frontend cache env, and the approval payload so policy and the spawned child agree on the working directory.
- **spawn.rs** — propagates `ctx.session_scope` into both the sync and async sub-agent branches via `Agent::with_session_scope`, so the child Agent's execution loop stamps the scope onto every child `ToolContext`. The child shares the parent workspace (no per-child subdir carve-out yet).
- 6 new tests covering scope present, scope absent, and the codex-anticipated `SafePolicy` denylist regression (`rm -rf /` still denied with a scope on the context).

## Test plan

- [x] `cargo test -p octos-agent --lib tools::shell tools::spawn` — 12 shell + 42 spawn pass, 0 new failures
- [x] `cargo test -p octos-agent --lib --no-fail-fast` — 1624 passed, 3 pre-existing ignored
- [x] `cargo build --workspace` clean
- [x] `cargo clippy --workspace --lib --tests -- -D warnings` clean

## Out of scope

- Phase 2-A: pipeline workers
- Phase 2-B: plugin tool
- Phase 2-C: read_file / write_file / edit_file

These ship in their own PRs.

### DIFF
--- a/crates/octos-agent/src/tools/shell.rs
+++ b/crates/octos-agent/src/tools/shell.rs
@@ -222,18 +222,28 @@ impl Tool for ShellTool {
             serde_json::from_value(args.clone()).wrap_err("invalid shell tool input")?;
 
         // Phase 2-D of the SessionScope migration: when the host has
-        // threaded a scope through `ToolContext`, use `scope.workspace()`
-        // as the child process CWD so every tool in the session shares
-        // the single filesystem contract. Without a scope (e.g.
-        // `octos chat`, unit tests, legacy callers) we keep the
-        // construction-time `self.cwd` byte-for-byte so behaviour is
-        // unchanged. The same effective CWD also feeds the
-        // `CommandPolicy::check` call so a policy that consults the
-        // working directory sees a consistent value with what the child
-        // process will observe.
+        // threaded a scope through `ToolContext`, prefer the scope's
+        // workspace so every tool in the session shares the single
+        // filesystem contract.
+        //
+        // Codex P1 (round-1, this PR): respect a hinted workspace
+        // override. `SessionRuntime` builds `session_scope` from
+        // `<data_dir>/users/<id>/workspace` independent of any
+        // `workspace_hint` supplied by the coding-agent flow, while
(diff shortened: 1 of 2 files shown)


ymote:
## Codex round-1 review

> [P1] Preserve hinted workspaces when selecting shell CWD — `shell.rs`. For sessions bootstrapped with a `workspace_hint`, the registry's tools are rebound to the hinted `workspace_root`, but `SessionRuntime` still builds `ctx.session_scope` from `<profile.data_dir>/users/<session_id>/workspace`. This unconditional `Some(scope) => scope.workspace()` makes `shell` run builds/tests in that default data-dir workspace instead of the user's hinted repo, and the same scope is propagated to spawned children. Use a scope whose workspace is the effective workspace root, or fall back to `self.cwd` when the scope does not match the tool registry's CWD.

Addressed in `6bbcf43f` — when `scope.workspace() != self.cwd`, the caller deliberately rebound the tool to a different workspace (the hinted coding-agent path), so we keep `self.cwd`. SPA sessions (no hint) have matching scope and `self.cwd`, so the migration still takes effect. Phase 3 will reconcile SessionScope construction with hinted workspaces.

New regression test: `shell_respects_hinted_workspace_over_session_scope_default` exercises the exact divergence case.

Also: codex's "the same scope is propagated to spawned children" concern is closed by symmetry — the child shell tool's `self.cwd` derives from the parent SpawnTool's `working_dir` (the rebound hint), so the same scope-vs-self.cwd divergence detection triggers on the child and the child also lands on the hinted workspace.

Issue #257, opened 2026-04-02
Add Feishu reply threading regression tests


ymote:
## Problem

The Feishu test block covers base URL, allowlists, dedup, and content parsing, but nothing that exercises:
1. `parse_event()` → `message_id` preservation
2. Outbound reply routing when `reply_to` is present

This is why the threading divergence from Telegram went unnoticed.

## Fix

Add unit tests:
- `should_preserve_message_id_from_inbound_event` — verify `parse_event()` sets `InboundMessage.message_id`
- `should_use_reply_endpoint_when_reply_to_present` — verify outbound routing

ymote:
Closure coordination update (2026-05-26):

Two open PRs currently claim the #257 closure path:

- #1256 (`Closes #257`), head `c12c317870995323dc509cab3faad0ed093bbb37`, `mergeable=MERGEABLE`, CI green, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`.
- #1286 (`Closes #257`), head `04bf3b3b078544e9717e4eb914c2c31b8537f137`, `mergeable=MERGEABLE`, CI green, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`.

Both PRs add Feishu reply-threading regression coverage for `reply_to` routing. #1256 also reports `cargo clippy -p octos-bus --features feishu --all-targets -- -D warnings`; #1286 reports focused Feishu tests and clippy with existing warnings.

Decision: not manually closing #257. A maintainer should choose one closure PR, merge it after required review, and close or retarget the duplicate so the issue is not closed by two overlapping PRs.

ymote:
Current-main coordination update for #257 on 2026-05-26:

Both closure PRs were revalidated against current `origin/main` `21383341c5ee486bc010dcd345ddedded1e67303`:

- #1286 (`04bf3b3b078544e9717e4eb914c2c31b8537f137`) is the smaller patch. It adds the missing outbound reply-threading regression and passes focused/full Feishu tests. `cargo clippy -p octos-bus --features feishu --tests --no-deps` exits 0 but leaves existing Feishu warning debt.
- #1256 (`c12c317870995323dc509cab3faad0ed093bbb37`) covers the same #257 regression and also removes the existing Feishu warning debt enough for `cargo clippy -p octos-bus --features feishu --all-targets -- -D warnings` to pass.

[...]

PR #1230, opened 2026-05-24
test(feishu): cover reply threading

ymote:
## Summary
- Add a local Feishu REST test server to verify outbound `reply_to` messages use `/im/v1/messages/{message_id}/reply` rather than the top-level send endpoint.
- Keep the existing inbound `message_id` preservation regression in the same focused lane.
- Remove stale unused Feishu frame helpers/constants and simplify a no-op wire-type expression so the focused `feishu` clippy lane passes with `-D warnings`.

Related to #257

## Validation
- `cargo test -p octos-bus --features feishu feishu_channel::tests::should_ -- --nocapture`
- `cargo clippy -p octos-bus --features feishu --tests -- -D warnings`
- `git diff --check`
- `git ls-files --others --exclude-standard`

### DIFF
--- a/crates/octos-bus/src/feishu_channel.rs
+++ b/crates/octos-bus/src/feishu_channel.rs
@@ -32,8 +32,6 @@ use futures::SinkExt;
 
 /// Token refresh interval (slightly under 2 hours).
 const TOKEN_TTL_SECS: u64 = 7000;
-/// Maximum message IDs to track for dedup.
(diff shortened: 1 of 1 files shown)


ymote:
Closing as superseded by #1256/#1286. This refs-only Feishu threading test branch covers the same focused file as the newer closing #257 PRs. Leaving this branch undeleted for traceability.

PR #1278, opened 2026-05-25
fix(email): isolate inbound email threads

### LINKED ISSUE #1142: 【BUG】Email Channel Context Bloat
**Problem:** The Email channel uses the sender's address as the chat_id, so all emails from the same person go into a single session. Without context compaction, the token count keeps growing until it exceeds the model's limit (131K), causing all subsequent LLM requests to fail with 400 errors.

Observed growth: 220K → 437K → 873K → 1.7M tokens

**Suggested fixes:**
1. Enable session compaction for the Email channel
2. Isolate sessions by Message-ID or subject
3. Auto-truncate early history when exceeding the token limit

ymote:
## Summary
- derive a deterministic gateway-safe email thread topic from normalized subject or Message-ID headers
- attach the topic to inbound email metadata so gateway routing no longer keys every sender email to one session
- preserve the sender email as chat_id for outbound delivery while retaining Message-ID/In-Reply-To/References metadata

Closes #1142

## Validation
- cargo test -p octos-bus --features email email_channel::tests
- cargo test -p octos-bus --features email
- cargo clippy -p octos-bus --features email --all-targets -- -D warnings
- git diff --check
- git ls-files --others --exclude-standard

### DIFF
--- a/crates/octos-bus/src/email_channel.rs
+++ b/crates/octos-bus/src/email_channel.rs
@@ -9,6 +9,7 @@ use chrono::Utc;
 use eyre::{Result, WrapErr};
 use futures::StreamExt;
 use octos_core::{InboundMessage, OutboundMessage};
+use sha2::{Digest, Sha256};
 use tokio::sync::mpsc;
 use tracing::{info, warn};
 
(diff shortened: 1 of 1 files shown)


ymote:
Closing this as a duplicate of #1242. #1242 is already green, already carries the Closes #1142 reference, and has the stronger email-thread key order using References/In-Reply-To before subject fallback. Keeping one closure PR avoids duplicate closing refs for #1142.

PR #1220, opened 2026-05-24
fix(gateway): stop actors on session delete

ymote:
## Summary
- Wire gateway slash `/delete` handling into the existing session-delete event channel used to stop in-memory session actors.
- Notify the runtime with the exact profiled deleted session key after the disk session clear succeeds.
- Add gateway dispatcher regressions for named deletion and current-topic deletion emitting actor-stop signals.

Closes #1219

## Current-main refresh
- Base: current GitHub `main` / `origin/main` `4adbe05fff212cb85d1f0a6d6225da0713446016`.
- Head: `a66a2dd851659989c203f7c981ce2364e7b09dbe`.
- Isolated checkout: `/private/tmp/octos-1220-current.f8pao4/octos`.
- Environment: Darwin 25.5.0 arm64; `rustc 1.95.0`; `cargo 1.95.0`.
- Root dirty checkout was not modified.

## Local validation
- `git diff --check origin/main...HEAD` passed.
- `cargo fmt --all -- --check` passed.
- `rustfmt --check crates/octos-cli/src/commands/gateway/gateway_runtime.rs crates/octos-cli/src/gateway_dispatcher.rs` passed.
- `typos crates/octos-cli/src/commands/gateway/gateway_runtime.rs crates/octos-cli/src/gateway_dispatcher.rs` passed.
- `CARGO_TARGET_DIR=/private/tmp/octos-1220-current-4adbe-target-cli CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo test -p octos-cli gateway_dispatcher::tests::should_signal_actor_stop -- --nocapture` passed: 2/2 matching tests.
- `CARGO_TARGET_DIR=/private/tmp/octos-1220-current-4adbe-target-cli CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo test -p octos-cli gateway_dispatcher::tests::should_delete -- --nocapture` passed: 3/3 matching tests.
- `CARGO_TARGET_DIR=/private/tmp/octos-1220-current-4adbe-target-cli CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo test -p octos-cli gateway_dispatcher::tests -- --nocapture` passed: 34/34 gateway dispatcher tests.
[...]

PR #1282, opened 2026-05-25
Stop gateway actors after slash delete

### LINKED ISSUE #1219: Bug: Gateway `/delete` fails to stop the in-memory session actor
## Description

When a client sends the `/delete` command over a WebSocket gateway connection, the `handle_delete_command` path correctly calls `session_mgr.clear()` to wipe persisted topic/history data from disk. However, **it does not stop the in-memory `SessionActor`** that was spawned by the dispatch registry for that session.

As a result, the actor continues running with stale internal state (context, tools, pending LLM calls). If the same `channel:chat_id` pair sends new messages, the **old actor** processes them — it has no way of knowing its backing data was deleted. This violates user expectations: after `/delete`, the user's session should be fresh, with no leftover conversational state.

---

## Root Cause

The `/delete` handler in `GatewayDispatcher` (in `crates/octos-cli/src/commands/gateway/mod.rs`) only performs a disk-level clear:

```rust
async fn delete_session_handler(
    &self,
    session_mgr: &mut SessionManager,
    session_key: &SessionKey,
) -> Result<()> {
    session_mgr.clear(session_key).await?;
    // Missing: self.registry.remove_session(session_key)
    Ok(())
}
```

No call is made to the dispatch registry's `cancel()` / `remove_session()` / equivalent method to abort the actor task and remove it from `self.actors`.

Meanwhile, the dispatch registry also has a `cancel(session_key)` method that properly shuts down the actor and removes it from the `self.actors` map — but this is only ever called for the explicit `/session/stop` API endpoint, not for `/delete`.

---

## Steps to Reproduce

1. Connect via WebSocket with channel `slack` and chat-id `C123`.
2. Send a few messages so a `SessionActor` is spawned.
3. Send `/delete` (which calls `handle_delete_command`).
4. Observe that the dispatch registry (`self.actors`) still contains the entry for `slack:C123`.
5. Send another message for `slack:C123`.
6. The **old** actor handles it — no fresh session is created.

The verification test `test_gateway_delete_leaves_actor_alive` asserts step 4 programmatically. Run it:

```bash
cargo test -p octos-cli test_gateway_delete_leaves_actor_alive
```

---

## Expected Behaviour

After a `/delete` command:

1. Persistent data is cleared (already works via `session_mgr.clear()`).
2. The in-memory `SessionActor` is cancelled and removed from the dispatch registry.
3. The next message for the same `channel:chat_id` spawns a **fresh actor** with no prior history.

---

## Proposed Fix

Add a call to the dispatch registry's `cancel()` (or equivalent session-removal method) in the `/delete` handler. Something like:

```rust
async fn delete_session_handler(
    &self,
    session_mgr: &mut SessionManager,
    session_key: &SessionKey,
) -> Result<()> {
    // 1. Wipe persisted data
    session_mgr.clear(session_key).await?;
    
    // 2. Stop the in-memory actor so the next message starts fresh
    self.dispatch_registry.cancel(session_key).await;
    
    Ok(())
}
```

The exact API depends on how the dispatcher stores a reference to the registry. See the existing `/session/stop` handler for the canonical pattern of cancelling an actor by session key.

---

## Impact

- **Data integrity:** Users see stale conversational state after deleting their session.
- **Fresh-start guarantees broken:** Any feature that relies on `/delete` to reset a conversation is unreliable.
- **Resource leak:** Abandoned actor tasks consume memory and channel slots until the gateway connection closes.

ymote:
## Summary
- wire gateway slash-command deletion into the existing session-delete actor removal channel
- notify the runtime only after `SessionManager::clear` succeeds
- add dispatcher regressions for named deletion and current-topic deletion notifications

Closes #1219

## Validation
- `CARGO_TARGET_DIR=/private/tmp/octos-1219-target cargo test -p octos-cli --features api should_notify_runtime_when -- --nocapture`
- `CARGO_TARGET_DIR=/private/tmp/octos-1219-target cargo test -p octos-cli --features api gateway_dispatcher::tests::should_delete -- --nocapture`
- `CARGO_TARGET_DIR=/private/tmp/octos-1219-target cargo clippy -p octos-cli --features api --all-targets`

Notes: clippy passed with existing warnings in untouched code.

### DIFF
--- a/crates/octos-cli/src/commands/gateway/gateway_runtime.rs
+++ b/crates/octos-cli/src/commands/gateway/gateway_runtime.rs
@@ -1511,6 +1511,7 @@ impl GatewayRuntime {
             pending_messages.clone(),
             out_tx.clone(),
         )
+        .with_session_delete_notifier(session_delete_tx.clone())
         .with_data_dir(data_dir.clone());
 
         // Drop the original out_tx — factory and registry hold their own clones.
--- a/crates/octos-cli/src/gateway_dispatcher.rs
+++ b/crates/octos-cli/src/gateway_dispatcher.rs
@@ -37,6 +37,7 @@ pub struct GatewayDispatcher {
     pub(crate) active_sessions: Arc<RwLock<ActiveSessionStore>>,
     pub(crate) pending_messages: PendingMessages,
     pub(crate) out_tx: mpsc::Sender<OutboundMessage>,
+    pub(crate) session_delete_tx: Option<mpsc::UnboundedSender<String>>,
     /// Profile ID used for building profiled session keys (None = main profile).
     pub(crate) dispatch_profile_id: Option<String>,
     /// Profile data directory for per-user files (soul.md, etc.).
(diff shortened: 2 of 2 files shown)


ymote:
Closing this as a duplicate of the earlier #1220, which already targets #1219 with the same gateway /delete actor-stop fix and validation. Keeping #1220 as the active review/merge vehicle avoids duplicate closure PRs for the same issue.

Issue #1380, opened 2026-05-31
AppUI WS task-control advertised but not executable in `octos serve` (returns runtime_unavailable)


ymote:
## Summary
`octos serve` **advertises** the `harness.task_control.v1` capability (so `task/cancel`, `task/list`, `task/restart_from_node` appear in `supported_methods`), but its AppUI **WebSocket** handler **cannot execute them** — `task/cancel` returns `runtime_unavailable: "task supervisor not wired for AppUI task commands"`. Capability advertisement is decoupled from runtime wiring.

## Evidence (live)
Reproduced end-to-end on 2026-05-31 via a real DeepSeek-backed `octos serve` + the octos-tui task-cancel work (octos-tui#156):
- WS capability probe against `octos serve`: **with** `X-Octos-Ui-Features: harness.task_control.v1` → 37 methods incl. `task/cancel`; **without** → 34 (so serve *does* advertise it).
- Prompted the model to `spawn` a background subagent (`sleep 200`) → a real Running task appeared.
- Client sent `task/cancel` → server replied `runtime_unavailable: task supervisor not wired for AppUI task commands`; the task kept running.

## Root cause
- `crates/octos-cli/src/commands/serve.rs:653-660` sets `task_query_store: None` **by design** — *"the API server proper — all session actors live in gateway processes … cancel/restart handlers proxy via `resolve_api_port`; the gateway runtime sets its own store on the embedded api channel."*
- The **REST** `/api/tasks/{id}/cancel` proxies to the gateway, but the **WS** AppUI handler (`crates/octos-cli/src/api/ui_protocol.rs:11889` `task_query_store_or_error`) has **no proxy fallback** — it just errors when `task_query_store` is `None`.
- Meanwhile the capability is advertised from the client's requested features only (`ConnectionUiFeatures::harness_task_control`, `ui_protocol.rs:1031/1133`), never gated on whether this connection can actually run it.

## Impact
Any AppUI client (TUI/app/web) connecting to a bare `octos serve` is offered task-control affordances that always fail with a typed runtime error. (Clients surface the error correctly — this is a server-honesty gap, not a client bug.)

## Proposed fix
[...]

PR #1381, opened 2026-05-31
fix(api): proxy WS task/cancel + task/restart_from_node to the gateway (#1380)

ymote:
Fixes the server side of #1380 without changing AppUI capability advertisement.

## Problem
Bare octos serve advertises harness.task_control.v1 to AppUI WebSocket clients, but the WS task-control handlers previously returned runtime_unavailable when no local task_query_store was wired. That is the normal serve topology: session actors live in gateway processes. The REST task cancel path already proxies to the gateway; the WS path did not.

## Fix
Mirror the REST gateway behavior for WS task control when no local store is available:

- task/cancel resolves the gateway API port for the request profile, proxies POST /tasks/{id}/cancel, and maps gateway status into the JSON-RPC task cancel outcome.
- task/restart_from_node resolves the same gateway route, proxies POST /tasks/{id}/restart-from-node with node_id, and maps gateway status/body into the relaunch outcome.
- The 2026-05-31 review refresh also keeps gateway port fallback and session scoping on the proxy path.
- task/list is not speculatively remapped here; gateway-mode listing remains available through the existing session/tasks.list auxiliary method.

A bare serve with no gateway still returns the typed runtime_unavailable error. Advertisement is unchanged, preserving UPCR-2026-007 first-server echo and WS/stdio parity.

## Validation
Base: origin/main 01c91d5ee399ea0b6a1b030c3c04fa0164321ca3.
Head: 869905cf.
Environment: ignored clone under target/octos-1338-fix.j7YNgn/octos; CARGO_TARGET_DIR=/private/tmp/octos-1381-test-target; CARGO_INCREMENTAL=0; CARGO_PROFILE_DEV_DEBUG=0.

Passed locally:

- cargo fmt --all -- --check
- git diff --check
- git ls-files --others --exclude-standard
- cargo test -p octos-cli --features api proxied_task -- --nocapture -> 2/2 proxy tests passed
- cargo clippy --workspace --all-targets -- -D warnings

No generated artifacts are included.

## Remaining gap
The live serve plus gateway proxy round-trip still needs an integration check before closing #1380 from evidence. This PR intentionally uses Refs rather than Closes until that proof exists.

Refs #1380.

### DIFF
--- a/crates/octos-agent/src/agent/execution.rs
+++ b/crates/octos-agent/src/agent/execution.rs
@@ -172,11 +172,11 @@ pub(super) fn satisfied_completion_content(output_files: &[String], tool_output:
 /// slides-1780013669236-8w2ime, the production failure that motivated
 /// this fix):
 ///   - kimi-k2.5 + only `check_workspace_contract`:
-///       loop rate 5/5 → 3/5 with this block (40% break out)
+///     loop rate 5/5 → 3/5 with this block (40% break out)
 ///   - kimi-k2.5 + check + read_file + list_dir:
-///       no consistent change (within noise)
+///     no consistent change (within noise)
 ///   - claude-opus-4.7:
-///       already 0/5 loops in both arms; block has no effect on Opus
+///     already 0/5 loops in both arms; block has no effect on Opus
 ///
 /// The block follows hermes-agent's prompt-time-injection pattern (in
 /// `agent/prompt_builder.py`), but applied universally rather than
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -2058,9 +2058,7 @@ impl Agent {
                 let Some(&(name, args)) = id_to_call.get(id) else {
                     continue;
                 };
-                if let Some(hint) =
-                    loop_detector.record_result(name, args, &message.content)
-                {
+                if let Some(hint) = loop_detector.record_result(name, args, &message.content) {
                     message.content.push_str(&hint);
                 }
             }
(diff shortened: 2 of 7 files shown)


ymote:
Extended the same gateway-proxy fallback to **`task/restart_from_node`** (commit `53b5bfc2`): forwards `POST /tasks/{id}/restart-from-node` with `{node_id}`, reads the relaunched task id from the gateway body, maps 404/409 to typed relaunch errors. Unit-tested; full `ui_protocol` module green (299, single-threaded).

**`task/list` intentionally not proxied:** gateway-mode listing already works via the `session/tasks.list` auxiliary method (which proxies `/sessions/{id}/tasks`), and a *typed* `task/list` proxy would need an unverifiable `Value`→`Vec<TaskListEntry>` deserialize. Documented rather than shipping a speculative mapping. So the two task-control **actions** (cancel, restart) now proxy; the **read** has an existing path.

ymote:
**Ran `codex review`** (vs origin/main) — it found 2 P2s in the gateway-proxy path, both now fixed in `179af008`:

1. **Port resolution required `connection_profile_id`** → the proxy never ran for connections without a bound profile (no-auth/first-gateway/admin-routed), dead-ending at `runtime_unavailable`. Now falls back to `ProcessManager::first_api_port`, mirroring `resolve_api_port`.
2. **Session scoping bypassed** → the proxy forwarded the (task-id-scoped) gateway cancel/restart without checking the task belongs to the supplied session. Added `ensure_task_in_gateway_session` (proxy `GET /sessions/{id}/tasks`, confirm membership) before forwarding, mirroring the local `ensure_task_in_session`; a foreign task id now returns `unknown_task_id`.

Unit-tested the membership-list parser; full `ui_protocol` module green (300, single-threaded). Verification caveat unchanged: the live proxy round-trip still needs a `serve`+`gateway` integration check.

PR #1449, opened 2026-06-09
fix(api): keep approvals pending across ws reconnect

alan0x:
## Summary

Keep a tool-approval request pending when the WebSocket that delivered it disconnects, so the runtime waiter survives a reconnect instead of being dropped/auto-resolved.

## Problem

When the UI WebSocket closed (e.g. a browser reconnect) while an approval was outstanding, the pending approval and its runtime waiter were lost — the agent could be left blocked or the approval silently resolved, breaking the human-gate flow across reconnects.

## Changes

- `crates/octos-cli/src/api/ui_protocol.rs`: keep the approval pending and its runtime waiter alive across a WS disconnect, so a reconnecting client can still respond to the outstanding request.

## Tests

`api::ui_protocol::tests::approval_request_closed_ws_keeps_pending_runtime_waiter` (run with `--features api`) — passing.

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol.rs
+++ b/crates/octos-cli/src/api/ui_protocol.rs
@@ -148,6 +148,14 @@ const INTERRUPT_ACK_TIMEOUT: std::time::Duration = std::time::Duration::from_sec
 /// traffic. Tunable per session size.
 const WS_WRITER_CHANNEL_CAPACITY: usize = 1024;
 const APPROVAL_CANCELLED_REASON_REQUEST_SEND_FAILED: &str = "request_send_failed";
+/// Reason recorded when a pending APPROVAL entry is cancelled because the
+/// approval-requesting tool's waiting future was dropped (turn interrupt/abort,
+/// per-tool timeout, panic, connection close) before a client decision arrived
+/// — the approval-store analogue of [`USER_QUESTION_CANCELLED_REASON_WAITER_DROPPED`].
+/// This is what gives a kept-pending approval (a `Closed`/`FatalClosed` direct
+/// send that waits for reconnect/replay, #1449) a guaranteed release path
+/// instead of hanging the turn forever when no client ever answers.
+const APPROVAL_CANCELLED_REASON_WAITER_DROPPED: &str = "waiter_dropped";
 /// Reason recorded when a pending structured-USER-QUESTION entry is cancelled
 /// because the `ask_user_question` tool's waiting future was dropped (turn
 /// interrupt/abort, panic, connection close) — UPCR-2026-023 drop-guard. This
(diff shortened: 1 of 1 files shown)


rpelevin:
Keeping the waiter alive across reconnect is the right fix. I would make sure the reconnect path preserves not just "there is a pending approval," but the exact approval request the user is allowed to answer.

The pending record should probably carry:

- approval id;
- run id / session id;
- tool call id;
- tool name;
- canonical args digest;
- redacted display snapshot hash;
- requested_at / expires_at;
- delivered_to_connection_id;
- disconnected_at, if the delivery socket drops;
- reconnect generation or delivery attempt;
- current state: pending, re-delivered, approved, denied, expired, runtime_cancelled;
- decided_by / decided_at;
- consumed_at;
- terminal reason.

The subtle failure mode is a stale client resolving a prompt after a reconnect or after a newer decision. I would make the approval id stable across reconnect, but require the decision to match the still-pending approval id and fail closed if the runtime cancelled, expired, or consumed it.

Tests I would add:

1. disconnect after prompt delivery keeps the pending approval and runtime waiter alive.
2. reconnect re-emits the same approval id and same redacted display payload.
3. approval after reconnect resumes exactly one waiter.
4. duplicate approve/deny messages are idempotent after the first terminal decision.
5. stale response from an old socket cannot override a newer decision.
6. expiry during disconnect produces a terminal denial/expiry and cannot execute later.
7. changed tool args digest cannot reuse the pending approval.
8. runtime cancellation clears the waiter and rejects any later approval response.

That turns the reconnect behavior into a real approval lifecycle guarantee instead of just a UI retry.

alan0x:
Thanks for the detailed lifecycle breakdown. Most of what you're describing is already enforced by `PendingApprovalStore` today, so I want to map your list onto the existing state machine rather than re-introduce it.

**Already enforced (with tests):**

`ApprovalEntry` is a state machine — `Pending → Responded { decision }` or `→ Cancelled { reason }` — keyed on `(session_id, approval_id)`:

- **Idempotent terminal decision / stale-socket override (your items 4 and 5):** once an entry is `Responded`, any later `respond` (from *any* socket) returns a typed `APPROVAL_NOT_PENDING` carrying the `recorded_decision`. `respond` doesn't care which connection it came from — first decision wins, the rest are no-ops. See `known_pending_approval_accepts_once` and `reconnect_retry_preserves_recorded_approval_decision`.
- **Exactly one waiter resumed (your item 3):** the runtime waiter is a `oneshot` and resumed via `take()`, so it's structurally once-only. See `runtime_approval_response_resumes_waiting_tool`.
- **Runtime cancellation clears the waiter and rejects later responses (your item 8):** cancel sets `response_tx = None` (receiver errors → agent treats as Deny) and moves the entry to `Cancelled`, so a late `respond` gets a typed `APPROVAL_CANCELLED`. See `cancel_drops_runtime_waiter_so_it_resolves_to_deny`, `respond_to_cancelled_approval_returns_typed_error`, `exact_cancel_preserves_cancelled_error_for_late_respond`.
- **Stable approval id + identical replay on reconnect (your item 2):** the `approval_id` is a fresh UUID per request and never reused; the durable ledger replays the original `ApprovalRequestedEvent` verbatim, so the reconnect re-emit is byte-identical by construction. This PR's test asserts the replayed event carries the same `approval_id` and that no `ApprovalCancelled` was emitted.

So this PR is the missing piece — keeping the entry `Pending` (instead of cancel+deny) when the *delivery socket* is `Closed`/`FatalClosed`, so the already-existing lifecycle can resume on reconnect. That's your item 1.

**Doesn't apply to the current protocol:**

- **Canonical args digest (your item 7):** `respond` carries only `(session_id, approval_id, decision)` — it never carries tool args. The args are fixed server-side on the `ApprovalRequestedEvent` at request time, so there's no client-supplied surface to re-bind a digest against.
- **Connection binding (`delivered_to_connection_id`, `disconnected_at`, reconnect generation):** approvals are intentionally scoped to the *session*, not a connection. "Any client in the session may answer, first-wins, rest idempotent" is exactly what makes answering-after-reconnect work; binding to a connection id would defeat the fix.
- **Redacted display snapshot hash:** there's no separate redaction layer here — the ledger replays the event itself, so the re-emit is already identical; there's nothing distinct to hash.

**The one genuinely new capability — expiry (your item 6):** approvals currently have no TTL; a pending entry lives until the turn is cancelled. That's a reasonable hardening, but it's a separate lifecycle feature (timer + terminal expiry event + ledger record) and out of scope for this reconnect fix. Tracked in #1451 rather than expanding this PR.

PR #1461, opened 2026-06-13
feat(voice): per-tenant selectable reply voice (backend)

alan0x:
Lets each tenant choose their TTS timbre — sticky per-profile and effective on the next turn without a restart.

## What
- `octos-llm`: `VoicesRegistry` parses `~/.OminiX/models/voices.json`; `synthesizable()` lists only voices whose ref audio exists; `resolve()` canonicalises id/alias.
- `GET /api/voices` lists synthesizable voices + the caller's current voice (degrades gracefully when the registry is missing).
- `PUT /api/my/voice` validates against the registry, persists the canonical id to `profile.config.voice_default` (per-tenant, survives restarts), and sets a live override.
- Timbre is now per-tenant: `ProfileConfig.voice_default` overlays the serve-level `VoiceConfig` at bootstrap via `with_default_voice_override()`; route/ASR settings stay platform-level.
- Turn handler resolves the reply voice through the override at both synthesis sites.

## Design notes
- The live override is a process-wide `LazyLock` map keyed by profile id (not an `AppState` field — `AppState` has ~50 construction sites; the map matches the single-serve-process scope).
- OminiX-API needs no code change; voices are registered as data in the runtime `voices.json`.

## Tests
- octos-llm registry (2), voices module (3), config override (2); profiles/runtime suites pass; fmt + clippy clean.

### DIFF
--- a/crates/octos-cli/src/api/auth_handlers.rs
+++ b/crates/octos-cli/src/api/auth_handlers.rs
@@ -1110,6 +1110,139 @@ pub async fn update_my_profile(
     }))
 }
 
+// ── Voice selection endpoints ────────────────────────────────────────
+
+#[derive(Serialize)]
+pub struct VoicesResponse {
+    /// Voices the engine can actually synthesize (ref audio present).
+    pub voices: Vec<octos_llm::ominix::VoiceInfo>,
+    /// This user's currently effective reply voice (live override > persisted
+    /// per-profile default > serve default).
+    pub current: String,
+}
+
+/// GET /api/voices — list synthesizable voices + this user's current choice.
+///
+/// Reads the platform registry (`~/.OminiX/models/voices.json`). A missing or
+/// unreadable registry degrades to a single-entry list of the current voice
+/// rather than failing the request.
+pub async fn list_voices(
+    State(state): State<Arc<AppState>>,
+    headers: HeaderMap,
+    axum::Extension(identity): axum::Extension<AuthIdentity>,
+) -> Result<Json<VoicesResponse>, StatusCode> {
+    let ps = state
+        .profile_store
+        .as_ref()
+        .ok_or(StatusCode::SERVICE_UNAVAILABLE)?;
+    let profile = resolve_my_profile(&identity, ps, &state, &headers)?;
+    let profile_id = profile.id.clone();
+
+    let registry_path = crate::api::voices::registry_path();
+    let (mut voices, registry_default) =
+        match octos_llm::ominix::VoicesRegistry::load(&registry_path) {
+            Ok(reg) => (reg.synthesizable(), reg.default_voice),
+            Err(e) => {
+                tracing::warn!(
+                    error = %e,
+                    path = %registry_path.display(),
+                    "voices.json unavailable; returning current default only"
+                );
+                (Vec::new(), String::new())
+            }
+        };
+
+    // Fallback chain for the "current" voice when no live override is set: the
+    // bootstrapped per-profile default (already overlays the serve default),
+    // then the registry default, then the first listed voice.
+    let runtime_default = state
+        .profiles
+        .get(&profile_id)
+        .map(|r| r.voice.default_voice.clone())
+        .filter(|s| !s.is_empty());
+    let fallback = runtime_default
+        .or_else(|| {
+            profile
+                .config
+                .voice_default
+                .clone()
+                .filter(|s| !s.is_empty())
+        })
+        .or_else(|| Some(registry_default).filter(|s| !s.is_empty()))
+        .or_else(|| voices.first().map(|v| v.id.clone()))
+        .unwrap_or_default();
+    let current = crate::api::voices::resolve_reply_voice(&profile_id, &fallback);
+
+    // Degrade gracefully: an unreadable registry still shows the current voice.
+    if voices.is_empty() && !current.is_empty() {
+        voices.push(octos_llm::ominix::VoiceInfo {
+            id: current.clone(),
+            aliases: Vec::new(),
+        });
+    }
+
+    Ok(Json(VoicesResponse { voices, current }))
+}
+
+#[derive(Deserialize)]
+pub struct SetVoiceRequest {
+    pub voice: String,
+}
+
+#[derive(Serialize)]
+pub struct SetVoiceResponse {
+    pub ok: bool,
+    pub voice: String,
+}
+
+/// PUT /api/my/voice — set this user's sticky reply-voice default.
+///
+/// Validates the voice against the registry, persists the canonical id to the
+/// profile (sticky across restarts), and records a live override so the switch
+/// takes effect on the next turn without a runtime reload.
+pub async fn set_my_voice(
+    State(state): State<Arc<AppState>>,
+    headers: HeaderMap,
+    axum::Extension(identity): axum::Extension<AuthIdentity>,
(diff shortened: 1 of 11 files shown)


ymote:
Nice work. Two notes from a review pass:

- **CI**: the red `check` is a pre-existing rustfmt drift on `session_actor.rs` — a file this PR doesn't touch — already resolved on current `main`. A rebase should clear it; nothing wrong with your code.
- **Tenant isolation looks correct**: the live-override map is keyed by `profile_id`, `resolve_reply_voice` looks up that exact key, and the setting persists to the per-profile `ProfileConfig.voice_default`. No cross-tenant path. `PUT /api/my/voice` resolves the caller's own profile, so a tenant can only set their own.

One nit: the `voices.rs` module doc references `profile.config.voice.default_voice`, but the field is `voice_default` (code is consistent everywhere else — just the comment). Will review in earnest once it's green.

ymote:
Ran this through the codex gate (CI was green first). The core design is sound — the live-override map is correctly keyed by `profile_id` and the persisted setting is per-profile. But the gate surfaced a tenant-scoping gap (which I'd missed in my first pass — I'd only checked the override-map keying, not the selectable-voice set) plus a path-resolution bug. Both verified against the code:

**[P1 — blocker] Voice selection isn't scoped to the tenant.** The registry is a single global `voices.json`, and `scripts/register-fleet-voices.sh` (run as part of `deploy.sh` post-deploy) writes *every* per-profile voice clone from `~/.octos/profiles/*/data/voice_profiles/` into it — its own header says it makes `/v1/voices` "enumerate every saved voice profile." So on a deployed fleet host:
- `GET /api/voices` → `reg.synthesizable()` lists *all* tenants' voice ids (auth_handlers.rs:1146), and
- `PUT /api/my/voice` → `registry.resolve(req.voice)` validates against the global set with no profile filter (auth_handlers.rs:1227).

So any authenticated tenant can enumerate and set another tenant's cloned voice as their reply voice. Since a clone is derived from someone's uploaded audio, that's a cross-tenant privacy issue. Scope both endpoints to shared presets + voices owned by the resolved profile before listing/resolving.

**[P2] Registry path resolution can hide valid voices.** `ref_exists` does `Path::new(&models_base_path).join(&ref_audio).exists()` (ominix.rs:230-233) with no `~` expansion. When `models_base_path` is `~/.OminiX/models` (literal tilde) or empty while `ref_audio` is relative to the registry dir, it stats a literal `~`/CWD path → `synthesizable()` drops valid voices and `PUT /api/my/voice` rejects names Ominix can actually synthesize. Expand `~` and resolve a missing/relative base against the `voices.json` parent before the existence check.

P1 is the blocker. Holding the merge until the scoping is addressed — happy to re-review once it is. CI is green and the rest of the change looks good.

alan0x:
Pushed c2555be4 — P1 and P2 addressed, plus the `voice_default` doc nit from your first pass.

**P1 (blocker) — cross-tenant voice exposure.** Both endpoints are now scoped to the resolved profile: shared presets for everyone, a clone only for its owner. The fleet script registers each clone's absolute path (`.../profiles/<id>/data/voice_profiles/<name>.wav`) as `ref_audio`, so `owning_profile_of` derives the owner from that path segment — gated on a trailing `voice_profiles` segment so a stray `profiles/` in some other path isn't mistaken for ownership — and `voice_visible_to` allows shared-or-owned only. New `VoicesRegistry::synthesizable_visible` / `resolve_visible` take an `is_visible(ref_audio)` predicate; `GET /api/voices` and `PUT /api/my/voice` pass it scoped to the caller's own profile. A preset (relative ref audio, no profile segment) stays visible to all.

**P2 — ref-audio path resolution.** `resolved_ref_path` now expands a leading `~`, uses an absolute `ref_audio` verbatim (the per-profile clones), and resolves a relative one against the (expanded) `models_base_path`, falling back to the `voices.json` parent dir when the base is empty or relative. So the literal `~/.OminiX/models` base the script writes no longer stats a bogus path and drops synthesizable presets.

**Doc nit.** Fixed the `voice.default_voice` → `voice_default` reference in the `voices.rs` module doc.

Tests cover ownership parsing + visibility (`owning_profile_of`, `voice_visible_to`), the registry-level visibility filter, relative-ref resolution against the voices.json dir, and tilde expansion; existing ominix / auth_handlers suites stay green. `build` / `clippy --features api` / `fmt` clean. Branch is already up to date with main.

ymote:
Your `c2555be4` resolves both original findings cleanly (per-tenant scoping via `owning_profile_of` + `voice_visible_to`, and the `~`/registry-dir path resolution). I re-ran the gate — three follow-ups, all edges rather than the core:

**[P1, but low practical exposure] `owning_profile_of` only splits on `/`.** On a Windows deployment where `voices.json` stores a clone as `C:\Users\…\profiles\alice\data\voice_profiles\x.wav`, the `split('/')` (voices.rs:70) returns `None`, so Alice's clone is treated as a shared preset and other tenants can list/select it — the exact cross-tenant leak this PR closes, but on `\`-separated paths. Normalize both separators (or split on `std::path::MAIN_SEPARATOR` / use `Path` components) before the tenant check. (Flagging severity honestly: the voice fleet is macOS/Linux, so this isn't an active leak today — but it's a real portability gap worth closing.)

**[P2] `~` expansion is `$HOME`-only.** `expand_tilde` (ominix.rs:~213) checks only `$HOME`; under a native Windows service `%USERPROFILE%` is set but `HOME` often isn't, so `~/.OminiX/models` stays literal and every relative `ref_audio` check fails (empty `/api/voices`, valid voices rejected). `dirs` is already a workspace dep — `dirs::home_dir()` handles the platform fallback. (octos-llm doesn't currently pull `dirs`; either add it there or read `%USERPROFILE%` as a fallback.)

**[P2] `voice_default` can be cleared by an unrelated profile update.** The choice persists only to `profile.config.voice_default`, but `PUT /api/my/profile` still replaces the whole `ProfileConfig`; an older client sending a config without this new field serde-defaults it to `None` and silently clears the sticky voice (auth_handlers.rs:1246). This is the same class as #1467 — if that PR's config-merge lands it'd preserve this field; otherwise preserve it explicitly on the self-update path.

Nice work overall — the core scoping is right. These are hardening/portability.

alan0x:
Pushed (rebased onto current main). All three follow-ups handled.

**Windows path separators in `owning_profile_of`.** It now splits on both `/` and `\`, so a `C:\…\profiles\alice\…\voice_profiles\x.wav` clone is still recognised as owned rather than treated as a shared preset (which would re-open the cross-tenant leak). Test covers the backslash path.

**`~` expansion was `$HOME`-only.** `expand_tilde` now resolves the home dir via `$HOME` then `%USERPROFILE%` (set on native Windows where `HOME` often isn't), and accepts a `~\` prefix — so a literal `~/.OminiX/models` base resolves there too instead of dropping every relative voice. Split the core into `expand_tilde_with(path, home)` for env-free unit tests.

**`voice_default` cleared by an unrelated profile update.** Went the #1467 route, as discussed: rebased onto current main, which now carries that PR's config-merge, so `update_my_profile` patches the config instead of replacing it wholesale — an omitted `voice_default` keeps its stored value rather than serde-defaulting to `None`.

`cargo fmt --all -- --check`, `cargo check` (default + `--features api`), and the `octos-cli` / `octos-llm` lib suites are green. Ready for another look when you get a chance.

PR #213, opened 2026-03-31
perf: replace Mutex<ModelType> with AtomicU8 in adaptive router

heyong4725:
## Summary
- Replace `Mutex<ModelType>` with `AtomicU8` in `AdaptiveSlot` for lock-free reads in the routing hot path
- Upgrade `Mutex<StatusCallback>` to `RwLock<StatusCallback>` for better read concurrency
- Add `to_u8()`/`from_u8()` helpers on `ModelType` enum

Closes #189

Generated with [Claude Code](https://claude.com/claude-code)

PR #1462, opened 2026-06-14
Add profile-backed home dashboard config

BH3GEI:
## Summary
- add typed `config.home` support to profile config patches
- persist Home settings, calendar events, photo URLs, widget ordering, and Metro tile layout in the profile contract
- mirror the Home config contract in dashboard TypeScript types
- include rustfmt-only cleanup required by current `origin/main` for two existing test blocks

## Why
This is the backend schema companion for octos-web PR octos-org/octos-web#213. Without this schema, the web dashboard can still use local fallback state, but canonical cross-device Home persistence cannot be accepted by `/api/my/profile`.

## Verification
- `cargo fmt --check` → passed
- `cargo test -p octos-cli profile_config_patch --lib` → 9 passed, 0 failed

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol.rs
+++ b/crates/octos-cli/src/api/ui_protocol.rs
@@ -25880,7 +25880,10 @@ ignore = []
             .collect::<Vec<_>>();
         let visible_refs = visible.iter().map(String::as_str).collect::<Vec<_>>();
         let disabled_refs = disabled.iter().map(String::as_str).collect::<Vec<_>>();
-        let deferred_refs = recoverable_deferred.iter().map(String::as_str).collect::<Vec<_>>();
+        let deferred_refs = recoverable_deferred
+            .iter()
+            .map(String::as_str)
+            .collect::<Vec<_>>();
         let payload = coding_tool_contract::tool_status_list_payload(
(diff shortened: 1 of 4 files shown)


ymote:
Thanks @BH3GEI — closing this in favor of #1467, which we're consolidating on.

#1467 covers the same `config.home` persistence and additionally:
- fixes a live partial-PATCH data-loss bug in the `/api/my/*` handlers (the current full `profile.config = config` replace wipes `plugins`/`channels`/`env` on any partial update),
- completes `ProfileConfigPatch` (`plugins`/`lane_routing`),
- and uses an opaque `serde_json::Value` for `home`, which matches what the already-merged octos-web #213 ("Replace mock home state with profile-backed data") was built against — Home stays a web-owned surface with typed validation in the web app.

The typed `HomeConfig` schema here, with `deny_unknown_fields`, would couple every web release to a backend release (any new Home field the web sends would 400 until the Rust struct caught up). Both PRs also add `home` to the same `ProfileConfig`/`ProfileConfigPatch` structs, so they can't co-exist without a duplicate-field compile error.

Really appreciate the work — the schema modeling here was useful for confirming the shape. Let's land it via #1467.

PR #1464, opened 2026-06-15
perf(voice): lean prompt on voice turns — defer tools + cap history

alan0x:
## Problem

Voice-turn LLM latency was dominated by prompt **prefill**. A spoken turn sent the **full tool registry (52 specs)** and the **full conversation history (142 messages)** on every call — ~**34k input tokens**, of which tools and history were each roughly half. Time-to-first-token was ~9s.

The existing LRU tool deferral (target 15 active) never helped here: it only evicts on iteration ≥ 2 after an idle threshold, but conversational voice turns are **single-iteration** (zero tool calls, `EndTurn`). So every spoken turn paid the full tool tax.

## Fix

Two per-turn levers, both gated on a cheap `voice_turn_hint` — derived from the turn carrying an audio attachment (`params.media`), available **before** STT so it can shape the registry and the prompt bridge:

**1. Lean tool set** — defer every tool except `activate_tools` on the **per-turn registry snapshot** (private to the turn; never leaks to other sessions/text turns). Deferred tools stay recoverable via `activate_tools`, so a voice query that genuinely needs a tool pays one extra round-trip instead of every turn paying the full prefill.

**2. Capped history projection** — set `max_prompt_token_estimate` (default **3000**, `OCTOS_VOICE_MAX_PROMPT_TOKENS`) on the **outgoing** prompt projection only. The record / coverage / compaction passes keep the **uncapped** policy, so the persisted transcript is never re-recorded from a trimmed view (no duplication / corruption). Trimming drops oldest-first while preserving any compaction summary and recent turns.

## UX trade-off (intended, short-term)

Both levers fire on **every** audio-attachment turn. So a tool-bearing voice query ("what's the weather") pays one extra `activate_tools` round-trip, and a follow-up spoken turn ("and the one before that?") can lose mid-conversation context once history exceeds the cap. Both are env-tunable (`OCTOS_VOICE_MAX_PROMPT_TOKENS`). This is the accepted trade-off for voice turns for now; the plan is to replace the hard history cap with a Claude-Code-style context-compaction pass for voice turns, so follow-ups keep a compressed view of earlier context instead of dropping it oldest-first.

## Result (measured on a real voice turn)

| | before | after |
|---|---|---|
| tools | 52 | **1** |
| messages | 142 | **4** |
| input_tokens | 34,390 | **~8,800** |
| LLM stage | ~10.2s | **~1.4s** |

## Tests

- `voice_turn_defers_everything_but_the_keep_list`, `voice_turn_defer_is_noop_when_only_keep_tools_present` — pure deferral helper
- `voice_turn_bridge_caps_outgoing_prompt_only` — cap on outgoing policy only
- existing `appui_prompt_context_bridge_preserves_current_user_turn` — green (no regression)
- `cargo build` / `cargo clippy -p octos-cli --features api` clean

### DIFF
--- a/crates/octos-cli/src/api/context_manager.rs
+++ b/crates/octos-cli/src/api/context_manager.rs
@@ -1669,6 +1669,19 @@ fn first_removable_prompt_group(entries: &[PromptMessageEntry]) -> Option<(usize
     let mut end = start + 1;
 
     if matches!(entries[start].message.role, MessageRole::User) {
+        // Never trim away the most recent user turn: it is the current request
+        // the model must answer. When no later user message exists, this group
+        // IS that request (plus any in-turn tool-recovery replies), so it stays
+        // even under a tight cap — `max_prompt_token_estimate` is a soft latency
+        // target, and dropping the live request (e.g. a voice turn whose
+        // protected compaction summary already fills the budget) would leave the
+        // model answering nothing. See `voice_cap_keeps_current_request_*`.
+        let has_later_user = entries[start + 1..]
+            .iter()
+            .any(|entry| matches!(entry.message.role, MessageRole::User));
+        if !has_later_user {
+            return None;
+        }
         while end < entries.len()
             && !entries[end].protected
             && !matches!(entries[end].message.role, MessageRole::User)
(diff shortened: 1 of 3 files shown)


ymote:
The latency win is compelling and the correctness checks out — I confirmed the history cap is applied only to the **outgoing** projection (`outgoing_prompt_policy`), not the record/coverage/compaction passes, so the persisted transcript isn't trimmed or corrupted. Deferring tools doesn't break tool-calling either, since `activate_tools` stays active and the agent can recover a deferred tool in one round-trip.

- **CI**: the red `check` is the same pre-existing `session_actor.rs` rustfmt drift (rebase clears it); the `cli_agent_adapter::times_out_and_kills_child` failure is an unrelated `Text file busy` exec flake in a file this PR doesn't touch.
- **One product question before merge**: defer-all-tools + the 3000-token cap fire on *every* audio-attachment turn. So a tool-bearing voice query ("what's the weather") pays an extra `activate_tools` round-trip, and a follow-up spoken turn ("and the one before that?") can lose mid-conversation context. Both are env-tunable (`OCTOS_VOICE_MAX_PROMPT_TOKENS`), so this is fine if intended — can you confirm that's the UX trade-off you want for voice turns? Worth a sentence in the PR body.

alan0x:
Confirmed — that's the intended trade-off for voice turns, at least for now.

The extra `activate_tools` round-trip on a tool-bearing query is an acceptable cost for dropping the full-tool prefill off every spoken turn, and the history cap is deliberately aggressive to keep TTFT low. Both stay env-tunable, so anyone who wants richer context on voice can raise `OCTOS_VOICE_MAX_PROMPT_TOKENS`.

The follow-up context loss is the part I do want to improve, but not by bumping the cap — I'd rather replace the hard oldest-first trim with a Claude-Code-style compaction pass on voice turns, so earlier context is compressed instead of dropped. That's a separate change; this PR keeps the simple cap.

I've added a sentence to the PR body covering the trade-off, and I'll rebase to clear the rustfmt drift.

ymote:
Pre-merge review (codex gate) surfaced two functional regressions in the new voice-turn paths. I pushed a fix commit (`8f6797ca`) on top:

- **[P1] current request dropped under the voice cap.** After compaction, `for_prompt` keeps the compaction summary as a *protected* entry. With the 3k voice cap, `trim_prompt_entries_preserving_invariants` drained every *unprotected* group — including the current voice transcript — once the protected summary alone neared the cap, leaving the model only system + summary. Fix: trim never drops the most recent user turn (the live request); the cap stays a soft latency target.
- **[P2] all tools deferred with no recovery path.** If a profile's tool surface lacks `activate_tools`, `defer_tools_for_voice_turn` hid every tool with no way to call `activate_tools` back. Fix: defer only when the keep-list recovery tool is actually registered.

Added regression tests for both (`voice_cap_keeps_current_request_even_when_protected_context_fills_budget`, `voice_turn_skips_defer_when_no_recovery_tool_present`). Existing cap/voice tests stay green; clippy/fmt clean.

PR #1468, opened 2026-06-16
feat(llm): Vertex AI auth for Gemini (service-account JSON, keychain-backed)

alan0x:
## Summary
Adds a **Vertex AI** auth path to the Gemini provider so it can be driven by a Google **service-account JSON** instead of an AI Studio API key, and makes it configurable from the dashboard with the private key stored in the macOS Keychain (never in plaintext config).

## What's included
- **`vertex_auth.rs`** — parse SA JSON, sign an RS256 JWT assertion, exchange it for an OAuth2 access token, cache + auto-refresh (`TokenSource`).
- **`gemini.rs`** — `GeminiAuth { ApiKey | Vertex }`; Vertex mode targets `aiplatform.googleapis.com/.../locations/global` with `Authorization: Bearer`. AI Studio path unchanged (regression-tested).
- **`registry/vertex`** — new `vertex` provider (aliases `vertex-ai`/`vertexai`). Credential is the SA JSON *content*, resolved through the `VERTEX_SA_JSON` key channel; project read from the JSON, region fixed to `global`.
- **`update_my_profile`** — relocate a freshly-entered SA JSON out of `env_vars` into the OS keychain, storing only the `keychain:` marker. macOS-only; a raw secret on another OS is rejected, not persisted.
- **`test_provider` / `resolve_saved_key`** — Test Connection resolves the keychain marker and uses a 128-token budget for Vertex (gemini-2.5 thinking models return empty at 16).
- **`keychain::get_secret`** — decode the hex output `security -w` emits for multi-line secrets (the SA JSON newlines), which otherwise corrupted the resolved credential. Ordinary single-line keys pass through untouched.

## Notes
- Pairs with octos-web PR (`feat/vertex-admin-config`) for the settings UI.
- Keychain config is **macOS-only** (by design — no cross-platform fallback).
- New dep: `jsonwebtoken` (RS256 via `ring`, already in tree via rustls — no OpenSSL).

## Test plan
- `cargo test -p octos-llm` (unit: SA parsing, JWT claims, token cache/refresh, endpoint/header per mode).
- `cargo test -p octos-cli --features api` (keychain hex decode, profile keychain relocation).
- Live: `VERTEX_SA_JSON=… VERTEX_TEST_IMAGE=… cargo test -p octos-llm --test vertex_integration -- --ignored`.
- End-to-end verified: configure Vertex in the dashboard → Test Connection OK → chat replies via Gemini/Vertex.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3122,6 +3122,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "jsonwebtoken"
+version = "9.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a87cc7a48537badeae96744432de36f4be2b4a34a05a5ef32e9dd8a1c169dde"
+dependencies = [
+ "base64",
+ "js-sys",
+ "pem",
+ "ring",
+ "serde",
+ "serde_json",
+ "simple_asn1",
+]
+
+[[package]]
 name = "keyring"
 version = "3.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3675,10 +3690,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "num-bigint"
+version = "0.4.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a5e44f723f1133c9deac646763579fdb3ac745e418f2a7af9cd0c431da1f20b9"
+dependencies = [
+ "num-integer",
+ "num-traits",
+]
+
+[[package]]
 name = "num-conv"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf97ec579c3c42f953ef76dbf8d55ac91fb219dde70e49aa4a6b7d74e9919050"
+
+[[package]]
+name = "num-integer"
+version = "0.1.46"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7969661fd2958a5cb096e56c8e1ad0444ac2bbcd0061bd28660485a44879858f"
+dependencies = [
+ "num-traits",
+]
 
 [[package]]
 name = "num-traits"
@@ -3914,6 +3948,7 @@ dependencies = [
  "chrono",
  "eyre",
  "futures",
+ "jsonwebtoken",
  "metrics",
  "octos-core",
  "redb",
@@ -4116,6 +4151,16 @@ dependencies = [
  "postscript",
  "type1-encoding-parser",
  "unicode-normalization",
+]
+
+[[package]]
+name = "pem"
+version = "3.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1d30c53c26bc5b31a98cd02d20f25a7c8567146caf63ed593a9d87b2775291be"
+dependencies = [
+ "base64",
+ "serde_core",
 ]
 
 [[package]]
@@ -5443,6 +5488,18 @@ name = "similar"
 version = "2.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bbbb5d9659141646ae647b42fe094daf6c6192d1620870b449d9557f748b2daa"
+
+[[package]]
+name = "simple_asn1"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d585997b0ac10be3c5ee635f1bab02d512760d14b7c468801ac8a01d9ae5f1d"
+dependencies = [
+ "num-bigint",
+ "num-traits",
+ "thiserror 2.0.18",
+ "time",
+]
 
 [[package]]
 name = "siphasher"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,9 @@ cbc = { version = "0.1", features = ["alloc"] }
 rand = "0.8"
 secrecy = { version = "0.10", features = ["serde"] }
 subtle = "2"
+# JWT signing for Google service-account (Vertex AI) auth. Uses `ring`
+# (already in tree via rustls) for RS256 — no OpenSSL.
+jsonwebtoken = "9"
 
 # Error handling (dora-rs pattern)
 eyre = "0.6"
--- a/crates/octos-cli/src/api/admin.rs
+++ b/crates/octos-cli/src/api/admin.rs
@@ -226,6 +226,38 @@ pub async fn get_profile(
     ))
 }
 
+/// Move any keychain-backed secrets (e.g. the Vertex service-account JSON, which
+/// carries a private key) out of plaintext profile config and into the OS
+/// keychain. Shared by every profile/sub-account create/update save path so they
+/// all uphold the same keychain-only contract as `PUT /api/my/profile`. A raw
+/// secret on a non-macOS host is rejected rather than silently persisted.
+///
+/// Detection is by **content** (a raw service-account JSON), not just the
+/// declared `VERTEX_SA_JSON` name — so a private key pasted under a custom env
+/// var (e.g. a dashboard "Custom" provider deriving `VERTEX_API_KEY`) can't slip
+/// past into plaintext config.
+pub(crate) fn relocate_keychain_backed_secrets(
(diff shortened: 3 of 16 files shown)


ymote:
Ran this through a deeper gate (codex + manual verification of each finding against the diff). The crypto path itself is solid — RS256 via `ring` (no OpenSSL), key held in `SecretString`, correct OAuth grant/audience/3600s expiry, good unit coverage. But there are 5 issues to resolve before merge, four around how the credential is handled *after* it leaves the auth path. I verified each:

1. **[P2 — blocker] SA private key written to disk on admin saves.** `relocate_secret_to_keychain` runs only in `PUT /api/my/profile` (auth_handlers.rs:1131). `create_profile`/`update_profile` in `admin.rs` (230/289) persist `req.config` directly, so an admin creating or editing a profile with a raw `VERTEX_SA_JSON` writes the private key straight into the profile file, bypassing the keychain-only contract. Apply the same relocation in both admin handlers.

2. **[P2 — blocker] SA credential forwarded to plugins as `OPENAI_API_KEY`.** `build_plugin_env` (profile_factory.rs:334-338) maps unknown providers via `_ => "OPENAI_API_KEY"`. With `vertex` active it resolves the Vertex credential and pushes it under `OPENAI_API_KEY` to any plugin that allowlists that var — handing plugins an unusable "OpenAI key" and forwarding a private credential under the wrong name. Vertex needs explicit handling here (or exclusion from plugin key injection).

3. **[P2] Vertex provider reports `provider_name() = "gemini"`.** vertex.rs returns a `GeminiProvider`, whose `provider_name()` is hardcoded `"gemini"` (gemini.rs:404). Routing / adaptive-lane matching / metrics keyed on provider name can't tell `vertex` apart from AI Studio Gemini, so a route configured for `vertex` may not match its slot.

4. **[P2] Token-exchange client is unbounded.** `VertexTokenProvider::from_service_account` builds its fetcher with `reqwest::Client::new()` (vertex_auth.rs:188), which isn't covered by the provider's `with_http_timeout`. If the OAuth token endpoint stalls, a Vertex chat blocks past the configured LLM/connect timeouts before it ever reaches `generateContent`. Thread the timeout into the token client.

5. **[P3] `decode_security_hex` can corrupt unrelated secrets.** It runs for every keychain-backed env var, so a legitimate secret that's even-length ASCII-hex and valid UTF-8 (e.g. `41424344`) is silently decoded to `ABCD`. Scope the hex-decode to the multi-line SA-JSON case or gate it on the key name.

Pre-existing (not introduced here, but compounds with #1 and #2): the keychain *write* uses `security add-generic-password -w <secret>`, which exposes the secret in argv (`ps`). Worth switching `set_secret` to stdin/`-X` in a follow-up now that it carries an RSA private key.

#1 and #2 are the blockers — both put the SA private key somewhere it shouldn't be. Happy to re-review once those are addressed; not merging until then. Thanks for the careful crypto work on the core path.

alan0x:
Pushed 30774772 — all five addressed.

**1 (blocker) — SA key written to disk on admin saves.** `create_profile` and `update_profile` now relocate keychain-backed secrets before persisting, via a shared `relocate_keychain_backed_secrets` helper that wraps the same `relocate_secret_to_keychain` + `KEYCHAIN_BACKED_ENV_VARS` path `PUT /api/my/profile` already uses. A raw secret on a non-macOS host is rejected rather than written in plaintext, same as the my-profile path.

**2 (blocker) — SA credential forwarded as `OPENAI_API_KEY`.** `build_plugin_env` now special-cases `vertex`/`vertex-ai`/`vertexai` and injects nothing — the SA JSON isn't a usable API key and shouldn't leave under any `*_API_KEY` name. The catch-all `OPENAI_API_KEY` mapping stays for openai-compatible providers. Covered by `build_plugin_env_does_not_forward_vertex_sa_json_as_api_key`.

**3 — `provider_name()` was `"gemini"`.** It now matches on the auth mode: `"vertex"` for Vertex, `"gemini"` for AI Studio, so routing / adaptive-lane matching / metrics can tell them apart. Test: `provider_name_distinguishes_vertex_from_studio_gemini`.

**4 — unbounded token client.** Added `VertexTokenProvider::from_service_account_with_timeout`, threaded through `vertex_from_service_account_with_timeout` and `registry::vertex::create`, so the OAuth token-exchange client is built with the same HTTP timeout as the chat client instead of a bare `reqwest::Client::new()`.

**5 — `decode_security_hex` over-reach.** It now only decodes when the decoded text contains a newline — the actual signal that `security` hex-encoded a multi-line secret — so a single-line value like `41424344` is left verbatim instead of being corrupted into `ABCD`. Test: `leaves_single_line_ascii_hex_secret_untouched`.

On the pre-existing `add-generic-password -w <secret>` argv exposure: agreed it's worth fixing now that an RSA private key flows through it. I'll switch `set_secret` to stdin in a separate follow-up so it doesn't widen this PR.

`cargo build` / `clippy --features api` / `fmt` clean; touched test suites green. Left the branch unrebased since CI is green, but happy to rebase onto main if you'd prefer to merge up-to-date.

ymote:
Thanks for the fast turnaround — your fix resolves all five original findings (admin-path relocation ✓, no `OPENAI_API_KEY` forwarding ✓, `provider_name()` distinguishes vertex ✓ with a test, token client now timeout-bound ✓, hex-decode scoped ✓). I re-ran the gate; two issues remain, one introduced by the relocation fix:

**[P1 — blocker] Keychain entries aren't scoped per profile.** `relocate_secret_to_keychain` stores the SA JSON via `set_secret(key, value)` where `key` is the env-var name `VERTEX_SA_JSON` (auth_handlers.rs:1079) — i.e. the same keychain account for every profile. If two profiles/users save different Vertex service accounts, the second `add-generic-password` overwrites the single keychain item, and the first profile then resolves to the *second* profile's private key on its next run. The persisted marker is the generic `keychain:` with no profile in it either. Scope the keychain account name (and the marker) by profile/user — e.g. an account like `VERTEX_SA_JSON::<profile_id>`, resolved symmetrically on read — so saves can't collide.

**[P2] `provider_metadata()` still labels vertex as gemini.** You fixed `provider_name()` to return `"vertex"`, but the adjacent `provider_metadata()` still does `ProviderMetadata::new("gemini", …)` (gemini.rs:442), so response provenance surfaced via `provider_metadata_for_index` mislabels Vertex calls — defeating the distinction you just added. Derive the metadata name from the auth mode too.

P1 is the blocker (cross-profile private-key resolution). Otherwise the credential handling is in good shape now. Happy to re-review.

alan0x:
Pushed (rebased onto current main). Both addressed, plus some hardening the fix surfaced.

**P1 (blocker) — keychain entries weren't scoped per profile.** Secrets now live under a per-profile account `VERTEX_SA_JSON::<profile_id>`, and the stored marker carries that account (`keychain:<account>`); `resolve_value` reads the account back from the marker. Legacy bare `keychain:` markers still resolve to the env-var name, so existing single-profile entries keep working with no migration. Two profiles saving different SAs no longer collide — covered by `two_profiles_get_distinct_keychain_accounts`.

**P2 — `provider_metadata()` labelled vertex as gemini.** It now derives the name from the auth mode (same as `provider_name`), so Vertex provenance/metrics read `vertex`. `provider_name_distinguishes_vertex_from_studio_gemini` asserts both `provider_name()` and `provider_metadata().provider`.

While wiring the per-profile relocation in, I also closed a few related gaps so the keychain-only handling stays consistent across the save paths and env-var names:

- The relocation runs on every profile-config save path via one shared helper: `PUT /api/my/profile`, admin create/update profile, admin create sub-account, self-service create/update sub-account, and the AppUI `profile/llm/upsert` RPC.
- Detection is by **content** (a raw service-account JSON), not just the `VERTEX_SA_JSON` name — so a key pasted under a custom env var (e.g. a dashboard "Custom" provider deriving `VERTEX_API_KEY`) is still relocated/rejected rather than landing in plaintext. A raw secret on a non-macOS host is rejected, not saved.
- Read paths (`Config::get_api_key`, REST `test_provider`, AppUI provider test / fetch-models) resolve the marker via `keychain::resolve_value`.
- `octos auth set-key` applies the same content-aware scoping; `keys`/`remove-key` understand the scoped markers, and `remove-key` won't delete a shared bare account another profile still references.

`cargo fmt --all -- --check`, `cargo check` (default + `--features api`), and the full `octos-cli` / `octos-llm` lib suites are green. Ready for another look when you get a chance.

PR #1476, opened 2026-06-18
feat(voice): live-camera context hint on audio+image turns

alan0x:
Backend half of the voice **camera / video chat** feature (frontend landed in octos-web #233).

When a voice turn carries both audio and a camera frame, inject a short context hint so the model knows it is seeing a *live* camera frame (not an uploaded photo) and can answer about what's in view.

## Notes
- Pairs with the merged frontend (octos-web #233) that sends the per-turn camera frame.
- Independent of the rich-output PR; if both land, merge this first (they touch adjacent voice-turn code).

### DIFF
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -768,8 +768,15 @@ impl Agent {
                 // transcribed into `user_content` by this point), and image
                 // detection looks at the outgoing `media`.
                 let has_image = media.iter().any(|p| octos_llm::vision::is_image(p));
-                let had_audio = TURN_ATTACHMENT_CTX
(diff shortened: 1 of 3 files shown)


ymote:
Merged (`839f4642`). The codex P2 (unreliable `had_audio && has_image` heuristic) is fixed: the video-call note is now driven by an explicit `TurnAttachmentContext.live_video` read from `inbound.metadata.live_video` (never inferred from attachments), so it no longer mis-fires on a voice-note+image turn or depends on the audio that the AppUI voice path strips.

Trade-off (intentional): with the heuristic gone and no ingress setting the flag yet, the note is **dormant** until the signal is wired end-to-end (TurnStartParams + `/chat` metadata mapping + the AppUI `turn/start` path + clients). That's tracked in #1478. CI green (9/9), tests cover the explicit-flag behavior + the voice-note+image regression.

Issue #239, opened 2026-04-01
feat: deploy.sh support for Windows servers


ymote:
## Problem

`deploy.sh` is a bash script with Unix-only assumptions. It cannot deploy to Windows servers.

## Scope

This is lower priority than Linux support (#237) since Windows server deployments are rare for this project. Two approaches:

### Option A: PowerShell deploy script (`deploy.ps1`)
- Mirror `deploy.sh` functionality in PowerShell
- Use Windows Services (via `sc.exe` or `New-Service`) instead of launchd/systemd
- Use `winget` or `choco` for dependency installation
- Use `scp`/`ssh` (OpenSSH is built into Windows 10+)

### Option B: Cross-platform deploy via SSH
- Extend `deploy.sh` to detect Windows remote (via `uname` or `cmd /c ver`)
- Upload binaries via SCP (works with Windows OpenSSH)
- Create Windows Service via remote `sc.exe create` commands
- Skip codesign, use `icacls` for permissions

### Windows service template
```powershell
sc.exe create OctosServe binPath= "C:\octos\bin\octos.exe serve --port 3000" start= auto
sc.exe start OctosServe
```

Or use NSSM (Non-Sucking Service Manager) for better log handling.

## Dependencies
- Depends on #238 (Windows CI + release builds)
- Depends on #237 (Linux deploy.sh — establishes the multi-platform pattern)

ymote:
Current-main closure evidence refresh for PR #1254; not closing this issue manually.

- PR: https://github.com/octos-org/octos/pull/1254 (`Closes #239`)
- Base rehearsed: current `origin/main` / GitHub `main` `f6936c452389c5172496ee0c3e13393956086d92`
- PR head: `561c8d9cccce1002f82ed749acf1ded20131c016`
- Environment: isolated clone `/private/tmp/octos-1254-f693.q9Dz1j/octos`; root dirty checkout left untouched; macOS/Darwin arm64 host.
- Merge rehearsal: `git merge --no-ff --no-commit pr-1254` applied cleanly onto current main.
- Changed files in the rehearsal: `.github/workflows/ci.yml`, `.github/workflows/release.yml`, `README.md`, `scripts/deploy.ps1`, and `scripts/test-windows-deploy.sh`.

Local validation passed:
- `git diff --check origin/main`
- `bash -n scripts/test-windows-deploy.sh scripts/test-cloud-host-deploy.sh`
[...]

Issue #238, opened 2026-04-01
feat: Windows build in CI + release artifacts


ymote:
## Problem

No Windows CI or release builds exist. The codebase has Windows support (`#[cfg(windows)]` guards, `cmd /C` fallback, `taskkill` for process management), but it is never tested or built in CI.

## CI (already added in check-windows job)

```yaml
check-windows:
  runs-on: windows-latest
  steps:
    - uses: actions/checkout@v4
    - uses: dtolnay/rust-toolchain@stable
    - uses: Swatinem/rust-cache@v2
    - name: Build
      run: cargo build --workspace
    - name: Test
      run: cargo test --workspace
```

## Release builds needed

Add `build-windows` job to `release.yml`:
- `runs-on: windows-latest`
- Build `octos.exe` + app-skill binaries
- Produce `octos-bundle-x86_64-pc-windows-msvc.zip`
- Upload as release artifact

## Windows-specific concerns

1. **Sandbox**: Currently `NoSandbox` on Windows. Future work could use Windows Job Objects or AppContainers for isolation.
2. **Shell**: Uses `cmd /C` (existing support in `NoSandbox`)
3. **Process kill**: Uses `taskkill /F /T /PID` (existing support in shell.rs, plugins/tool.rs)
4. **Binary discovery**: Uses `where` instead of `which` (existing support)
5. **Feature flags**: Email (`async-imap`) and browser (`chromiumoxide`) features may need testing on Windows
6. **Path separators**: `\` vs `/` — may surface bugs in path handling

## Install script

`install.sh` is bash-only and won't work on Windows. Options:
- PowerShell install script (`install.ps1`)
- Or document `cargo install --git` as the Windows install method

ymote:
Closing as implemented after audit against origin/main 92a5bc9. Evidence: CI has Windows check/build jobs, release publishes the Windows MSVC bundle, and install.ps1 maps the Windows artifact.

Issue #232, opened 2026-04-01
fix: analyze node prompt lacks output structure guidance


ymote:
## Problem

The analyze node in the pipeline has no built-in prompt guidance in `handler.rs`. It relies entirely on the DOT graph's prompt attribute. The example DOT template says:

> "analyst role prompt. You receive all search findings as input. Cross-reference, identify gaps..."

But doesn't instruct the analyzer to produce structured, concise output that the synthesize node can efficiently consume. This leads to verbose analyze output that inflates the synthesize node's input context.

## Fix

Update the example DOT template to include clearer analyze prompt guidance — structured findings with clear sections, key data points, and sourced claims — so the synthesize node receives well-organized input.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1477, opened 2026-06-18
feat(voice): rich output — illustrated HTML / image in voice turns

alan0x:
Lets a voice turn reply with an interactive visual (illustrated HTML / image /
infographic), not just speech, when the model decides one helps.

## Design (two layers)

**LLM-facing — in-band marker, no tool call.** The model appends a trailing
`[[VISUAL:kind|brief]]` marker after its spoken reply. It emits **no tool call**,
so the Gemini-3 `thought_signature` 400 never arises (voice turns also defer all
tools for a lean prefill). The marker is an internal control protocol and is kept
OFF every model-/client-facing surface:
- stripped from `response.content` + assistant carriers before persist/`done`;
- held back from live `message/delta` (marker-aware `VisibleDeltaFilter`);
- sanitized in the `ContextManager` at the record chokepoint AND at
  snapshot/fork import boundaries, so neither rebuilt nor loaded (pre-fix)
  ledgers replay it into a later prompt; compaction summaries stay clean too.

**Client-facing — typed visual lifecycle.** A structured event stream replaces
marker-scraping: `visual/generating{kind}` on dispatch, `visual/succeeded{kind,
files}` on success, `visual/failed{reason}` on failure/timeout/cancel.
`file/attached` stays a pure artifact-delivery signal, so the lifecycle survives
a future `projection.envelope.v1` cutover.

## Dispatch by kind
- `html` → `octos-agent::rich_output::author_html` writes a self-contained
  interactive document (focused, tool-less authoring call).
- `illustrated` → two-stage: `mofa_image` PNG → inlined as a `data:` URI into
  interactive HTML (single self-contained file).
- `image` / `infographic` → backend-orchestrated mofa skill.

## Governance & correctness
- Background visual runs through `ToolRegistry::execute` (provider policy +
  arg-size limit + deferred-tool auto-activation), registered as a cancelable
  `TaskSupervisor` task (terminal status, full session key so the interrupt path
  finds it), with a 180s timeout and token-usage logging.
- **Camera-frame grounding** (illustrated): the turn's live camera frame is
  forwarded to `mofa_image` as a reference image. Gated on the explicit
  `live_video` signal (#1476), never inferred from attachment types.
- **Unique output filename per dispatch** so the mofa skill's path cache can't
  return a stale image across turns.
- CI runs the `--features api` `voice_turn` tests (the module is feature-gated,
  so the default `--workspace` test never compiled them).

## Related
- Frontend (octos-web): #232, #238, #239 — all merged.
- Built on #1476 (live-camera context hint) — merged.
- `illustrated` / `image` paths require the `mofa_image` / mofa skills installed.

### DIFF
--- a/api/OCTOS_UI_PROTOCOL_V1_SPEC_2026-04-24.md
+++ b/api/OCTOS_UI_PROTOCOL_V1_SPEC_2026-04-24.md
@@ -474,6 +474,17 @@ Projection and session bridging (accepted `UPCR-2026-014`):
 - `file/attached`
 - `session/event`
 
+Voice rich-output visual lifecycle (#1477, ungated; accepted
+`UPCR-2026-024`):
+
+- `visual/generating`, `visual/succeeded`, `visual/failed` — typed
+  lifecycle for a background visual artifact (illustrated HTML / image /
+  infographic) produced by a voice turn. Emitted on the same
+  ledger-backed live path as `file/attached`, but kept distinct from it:
+  `file/attached` stays a pure artifact-delivery signal while these carry
+  the placeholder lifecycle, so the split survives a future
+  `projection.envelope.v1` cutover. See § 8.
+
 Router and queue (Wave4-A):
 
 - `router/status`, `router/failover`, `queue/state`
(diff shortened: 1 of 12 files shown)


ymote:
Reviewed (codex + read) — solid feature: the protocol additions are backward-compatible (new `Visual*Event` types + `visual/*` methods, optional fields), the `[[VISUAL:...]]` marker is stripped before TTS, and it's well-tested (28 markers). CI green, clean vs main.

**One P2 to fix before merge** (projection-envelope compat):
- `crates/octos-cli/src/api/ui_protocol.rs` (~18737-18746): the new sanitized-voice-delta path does a direct `MessageDelta` send, which `send_notification_ephemeral` filters out for connections that negotiated `projection.envelope.v1`, and the `continue` skips `forward_progress_event` — the path that dual-emits the `AssistantDelta` projection envelope. So envelope clients get **no streaming assistant deltas on voice turns**. Route the sanitized `visible` text through the same envelope-emitting path (and the flush path) instead of only the legacy direct-send. (Context: [[stdio projection_envelope]] — envelope vs legacy dual-emit.)

Also a non-blocking nit: since this adds `visual/*` protocol methods, a UPCR doc + the api/ §6 catalog entry would keep the protocol docs in sync (CI doesn't gate it).

Holding the merge for your fix on the P2 rather than pushing over your branch (lesson learned from #1461 🙂). Ping me when updated and I'll re-gate + merge.

ymote:
Pushed a fix for the P2 as #1496 (targets this branch, `feat/voice-rich-output`, so it's yours to merge/cherry-pick — not pushing over your work). It dual-emits the projection envelope for the voice deltas in both the streaming and flush paths. Verified green (api lib 1748, voice_turn 33). Once you take it in, ping me and I'll re-gate + land #1477.

alan0x:
Done — both items are in:

- Took in #1496 onto the branch.
- Added the `visual/*` protocol docs (UPCR-2026-024 + spec §6/§8).

Branch head is `75c42bc4`. Ready for re-gate + merge whenever you have a moment. Thanks!

PR #1497, opened 2026-06-23
perf(voice): reuse a process-wide HTTP client for Volcano cloud TTS

alan0x:
## What

Per-sentence reply synthesis built a fresh `reqwest::Client` on every call, so each sentence paid a new TCP+TLS handshake to bytedance. Share one process-wide client via `OnceLock` so reqwest pools connections (keep-alive) across the sentence-streamed cloud TTS path.

## Why

Cheap, zero-risk latency win on the cloud TTS path: a voice reply is synthesized sentence-by-sentence, and each sentence was opening a brand-new HTTPS connection. Reusing the client removes the per-sentence handshake (tens–hundreds of ms each).

## Scope

Single file (`voice_turn.rs`), 22 lines. No behaviour change, no API change — only connection reuse. Unit test asserts the client is a shared instance (`volcano_client_is_reused_across_calls`).

This is the first, independently-mergeable slice of a larger cloud-TTS latency effort; the remaining streaming work is parked on a separate branch pending a baseline first-audio measurement.

### DIFF
--- a/crates/octos-cli/src/api/voice_turn.rs
+++ b/crates/octos-cli/src/api/voice_turn.rs
@@ -299,6 +299,14 @@ fn volcano_from_env() -> Option<VolcanoTts> {
     })
 }
 
+/// Process-wide HTTP client for Volcano TTS. Reused across per-sentence
+/// syntheses so each sentence doesn't pay a fresh TCP+TLS handshake to
+/// bytedance — reqwest pools connections (keep-alive) on a shared client.
+fn volcano_client() -> &'static reqwest::Client {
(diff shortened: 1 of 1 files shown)


alan0x:
Closing in favor of #1504. The process-wide Volcano TTS client change here is subsumed by that PR (via the voice-latency branch) and has already been reconciled with main's current TTS security handling — HTTPS-only endpoints, the endpoint allowlist, and redirects disabled. This branch predates that reconciliation and now conflicts with main, so there's no need to land it separately. Tracking the shared-client work in #1504 instead.

PR #1499, opened 2026-06-23
feat(api): carry the real per-model context window on cost updates

ymote:
## Why
The octos-tui ctx gauge has no per-model context window to divide by, so it uses a hardcoded **128k** denominator — reading ~100% long before the model's real window fills. That also makes the existing `ContextCompactionCompleted` notice look like it *never fires*: compaction triggers off the model's **real** window (`appui_context_compact_threshold_tokens`), so at a gauge-100% that's really ~50% of a larger window, nothing has compacted yet.

## What
Carry the real window on the wire so clients can render an honest gauge:
- agent `ProgressEvent::CostUpdate` gains `context_window` (emitted from `LlmProvider::context_window()`).
- `event_to_json` adds it to the `cost_update` frame.
- `UiTokenCostUpdate` gains `context_window: Option<u64>`; `map_cost_update` extracts it → `metadata.token_cost.context_window`.

Additive + optional; clients without the field ignore it. **No new compaction mechanism** — the existing `ContextCompactionCompleted` notice already renders; this just makes the gauge honest so 100% coincides with compaction actually firing.

## Consumer
octos-tui PR (octos-org/octos-tui#TBD) consumes `metadata.token_cost.context_window` as the gauge denominator (falls back to 128k until the first cost update).

## Tests
- `event_to_json_cost_update_carries_context_window`
- `ui_protocol_progress_cost_update_carries_context_window_into_token_cost_metadata`
- octos-agent progress (29) + compaction (39) green; api test-compile green; fmt clean.
- clippy: zero findings in touched files (pre-existing workspace clippy debt is unchanged/out of scope).

### DIFF
--- a/crates/octos-agent/src/agent/streaming.rs
+++ b/crates/octos-agent/src/agent/streaming.rs
@@ -483,12 +483,25 @@ impl Agent {
         } else {
             Some(metadata.model.clone())
         };
+        // Resolve the context window from the SAME slot that answered. For the
+        // active slot (`provider_index == None`) use the provider's configured
+        // window (honors any context override); for a failover/routed slot
+        // derive it from the model that actually answered, parallel to how
+        // `model` and `pricing` are resolved above — otherwise the gauge could
+        // report the right model with the primary slot's window.
+        let context_window = match response.provider_index {
+            Some(_) if !metadata.model.is_empty() => {
+                octos_llm::context::context_window_tokens(&metadata.model)
+            }
+            _ => self.llm.context_window(),
+        };
         self.reporter().report(ProgressEvent::CostUpdate {
             session_input_tokens: total_usage.input_tokens,
             session_output_tokens: total_usage.output_tokens,
             response_cost,
             session_cost,
             model,
+            context_window: Some(context_window),
         });
     }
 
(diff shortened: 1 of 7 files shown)


ymote:
## codex review — round 1 (all 3 addressed in `d75105d5`)

- **[P1]** `cargo check --workspace --all-targets` failed: `octos-pipeline/src/handler.rs:1259` also constructs `CostUpdate`. Fixed (`context_window: None` on the test construct). Added `cargo check --workspace --all-targets` to my gate.
- **[P2]** `ChannelStreamReporter` dropped `context_window` via `..`. Now threaded into the RawSse payload (+ test `cost_update_carries_context_window_into_raw_sse_payload`) so channel/web clients get it too, matching the BoundedChannel/`event_to_json` path.
- **[P2]** Failover/routed responses could report the responding model with the **primary** slot's window. Now resolved per slot: active (`provider_index == None`) → `self.llm.context_window()` (honors overrides); failover → `context_window_tokens(metadata.model)`, parallel to how `model`/`pricing` are derived.

Re-gate green: fmt; `cargo check --workspace --all-targets`; octos-agent progress(29)+compaction(39); api `context_window` tests (events/progress/stream_reporter); clippy clean in touched files.

PR #1513, opened 2026-06-30
feat(api): per-tenant voice-pipeline readiness probe

alan0x:
## Summary

Replaces the OminiX-only `voice_models_ready` gate — which hardcoded `qwen3-asr-1.7b` + `qwen3-tts` and required both to be "downloaded" in OminiX-API's catalog — with a per-profile pre-flight that confirms the **whole voice path is usable under the caller's current config**.

The old gate checked a TTS model the pipeline never actually uses (on-device TTS is GPT-SoVITS `engine="sovits"`, cloud TTS is Volcano — never `qwen3-tts`), and ignored both the LLM and the cloud/local TTS distinction. This caused `/voice` to report "Voice models not ready" while voice actually worked.

## What it checks

New endpoint `GET /api/voice/readiness` (authenticated `my_api` group), returning three legs + per-leg detail:

- **ASR** — on-device model ready. Always required; there is no cloud ASR route (only TTS has a cloud option), so ASR is on-device in every mode.
- **LLM** — the profile's provider chain is constructed (running runtime with a non-empty `provider_name`).
- **TTS** — route-aware, mirroring `voice_turn::synthesize_reply`: cloud credentials resolve for `cloud`/`volcano` (and `auto` when configured), otherwise on-device GPT-SoVITS is available. Explicit cloud with missing creds reports **not-ready** instead of silently falling back to on-device, so the chosen route is what gets validated.

## Implementation

- `voice_turn.rs` — pure `classify_tts_route` + `cloud_tts_configured` (unit-tested).
- `ominix_runtime.rs` — pure `asr_ready` (unit-tested).
- `auth_handlers.rs` — `voice_readiness` handler + `VoiceReadiness`/leg types.
- `router.rs` — route wired into the authenticated group.
- Resolves the profile runtime via the shared `resolve_session_profile_runtime` (now `pub(crate)`), which checks **both** `state.profiles` and dynamically bootstrapped runtimes (onboarding / `profile/llm/upsert`), so readiness can't report "not started" while voice turns work. Read-only — no bootstrap side-effects on the polling endpoint.

## Known limitations

- LLM readiness keys off a constructed provider chain (non-empty `provider_name`); it does not verify the API key still resolves, so an expired/revoked key would still report ready. Acceptable for a pre-flight; flag if a live probe is wanted.
- The handler calls `ominix_runtime::runtime_status` (HTTP probe of OminiX + catalog) per poll (frontend polls every 10s).

## Testing

- `cargo test -p octos-cli --features api` — note the `api` feature is **required** or the module compiles out. New: `classify_tts_route` ×3, `asr_ready` ×1.
- `cargo clippy` clean on touched files; `cargo fmt --check` clean.

## Companion PR

Frontend consumer: octos-org/octos-web `feat/voice-pipeline-readiness`.


ymote:
Pushed fe134e720 (plus a merge of current main) addressing the two codex-review correctness findings — both were false-positive `ready: true` paths on the Local TTS leg:

**1. On-device TTS model is now part of the Local leg** (`ominix_runtime.rs`, `auth_handlers.rs`)
New `tts_engine_ready(health_healthy, &runtime.voice_models)` — `role == "tts" && ready`, mirroring `asr_ready`'s shape and unit test — AND'd into the Local leg. Previously a ready ASR + an existing voice with a missing GPT-SoVITS model reported `ready: true` and the real `synthesize_reply` then failed. New detail string: `"On-device TTS model not ready"`.

**2. The leg now checks the profile's EFFECTIVE voice, not any voice** (`auth_handlers.rs`)
`local_tts_voice_available` used to pass if *any* visible synthesizable voice existed. It now takes the effective voice, resolved exactly like a real turn does: `resolve_reply_voice(profile_id, rt.voice.default_voice)` (live `PUT /api/my/voice` override > bootstrapped per-profile default, which already overlays the serve default; when the runtime isn't started — LLM leg already not-ready — it falls back to the persisted `profile.config.voice_default` the bootstrap would apply). A non-empty selection must itself resolve (id or alias) as synthesizable **and** tenant-visible; only an empty effective selection degrades to the old any-visible-voice check, matching the engine-side default an empty voice name gets on a real turn. Detail: `"Selected voice '<id>' not available on device"`.

The registry check is factored into a pure `voice_available_in_registry` core with unit tests: selected voice must itself be synthesizable (the false-positive case), alias resolution, cross-tenant clone invisibility (another tenant's clone can't satisfy readiness), and the empty-selection fallback.

Gates: `cargo fmt --all -- --check` clean; `cargo clippy -p octos-cli --all-targets -- -D warnings` clean; `cargo test -p octos-cli --features api --lib voice` 81 passed; `asr_ready`/`tts_engine_ready`/`classify_tts_route`/`voice_available_in_registry` suites green; `cargo test -p octos-cli --features api --no-run` compiles.

PR #1515, opened 2026-07-01
Fix chat refresh media and tool rehydration

alan0x:
## Summary

- Persist message media in the session messages API response so refreshed chat history can restore uploaded image/video bubbles.
- Add `replayed_tool_envelopes` to `session/hydrate` for negotiated clients so refreshed sessions can rebuild tool-call UI state.
- Keep the new hydrate replay lane additive and capability-gated to avoid changing the legacy wire shape.

## Root Cause

Live chat state received media and tool-call events through websocket notifications, but refreshed sessions rebuilt their UI from persisted message rows and hydrate data that did not expose the same facts. Uploaded media was present in the underlying message rows, but omitted from the API shape. Tool-call state existed in ledger envelopes, but hydrate only replayed spawn-complete envelopes, leaving legacy ThreadStore clients unable to reconstruct tool cards after reload.

## Impact

After refresh, clients can restore:

- user-uploaded media attachments from persisted message rows
- tool start/progress/end UI state from hydrate replay envelopes

Legacy clients remain compatible because empty `media` fields are omitted and `replayed_tool_envelopes` is only returned for negotiated clients that include messages.

## Validation

- `cargo test -p octos-cli --features api session_hydrate_surfaces_replayed_envelopes_for_negotiated_client`
- `cargo test -p octos-cli --features api read_profile_session_messages_returns_bare_key_history`
- `cargo test -p octos-core golden_session_hydrate_result_serde`
- `cargo fmt --all --check`


ymote:
Merged origin/main (5948902e7) into the branch to clear the conflicts from today's main churn — the only textual conflict was the `octos_core::ui_protocol` import list in `crates/octos-cli/src/api/ui_protocol.rs` (this branch's `Envelope` vs main's `ContextCompactionStartedEvent` from #1558; resolved as the union). Also pushed a one-line follow-up (4ff5662) adding the missing `replayed_tool_envelopes: None` in the `session_rollback_command_and_result_round_trip` initializer — that E0063 is what failed this PR's previous `check` run. Verified locally on the merged tree: `cargo fmt --all -- --check` clean, `cargo test -p octos-core` passes, and the full `cargo test -p octos-cli --features api` suite passes (the only two failures under a heavily loaded box were known-flaky main-owned timing tests — session_actor master-continuation tick and the ACP 3s spawn-window test — both pass in isolation and are untouched by this PR). Heads-up unrelated to this PR: workspace clippy currently trips on `field_reassign_with_default` in `commands/profile.rs` that landed with #1569, so the `check` job may stay red until main gets that one-liner.

PR #1520, opened 2026-07-04
Forward Vertex service account env to skills

alan0x:
Summary:
- Forward profile/keychain Vertex service-account env vars to first-party skills under their original VERTEX_SA_JSON name.
- Add regression coverage for Vertex skill env forwarding, including host GOOGLE_APPLICATION_CREDENTIALS isolation.

Verification:
- cargo test -p octos-cli commands::gateway::profile_factory::tests -- --nocapture
- env GOOGLE_APPLICATION_CREDENTIALS=/tmp/fake-service-account.json cargo test -p octos-cli profile_plugin_env_forwards_vertex_service_account_by_its_own_name -- --nocapture
- cargo fmt --all -- --check


ymote:
Pushed 0bc4ab807 (plus a merge of current main) addressing the two codex-review security findings:

**1. Vertex/Google credential forwarding is now provider-gated + always secret-registered** (`profile_factory.rs`)
- `GOOGLE_APPLICATION_CREDENTIALS`, `VERTEX_SA_JSON`, `VERTEX_ACCESS_TOKEN`, `GOOGLE_OAUTH_ACCESS_TOKEN` moved out of the unconditional `FIRST_PARTY_SKILL_ENV_VARS` list into a new `GOOGLE_VERTEX_CREDENTIAL_ENV_VARS` group that `profile_plugin_env` forwards **only when the profile's provider chain (primary or fallback; `family_id` else `detect_provider(model_id)`) resolves to a Google-family provider** (gemini/google/vertex/vertex-ai/vertexai). A profile that merely has Vertex creds in `env_vars` but routes through another provider no longer hands its raw SA JSON to every skill subprocess.
- The four names are also force-registered via `register_secret_env_names` on every `profile_plugin_env` call: `VERTEX_SA_JSON` doesn't trip the `is_secret_env_name` heuristic, and the provider-build-time registration in `Config::resolve_api_key` only fires when Vertex is the *active* provider — so the SA JSON was previously echo-able from empty-allowlist subprocess envs (shell/hooks) whenever Vertex wasn't active.
- PR goal preserved: a Vertex-primary profile still gets the SA JSON under its own name (test updated), and a Google-family *fallback* in the chain also qualifies (new test). New negative test proves a deepseek/openai-chained profile leaks none of the four names. Non-secret Google config (`GOOGLE_CLOUD_PROJECT/LOCATION`, `VERTEX_BASE_URL`) still forwards unconditionally.

**2. `workspace_root` is now host-authoritative** (`plugins/tool.rs`)
- When the manifest opts in via `x-octos-host-config-keys: ["workspace_root"]`, the host-computed value **always overwrites** a caller-supplied one (with a `tracing::warn` when they differ), and a caller-supplied value is **stripped** when the host cannot compute one. The old `!obj.contains_key("workspace_root")` gate let a spoofed tool call point the plugin at a workspace outside the session root.
- `prepare_effective_args_does_not_overwrite_explicit_workspace_root` renamed/inverted to `prepare_effective_args_overwrites_caller_supplied_workspace_root_with_host_value`, plus a new strip-case test.

Gates: `cargo fmt --all -- --check` clean; `cargo clippy -p octos-agent -p octos-cli --all-targets -- -D warnings` clean; `octos-agent plugins::tool` 119 passed; `octos-cli profile_factory` 8 passed; `cargo test -p octos-cli --features api --no-run` compiles.

PR #1555, opened 2026-07-08
fix(voice): handle empty speech turns

alan0x:
## Summary
- emit voice_transcript progress before agent/TTS so clients can show ASR early
- emit voice_no_speech and complete empty/noise audio turns without entering the agent
- persist voice user rows using the real transcript instead of prompt scaffolding
- stop treating non-image media as [User sent an image]
- raise voice prompt context budget to keep short-term voice memory usable

## Tests
- cargo test -p octos-agent should_not_call_empty_non_image_media_an_image
- cargo test -p octos-cli --features api voice_user_message_persist_uses_transcript_not_prompt_scaffolding


ymote:
Pre-merge codex review flagged two issues, fixed in b8bf82fd8: (1) the `voice_no_speech` early return fired for ANY media-bearing turn with no transcript — `asr_media` holds all materialized media, so a text+image turn was silently completed without running the agent; the gate is now a pure `should_short_circuit_no_speech()` requiring audio-present + no transcript + empty typed prompt + no non-audio media. (2) mixed typed-text+audio turns persisted transcript-only history, erasing the typed text; persist now captures the combined pre-scaffold user content. Unit tests added for both.

PR #1564, opened 2026-07-09
feat(api): add ACP WebSocket endpoint with JSON-RPC 2.0 support

alanpoon:
## Summary

- Adds a new `/acp` WebSocket endpoint implementing the **Agent Communication Protocol (ACP)** over JSON-RPC 2.0
- Introduces `acp_websocket.rs` with a `WebSocketStreamReporter` that forwards `ProgressEvent` stream chunks as `session/update` notifications to connected clients
- Registers the route in the existing API router alongside the `session_ingress` WebSocket handler
- Moves the `url` crate dependency to the shared workspace dependencies section in `Cargo.toml`

## Test plan

- [ ] `cargo build --workspace` passes
- [ ] `cargo clippy --workspace` passes
- [ ] Connect a WebSocket client to `ws://localhost:<port>/acp`, send a JSON-RPC 2.0 `tasks/send` request, verify `session/update` streaming notifications are received
- [ ] Verify that invalid/malformed JSON-RPC messages return a proper JSON-RPC error response (code -32700 / -32600)
- [ ] Verify that existing `/v1/session_ingress/ws/{session_id}` endpoint is unaffected


alanpoon:
Addressed all review comments in dcb39390:

**Critical (build failure)**
- Removed `use super::acp` (module doesn't exist — only `acp_websocket` does). Route now uses the directly-imported `acp_websocket_handler`. This fixes the `--features api` compile error.

**Security**
- `/acp` is no longer in the unauthenticated `public` router. It now follows the same `user_auth_middleware` conditional pattern as `/metrics` — protected when `auth_token`/`auth_manager` is configured, open when auth is disabled.

**Protocol**
- Invalid/unparseable JSON now returns a proper JSON-RPC 2.0 `-32700 Parse Error` response instead of a silent `warn!()`.

**Architecture**
- `SessionKey` is now generated once per WebSocket connection (`Uuid::new_v4()` in `handle_acp_socket`) and threaded down to `handle_chat_request`, so concurrent connections no longer share session history.
- `recv_task` now breaks the loop when a channel send fails (send_task has exited → connection is dead).

**Code quality**
- All JSON-RPC types narrowed from `pub` to `pub(crate)`.
- Removed unused `tokio-tungstenite` dep and redundant `futures-util = "0.3"` dep (the latter is already re-exported by the workspace `futures` crate).

ymote:
Hey @alanpoon — thanks a lot for this, and apologies for the collision! We were both building ACP support for octos at the same time — your first commit actually predates ours by a couple of hours. The stdio side just merged as **`octos acp`** in #1563: octos as a real [Agent Client Protocol](https://agentclientprotocol.com) agent over stdin/stdout, which is the transport Zed and other ACP editors actually spawn. A few notes to reconcile before you invest more here:

**Real ACP handshake/methods.** ACP clients drive `initialize` → `session/new` → `session/prompt` (+ `session/cancel`), with the agent side built on the `agent-client-protocol` crate's `Agent` builder. This PR's `chat` / `user_input` / `status` / `list_sessions` are a custom JSON-RPC surface, so a real ACP client (Zed) can't actually connect. If the goal is genuine **ACP-over-WebSocket**, the cleanest path is to reuse the handlers we already have in `crates/octos-cli/src/commands/acp.rs` — the `initialize`/`session/new`/`session/prompt`/`cancel` dispatch plus `AcpProgressReporter`, which maps `ProgressEvent` → `session/update` for streamed text, thinking, **and** tool calls — behind a WebSocket transport adapter, rather than a parallel protocol. That gets you real client interop and the full event mapping (today `WebSocketStreamReporter` only forwards `StreamChunk`, so tool/thinking events and non-streaming providers' output are dropped).

**Overlap with UI Protocol.** If the intent is more "drive an octos agent from a browser/remote client over WebSocket," octos already has that: **UI Protocol v1** over `/api/ui-protocol/ws` (used by octos-web and octos-tui) — it covers streamed turns, tool activity, approvals, rollback, and cursor replay. A second WS agent protocol on the same server would overlap it.

Two small things regardless of direction: `state.profiles.values().next()` picks an arbitrary profile (no tenant selection), and the JSON-RPC `id` should allow numeric/null (real ACP uses numeric ids).

Really appreciate the initiative — could you share which use case you were aiming for (a non-stdio ACP client? a remote WS client?)? Happy to help wire the reuse if there's a genuine ACP-over-WS need. 🙏

anantheparty (changes requested):
Blocking this as an ACP implementation in its current form. The endpoint does not implement the Agent Client Protocol method/state machine (`initialize` → `session/new` → `session/prompt`, plus `session/cancel`); it exposes a separate `chat` / `user_input` / `status` / `list_sessions` JSON-RPC API, so standard ACP clients cannot interoperate with it. The canonical ACP execution and event-mapping path is already merged in #1563 (`crates/octos-cli/src/commands/acp.rs`).

Please first clarify the intended transport/use case. If this is ACP-over-WebSocket, adapt the existing ACP handlers and full progress mapping rather than creating a parallel protocol. If this is a browser/remote Octos transport, the existing UI Protocol endpoint is the appropriate path. The current implementation also only forwards `StreamChunk`, so thinking/tool events and non-streaming output are lost, and it selects an arbitrary first profile.

Because the branch is also conflict-dirty against current `main`, I recommend closing it unless a concrete ACP-over-WebSocket consumer is identified; otherwise rebase and rework it around the merged ACP implementation before requesting review again.

anantheparty:
2026-08-15 巡检结论：与规范 ACP 实现（#1563 已合）重复且不兼容（无 ACP 状态机、事件映射残缺、零测试、未认证暴露面）。除非有明确的 ACP-over-WebSocket 消费者，否则关闭；浏览器/远程传输场景应走既有 UI Protocol 端点。关闭处理。

PR #1250, opened 2026-05-25
feat(spawn): add worker git worktree isolation

ymote:
## Summary
- refresh #294 closure branch onto current `origin/main` `f6936c452389c5172496ee0c3e13393956086d92`
- add `isolation: "worktree"` to builtin spawn calls, allocating each worker under `.octos/work/<agent_id>` on a fresh `octos/worker/<agent_id>` branch
- rebind child tool registries, task CWD, plugin work dirs, and inherited session scopes to the worker worktree while preserving parent scope policy
- write worker worktree status tombstones and teach `octos clean` to remove orphaned `.octos/work/*` directories
- keep current-main turn-ledger behavior while applying the spawn worktree isolation changes

Closes #294

## Current-main refresh
- Base: `f6936c452389c5172496ee0c3e13393956086d92`.
- Head: `8439df924fd6845f68ff165d5509d26ef38a0c99`.
- Isolated checkout: `/Users/yuechen/home/octos/target/octos-1250-fresh.epPHQW/octos`.
- Environment: Darwin arm64; `rustc 1.95.0`; `cargo 1.95.0`.
- Refresh conflict resolved only in `crates/octos-agent/src/agent/loop_runner.rs`, preserving current-main turn-ledger recording and applying the worktree isolation slice.
- Final refreshed diff is limited to `crates/octos-agent/src/tools/spawn.rs`, `crates/octos-cli/src/commands/clean.rs`, and `crates/octos-core/src/session_scope.rs`.

## Validation
- `git status --short --branch` showed a clean isolated checkout at push time.
- `git ls-files --others --exclude-standard` was empty at push time.
- `git diff --check origin/main...HEAD` passed.
- `cargo fmt --all -- --check` passed.
- `typos crates/octos-agent/src/tools/spawn.rs crates/octos-cli/src/commands/clean.rs crates/octos-core/src/session_scope.rs` passed.
- `CARGO_TARGET_DIR=/private/tmp/octos-1250-f693-target CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo test -p octos-core with_workspace -- --nocapture` passed: 2 tests.
- `CARGO_TARGET_DIR=/private/tmp/octos-1250-f693-target CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo test -p octos-agent worker_worktree_slug_validation -- --nocapture` passed: 1 matching test.
[...]

PR #1577, opened 2026-07-09
fix(profile): clippy field_reassign_with_default in test helper (unbreak main check)

ymote:
main's `check` job (Linux clippy `--all-targets -D warnings`) fails at `crates/octos-cli/src/commands/profile.rs:300` — the `config_with_provider` test helper from the #1569 squash trips `field_reassign_with_default`. The squash landed before its refreshed CI run completed, so this was never caught on the PR (#1574 merged on top the same way; its branch CI failed with the same single inherited error + the known flaky ACP cancel test, nothing new).

Every open PR's `check` job fails against current main until this lands — proof: #1250's fresh run (main-merged tree, unrelated spawn-only diff) fails at exactly this line.

Fix: struct init with `..Config::default()`. Test-code only.

### DIFF
--- a/crates/octos-cli/src/commands/profile.rs
+++ b/crates/octos-cli/src/commands/profile.rs
@@ -297,10 +297,12 @@ mod tests {
     use super::*;
 
     fn config_with_provider() -> Config {
-        let mut config = Config::default();
-        config.provider = Some("deepseek".into());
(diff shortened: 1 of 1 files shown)


ymote:
Superseded: #1549's merge included the identical struct-init fix for the `config_with_provider` helper (main tip 44ae43108 verified fixed). Closing without merge.

Issue #1602, opened 2026-07-09
[bug] mcp_agent backend robustness: HTTP total-timeout/body-cap, stdio grandchild reap, stderr leak


ymote:
From the 2026-07-09 deep review of `crates/octos-agent/src/tools/mcp_agent.rs` (the dispatch backend for swarm + spawn). NOTE: the generic MCP *client* (`mcp.rs`) was migrated to rmcp in #1584 — this is the separate agent-dispatch backend, still hand-rolled.

- **HTTP `read_timeout` doesn't bound total body; no body-size cap (P2).** `HttpMcpAgent` sets `.read_timeout(...)` (per-read, not whole-response), so a slow-drip remote (1 byte per <10s) keeps `resp.text().await` alive until the outer `dispatch_timeout` (180s default). With swarm fan-out that's `concurrency` sockets pinned ~180s each, plus retries. No `MAX` body cap either (unlike the stdio path's `MAX_LINE_BYTES`) → unbounded buffering. Fix: add a whole-request `.timeout()` and cap the body via `content_length()` / bounded reader.
- **stdio child kill on timeout doesn't reap grandchildren (P2).** On timeout `ChildGuard::drop` calls `start_kill()` (fire-and-forget, no `wait()`); reaping relies on `kill_on_drop`. Only the *direct* child is signaled — a sub-agent that spawned its own tool subprocesses (e.g. shells) leaves **grandchildren** orphaned (no process-group kill). The module doc ('enforces an explicit kill … so the child is reaped') overstates this. Fix (Unix): `process_group(0)` + `killpg` on timeout, or keep the handle and `.wait()` after `start_kill()`.
- **`stderr(Stdio::inherit())` leaks sub-agent stderr into the parent (P2).** For a backend named 'MCP over stdio', an untrusted sub-agent writes directly to octos's stderr — leaks internal output/secrets into operator logs and a chatty child can flood the log pipeline. Fix: `Stdio::piped()` + bounded drain, or `Stdio::null()`.

Minor (P3): `endpoint_label()` may surface URL userinfo into events; `allow_loopback_for_tests` is a runtime `pub` bool (not `#[cfg(test)]`-gated) that, if flipped, opens SSRF. Verified non-issues: gate fail-closed, redirect `Policy::none()` + explicit 3xx-reject, no unwrap on remote data, cost/validators locally derived (no remote injection).

ymote:
[...]

Issue #1597, opened 2026-07-09
[security][P0] octos mcp-serve exposes an unsandboxed, un-approved agent (RCE)


ymote:
From the 2026-07-09 swarm+MCP deep review. **Live-proven RCE.**

## Problem
`octos mcp-serve`'s `run_octos_session` builds its inner agent registry with `ToolRegistry::with_builtins(&self.config.cwd)` (`crates/octos-cli/src/commands/mcp_serve.rs:322`), which resolves to `with_builtins_and_sandbox(cwd, Box::new(NoSandbox))` + `EffectivePermissions::workspace_write()` (`ApprovalPolicy::Ask`). So the MCP-exposed agent runs `shell`/`bash`/`write_file`/`spawn`:
- with **no sandbox** (raw `sh -c` on the host),
- with **no config `tools`/`tools.byProvider` deny-list**,
- with `ApprovalPolicy::Ask` but **no interactive approver** in server mode (so `Decision::Ask` commands are undefined-behavior; `SafePolicy` is explicitly *not* a security boundary).

Contrast the chat path (`chat.rs`), which uses `create_sandbox(&config.sandbox)` + the profile's `ToolPolicy`.

## Proof (live)
An external MCP client called `run_octos_session` with an attacker-controlled prompt; the inner agent ran `bash` and wrote a file **outside** the `--cwd` workspace (`PWNED-BY-MCP`). Over `--transport http` the only barrier is `OCTOS_MCP_SERVER_TOKEN` (weak tokens accepted with no warning; `--bind 0.0.0.0` allowed) → token leak = host RCE with the operator's configured sandbox silently disabled. Shell commands are also **not** confined by `FilesystemScope` (review suspicion S-1, confirmed by the out-of-cwd write).

## Fix direction
- Build the MCP registry with `with_builtins_and_sandbox(&cwd, create_sandbox(&config.sandbox))` and apply the same `ToolPolicy`/profile tool-filter the chat path uses — the MCP boundary must be **≥** as restrictive as local dispatch, not looser.
- Set `ApprovalPolicy::Never` (fail-closed) + an explicit allow-list on the server path, or wire approvals into the MCP protocol.
- Enforce a minimum token entropy and warn on `0.0.0.0` binds.

Do not expose the HTTP server beyond a trusted-parent stdio pipe until fixed.

ymote:
[...]

PR #1605, opened 2026-07-10
fix(security): harden the mcp-serve sandbox (fail-closed, tool policy, sandboxed validators)

ymote:
## Summary

Hardens the sandbox fix **#1604** landed for the `run_octos_session` RCE (**#1597**). #1604 confines the MCP-served agent's tools via `with_builtins_and_sandbox(Auto)` — the core fix — but an external MCP caller is only parent-trusted (stdio) or bearer-token authenticated (http), never fully trusted like a local chat user, and several gaps remained. This PR closes them.

> Rebased onto #1604; this is a pure hardening delta on top of it.

## What this adds on top of #1604

- **Fail closed on the mcp-serve path.** `Auto` falls back to `NoSandbox` when a host has no backend (e.g. Linux without `bwrap`, or Windows without the `octos-sandbox` helper) — which leaves the RCE open there. Unlike chat/gateway (where the human runs their own commands), the mcp-serve session now **refuses to start** when a sandbox was requested but no backend is available. An explicit `sandbox.mode = "none"` opt-out is respected. *(Deliberate divergence from chat/gateway — flagged here for review.)*
- **Fail-closed approval policy.** Tools run under `EffectivePermissions::workspace_write().with_approval_policy(Never)`. There is no interactive approver in server mode, so any tool call that would prompt fails at the boundary instead of silently proceeding (parity with local chat).
- **Honour operator tool policy.** The global `tool_policy` and the per-provider policy are applied to the MCP registry, so a server that denies `shell`/`bash` locally doesn't re-expose them to an external caller. The per-provider policy is resolved **per session against the built provider's `model_id()`**, so a policy keyed to a provider's *default* model is honoured.
- **Sandbox completion-command validators.** #1604 left `ValidatorRunner` running workspace command validators **unsandboxed**. They now run *through* the sandbox (`sh -c` + `shlex` argv quoting on POSIX). No-op sandboxes and Windows (`cmd /C`, which ignores POSIX quoting) run the argv **directly** via `Command`, so a caller-influenced argument cannot inject shell metacharacters.
- **Windows fail-closed.** `AppContainerSandbox::is_noop()` reports `true` when the `octos-sandbox` helper is unavailable (its `wrap_command` otherwise falls back to unsandboxed `cmd /C`), so the fail-closed check refuses the session — mirroring the Linux Landlock backend.

## Tests

- New `is_noop` contract unit tests (NoSandbox + disabled/none modes → no-op) — the fail-closed check depends on this.
- Scripted mcp-serve dispatch tests opt out of the sandbox (`mode = none`) so the no-backend fail-closed check is host-independent; the shell-escape regression (`should_block_shell_write_outside_workspace_via_sandbox`) opts into a real backend and self-skips when none is available.
- `octos-agent` sandbox + validators suites, `mcp_serve` unit + integration suites all green; `cargo build` / `clippy --all-targets -D warnings` / `fmt` clean.

## Review

Hardened across four `codex review` rounds (fail-closed sandbox → apply tool policy → sandbox validators → argv quoting → per-provider policy → Windows AppContainer + provider-default model). One of the swarm/mcp-serve findings tracked in #1597–#1602.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4293,6 +4293,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "shlex",
  "similar",
  "symphonia",
  "tar",
--- a/crates/octos-agent/Cargo.toml
+++ b/crates/octos-agent/Cargo.toml
@@ -26,6 +26,9 @@ tracing = { workspace = true }
 metrics = { workspace = true }
 glob = { workspace = true }
 globset = { workspace = true }
+# Shell-quote validator argv when wrapping command validators for a POSIX
+# sandbox (so `sh -c` reconstructs the exact argv rather than word-splitting).
+shlex = "1"
 which = { workspace = true }
 regex = { workspace = true }
 ignore = { workspace = true }
--- a/crates/octos-agent/src/sandbox/mod.rs
+++ b/crates/octos-agent/src/sandbox/mod.rs
@@ -221,12 +221,24 @@ pub enum SandboxMode {
 pub trait Sandbox: Send + Sync {
     /// Wrap a shell command string into a sandboxed `Command`.
     fn wrap_command(&self, shell_command: &str, cwd: &Path) -> Command;
+
+    /// Whether this sandbox provides no confinement (runs commands directly).
+    /// Lets callers that require confinement (e.g. the `mcp-serve` server path)
+    /// fail closed when `SandboxMode::Auto` resolves to no backend. Real
+    /// backends inherit the default `false`.
+    fn is_noop(&self) -> bool {
+        false
+    }
 }
 
 /// No-op sandbox: executes commands directly.
 pub struct NoSandbox;
 
 impl Sandbox for NoSandbox {
+    fn is_noop(&self) -> bool {
+        true
+    }
+
     fn wrap_command(&self, shell_command: &str, cwd: &Path) -> Command {
         #[cfg(windows)]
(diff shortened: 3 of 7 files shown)


ymote:
**codex round 5** (on the reconciled delta) surfaced the validator subsystem's three dispatch paths:

- **[P1] Windows command validators bypassing AppContainer** — *fixed* in c341a7b08: a real (non-no-op) sandbox on Windows now fails command validators **closed** rather than falling to direct-argv (which would bypass the AppContainer, since a policy can name an arbitrary executable). POSIX keeps `sh -c` + `shlex`; no-op / no-sandbox still run argv directly.
- **[P1] project-root command validators run unsandboxed** + **[P2] project-root ToolCall validators bypass per-provider policy** — *deferred to #1607*. These are the **Agent-internal** `run_project_root_validators` path (`build_validator_runner` / `MapToolDispatcher`), shared by chat/gateway/spawn/mcp-serve. The clean fix is a `sandbox` field on `ToolRegistry` threaded through the single `build_validator_runner` chokepoint — a cross-cutting change that alters project-root validator behavior on every path (slides/sites generation), so it needs a fleet-level test sweep and its own PR rather than riding along here.

The **primary #1597 RCE** (MCP caller drives octos tools unsandboxed) is closed by #1604 + this PR's registry-level sandboxing; #1607 covers the narrower secondary escape via project-root validators.

Issue #1594, opened 2026-07-09
Security: session-ingress work-secret bypasses scope on the raw RPC surface (cross-session + profile mutation)


ymote:
## Summary

A session-ingress **work-secret** — issued for and scoped to a single session via `/v1/session_ingress/ws/{session_id}` — can execute **raw** UI-Protocol RPCs for a *different* session, and can perform **profile-level mutations**, because the raw dispatch path runs **before** scope validation.

## Root cause

In `ui_protocol_connection` (`crates/octos-cli/src/api/ui_protocol.rs`), the per-frame order is:

1. `handle_client_hello_rpc(...)`
2. **`handle_raw_appui_rpc(...)`**  ← dispatches & executes raw methods, returns `true` (handled)
3. `route_rpc_command(...)`
4. **`validate_session_ingress_command_scope(&command, &scope.session_id)`**  ← the scope gate

The scope gate at step 4 only runs for *typed* commands produced by `route_rpc_command`. Raw methods handled at step 2 (`session/goal/set`, `loop/create`, `profile/llm/upsert`, `profile/llm/delete`, `agent/*`, `review/start`, …) **never reach it**.

Worse, the raw profile handlers resolve the target profile from **request params** independent of the credential: `raw_profile_id()` returns `params.profile_id` first, so `raw_profile_llm_upsert` writes to whatever `profile_id` the caller names (no equality check against `connection_profile_id`).

## Impact

A holder of a work-secret scoped to session *A* (profile *P*) can, over the same socket:

- **Cross-session autonomy**: `session/goal/set` / `loop/create` / `agent/interrupt` targeting session *B*.
- **Profile mutation / secret write**: `profile/llm/upsert` with `params.profile_id` = any profile — persisting an LLM route + API-key env var into a profile the credential was never scoped to.

The socket is already work-secret authenticated (revalidated per frame), so this is a **privilege escalation within an authenticated but session-scoped credential**, not an unauthenticated RCE. The doc comment on `ws_handler_for_session_ingress` claims the dispatcher "applies a session-scope gate before executing any RPC" — which is only true for the typed surface.

## Fix (PR attached)

Enforce ingress scope on the raw surface **before** `handle_raw_appui_rpc`, only for raw-dispatched methods (typed methods keep flowing to `validate_session_ingress_command_scope`):

[...]

PR #1609, opened 2026-07-10
feat(appui): session/btw — quick aside question answered while the turn runs

ymote:
## What

New UI-protocol method **`session/btw`** — the server half of the TUI `/btw` feature: the user asks a quick aside question **while the session's live turn keeps running**, and the server answers out-of-band with one restricted LLM call.

```
/btw what are you working on

  ✽ Answering…
```

## Design

- **No tools, by construction**: the provider gets an empty tool-spec list + `tool_choice=None`.
- **Bounded**: transcript tail (20 messages / 16KB), a live-activity digest from the ledger replay tail (tool/task/turn lifecycle), `max_tokens=700`, 30s timeout.
- **Ephemeral**: nothing is appended to session history and no notification is emitted — the live turn never sees the aside; the answer rides only on the RPC result.
- **One in-flight aside per session** (`btw_busy` rejection), released via a Drop guard on every exit path.
- **Mid-turn concurrency is the existing dispatch model** — same path that serves `approval/respond` / `task/cancel` during turns. Provider = the profile's shared chain (`ProfileRuntime.llm`), so retry/failover/adaptive routing apply.
- **Session-ingress scope**: classified session-scoped (callable with a matching ingress credential), consistent with #1594/#1596.
- Advertised through `APPUI_EXTRA_METHODS`, so clients capability-gate their `/btw` command; old servers simply don't advertise it.

## Testing

- 7 new tests: prompt shape + no-tools restriction, ingress scope accept/reject, capability advertisement, empty-question reject, unknown-session reject, busy rejection, full happy path through a stub provider (asserts empty tool list, question + transcript reach the provider, in-flight slot released).
- The `advertised_stdio_methods_have_dispatch_path` drift guard caught the missing probe fixture — added, so stdio dispatch parity is pinned.
- `cargo test -p octos-cli --features api --lib`: 2226 passed / 0 failed. fmt clean; clippy warning count identical to origin/main baseline (48; none in new code).

Client half (octos-tui `/btw` command + `✽ Answering…` UI) follows as a separate PR gated on this capability.


ymote:
**Codex round 1** returned 3 P1 + 4 P2 — all verified real, fixed in 752edccc0 + aa326c9c3:

| Finding | Fix |
|---|---|
| P1: handler awaited inline → 30s provider call blocked the connection's read loop (interrupts/approvals stalled; busy gate never raced) | validation + transcript snapshot stay synchronous; the provider call runs in a detached task with the in-flight guard moved in |
| P1: topic not folded → a `#topic`-scoped ingress credential could read the bare session's transcript | `session_key_with_optional_topic` applied before scope validation and every lookup/guard |
| P1: profile precedence reversed vs `resolve_sessions_for_lookup` + missing session-open fallback → transcript and provider could come from different profiles | ONE resolution (session key → connection → routed, ws arm threads `session_open_profile_id` through routed) shared by both |
| P2: missing from canonical core inventories | added to `UI_PROTOCOL_COMMAND_METHODS` + `UI_PROTOCOL_FIRST_SERVER_METHODS` + the golden v1 wire-contract fixtures (the goldens firing was the guard working) |
| P2: paid asides invisible to usage analytics | `UsageEvent::completed_run` on the aside-specific `appui_btw` channel, catalog-priced |
| P2: model reported from the chain, not the answering slot | `provider_metadata_for_index(response.provider_index)` |
| P2: in-flight lock keyed by bare session | keyed by (profile, session) |

Note for the web repo: `validateNegotiatedCapabilities`' route inventory (octos-web) will need `session/btw` when the web client picks this up — out of scope here.

8 btw tests (+ topic-folding regression); test provider slot serialized (two slot users raced). octos-core 304 green, octos-cli 2227 green. Round 2 running.

ymote:
**Codex round 2** (on 752edccc0) found 1 P1 + 1 P2, both verified and fixed in ec48ff89a — plus a live-soak finding fixed in the same commit:

- **P1: detached asides outlived the connection** — an orphaned aside held a `WsConnection` writer sender, blocking stdio EOF shutdown up to 30s while paid provider work kept running. The handler now returns its `JoinHandle`; both dispatch loops register it and teardown aborts+awaits (abort drops the future → the guard's `Drop` frees the busy slot).
- **P2: busy key `""` vs `_main`** — canonicalized to `MAIN_PROFILE_ID`, matching sessions lookup and runtime resolution.
- **Live soak:** a mid-turn "what are you working on?" answered from the *committed* transcript, not the answer being written — `message/delta` is ephemeral per spec §9 and never ledgered. The aside context now includes a **live-draft tail** fed at the ephemeral send choke point (1.2KB/session, reset at every turn lifecycle edge, 128-session crash-leak cap). Verified live against a real provider: the aside described the in-flight turn's topic *and its current section*.

(The small golden-fixture commit aa326c9c3 was separately reviewed clean.) 2227 lib tests green. Round 3 running on the new commit.

ymote:
**Codex round 3** (on ec48ff89a) found 1 P1 + 1 P2 on the new live-draft store, both verified and fixed in 8d89d2d4d:

- **P1: cross-profile draft leakage** — the store now keys by (profile, session) and is fed from the mapped-progress path (`forward_progress_event`), where the turn's resolved profile is in scope via `ProgressMappingContext::with_draft_profile(usage_profile_id)`. The profile-blind feed at the ephemeral choke point is gone; two profiles on the same bare session id can no longer read each other's stream.
- **P2: resets were dead code for regular turns** (they emit via `send_notification_lifecycle`, not `_durable`) — the reset moved to the real turn start beside the context construction, and instead of chasing every terminal emission site, the **aside read is gated on the session having a live entry in the active-turns registry**: a finished turn's leftover tail can never masquerade as in-flight.

2 new tests (profile-scoped isolation; draft rides the prompt only while a turn is live, via an active-turns fixture). 2229 lib tests green.

**Full live checklist re-verified after the redesign** (real deepseek provider, tmux): mid-turn `/btw` answers with the in-flight turn's topic, busy rejection, card persists past settle, next-prompt dismissal. Round 4 (closure) running.

ymote:
**Codex round 4** (on 8d89d2d4d) found 2 P1 + 1 P2 — all holes in round 3's live-draft gate/reset/feed. Per the treadmill rule that was a redesign signal, not a patch signal: the store is now **TurnId-keyed** (fed once at the ephemeral send choke point where every `message/delta` — including voice/review/fixture direct sends — passes with its turn_id; no resets anywhere; reads resolve the registry's **non-terminal** turn and read exactly that turn's tail). This eliminates the hole class: terminal-retained entries can't pass the gate, the admission window reads the new turn's fresh empty key, and coverage is complete by construction. `ProgressMappingContext` reverted to stock. Fixed in 06fa60759. 2229 tests green; live re-verified both directions (in-flight topic mid-turn; no stale draft post-settle). Final closure round running.

ymote:
**Codex round 5** (on 06fa60759) found 1 P1 + 1 P2, both verified real and fixed in 54690225c:

- **P1: bare-session registry collision** — `ActiveTurn` now records the profile it was admitted under (session-key-first precedence, canonical `_main`), and the aside verifies it before using the entry's turn id. Two profiles' same-named bare sessions (supported `web-*` ids) can no longer cross-read a live draft. Gate test gains a mismatched-profile leg.
- **P2: `turn_id` is client-supplied** — the draft store keys by (SessionKey, TurnId) so cross-session id collisions can't merge streams, and every ActiveTurn admission site clears its (session, turn) slot synchronously, so a reused id starts clean.

**Review closed per the stop criterion.** Five rounds, 19 findings, all premise-verified and fixed. Rounds 3–5 were all about the live-draft store this feature introduced; its design has structurally converged: one feed choke point (every `message/delta` passes with session+turn), (session, turn) keys, admission-time clears, and reads that require a **non-terminal, profile-matching** registry turn. The remaining bare-session `active_turns` keying is a pre-existing server-wide semantic (shared by `turn/interrupt`, `turn/state/get`) and out of this PR's scope.

Final state: 2229 lib tests green (11 btw tests incl. profile-mismatch and terminal-vs-active gate fixtures), octos-core 304 green, fmt clean, clippy at the origin/main baseline. **Live-verified after every redesign** — last check moments ago: mid-turn `/btw what are you explaining right now` → "I'm explaining Bloom Filters and Cuckoo Filters…" while the turn streamed. Ready for merge; the TUI client PR (`feat/btw-command`, 4 commits, 1063 tests green) follows with the core rev bump once this lands.

ymote:
**On the red `test-octos-agent (integration)` check — infra outage, not this PR.** Evidence:

1. **Dependency graph**: the job runs `cargo test -p octos-agent --tests`; every commit since the last green run touched only `crates/octos-cli/**`, and octos-cli sits *above* octos-agent (cli depends on agent, never the reverse) — the tested code is bit-identical.
2. **Local run**: the full agent integration suite passes on this exact branch head (every test binary green, no hangs).
3. **Time boundary, not code boundary**: *main's own* integration jobs fail identically — pass at 02:04Z (bf6be8cca), fail at 03:04Z (89512aeed) and 04:02Z (b222bd72b). Every integration job starting after ~02:45Z, on any branch, hangs ~52m until the runner is killed (no logs uploaded — the runner-termination signature this workflow's comments already document). This PR's failures (02:45, 03:06, rerun) all sit inside that window.

Merging with admin override per the above; main's next runs will confirm once the runner fleet recovers.

PR #1627, opened 2026-07-10
feat(appui): real multi-model switching — list fallbacks, implement profile/llm/select, fresh onboarding catalog

ymote:
## What

Three gaps kept a profile pinned to its primary model (user report: "/model and /onboard only ever offer deepseek"):

1. **`profile/llm/list`-backed model lists exposed only the primary** → the TUI `/model` menu had one row. Now every configured model rides the list — primary as `selected`, fallbacks as switchable alternatives.
2. **`profile/llm/select` was a stub** — it echoed the primary back and ignored the request. Now it promotes the requested *configured* model to primary, demotes the old primary to a fallback (lossless, reversible), persists to the profile store, honors the caller's `session_id` in the result, and rejects unconfigured models with a save-it-first error. **The switch takes effect on the next turn, no restart**: the profile's cached `SessionRuntime`s are evicted (new `SessionRuntimeCache::invalidate_profile`) and the dynamic `ProfileRuntime` is dropped + re-bootstrapped. Startup-config profiles persist the change with a restart-only note.
3. **The onboarding catalog lagged provider releases** (zai stopped at glm-5.1). `profile/llm/catalog` now unions the QoS model catalog (`model_catalog.json`) into the bundled families — new provider models appear without a code change; families lacking onboarding metadata (key env/route) are skipped. `providers.json` also gains the current zai lineup (`glm-5.2`, `glm-5-turbo`) as a 2-line refresh.

Multi-model *configuration* already existed (`profile/llm/upsert` `set_primary:false` → fallbacks; the onboarding wizard's "Add as fallback" row) — this PR makes those configurations listable and switchable.

## Testing

- 3 new tests: select promote/idempotent/reject + full list shape; catalog union (QoS-only model appears, unknown families skipped, bundled lineup intact); cache profile sweep spares other profiles.
- `cargo test -p octos-cli --features api --lib`: 2284 green.
- **Live-verified** (tmux, TUI ↔ `serve --stdio`, real deepseek key): `/model` lists both configured models, selecting flips the composer footer instantly, the next turn answers on the switched model, the swap survives a restart and is reversible, and the onboarding zai list shows `glm-5.2`.

Companion TUI PR (sends `session_id` in select params + auto-loads the onboarding catalog) follows; both directions are wire-compatible with the other side's current main.

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol.rs
+++ b/crates/octos-cli/src/api/ui_protocol.rs
@@ -7116,11 +7116,14 @@ fn raw_catalog_result(state: &AppState, profile_id: Option<&str>) -> Result<Valu
     // the home seed (`load_seed_qos_catalog` also falls back to `~/.octos`).
     let mut sources = Vec::new();
     if let Some(store) = state.profile_store.as_ref() {
-        let home = store.octos_home_dir().to_path_buf();
-        if let Some(profile_id) = profile_id {
-            sources.push(home.join("profiles").join(profile_id).join("data"));
+        if let Some(profile) =
+            profile_id.and_then(|profile_id| store.get(profile_id).ok().flatten())
+        {
+            // Honors `UserProfile.data_dir` overrides — the same resolution
+            // `ProfileRuntime::bootstrap` uses when it exports the catalog.
+            sources.push(store.resolve_data_dir(&profile));
         }
-        sources.push(home);
+        sources.push(store.octos_home_dir().to_path_buf());
     }
     for qos in sources
         .iter()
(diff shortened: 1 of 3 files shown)


ymote:
**Codex round 1** found 1 P1 + 3 P2 — all verified real, fixed in fcc2dc4fa: profile-scope guard on select **and** the sibling llm upsert/test/fetch_models (same unguarded resolver; test/fetch spend the target profile's keys), exact-route discrimination (ambiguous "official" wildcard now rejects), per-profile QoS catalog merge (profiles/<id>/data), and a generation guard so an in-flight bootstrap can't re-cache the pre-switch runtime. 3 new tests; 2286 green. Round 2 running.

ymote:
**Codex round 2** (on fcc2dc4fa): 2 P2s, both verified and fixed in b3a3011f3 — the generation re-check now happens inside `insert_with_eviction` under the same lock the invalidation sweep serializes on (no window for a stale insert), and the per-profile catalog merge resolves `UserProfile.data_dir` overrides via `ProfileStore::resolve_data_dir`. 2286 green. Final closure round running.

PR #1629, opened 2026-07-10
feat(api): retire /api/my/memory* + /api/my/cron* REST routes

ymote:
## Summary

Final server step of the #1621 consolidation sequence: the memory and cron panels are served over ui-protocol (`memory/overview`, `memory/entity`, `cron/list`, `cron/toggle` — #1621), and the web SPA re-pointed in octos-web#268. This retires the four REST panel routes in the M12 Phase D-5 `content/*` style, with the router comment documenting the replacement mapping.

**The panel handler cores stay** (`memory_panel.rs` / `cron_panel.rs`) — the RPC dispatchers call them directly. Only the axum route registrations and the router's now-unused module imports are removed.

## Tests

Full `--features api` lib suite green (2287) — the panel-core and RPC-forwarding tests all exercise the handlers directly, so coverage is unchanged by route removal. clippy + fmt clean.

Follow-up: TUI Memory/Schedule panes over the same methods.

### DIFF
--- a/api/OCTOS_UI_PROTOCOL_V1_SPEC_2026-04-24.md
+++ b/api/OCTOS_UI_PROTOCOL_V1_SPEC_2026-04-24.md
@@ -1591,9 +1591,16 @@ Request/response Rust types live in `crates/octos-core/src/ui_protocol.rs`
 - Params type: `CronListParams` — `{}` (accepts `params: {}` or
   `params: null`; the `params` member itself must be present, as above).
 - Result type: `CronListResult` — `{ jobs: CronJobRow[], count: number,
-  gateway_running: bool }`. Mirrors the REST body minus the redundant
-  `ok` flag; `gateway_running` reports whether a spawned gateway child
-  owns `cron.json` (toggles are refused while it does).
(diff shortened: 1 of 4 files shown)


ymote:
Superseded by #1631, which landed the same retirement (gitlink bump to 1e98538 + dropping the four REST routes + cron/list bounding via count+byte budget with a declared `truncated` flag).

Residual items from this PR's codex round-2 that #1631's scope does not mention, for a follow-up:
- **P1**: the sessionless aux bridge (web /settings) sends no selected-profile signal — `resolve_my_profile_id` host/Admin/User resolution means a multi-profile admin's `cron/toggle` can hit the wrong profile. Needs a `?profile_id=` on the WS upgrade validated via `is_authorized_for_profile` (and the web bridge appending it), plus profile-keyed aux-bridge lifecycle.
- **P2**: verify #1631's per-row field caps exclude `id` (capping the exact-match key breaks `cron/toggle`) and cover the nested `schedule.expr` (CronSchedule::Cron).
- **P2**: web Schedule tab does not yet render the truncation signals.

PR #1642, opened 2026-07-11
fix(llm): retry transport send-failures + prompt-cache kill-switch

ymote:
## Diagnosis: did prompt caching (#1640) break glm-5.2? **No.**

A user's `glm-5.2` turn hard-failed with `runtime_error: failed to send streaming request to Anthropic` after 20 successful tool actions (installing an Android AVD), with a partial response. glm-5.2 **is** routed through `AnthropicProvider` at `https://api.z.ai/api/anthropic` with prompt-caching ON, so caching was a fair suspect. It is exonerated by evidence:

| Evidence | Finding |
|---|---|
| **Error signature** | The string comes from `wrap_err("failed to send streaming request to Anthropic")` on the reqwest `.send()` — a **transport** failure. A `cache_control` payload rejection would instead **succeed the send** and return a non-2xx *response* (a different, HTTP-status-carrying error). |
| **Live cache-acceptance probe** | z.ai glm-5.2 returns **200 with `cache_read=7232`** when `cache_control` is present — caching works, z.ai does not reject it or drop the connection. |
| **Live A/B (10 rounds each)** | caching **ON**: 10 ok / 0 send-failures. caching **OFF**: 10 ok / 0 send-failures. No differential. |
| **Timing** | Failed at action 20 (not action 1) and only once. A systematic caching incompat would fail immediately and every turn. |

The failure is a **transient transport drop** — reqwest reusing a pooled keepalive socket that z.ai's load balancer had already closed ("connection closed before message completed"). `build_http_client` keeps idle sockets for 90s, so a later request in a long turn grabs a stale one.

## The real bug: that transport error was neither retried nor failed over

`RetryProvider::is_retryable_error` / `should_failover` are the single classification chokepoint for **both** same-provider retry (`RetryProvider`) and failover (`ProviderChain`/`FallbackProvider`). Proven empirically with a local TCP server:

| Failure mode | reqwest | old is_retryable | old should_failover |
|---|---|---|---|
| connection refused | `is_connect=true` | true | true |
| **connection closed before message completed** | `is_connect=false, is_request=true`, no status | **false** | **false** |

So the user's turn hard-failed with **no retry and no failover** to the configured deepseek lanes.

## Changes (octos-llm only)

- **`retry.rs`** — statusless transport send/body failures (`is_request()` / `is_body()`) are now retryable. Status-bearing errors already return above, so only true transport faults reach the new branch. Also **reorder `is_timeout` before `is_connect`** so a connect *timeout* (which is both) fails over instead of hammering the same unreachable lane. String fallback also matches `connection closed` / `broken pipe` / `error sending request`.
- **`anthropic.rs`** — **`OCTOS_PROMPT_CACHING` kill-switch**: disable prompt caching at startup without a rebuild for any Anthropic-compatible proxy that rejects `cache_control`. Default stays **ON**; only falsy values (`0`/`false`/`off`/`no`) disable. Pure, unit-tested seam (no `unsafe` env mutation; workspace is `deny(unsafe_code)`).

No per-endpoint cache gating was added — caching is exonerated for glm-5.2, so it would be unwarranted.

## Tests / gates

- Real-socket transport-failure classification: connection-closed, immediate-close, refused, request-timeout, connect-timeout.
- Env kill-switch parsing.
- Ignored live z.ai A/B (`glm_send_failure_ab.rs`).
- `cargo test -p octos-llm` green (455 lib); `cargo fmt --all --check` + `cargo clippy -p octos-llm --all-targets -D warnings` clean.
- **codex review**: 2 rounds. R1 → P2 (connect-timeout ordering) fixed. R2 → "Reorder is correct" + P3 (test proxy-robustness) fixed with `.no_proxy()`.

Note: the `test-octos-agent-integration` CI job is pre-existing-broken on main (unrelated).

### DIFF
--- a/crates/octos-llm/src/anthropic.rs
+++ b/crates/octos-llm/src/anthropic.rs
@@ -1226,7 +1226,10 @@ mod tests {
 
     #[test]
     fn test_build_request_filters_system() {
-        let provider = AnthropicProvider::new("test-key", "claude-test");
+        // Pin caching ON: the extraction assertion below reads block-form
+        // `system[0].text`, which only exists when caching is enabled — keep
+        // it hermetic w.r.t. an ambient `OCTOS_PROMPT_CACHING=0`.
+        let provider = AnthropicProvider::new("test-key", "claude-test").with_prompt_caching(true);
         let messages = vec![
             msg(MessageRole::System, "system prompt"),
             msg(MessageRole::User, "hello"),
@@ -1605,7 +1608,11 @@ mod tests {
 
     #[test]
     fn should_mark_system_last_tool_and_last_user_block_with_cache_control() {
-        let provider = AnthropicProvider::new("test-key", "claude-test");
+        // Pin caching ON so this wire-shape assertion is hermetic w.r.t. an
+        // ambient `OCTOS_PROMPT_CACHING=0` (the builder override wins over the
+        // env default). The default-resolution truth table is covered
+        // separately by `prompt_caching_env_*` unit tests.
+        let provider = AnthropicProvider::new("test-key", "claude-test").with_prompt_caching(true);
         let tools = vec![
             tool_spec("alpha", "first tool"),
             tool_spec("omega", "last tool"),
@@ -1672,7 +1679,9 @@ mod tests {
         // batch — the breakpoint must land on its LAST block only, so the
         // whole history including the current results is cached for the
         // next iteration.
-        let provider = AnthropicProvider::new("test-key", "claude-test");
+        // Pin caching ON so the assertion is independent of the ambient
+        // `OCTOS_PROMPT_CACHING` kill-switch (builder override wins).
+        let provider = AnthropicProvider::new("test-key", "claude-test").with_prompt_caching(true);
         let mut assistant = msg(MessageRole::Assistant, "");
         assistant.tool_calls = Some(vec![
             tool_call("toolu_a", "shell", serde_json::json!({"command": "ls"})),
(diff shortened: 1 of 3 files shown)


ymote:
## Follow-up: codex adversarial review resolved (commit b29a4487c)

Ran an independent adversarial codex pass on the transport-retry change. Findings and resolution:

- **P1 (statusless POST replay can double-charge)** — *accepted tradeoff, documented in `retry.rs`.* A statusless send failure is ambiguous ("no response" ≠ "not accepted"), so a replay can double-bill a non-idempotent chat POST in the rare mid-generation-drop sub-case. Accepted because: the dominant cause is a **never-serviced** reused-keepalive socket (safe to replay); for a **streaming** POST a server that had begun generating would have already flushed response headers, so `.send()` would have resolved and the drop would surface as a stream/body error *outside* this retry scope; the agent loop consumes exactly one `ChatResponse` (**no duplicated tool side-effects** — residual harm is provider-side billing only); and the alternative is hard-failing the whole turn on any transient drop. Future hardening noted in-code: short pool idle-timeout / client idempotency key.
- **P3 (env-sensitive tests)** — **fixed.** Pinned `.with_prompt_caching(true)` on the 5 wire-shape tests that assert caching-ON on a default provider. Proven hermetic: `OCTOS_PROMPT_CACHING=0 cargo test -p octos-llm` now passes **455/0** (was 2 failed).
- **P2 (string-fallback substring match)** — left as-is: pre-existing class (prior code already matched `connection refused`/`reset`/`timed out`/`overloaded`) and unreachable for real providers whose typed `LlmError` returns before the fallback.

Gates re-run on the branch: `cargo test -p octos-llm` 455/0 (both with and without the kill-switch set), `cargo fmt --check` clean, `cargo clippy --all-targets -D warnings` clean.

PR #1651, opened 2026-07-12
feat(matrix): mention-only gate counts bots from the room map

ZhangHanDong:
## Problem

Appservice bots in multi-bot rooms answered **every** message, even when nobody @-mentioned them. Two causes:

1. The mention gate derived DM-ness from the homeserver's `joined_members` alone — and Palpo omits appservice virtual users from that response, so the gate always saw "1 human, 0 bots" and treated any room as a 1:1 DM.
2. Robrix (client side) silently targeted the room-bound bot on every message in DM-flagged rooms, bypassing the gate entirely. That half is fixed in the companion Robrix2 PR (link below).

## Changes

- **Mention-only gate** (channel setting `mention_only`, default `true`): outside a true 1:1 DM, a bot only replies when explicitly addressed — an `m.mentions` entry, an MXID pill/text mention, or a client-supplied `org.octos.target_user_id`.
- **Bots are counted from the appservice's own `BotRouter` room map** (`bot_count_for_room`), merged with the membership-derived count via `merge_mapped_bot_count` — the room map is authoritative when the homeserver hides virtual users. The `UNKNOWN` sentinel survives the merge, so fail-closed behaviour is preserved.
- **A true 1:1 DM** is one human plus at most one managed bot; BotFather counts as a potential responder (it answers unrouted messages in admin mode). Unknown membership fails closed.
- **`org.octos.explicit_room` honored**: messages carrying the client marker go through the gate even when `mention_only` is disabled.
- Member-count cache is invalidated on room membership events; `docs/user-guide.md` documents the new field.

## Resulting behavior

| Room | Unaddressed message |
|------|---------------------|
| 1 human + 1 managed bot | bot replies (DM exemption) |
| 1 human + 2+ managed bots | all bots stay silent; @-mention picks the responder |
| 2+ humans | mention required (unchanged) |

## Testing

- `cargo test -p octos-bus --lib --features matrix` — 351 passed, including new `gate_blocks_unaddressed_when_room_map_has_multiple_bots`, `gate_allows_unaddressed_dm_with_single_mapped_bot`, `room_member_counts_uses_room_map_when_homeserver_hides_bots`.
- `cargo test -p octos-cli --lib` — 630 passed (config plumbing for `mention_only` / `mentionOnly`).

## Related

- Companion client-side PR (Robrix2): Project-Robius-China/robrix2#253 — removes the client-side implicit bound-bot targeting that bypassed this gate; deploy together
- Spec: `specs/task-agent-mention-gating.spec.md` in the Robrix2 repo (agent-spec contract, lint 100%)


ZhangHanDong:
Note on the `test-octos-agent (integration)` failure: it is **pre-existing on `main`**, not introduced by this PR. The last 6 completed CI runs on `main` all fail the same job (alongside check-arm / check-windows), and the job times out after ~51 minutes rather than failing an assertion. This PR does not touch `octos-agent` — changes are limited to `octos-bus` (matrix channel gate), `octos-cli` (config plumbing), and docs. All jobs relevant to this change pass: `check`, `check-matrix`, `test-octos-agent (lib)`, `test-octos-cli`.

ymote (dismissed):
Thanks for tackling this — the design (mention-only gate, counting bots from the appservice room map because Palpo hides virtual users, `UNKNOWN`-fails-closed sentinel, room-routing not counting as "addressed", safe-by-default `mention_only: true`) is the right shape, and the test coverage is substantial.

Before merge, though, an adversarial pass surfaced one correctness gap that lets the gate be bypassed in this PR's own threat model, plus a few smaller issues. Requesting changes on these:

## P1 — `merge_mapped_bot_count` undercounts because `max` is not a union

`crates/octos-bus/src/matrix_channel.rs:587` merges the membership-derived bot count with the room-map count via `max`. But the two sources count *different* bots:

- membership (`count_room_members`) counts BotFather (the appservice sender) **but not** homeserver-hidden child virtual users;
- the room map (`bot_count_for_room`, matrix_channel.rs:289) counts room-bound child bots **but not** BotFather.

So a room of **Alice + BotFather + one child bot, with the child hidden by the homeserver** yields:

- membership → `{humans: 1, managed_bots: 1}` (BotFather only)
- mapped_bots → `1` (the child)
- `max(1, 1) = 1` → `{humans: 1, managed_bots: 1}` → `is_direct_chat() == true` → **an unmentioned message is dispatched as a 1:1 DM**, even though the true composition is 1 human + 2 bots and the design wants a mention.

Since counting BotFather as a responder is intentional (per the `RoomMemberCounts` doc comment), and this is exactly the "homeserver hides virtual users" case the PR targets, the gate can be bypassed in practice.

Relatedly, `bot_count_for_room` returns `0` for a room absent from the map (`room_bots.get(room_id).map_or(0, …)`), so a map-load failure contributes `0` rather than failing closed.

**Suggested fix:** make the bot count a genuine union rather than a `max` — e.g. union the room-map child-bot set with the membership-derived managed set (add BotFather explicitly if present), or fail closed when the map lookup misses.

## P2 — private-bot rejection fires before the gate

`crates/octos-bus/src/matrix_channel.rs:1594` runs the "This is a private bot…" rejection before the mention gate at :1659. `route_by_room` (:1574) sets `METADATA_TARGET_PROFILE_ID` **without** marking the message `addressed`, so in a group room bound to a private child bot, any unmentioned message from a non-owner triggers the rejection reply — spamming the room and revealing the private bot, exactly the unmentioned-noise the gate is meant to suppress. Consider gating the private-bot reply behind `addressed` (or running it after the mention gate).

## P2 — standard `matrix.to` pills are not recognized

`crates/octos-bus/src/matrix_channel.rs:444`: `is_matrix_user_id_char` includes `'/'`. In a standard pill's `formatted_body` the MXID sits in an href like `https://matrix.to/#/@bot:server`, i.e. immediately preceded by `/`. `contains_exact_matrix_user_id_mention` (:448) therefore sees a "user-id char" before the MXID, fails its left-boundary check, and does not treat the pill as a mention. A user who pills the bot without their client also sending `m.mentions` won't be recognized as addressing it. Dropping `'/'` from `is_matrix_user_id_char` (it isn't a valid MXID character) fixes this.

## P2 — `mention_only` can't be expressed via the profiles config

`crates/octos-cli/src/profiles.rs:954`: the `ChannelCredentials::Matrix` variant has no `mention_only` field, so a matrix channel configured through this path can't carry `mention_only: false` — the documented opt-out silently reverts to the `true` default. The new field needs to be threaded through this representation too (and its conversion into channel settings).

## P2 — member-count cache race

`crates/octos-bus/src/matrix_channel.rs:649`: a membership fetch started before a membership-event invalidation can complete afterward and unconditionally reinsert stale DM counts, which then stay valid for the full `DM_MEMBER_CACHE_TTL` (~60s), letting subsequent unmentioned group messages through. A generation/epoch check (or dropping a write whose room was invalidated mid-flight) would close this.

## P3

- `crates/octos-bus/src/matrix_channel.rs:649`: the TTL is only checked on read; entries are removed on membership events but never evicted by age, so `dm_member_cache` grows unbounded across all rooms ever seen.
- `docs/user-guide-zh.md` documents the Matrix channel fields but wasn't updated with `mention_only` (parity with `docs/user-guide.md`).

## CI note

`test-octos-agent (integration)` is red, but that's the known pre-existing/flaky integration job (job-level timeout, no test failures) and this PR doesn't touch `octos-agent` — not a blocker.

Happy to re-review once the P1 union-count and the P2s are addressed.

ZhangHanDong:
@ymote All items addressed in 6d6bceed — point by point:

**P1 (union, not max)** — Confirmed the bypass: membership and the room map count disjoint bot populations, so `max` undercounted exactly the "1 human + BotFather + hidden child" composition. `merge_mapped_bot_count` is replaced by `merge_room_bot_sources`: membership now reports `{humans, child_bots, botfather_joined}` and the room map reports `{child_bots, botfather_mapped}` (`BotRouter::room_bot_composition`, which resolves each mapped profile through the routes table to tell BotFather apart from children; an unresolvable profile counts as a child — over-counting fails safe). Child bots are maxed across the two views of the same population and BotFather is added separately when either source saw it. Regression test `merge_counts_botfather_and_hidden_child_as_two_responders` encodes your exact case.

On the related fail-closed point: a room **absent** from the map still contributes zero mapped bots (that's a legitimate state — e.g. a room with no bound bots), but a room map file that **exists and fails to read/parse** now poisons the router: `room_bot_composition` returns `None` and the gate fails closed (`merge_fails_closed_when_room_map_unavailable`). `reload()` similarly keeps the previous in-memory map on a parse failure instead of clobbering it with an empty one.

**P2 private-bot rejection** — Now gated behind `addressed`: unaddressed room chatter routed to a private bot via the room mapping is dropped silently (no reply, no dispatch, no reveal); an explicit mention/target still gets the rejection as legitimate feedback. Existing test updated to assert silence; new `test_private_bot_rejection_sent_when_explicitly_addressed` covers the addressed path.

**P2 matrix.to pills** — `/` removed from `is_matrix_user_id_char` (not a valid MXID character); a standard pill href now passes the left-boundary check. Covered in `mentions_user_detects_structured_and_text_mentions`.

**P2 profiles config plumbing** — `ChannelCredentials::Matrix` gains `mention_only` (serde default `true`), `channel_to_entry` emits it in the appservice branch, and `MatrixChannelSettings::from_entry` already reads it — round-trip covered by `test_matrix_appservice_channel_to_entry_carries_mention_only` (including the `false` opt-out).

**P2 cache race** — Added a generation counter bumped on every membership-event invalidation; a fetch that started before the bump serves its result but does not cache it, so stale counts can no longer be resurrected for a TTL.

**P3** — Expired entries are now age-evicted on every cache write (bounded growth), and `docs/user-guide-zh.md` documents `mention_only` (config example + field table) in parity with the English guide.

Tests: `cargo test -p octos-bus --lib --features matrix` → 352 passed; `cargo test -p octos-cli --lib` → 631 passed; `cargo fmt --all -- --check` clean.

ZhangHanDong:
Gentle re-review nudge — this has been ready since `6d6bceed` (2026-07-12) with no changes needed since.

@ymote All six items from your review are addressed in `6d6bceed`; the point-by-point walkthrough is [above](https://github.com/octos-org/octos/pull/1651#issuecomment-4952285384). Summary:

- **P1** — `max`-merge replaced with a genuine union (`merge_room_bot_sources`): child bots maxed across the membership/room-map views of the same population, BotFather counted once when either source sees it. Your "1 human + BotFather + hidden child = 2 responders" case is now gated (`merge_counts_botfather_and_hidden_child_as_two_responders`). A room-map file that exists but fails to load poisons the router → gate fails closed.
- **P2** — private-bot rejection only replies when `addressed` (silent drop otherwise); `/` removed from `is_matrix_user_id_char` so standard matrix.to pills register; `mention_only` threaded through `ChannelCredentials::Matrix` + `channel_to_entry`; member-count cache guarded against the invalidation race with a generation counter.
- **P3** — age-eviction on cache write; `mention_only` documented in `user-guide-zh.md`.

CI: every check relevant to this change is green (`check`, `check-matrix`, `test-octos-agent (lib)`, `test-octos-cli`). The only red is `test-octos-agent (integration)`, which is the pre-existing main-branch failure (job-level ~52min timeout, no test failures) — `main` itself is still red on that same job as of 2026-07-18, and this PR touches only `octos-bus`/`octos-cli`/docs.

Could you re-review when you have a moment? And if the integration job is the only thing blocking, it'd need a maintainer override since it's unrelated to this change.

anantheparty:
I reproduced the conflict against current `main` (`1f4370f`). All production code merges automatically; the only conflict is the old inline `matrix_channel` test module versus main's extracted `matrix_channel_tests.rs`.

I prepared a Draft helper PR targeting the existing head branch (not a replacement PR to `main`): https://github.com/ZhangHanDong/octos/pull/3

The resolution preserves main's test extraction, ports the mention-gate tests into the external test file, and keeps the automatically merged production implementation. `git diff --check` and `cargo fmt --all -- --check` pass. Targeted tests are left to CI because the isolated local dependency fetch timed out.

anantheparty:
The conflict helper has now been refreshed through current `main` (`44ee3eae`), marked Ready, and fully revalidated: https://github.com/ZhangHanDong/octos/pull/3

Its head is `8ce396c8`. The base `octos-bus` suite passes 225/225; the focused Matrix mention/union/pill/private-bot coverage passes 20/20; Matrix gateway settings pass 5/5; formatting and diff checks are clean. The helper targets this PR's existing source branch, so merging it will update #1651 in place and allow current CI/re-review without opening a replacement PR.

anantheparty (approved):
2026-08-15 复核通过：ymote 的 P1（merge_room_bot_sources 计数）与 fail-closed（load_rooms poisoned → UNKNOWN）均已在当前 head 落实并有专属测试；缓存 generation 防回填、私有 bot 防暴露等设计到位。合并前请：① rebase 到最新 main；② 重跑 integration 测试（上次 run 日志已过期）。

anantheparty (approved):
2026-08-15 复核通过（详见上一条评审）：P1 计数修复与 fail-closed 均已落实并有测试。合并前请 rebase + 重跑 integration 测试。

anantheparty:
分支关联丢失（原 head 分支被删除，重建后 GitHub 未恢复）。已由维护者 rebase 到最新 main（解决 matrix_channel 测试模块结构冲突，测试文件合并保留全部 51 个测试；cargo test --features matrix 384 passed）。关闭本 PR，由新 PR 承接。

PR #1661, opened 2026-07-13
feat(catalog): make model_catalog.json the SSOT for model provisioning

ymote:
## Summary

Reconciles the hand-maintained provider list and the runtime QoS catalog into a single canonical `model_catalog.json` — **the single source of truth for model provisioning** — and wires every provisioning path to it. Providers configured in the dashboard become a *subset* of this catalog.

The catalog now carries **148 curated models** with researched context windows + max-output (comprehensive per-model research across Anthropic, OpenAI, Gemini, DeepSeek, GLM/zhipu, Kimi/moonshot, MiniMax, Qwen/dashscope, Llama/Mistral/Nemotron, etc.).

## What changed

**Onboarding (`raw_catalog_result`) is sourced only from the canonical catalog + registry.**
- Family list + models come from the compiled-in canonical catalog.
- Family key-env comes from `octos_llm::registry` (alias-aware: `google→gemini`, `qwen→dashscope`, `glm→zhipu`), not a hand-maintained env map.
- `providers.json` is retired as a server source (now a derived, web-only artifact consumed by the admin SPA).
- **The runtime QoS catalog is no longer unioned into onboarding.** Configured providers are a *subset* of the catalog, so a live QoS catalog could only ever (a) re-introduce a curated-out model — e.g. `deepseek-chat`, which still lingered in `~/.octos/model_catalog.json` — or (b) zero out a researched context window. New models are added by editing the catalog, which is the point of an SSOT. Live QoS still drives **routing** via the adaptive chain; that path is untouched.

**One embedded copy of the catalog.** `qos_catalog` owns the single `include_str!`; onboarding (`ui_protocol`) and `octos init` alias it, so there is exactly one compiled-in `model_catalog.json`.

**Fresh-install seeding.** `load_seed_qos_catalog` falls back to the embedded canonical, so a clean install seeds the router / context-window / pricing tables with researched values instead of cold-start zeros.

**Exporter preserves all entries.** The QoS exporter now persists `merge(embedded_base, live_export)` (both the initial write and the 30s re-export), so the on-disk catalog stays a full superset with live scores overlaid — it never shrinks to just the configured lanes, and effectively seeds the data-dir with the full catalog on first run.

**Curation.** Add `zai/glm-5.2` (1M ctx) and `moonshot/kimi-k2.6`; keep `minimax-m3`; remove `deepseek-chat` / `deepseek-reasoner`; fix three `-nano` models mis-tagged `strong`.

**Packaging.** `model_catalog.json` ships next to the binary in both the local bundle (`scripts/build-local-bundle.sh`) and the CI release tarball (`release-dispatch.yml`), so installs have the full catalog + accurate context windows without a network fetch.

## Tests
- `catalog_result_sourced_from_registry_and_canonical_catalog` — registry env, glm-5.2 + kimi-k2.6 present, deepseek-chat absent, researched context window flows through.
- `catalog_result_ignores_runtime_qos_not_in_canonical` — regression guard: a runtime-QoS-only model must **not** leak into onboarding.
- `merge_qos_catalog_preserves_base_and_overlays_live` — base preserved, live export overrides matching providers, overlay-only lanes appended, deterministic order.
- `embedded_qos_catalog_is_curated_ssot` — the embedded canonical parses and reflects curation.
- Full `octos-cli` lib suite green (2352 passed / 0 failed); rebased on latest `main` (#1660).


ymote:
Independent adversarial review (7 correctness claims) came back clean — **no P1/P2 findings**:
- Exporter merge base is the embedded canonical captured once at startup, so the on-disk catalog is bounded (`|embedded| ∪ |configured lanes|`) — no unbounded growth across ticks/restarts.
- Onboarding reads the embedded canonical only, so a stale `~/.octos` catalog can't resurrect a curated-out model.
- `load_seed_qos_catalog` never-None only touches configured router slots (non-matching entries ignored) plus benign global context/pricing seeding — a strict improvement on fresh installs.
- `merge_qos_catalog` is overlay-wins-by-provider, base-preserved, deduped, deterministically ordered.
- No panics; module visibility correct; clippy clean in changed lines.

One P3 was actioned: the retired `providers.json` used to supply an `endpoints` array (alternative provisioning routes — AutoDL/WiseModel) for 4 models (`deepseek-v4-pro`, `deepseek-v4-flash`, `kimi-k2.5`, `MiniMax-M2.5-highspeed`) that the TUI onboarding route-picker consumes. Those endpoints are now carried in `model_catalog.json` and re-emitted by `raw_catalog_result`, riding the catalog JSON without polluting the routing `ModelCatalogEntry` type. Added a regression test asserting the AutoDL route survives.

ymote:
Ran the Codex CLI review (`codex exec review`) and addressed all four findings in `9e1ee85` (octos-llm 456/0, octos-cli 2352/0, clippy clean):

- **P2a — SSOT convergence** (`qos_catalog::merge_qos_catalog`): now a field-level merge. For a provider in both the canonical base and the live export, canonical STATIC fields (cost / context window / max output / type) win and only DYNAMIC QoS (score, stability, latency) comes from the overlay — so a stale value seeded from a pre-upgrade on-disk catalog can no longer re-persist itself forever.
- **P1 — pricing/context bare-alias collision** (`pricing::seed_pricing_catalog`, `context::seed_from_catalog`): the bare model-id alias (`kimi-k2.5`) is now awarded to the row with the fewest path segments (the native provider), so a re-host can't clobber the native rate under the shared alias. Deterministic, order-independent.
- **P2b — host-tagged lane seeding** (`AdaptiveRouter::seed_catalog`): falls back to the normalized bare-family key, so an OpenAI-compatible lane (`moonshot@api/kimi-k2.5`) is seeded from the untagged canonical entry (`moonshot/kimi-k2.5`) instead of being skipped.
- **P2c — dashboard parity**: added `scripts/sync-dashboard-providers.py`, an append-only generator that derives the `providers.json` lineup from `model_catalog.json` (existing entries + curated order preserved); ran it to add the 3 missing NVIDIA models. Full lineup parity restored.

Each fix has a dedicated regression test. A Codex verification pass on the fixes commit is running.

PR #1663, opened 2026-07-13
refactor: type JSON boundaries for profiles and MCP

chrislearn:
## Summary

Replace unstructured JSON values with strong Rust types at stable API boundaries.

- Use `OutputCursor` for agent output pagination.
- Use typed LLM, memory, and embedding configuration in profile QR payloads.
- Use `ValidatorOutcome` for MCP validation results.
- Add `McpSessionCost` for explicit token accounting.
- Preserve all five MCP token counters on the wire, including zero values.
- Simplify validator gate evaluation through `required_gate_passed()`.

## Motivation

These boundaries previously relied on manual JSON construction and field lookup, delaying validation until runtime and allowing inconsistent payload shapes. Reusing existing domain types provides compile-time validation and keeps producers, consumers, and tests aligned.

## Validation

- `cargo check -p octos-agent -p octos-cli --all-features --lib --bins`
- `cargo clippy -p octos-agent -p octos-cli --all-features --lib --bins --no-deps`
- MCP server acceptance tests: 10 passed
- MCP dispatch integration tests: 7 passed
- Profile QR tests: 10 passed
- Profile payload tests: 12 passed
- Agent output cursor regression test passed
- MCP token-cost regression test passed

Existing unrelated compiler and Clippy warnings remain outside the scope of this change.


anantheparty (changes requested):
Requesting changes for two stable-boundary regressions.

1. The published profile-QR payload still emits `v: 1`, but its previously verbatim `llm`, `memory`, and `embedding` JSON blocks become closed/lossy typed schemas. `LlmProfileConfig` and nested types deny unknown fields, so an extension-bearing payload that was valid under v1 now fails to decode. `MemoryConfig`/`EmbeddingConfig` accept but drop unknown keys during decode/re-encode, changing the payload silently. Please preserve the raw v1 wire representation, or add a separately decoded v2 while retaining a v1 raw decoder; add literal historical and extension-bearing fixtures. A new-type round trip does not test compatibility.

2. Deserializing `agent/output/read.cursor` as `OutputCursor` in `RawAgentOutputParams` moves failure into generic raw-parameter parsing. Malformed cursors still return `-32602`, but lose `data.kind=agent_output_cursor_invalid`, profile/session/policy context, and `recoverable: true`, which clients use to clear the cursor and retry. Keep the cursor raw until domain validation (or explicitly remap the serde failure to the existing domain error), and add dispatch-level malformed-cursor error-shape tests.

These are independent of the other useful typed MCP changes and should be fixed before merge.

anantheparty (approved):
Re-reviewed current head `7fbbf2da`; the two requested stable-boundary fixes are complete.

- Profile QR v1 keeps `llm`, `memory`, and `embedding` as raw JSON values. Literal historical and extension-bearing fixtures now prove lossless round trips for nested future fields.
- `agent/output/read.cursor` remains raw at the RPC boundary and is parsed explicitly before dispatch. Malformed values preserve `-32602`, `data.kind=agent_output_cursor_invalid`, policy/profile/session context, and `recoverable: true`, with a regression proving the orchestrator is not called.
- The typed MCP changes remain behaviorally equivalent after merging current main.

No new P0-P3 findings. Linux check/agent/CLI CI is green; the running Windows job remains the normal merge gate.

anantheparty:
@BH3GEI the two previously requested contract fixes are resolved on current head `7fbbf2da`; I re-reviewed and approved it. All CI has now converged green, including the 37m06s Windows job: https://github.com/octos-org/octos/actions/runs/30486829285/job/90694499222

The only remaining merge gate is the repository's additional maintainer review requirement.

Issue #1541, opened 2026-07-07
Quickstart friction: brew formula lacks embedded SPAs; auth login needs a provider picker; init's 'auto' model default breaks DeepSeek


ymote:
Found while validating the README Start-here end-to-end on a clean mac mini (octos 1.1.0 via brew AND npm). Docs worked around all three in #1540; these are the product fixes.

## 1. Homebrew formula ships without the embedded web app + dashboard (P1 for the brew path)
`brew install octos-org/tap/octos` → `octos serve --solo` →
- `GET /app/` → **503** `{"error":"web_bundle_missing","message":"Run ./scripts/build-web-app.sh + rebuild octos-cli…"}`
- `GET /admin/` → **503** `admin_bundle_missing`

The **npm package embeds both** (200/200, same version `1.1.0 (29b224c)`) — so the formula builds from source without running `build-dashboard.sh`/`build-web-app.sh`, while npm wraps the release binaries. Fix: make the formula consume release artifacts (or run the SPA build steps).

## 2. `octos auth login` requires `--provider` with no interactive fallback (newbie UX)
Bare `octos auth login` errors with clap usage. Since `octos init` just captured the provider choice, `auth login` could default to the configured provider — or offer the same numbered picker `init` uses.

## 3. `octos init`'s `Model [auto]` default is invalid for DeepSeek
DeepSeek has no catalog in 1.1.0 ("No catalog models found — enter model name manually"), and accepting the suggested `auto` produces a hard 400 on every turn: *"The supported API model names are deepseek-v4-pro or deepseek-v4-flash, but you passed auto."* Either ship a DeepSeek catalog, validate the model at init time (a cheap models call), or don't offer `auto` where the provider rejects it.

Also noted: the brew bottle is a month behind main (`29b224c` 2026-06-08) — worth a bump when the formula is fixed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ymote:
Item 2 shipped in #1542: `octos init` now offers to capture the API key into the global auth store (same paste_token shape as `auth login`), honoring `--defaults`/non-TTY runs and treating expired stored credentials as missing. Remaining: item 1 (brew formula lacks embedded SPAs) and item 3 (init's `auto` model default invalid for DeepSeek).

ymote:
[...]

PR #1673, opened 2026-07-14
fix: resolve 11 logistical/CLI defects from full-repo audit

heyong4725:
## Summary

Surgical fixes surfaced by a full-repo feature audit of octos 2.0.1. Every fix is verified end-to-end; regression tests were added for the non-obvious correctness bugs. **Workspace suite: 5295 passed / 0 failed (0 regressions).**

The audit triaged ~35 total findings into Tier A (surgical logistical/CLI — this PR) and Tier B (security/architecture — deferred, needs design decisions). Only Tier A ships here.

## Fixes (11)

| Area | Defect | Fix |
|---|---|---|
| `llm/provider` | `truncate_error_body` byte-sliced `&body[..200]` → **panic** on multibyte UTF-8 error bodies (e.g. Chinese providers) | char-boundary truncation |
| `pipeline/condition` | tokenizer used `byte as char` → non-ASCII string literals never matched | decode literals as UTF-8 |
| `llm/context` | catalog lookup did substring match over nondeterministic HashMap order | deterministic longest-key-wins (mirrors `pricing.rs`) |
| `cli/office` | `validate` didn't resolve absolute OPC rels targets (`/xl/...`); `PathBuf::join` replaced the base → false "does not exist" | re-root leading-slash targets at package dir |
| `cli/cron` | `--every 0`/negative accepted; `remove`/`enable` on unknown id printed an error but **exited 0** | reject `<= 0`; `bail` → non-zero exit |
| `cli/init` | `--custom-base-url` silently wrote model `"auto"` (banned #1541; real APIs 400) | require `--custom-model`, else bail |
| `cli/status,docs` | Moonshot advertised `MOONSHOT_API_KEY` but `init`+skills use `KIMI_API_KEY` → status always "not set" | align to `KIMI_API_KEY` |
| `cli/account` | empty-channel hint suggested `account create` with wrong flags | correct to `octos account update <id>` |
| `agent/grep` | invalid `file_pattern` glob silently ignored → searched **all** files | compile once, error on invalid |
| `agent/web_fetch` | DNS pinning via `resolve()` in a loop overwrote → only **last** address pinned | `resolve_to_addrs` pins all validated IPs |

## Test plan

- [x] `cargo test --workspace` → 5295 passed / 0 failed
- [x] Live re-drive of each fixed CLI behavior against `target/debug/octos` in isolated `HOME` sandboxes (cron exit codes, `--every 0`, `init` guard, moonshot status, office validate) — all PASS
- [x] Regression tests added: multibyte `truncate_error_body`, non-ASCII pipeline condition, deterministic catalog lookup

## Notes

Several of these are *inconsistent application of patterns the codebase already gets right elsewhere* (web_fetch not reusing `ssrf.rs`'s `resolve_to_addrs`; context.rs not reusing pricing.rs's tie-break). The deferred Tier B findings (multi-tenant web authz, channel robustness/crypto) are tracked separately and need a threat-model discussion.


heyong4725:
Thanks — all three addressed in b22c953b.

**[P1] Moonshot key consistency.** Rather than pick one name, I made resolution **accept both**, using the registry's existing `kimi` alias so nothing else has to change. `Config::resolve_api_key` now builds a candidate list of the primary env var + every alias-derived var (`moonshot` → `MOONSHOT_API_KEY` + `KIMI_API_KEY`) and tries each across the auth store, keychain/env_vars map, and process env; all candidates are secret-registered. So a hand-written moonshot config authenticates whether the operator set `MOONSHOT_API_KEY` (registry default) or `KIMI_API_KEY` (docs/init/skills). `octos status` mirrors the same rule (registry key var + aliases) and reports `set` for either — verified: `MOONSHOT_API_KEY=x` and `KIMI_API_KEY=x` each show `set`, neither shows `not set`.

**[P2] Catalog determinism.** `context.rs` now mirrors `pricing.rs` exactly: longest key the model id *contains* (family it extends) wins; otherwise the *shortest* key that contains the model id; lexical tie-break. I also extracted a pure `catalog_lookup_in(map, id)` and rewrote the test to run over a local map with ids that are **not** exact keys (`gpt-4o-mini-2024-07-18` → branch 1, `4o-mini` → branch 2), so it exercises the substring path instead of returning through `map.get()` — and it no longer races the shared `CATALOG`.

**[P2] `--every` overflow.** `secs * 1000` → `secs.checked_mul(1000)` with an explicit rejection on overflow. Reproduced: the `i64::MAX` case now exits 1 with a clear message; added unit tests for non-positive, overflow, and valid intervals.

Workspace suite: **5298 passed / 0 failed**.

heyong4725:
Good catches — both fixed in e798c156. The name-alias derivation was the wrong mechanism; replaced it with explicit registry metadata as you suggested.

**[P1] Custom overrides are now exclusive.** Added `ProviderEntry.key_env_aliases` (moonshot = `["KIMI_API_KEY"]`, every other provider `&[]`). Resolution expands to a provider's declared sibling key vars **only when the configured var is itself one of those known names**. An arbitrary custom `api_key_env` like `MY_PROXY_KEY` is not a known moonshot key name, so the candidate set stays `[MY_PROXY_KEY]` — a miss now errors instead of falling back to an ambient credential. (The old derive-from-`aliases` approach was unsafe beyond moonshot too: `gemini` aliases `google`, which would have derived `GOOGLE_API_KEY`.) New test `custom_api_key_env_override_stays_exclusive` reproduces your case and asserts it errors.

**[P2] Moonshot is symmetric.** Because `key_env_aliases` is consulted whenever the configured var is any known moonshot name, an init config (`api_key_env: KIMI_API_KEY`) now expands to `[KIMI_API_KEY, MOONSHOT_API_KEY]` and resolves when only `MOONSHOT_API_KEY` is set — and vice versa. `octos status` uses the same metadata and reports `set` for either. New test `moonshot_accepts_either_declared_key_name`; verified `MOONSHOT_API_KEY=x` and `KIMI_API_KEY=x` both show `set` via the binary.

Workspace suite: **5300 passed / 0 failed**; `cargo fmt --check` clean.

heyong4725:
Conflict resolved in 438978cf — merged `origin/main` into the branch. GitHub now reports the PR as **MERGEABLE**.

The `context.rs` collision was two `#[test]` fns added at the same insertion point. Kept both: #1661's `build_catalog_map_awards_bare_alias_to_native_then_lexicographically` and this PR's `should_deterministically_match_by_substring_mirroring_pricing`, each with its own `#[test]` attribute and closing brace. Production code auto-merged cleanly — `build_catalog_map` (from #1661) and `catalog_lookup_in` (this PR) coexist, and `catalog_lookup` delegates to `catalog_lookup_in`.

Verified on the merged tree: both context tests pass, `cargo fmt --check` clean, full workspace suite **5338 passed / 0 failed** (the higher count reflects main's newly merged tests).

heyong4725:
## Pre-landing review (multi-agent: testing / maintainability / security / adversarial + Codex cross-model)

**Scope: CLEAN.** Diff is exactly the 11 audit fixes + the reviewer-requested credential work.

Codex and the fresh-eyes adversarial pass **independently converged on three real bugs** — two introduced by the earlier credential-alias work — plus a set of missing regression tests. All fixed in `1fdc1a6b`.

### Fixed

**[Codex P1] `get_api_key_with_env` bypassed the sibling expansion.** The alias support I added only landed in `resolve_api_key`. The explicit-env entry point still discriminated custom-vs-default on `api_key_env` alone, so an init-generated Moonshot config (`api_key_env=KIMI_API_KEY`) resolved in `chat` (via `get_api_key`) but reported the key missing in `octos doctor`, the dashboard, and embedding config. Fix: route any registry-known key name (`api_key_env` **or** `key_env_aliases`) through the full chain.

**[Codex P1] Case-sensitivity leak.** The sibling-expansion gate used `eq_ignore_ascii_case`. On case-sensitive Unix a genuinely custom `kimi_api_key` (lowercase) matched the declared `KIMI_API_KEY`, so a missing custom key fell back to an ambient `MOONSHOT_API_KEY` — reopening the exact exclusivity leak round 2 closed. Now case-sensitive.

**[Codex P2] cron `--every` overflow moved downstream.** `checked_mul(1000)` passes for `i64::MAX/1000`, but `compute_next_run` does `now_ms + every_ms` unchecked → debug panic / release wrap to an immediately-due interval. Added a next-run-timestamp guard in `cmd_add`.

**[adversarial] Empty-string key consistency.** My round-2 empty-skip left `resolve_env_var_only` and `octos status` on the old policy — `status` showed a `KIMI_API_KEY=""` provider as "set" while `chat` failed. All three paths now treat empty as unset; error text says "not set or empty".

**[adversarial] office `validate` traversal.** The absolute-path fix used `strip_prefix('/')` (one slash) — `//etc/hosts` re-became absolute, and `..` escaped the package. Now strips all leading slashes and rejects `..`.

**maintainability.** Extracted `ProviderEntry::key_env_names` / `is_known_key_env` (registry) and used it in config + status, removing the duplicated key-name derivation; fixed a stale status comment.

### Verified safe (no change)
condition.rs UTF-8 decode (continuation bytes can't be mistaken for delimiters); provider.rs `char_indices().nth(200)` keeps exactly 200 chars; web_fetch `resolve_to_addrs` preserves the SSRF guarantee (all addrs pre-validated); context.rs matcher is byte-identical to pricing.rs and order-independent; grep erroring on an invalid glob is the right call.

### +13 regression tests
credential exclusivity / symmetry / case-sensitivity / process-env / `get_api_key_with_env` alias; cron overflow + not-found bail; init custom-model guard; office absolute / double-slash / traversal; grep invalid glob; catalog equal-length tie-break.

### Deferred to focused follow-ups (both pre-existing)
1. `resolve_api_key` checks the provider-keyed auth store **before** the candidate env chain, so provider=X + custom `base_url` + a prior `octos auth login -p X` sends the stored token to the custom endpoint. The embedding path already guards this; the primary path needs base_url awareness.
2. Load-time sanitizing of a cron store poisoned by a pre-fix release build (negative `every_ms` → hot-firing). Astronomically unlikely (interval near `i64::MAX`).

**Gate:** workspace **5350 passed / 0 failed**, `cargo fmt --check` clean, no new clippy.

heyong4725:
Both fixed in c81023fc — good catches, the first was a real regression I introduced.

**[P1] Valid parent-relative targets no longer rejected.** My blanket `..`-rejection was wrong: `../media/image1.png` and `../slideLayouts/...` are legal OPC and resolve *inside* the package — the CLI itself emits them (office.rs slide_rels). Replaced the check with the file's existing `normalize_path` helper, which lexically resolves `.`/`..` and returns the package-relative path (or `None` if it escapes). A target is now flagged only when its **normalized** path actually leaves `dir`. Added `validate_accepts_parent_relative_target_inside_package` (a `../media/image.png` package validates OK) alongside the existing escape (`/../outside.txt`) and double-slash tests — all green.

**[P2] Single `now` for cron.** `cmd_add` now captures one `now_ms` up front and reuses it for the overflow validation, `created_at_ms`, and `compute_next_run`, so a boundary-sized interval can't pass a check taken at one instant and overflow the addition taken at another.

Workspace suite: **5351 passed / 0 failed**; `cargo fmt --check` clean.

heyong4725:
Fixed in 1ee56168.

**[P1] Normalize the root before the containment check.** `normalize_path` normalized the target but stripped an un-normalized `dir`, so a `../media/image1.png` package passed under an absolute canonical root yet was falsely reported as escaping when validated via `octos office validate .` or `/tmp/../tmp/pkg`. Extracted `lexical_normalize` and now apply it to `base` as well as `path` before `strip_prefix`, so both sides are normalized. Added `validate_accepts_parent_relative_target_with_noncanonical_root` — builds a package under `<tmp>/pkg` and validates it via `<tmp>/pkg/../pkg`; the escape and absolute-path tests still pass.

Workspace suite: **5352 passed / 0 failed**; `cargo fmt --check` clean.

On the CI Clippy failure: agreed it's unrelated — it's from `doctor.rs` that PR #1675 added to main under Rust 1.97, not from this branch's changes. This branch merged main at 438978cf (before #1675 landed), so it doesn't carry that lint; a later rebase onto main would pick up #1675 and need its clippy fix separately.

heyong4725:
Fixed in 03cc7632 — you're right, the relative-root case slipped through.

`lexical_normalize` popped on every `..`, so an unmatched `..` (one with no preceding `Normal` component) silently vanished — under a relative root, `xl/../../secret` collapsed to `secret` and got accepted if `./secret` existed. Now:
- an unmatched `..` (climbing above a relative root) is **preserved**, not dropped;
- a `..` at a filesystem root is dropped (can't go above root);
- `normalize_path` rejects a normalized path that begins with `..`, since the empty relative-root base made a bare `strip_prefix` accept everything.

Added deterministic unit tests (no cwd games): `lexical_normalize_preserves_unmatched_parents` (`xl/../../secret` → `../secret`) and `normalize_path_rejects_relative_root_traversal` (`xl/../../secret` under `.` → `None`; a contained `../media` target under `.` still `Some`). The escape, absolute, double-slash, and non-canonical-root tests still pass.

Workspace suite: **5354 passed / 0 failed**; `cargo fmt --check` clean.

heyong4725:
Fixed in 237ee9b8.

The leading-`..` rejection ran before stripping the base, so a sibling-relative root (`octos office validate ../package`) broke: a contained `../media/image1.png` normalizes to `../package/media/image1.png`, which legitimately begins with `..`, and got rejected. Now `normalize_path` strips the normalized base first and rejects only if the **remaining** relative path begins with `..` — escapes still climb past the base and are caught, sibling-relative roots work.

Added `normalize_path_allows_containment_under_sibling_relative_root`: contained target under base `../package` → `Some(media/img.png)`; a target escaping past it → `None`. The relative-`.`-root, non-canonical, absolute, and double-slash tests still pass (9 office tests total).

Workspace suite: **5355 passed / 0 failed**; `cargo fmt --check` clean.

heyong4725:
Fixed in 40749d1c.

`lexical_normalize` suppressed `..` after any `Prefix`, but a bare Windows drive prefix `C:` is drive-RELATIVE (`C:package`, not `C:\package`) — a `..` after it can climb above the root and re-enter a same-named dir, which the drop masked as contained. Now `..` is suppressed only after `RootDir`; it's preserved after a bare prefix (and above a relative root, as before).

Added `normalize_path_rejects_drive_relative_escape_and_reenter` (`#[cfg(windows)]`): `C:package\xl\..\..\..\package\secret` under root `C:package` → `None`, while a contained `C:package\media\img.png` → `Some`. Unix has no `Prefix` components so its behavior is unchanged — all 9 Unix office tests still pass.

Workspace suite: **5355 passed / 0 failed**; `cargo fmt --check` clean. (The Windows-only test runs on Windows CI; native Windows checks were skipped in your run.)

PR #1679, opened 2026-07-16
fix(spawn): bind a native spawn delegate in spawned-child registries

ymote:
## Problem

A spawned subagent's tool registry was built from `with_builtins_and_sandbox` + plugins + the parent's `child_tool_factories` (`send_file`, `run_pipeline`) — but **never a native `spawn` tool**. So a child registry carried only the delegate-**less** builtin `spawn_agent`, and any *nested* spawn (a child agent spawning its own sub-agents) failed with:

```
No native Octos spawn tool is bound behind spawn_agent in this ToolRegistry.
```

`spawn_agent_without_delegate` registers the task then immediately `mark_failed`s it, leaving the child task **orphaned with empty `output_files`** — the exact symptom a live multi-round review session hit (its "second-round" agents completed-but-empty). This regressed nested spawn, which `MAX_SPAWN_DEPTH` (`parent → planner → coder → tts`, depth 3) treats as a supported production feature.

## Fix

Register a self-clone of the `SpawnTool` into each child registry (`child_spawn_clone`), at both the **sync** and **detached** construction sites. Registering a tool named `spawn` triggers the existing `ToolRegistry` swap that binds `spawn_agent` + `delegate` behind it, so a child can nest a further spawn — bounded by `MAX_SPAWN_DEPTH` via the child worker's incremented `ctx.spawn_depth`. The subagent policy still denies **direct** `spawn` (deny is exact-match, so `spawn_agent`/`delegate` stay allowed for a default child). The clone carries every wired field (llm, memory, routers, factories, sandbox, policies, plugin dirs, …) forward so nesting behaves identically at each level.

### Two subtleties, resolved after adversarial review

- **Supervisor consistency / isolation.** The spawn tool and the registry it lives in must share one supervisor (the top level wires `with_task_supervisor(tool_registry.supervisor())`). The clone is bound to the child registry's **own private** supervisor — the one `ctx.task_supervisor` exposes — so `SpawnAgentTool`'s before/after task lookup resolves the task the delegate registers. Keeping it **per-subtree and private** (rather than sharing the session supervisor) is deliberate: it stops `newest_spawned_task` from cross-correlating **concurrent sibling** spawners, keeps a child's `run_pipeline` node tasks out of the persisted session ledger, and confines the child's task-control aliases (`wait`/`close`/`resume_agent`) to their own subtree. **Trade-off:** grandchildren aren't surfaced in the session `task/list` and the fan-out cap is per-subtree; the grandchild's *result* still flows back via the inherited result sender / inline output. Deep-descendant dashboard visibility is a follow-up.
- **Context fork.** `child_prompt_context_manager_factory` is dropped in the clone. It captures the *original* parent session's context and keys the fork by `worker_id`; carried to a grandchild it would fork from the wrong ancestor and collide keys. Grandchildren run with a fresh context instead. Depth-1 children are unaffected (they use the un-cloned top-level tool).

## Tests

- `should_bind_native_spawn_in_child_registry_for_nested_delegation` — a child declaring `allowed_tools=["spawn"]` clears the `ensure_subagent_tools_available` preflight (RED before the fix: `required tool(s) not available on this host: spawn`).
- `child_spawn_clone_is_named_spawn_and_binds_spawn_agent_via_registry_swap` — the clone is named `spawn`, rebased on the child cwd, and its registration binds `spawn_agent` + `delegate`.
- `default_subagent_policy_allows_spawn_agent_and_delegate_but_denies_direct_spawn` — a default (empty allow-list) child keeps the nesting aliases.
- `nested_spawn_agent_resolves_a_real_agent_id_through_the_bound_delegate` — driving `spawn_agent` with `ctx.task_supervisor == the aligned registry supervisor` reaches the delegate (not "No native Octos spawn tool is bound") and resolves a concrete `agent_id`.

## Validation

- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean; full `octos-agent` lib suite (2107) + `delegate_tool`/`child_primitive_contracts` integration green.
- Four adversarial `codex` rounds; the design converged (supervisor isolation model) with the final state reported sound.

### DIFF
--- a/crates/octos-agent/src/tools/spawn.rs
+++ b/crates/octos-agent/src/tools/spawn.rs
@@ -983,6 +983,78 @@ impl SpawnTool {
         }
     }
 
+    /// Reconstruct an equivalent `SpawnTool` for a spawned child's own
+    /// registry, rebased on the child's working directory.
+    ///
+    /// Registering the returned tool under the name `spawn` triggers the
+    /// [`ToolRegistry::register`] swap that binds `spawn_agent` + `delegate`
+    /// behind it, so a subagent can nest a further spawn (bounded by
+    /// [`MAX_SPAWN_DEPTH`] via the child worker's incremented
+    /// `ToolContext::spawn_depth`). Without this the child registry carries
+    /// only the delegate-less builtin `spawn_agent`, and any nested spawn
+    /// fails with "No native Octos spawn tool is bound behind spawn_agent in
+    /// this ToolRegistry." — orphaning the child task with empty outputs.
+    ///
+    /// Every wired field (routers, factories, supervisor, sandbox, plugin
+    /// dirs, policies, …) is carried forward by `Arc`/value clone so nesting
+    /// works identically at each level; `worker_count` restarts at 0 because
+    /// each instance numbers only its own direct children, and `origin` is
+    /// snapshotted from the parent's current value.
+    fn child_spawn_clone(&self, working_dir: PathBuf) -> SpawnTool {
+        SpawnTool {
+            llm: self.llm.clone(),
+            memory: self.memory.clone(),
+            working_dir,
+            inbound_tx: self.inbound_tx.clone(),
+            origin: std::sync::Mutex::new(
+                self.origin
+                    .lock()
+                    .unwrap_or_else(|e| e.into_inner())
+                    .clone(),
+            ),
+            worker_count: AtomicU32::new(0),
+            provider_policy: self.provider_policy.clone(),
+            provider_router: self.provider_router.clone(),
+            worker_prompt: self.worker_prompt.clone(),
+            background_result_sender: self.background_result_sender.clone(),
+            child_session_sender: self.child_session_sender.clone(),
+            hooks: self.hooks.clone(),
+            hook_context_template: self.hook_context_template.clone(),
+            plugin_dirs: self.plugin_dirs.clone(),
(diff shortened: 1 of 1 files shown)


ymote:
## Live soak — mini5, real DeepSeek LLM ✅

Deployed the built binary (`octos 2.0.1 (8d57077e4)`) to mini5 and forced a 2-level nested spawn through the real agent loop:

```
root → spawn_agent(lead, sync) → lead → spawn_agent(worker, sync) → worker → shell(printf marker)
```

The **lead** (a child agent) calling `spawn_agent` to spawn the **worker** is exactly the operation that previously failed. Results:

- **Ground-truth marker written**: the nested worker (depth 2) ran a real shell tool and produced `/tmp/…-marker.txt` → `BANANA-NESTED-OK`. Pre-fix this file would not exist (the lead's `spawn_agent` would error before the worker ran).
- **`No native Octos spawn tool is bound` occurrences: 0.**
- Logs show the nesting explicitly: `spawning subagent … task=You are the LEAD…` → LEAD emits `tool_calls=1 tool_names=spawn_agent` → `spawning subagent … task=Use the shell tool …`.
- Both nested tasks logged `task completed … contract_failed=false` (distinct task_ids, worker completed before lead — correct inner→outer order); no orphaned/empty-output tasks.
- Root returned `ROOT-DONE`.

All soak artifacts + the test binary were scrubbed from mini5.

PR #1701, opened 2026-07-17
feat(spawn): surface shell-written deliverables via a per-task workspace contract

ymote:
## Summary

Closes the **primary** cause of the mini4 "sub-agent code reviews produced zero deliverables" incident: workers wrote their reports with a raw `shell` heredoc (`cat > octos-review.md <<'REPORT_EOF'`) because their `allowed_tools` had **no `write_file`** — and `output_files` is built *only* from tool-reported `write_file`/`edit_file` `file_modified`, with **no filesystem scan**, so the deliverables never surfaced (ledger `output_files: []`).

Rather than invent a bespoke scan, this reuses the **existing workspace-contract artifact mechanism** — which is already a hardened, glob-based, disk-resolving artifact hook — just wired for general spawns instead of only slides/sites workflows.

## What it does

Adds an opt-in `deliverable` glob to the spawn `Input`. When set, a background spawn:

1. runs in a **fresh per-task output directory** seeded with a workspace-contract policy declaring `primary = <glob>`;
2. is told (via the worker prompt) to write its deliverable into that cwd — *"a shell redirect/heredoc into the cwd works; you do NOT need write_file"*;
3. on completion, surfaces whatever matches the glob through the existing `resolve_workspace_contract_artifact_paths` — the **same** resolver the slides/sites contracts use — so a file written by **any** means (shell heredoc, `write_file`, a plugin) lands in `output_files`.

```jsonc
spawn({ "task": "review octos-tui, write findings to a markdown file",
        "allowed_tools": ["shell", "read_file", "grep", "glob"],
        "deliverable": "*-review.md" })   // <- new
```

## Design notes (why this shape)

- **Reuses the harness, doesn't fork it.** `resolve_workspace_contract_artifact_paths` reads the seeded policy's declared glob and resolves matches on disk. It is kind-agnostic, so it works for a plain document.
- **Deliberately skips `inspect_workspace_contract_at_root`.** That wrapper is slides/sites project-kind gated (`infer_kind_from_root` + `matches_project_kind`) and runs pptx/mp3 content validators that would wrongly reject a `.md`. We want the artifact-glob resolution, not the deck/audio validators.
- **Fresh dir ⇒ no mtime baseline needed.** The output dir is created per task, so a plain glob is exactly this run's outputs. (Codex, for comparison, is tool-record-only for turns and has the same shell-write blind spot; its only real dir-scan is scoped to the memories workspace.)
- **Pins where deliverables go.** Detection alone is unreliable (the model may write anywhere — mini4 wrote to `/tmp`); making the child's cwd the seeded, scanned dir + a prompt directive is what makes surfacing reliable.
- **Guards:** empty `deliverable` ⇒ `*` (top-level only, so a repo cloned into the output dir isn't swept in); `deliverable` + worktree isolation is rejected (the child's session scope would be rooted at the worktree while its cwd is the deliverable dir).

## Tests (6 new, all green)

- unit: glob normalization; contract surfaces a shell-written file that **no tool reported** (and the tool-record path is blind to it — the bug); a precise glob excludes a nested clone's `README.md`.
- **end-to-end**: a real background spawn drives a child that writes via `shell` (no `file_modified`) — `output_files` carries the file **with** `deliverable`, and is **empty without** it (control).

Full `octos-agent` suite **2115 pass**; clippy + fmt clean; `octos-cli` builds.

Companion to #1692 (the preflight / input-error non-cascade / budget-reminder fixes from the same mini4 forensic). This PR is the surfacing half.

### DIFF
--- a/crates/octos-agent/src/tools/spawn.rs
+++ b/crates/octos-agent/src/tools/spawn.rs
@@ -1504,6 +1504,19 @@ struct Input {
     /// Inline always wins.
     #[serde(default)]
     agent_definition_id: Option<String>,
+    /// Glob (relative to a dedicated per-task output directory) matching the
+    /// deliverable file(s) this spawn is expected to produce. When set, the
+    /// child runs in a FRESH output directory seeded with a workspace-contract
+    /// artifact declaration and is told to write its deliverable there; any
+    /// matching file it leaves — written by ANY means, including a raw `shell`
+    /// heredoc that reports no `file_modified` — is surfaced as the task's
+    /// `output_files` via the existing workspace-contract artifact resolver.
+    /// Empty string is treated as `*` (top-level files). Absent = legacy
+    /// behaviour: `output_files` come only from `write_file`/`edit_file` tool
+    /// records, so a shell-written deliverable would not surface (issue: the
+    /// mini4 code reviews wrote via shell and reported no output_files).
+    #[serde(default)]
+    deliverable: Option<String>,
 }
 
 fn default_backend() -> String {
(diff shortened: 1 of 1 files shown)


ymote:
Added the **auto-materialize salvage** on top of the seeded-deliverable contract (`1d921ff7a`).

**Live mini4 TUI soak (MiniMax-M3, two parallel reviewer children):** with the seeded per-task cwd + strengthened `DELIVERABLE_WORKER_DIRECTIVE`, both children now write their `*-review.md` **directly** into their seeded dir — auto-materialize never had to fire. It remains as a belt-and-suspenders salvage: a child that declares a deliverable but returns its work as inline final text (≥400B) gets that text written to a glob-matching path (`derive_deliverable_filename`) so the artifact still exists and the output pipeline surfaces it. Two earlier runs lost work exactly this way (a 14 KB review returned as final text with `files_modified=0`; another explore-looped to the iteration cap).

Both new tests green; full `octos-agent` lib suite 2117 passed.

Issue #1712, opened 2026-07-17
Truncated turns (finish_reason=length) on a tool-call are not recovered


ymote:
## Symptom

Models that stream reasoning in-band (inline `<think>`) spend output budget on reasoning; a large tool call can exceed `max_tokens` and truncate mid-JSON (`finish_reason=length`). octos consumes the truncated turn as-is — `loop_runner` only emits a `[partial output: max_output_tokens reached before a final answer]` marker — so the partial/truncated tool call flows downstream (and, absent repair, becomes a fatal provider 400; see the sibling issue).

## Root cause

There is no `StopReason::MaxTokens` recovery path for a turn that ends mid-tool-call. Separately, the chat path defaults `max_tokens` to `4096` when unset (`openai.rs` `config.max_tokens.or(Some(4096))`), which is too small for a reasoning model that emits its chain-of-thought in-band before a large tool call.

## Fix direction

- Detect `StopReason::MaxTokens` when the turn carries a (partial) tool call and **re-request/continue** rather than shipping the partial call. Bound the retries.
- Make the per-turn output budget model-aware (avoid the tiny 4096 fallback for reasoning models), analogous to the compaction path already using `provider.max_output_tokens()`.

The sibling repair issue makes truncation non-fatal even without this; this issue removes the truncation in the first place.

ymote:
## Reproduced live, and it's more serious than "just raise the budget"

Forced the tight-output-budget condition on a live serve run (Kimi-K3 reviewer children, per-turn output budget clamped low). The truncation reproduces deterministically and **kills the background task**:

```
WARN task: malformed tool call JSON — surfacing as StreamError::MalformedArgs
     tool=write_file tool_id=write_file_26 error=EOF while parsing a string at line 1 column 4973
WARN task: harness error classified variant="internal" recovery=bug
     error=malformed tool_call arguments (tool=write_file, id=write_file_26): EOF while parsing a string
→ task status: failed
```

### Two findings

1. **This is a DIFFERENT code path than #1711.** #1711 repairs the *inline* `<invoke>` tool-call path (`detection.rs`). But octos **streams**, and a truncated **native** streaming tool call fails hard at `agent/streaming.rs` — `serde_json::from_str(&args)` fails → `StreamError::MalformedArgs`. In the repro, the #1711 inline-repair path fired **0** times; native `MalformedArgs` fired every time. The native streaming path is the one that actually kills review children under budget pressure.

[...]

Issue #1690, opened 2026-07-16
[agents] "invalid spawn tool input" gives the model no repair hint (serde source dropped)


ymote:
## Symptom

When a model malforms `spawn` arguments, the tool result returned to it is the bare string:

```
Error: invalid spawn tool input
```

No field name, no schema, no hint. The model cannot self-correct and often repeats the same mistake (observed 5× in one session historically). In the mini4 code-review run, the very first turn's parallel `spawn` died this way — `WARN tool failed tool=spawn error=invalid spawn tool input variant="tool_execution" recovery=fail_fast duration_ms=0` — and (via the serial-batch cancellation, #1508) took its well-formed sibling spawn down with it, so **both** reviews failed to launch on turn 1 and had to be retried sequentially.

## Root cause

`crates/octos-agent/src/tools/spawn.rs:2037`:

```rust
let mut input: Input =
    serde_json::from_value(args.clone()).wrap_err("invalid spawn tool input")?;
```

The `Err` bubbles to the tool-error arm in `execution.rs`, which renders it to the model as `format!("Error: {e}")`. eyre's **non-alternate** `Display` prints only the top context (`invalid spawn tool input`) and drops the underlying serde **source** (e.g. `missing field 'task'`, `invalid type: string, expected a sequence`).

## Fix

Surface the serde detail and the expected schema so the model can repair on the next turn, e.g.:

```
invalid spawn tool input: missing field `task`.
Required: task. Optional: allowed_tools, role, mode ("sync"|"background"), context, ...
```

Either fold the source into the `wrap_err` message at the spawn site, or render the tool-error arm with `{e:#}` (alternate/chained) for input-validation failures. Pairs with the sibling-cancellation fix (#1508) so one malformed call no longer nukes a whole batch.

ymote:
Fixed by #1692 (`ToolInputError` carries a repair hint; malformed spawn input surfaces an actionable message the model can self-correct on). Deployed to mini5; pending merge to main.

PR #1713, opened 2026-07-17
fix(agents): repair malformed/truncated inline tool-call args instead of failing the task (#1711)

### LINKED ISSUE #1711: Malformed/truncated inline tool-call arguments cause a fatal provider HTTP 400 (should repair + coerce)
## Symptom

A spawned reviewer child on MiniMax-M3 failed hard with:

```
API error (minimax@api/MiniMax-M3): invalid request — HTTP 400
{"type":"error","error":{"type":"bad_request_error",
 "message":"invalid params, invalid function arguments json string, tool_call_id: call_inline_0_shell (2013)","http_code":"400"}}
```

The whole background task failed. HTTP 400 is non-retryable, so there is no recovery — one malformed tool call kills the task.

## Root cause (traced)

Some chat/completions models emit tool calls **inline in content** rather than as native `tool_calls`; octos recovers those via `agent/detection.rs` (`build_tool_call_from_invoke`, the `call_inline_<n>_<tool>` ids). When such a model also streams its reasoning inline (`<think>` in content), the reasoning competes with the tool call for the same output budget and can (a) truncate the tool-call JSON mid-string, or (b) emit slightly-malformed JSON (trailing commas, unescaped control chars).

`build_tool_call_from_invoke` (detection.rs ~188-193) currently does:

```rust
serde_json::from_str(raw_args)
    .unwrap_or_else(|_| serde_json::Value::String(raw_args.into()))  // bare string on parse failure
```

So an un-parseable arg is stored as a JSON **string**, not an object. On the next turn, `openai.rs:344` serializes it via `tc.arguments.to_string()` → `function.arguments` becomes a JSON string that does not decode to the parameters **object** the provider requires → the provider rejects the entire request with HTTP 400 → the task dies.

## Evidence

A 45-call tool-call-JSON probe matrix (3 reasoning models × simple/heredoc/big-report, streaming, 34 tools) reproduces this: models that stream reasoning **in-band** produce `structured_badjson` / `inline_in_content` (finish_reason=length) at a meaningful rate under a tight output budget, while a model that keeps reasoning in a **separate field** never trips it. Raising the budget makes it recover — but a malformed/truncated call must **never** be fatal.

## Fix

1. **Repair the inline args JSON** in `detection.rs`: lenient parse; strip trailing commas; **close unclosed `{}`/`[]`** (salvages truncation); re-serialize. If unrepairable, fall back to a valid empty object `{}`. **Never store a non-object as tool-call arguments.**
2. **Coerce at the serialization boundary** (`openai.rs:344`): if `arguments` is not a JSON object, emit `{}` — so no provider can 400 on it regardless of how the args were produced.

Net effect: a malformed/truncated tool call degrades to an ordinary tool-input error the model retries (tool-scoped, per #1690) instead of a fatal HTTP 400.

ymote:
Fixes #1711.

## Problem

Some chat/completions models emit tool calls **inline in content** (recovered by `agent/detection.rs`, the `call_inline_*` ids). When such a model streams its reasoning **in-band** and runs low on output budget, the tool-call JSON is truncated mid-object/mid-string or comes out slightly malformed. `build_tool_call_from_invoke` stored an un-parseable arg as a bare `Value::String`, and `openai.rs:344` serialized it back verbatim — so `function.arguments` no longer decoded to an object, and the provider rejected the **entire request** with a **non-retryable HTTP 400**. One malformed call killed the whole task:

```
invalid params, invalid function arguments json string, tool_call_id: call_inline_0_shell (2013)
```

## Fix (two layers)

1. **`detection.rs` — `repair_tool_arguments_to_object()`**: lenient parse, then a single string-aware repair pass that
   - escapes literal control chars inside string values (heredoc newlines/tabs),
   - drops trailing commas before `}`/`]`,
   - closes any object/array/string left open by truncation.

   It **always returns a JSON object**; unrepairable input degrades to `{}`.
2. **`openai.rs` — `tool_call_arguments_to_wire()`**: coerces any non-object `arguments` to `{}` at the request boundary (recovering a stringified object first). Objects pass through byte-identically, so it is a no-op on the normal path — but no provider can 400 on a recovered/malformed call anymore.

Net effect: a malformed or truncated tool call degrades to an ordinary tool-input error the model retries (tool-scoped, per #1690) — **never** a fatal request rejection.

## Why not just raise the token budget

Truncation is one trigger; a model can also emit complete-but-malformed inline JSON. This makes octos robust to **any** provider's tool-call quirks regardless of budget. Removing the truncation in the first place (`finish_reason=length` recovery + model-aware budget) is the sibling follow-up #1712.

## Tests
12 new: repair passes through valid objects; salvages truncated objects, truncated-mid-heredoc strings, unescaped control chars, trailing commas; degrades non-objects to `{}`; end-to-end through the inline detector; and the wire coerce (object passthrough, non-object → `{}`, stringified-object recovery). Full `octos-agent` (2119) and `octos-llm` (468) suites pass; clippy clean.

### DIFF
--- a/crates/octos-agent/src/agent/detection.rs
+++ b/crates/octos-agent/src/agent/detection.rs
@@ -170,6 +170,124 @@ fn extract_inline_invokes(content: &str) -> (String, Vec<ToolCall>) {
     (cleaned, calls)
 }
 
+/// Coerce inline tool-call argument text into a JSON **object**, repairing the
+/// malformations that in-band-reasoning models produce under a tight output
+/// budget: JSON truncated mid-object or mid-string (`finish_reason=length`),
+/// trailing commas, and literal (unescaped) control characters inside string
+/// values (common with heredoc `command` payloads).
+///
+/// The result is ALWAYS a JSON object. A valid object echoed back to a provider
+/// can never trigger the "invalid function arguments json string" HTTP 400 that
+/// a bare string or truncated fragment would (that 400 is non-retryable and
+/// kills the whole task). An unrepairable arg degrades to `{}`, which surfaces
+/// as an ordinary tool-input error the model can retry — never a fatal request
+/// rejection. See #1711.
+fn repair_tool_arguments_to_object(raw: &str) -> serde_json::Value {
+    let trimmed = raw.trim();
+    if trimmed.is_empty() {
+        return serde_json::json!({});
+    }
+    // Fast path: already a valid object.
+    if let Ok(value) = serde_json::from_str::<serde_json::Value>(trimmed)
+        && value.is_object()
+    {
+        return value;
+    }
+    // Repair pass: escape literal control chars, drop trailing commas, and
+    // close any structure/string left open by truncation, then re-parse.
+    let repaired = repair_json_fragment(trimmed);
+    if let Ok(value) = serde_json::from_str::<serde_json::Value>(&repaired)
+        && value.is_object()
+    {
+        return value;
+    }
+    // Unrepairable → empty object (never a fatal non-object on the wire).
+    serde_json::json!({})
+}
+
+/// String-aware JSON repair for [`repair_tool_arguments_to_object`]. Single
+/// left-to-right scan that (1) escapes literal control chars inside string
+/// values, (2) removes trailing commas before `}`/`]`, and (3) closes any
+/// string/object/array left open by mid-stream truncation. Operates only
+/// outside string literals for structural edits so it never corrupts content.
+fn repair_json_fragment(input: &str) -> String {
+    let chars: Vec<char> = input.chars().collect();
+    let mut out = String::with_capacity(input.len() + 8);
+    let mut stack: Vec<char> = Vec::new();
+    let mut in_string = false;
+    let mut escaped = false;
+    for i in 0..chars.len() {
+        let c = chars[i];
+        if in_string {
+            if escaped {
+                out.push(c);
+                escaped = false;
+            } else if c == '\\' {
+                out.push(c);
+                escaped = true;
+            } else if c == '"' {
+                out.push(c);
+                in_string = false;
+            } else if (c as u32) < 0x20 {
+                // Literal control char inside a string → escape it so the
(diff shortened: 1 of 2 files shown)


ymote:
Added `<parameter name="x">value</parameter>` inline-format parsing (`9bfa7a45b`).

**Why:** re-running the 3-model probe (bigreport@2048, streaming, max thinking, n=20) surfaced that Kimi-K3 emits ~15% of its tool calls in the Anthropic XML shape `<invoke name="shell"><parameter name="command">…</parameter></invoke>` — no JSON body. The inline detector only understood a JSON body / `arguments=` attr, so those recovered with empty `{}` and the command was silently dropped.

**Validated against the real captured Kimi output**, not just synthetic tests: the parser recovers the `command` (full heredoc intact) on every case that has complete `<invoke>…</invoke>` tags. Cases missing the closing tag (truncated / early-stop) remain #1712's domain.

**Updated effectiveness (Kimi-K3, n=20, the worst model):**
| | raw | with #1713 |
|---|---|---|
| valid tool call | 10/20 | **~16/20** (3 truncated-JSON + 1 inline-JSON repaired, 2 `<parameter>` recovered) |
| **fatal HTTP 400** | up to 10/20 | **0/20** |

Remaining ~4 are pure mid-`<think>` truncation (no call emitted) → #1712 (budget/continuation). DeepSeek stays 20/20; MiniMax 18/20 raw, 0 fatal. 5 new tests; `octos-agent` 2124 pass, clippy clean.

ymote:
**Scope note (coverage boundary):** this PR covers the **inline** tool-call path (`detection.rs` `<invoke>` recovery) + the request-wire boundary (`openai.rs`). A forced-truncation live soak (Kimi-K3, clamped output budget) surfaced that **native streaming** tool calls that truncate fail on a *separate* path — `agent/streaming.rs` `StreamError::MalformedArgs` — which this PR intentionally does not touch (that path is a deliberate #1355 invariant). That native-truncation case is tracked and reproduced in #1712, with the fix direction there (distinguish truncation from a genuine malformed call rather than silently repairing). This PR remains correct and complete for the inline path + wire coercion it targets.

PR #1716, opened 2026-07-17
feat(agents): per-spawn max_iterations override for repo-scale sub-agents

ymote:
## Problem (observed live)

A spawned sub-agent is hardcoded to `AgentConfig`'s default **50** iterations, with no per-spawn override — the orchestrator can set `allowed_tools`, `model`, `context_window`, `role`, etc. per spawn, but **not the turn budget**. A broad from-scratch repo review runs one tool call per iteration (read_file / shell / grep over big trees), so it needs ~100–150 turns. At 50, those lanes exhaust the cap and produce **nothing** — in a live 6-agent review, `review-extensibility` / `review-security` / `review-runtime-architecture` left no artifacts while the lanes that fit wrote **34–52 KB** reviews. The model *knew* it needed more turns but couldn't ask.

## Fix

Add an optional `max_iterations` to the spawn `Input` + JSON schema:
- Clamped to `MAX_SPAWN_MAX_ITERATIONS` (300) via `clamp_spawn_max_iterations` — repo-scale work is legitimate, but unbounded is a runaway-loop / cost footgun.
- Wired into **both** worker-config paths (sync and detached background).
- `None` (not requested) leaves the worker's default untouched — existing behavior is **byte-identical**; the no-config sync path still skips `with_config` entirely.
- An inline value survives manifest (`agent_definition`) layering (inline wins).

## Tests
5 new: clamp bounds (None→None, 0→1, over-ceiling→300, passthrough); JSON parse + default-None; schema exposes an `integer` `max_iterations`; inline value preserved through `apply_agent_definition`. Full `octos-agent` suite (2114) passes; clippy clean.

## Usage
The orchestrator can now issue e.g. `spawn(..., max_iterations: 150)` for a repo-scale review lane, so a capped-out agent is no longer the default failure mode for multi-agent code review.

### DIFF
--- a/crates/octos-agent/src/agent/budget.rs
+++ b/crates/octos-agent/src/agent/budget.rs
@@ -57,7 +57,11 @@ impl BudgetStop {
                      research/multi-step task, ask the agent to use \
                      `run_pipeline` (deep research) which delegates the \
                      work to specialised sub-agents instead of iterating \
-                     one tool call at a time."
+                     one tool call at a time. If this ran as a spawned \
+                     sub-agent (e.g. a repo-scale review), re-spawn it with a \
+                     higher `max_iterations` — the spawn tool accepts a \
+                     `max_iterations` field (up to 300); a broad from-scratch \
+                     review typically needs ~100–150."
                 )
             }
             Self::MaxTokens { used, limit } => {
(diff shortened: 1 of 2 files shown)


ymote:
Extended per a live finding (`e0a780e23`): the field alone wasn't enough — orchestrators don't set `max_iterations` unless told, so background sub-agents kept inheriting the bare `AgentConfig` default of **50** (an *interactive*-turn default) and starved on their first exploration pass (live: `budget exhausted ... iteration=50` across a dozen review subagents, producing nothing).

Two additions so sub-agent budgets are sensible **out of the box**:
1. **`DEFAULT_SPAWN_MAX_ITERATIONS = 150`** — a spawn that doesn't set `max_iterations` now gets a budget matched to bounded-but-substantive background work, not 50. Token budget + loop detection remain the real runaway guards; the iteration cap is a secondary backstop, so a generous spawn default trades a little worst-case headroom for first-try success on the common substantive-task case. Applied in both sync and background paths via `resolve_spawn_max_iterations` (caller value clamped to 300, else the default). **Interactive turns unchanged.**
2. **Self-correcting cap message** — the `MaxIterations` budget message now names the `max_iterations` lever ("re-spawn with a higher `max_iterations` … ~100–150"), so a parent whose child *still* caps out learns to raise the budget instead of the lever staying invisible.

Answers "why cap sub-agents at all": the cap is a legitimate runaway guard, but 50 was a legacy interactive default applied blindly to background work.

PR #1728, opened 2026-07-18
refactor(api): rename ui_protocol.rs -> ui_protocol_transport.rs (kill the same-name confusion)

ymote:
## Why
Two files were named `ui_protocol.rs`:
- `octos-core/src/ui_protocol.rs` — the protocol **type definitions** (203 structs/enums; the shared wire vocabulary every frontend + the server import).
- `octos-cli/src/api/ui_protocol.rs` — the axum/WebSocket **transport** that serves it (0 type defs, ~598 fns; imports its types from `octos_core::ui_protocol`).

That's correct layering, but the identical filename misleads. A recent architecture review read the two same-named files as a *"forked protocol definition"* — the exact opposite of reality (there is **one** definition, in core). Renaming the transport module to match its own doc comment (`//! UI Protocol v1 WebSocket transport`) and the sibling `ui_protocol_*` family makes the layering self-evident from the file tree.

## What
Pure rename, no behavior change:
- `git mv api/ui_protocol.rs api/ui_protocol_transport.rs` (tracked as a rename).
- Updated the 35 crate-local module path refs (`super::` / `api::` / bare `ui_protocol::`) + the `mod` / `use` declarations.
- **Left untouched:** 522 `octos_core::ui_protocol::` type imports (the core types) and 37 `"octos::ui_protocol::…"` tracing-target strings (log namespace stays stable for `RUST_LOG` filters).

## Verify
- `cargo build -p octos-cli --features api` — green (compiler validates every renamed ref).
- Diff is a rename + identifier updates only.


anantheparty:
The naming goal is still reasonable, but this branch is no longer a safe mechanical rename candidate. It is 136 commits behind current `main`, and the original ~51k-line transport file has since been rewritten/split across 31 commits (roughly +30.6k/-48.0k lines). The current conflict therefore cannot be resolved by preserving the old rename tree without replaying substantial stale structure.

I recommend closing this PR rather than preparing a conflict helper. If the naming improvement is still wanted, reopen it from current `main` as a fresh mechanical rename so the diff remains reviewable and behavior-neutral.

anantheparty (dismissed):
2026-08-15 巡检：纯机械重命名（14 文件 similarity 100%），CI 失败均为 main 漂移与 clippy 1.97 新 lint（非本 PR 文件）。合并前请 rebase 到最新 main 并重跑 CI，确认无残留旧引用。

Issue #1666, opened 2026-07-13
Per-project sessions (appui.sessions_in_cwd): harden the flag-ON path before enabling


ymote:
Per-project sessions (`<cwd>/.octos/sessions/`) merged **default-OFF** (#1662 server, octos-tui #306 client). The flag-OFF path is byte-identical (regression-tested); the following codex findings are all in the **flag-ON** path and MUST be fixed before `appui.sessions_in_cwd` is ever defaulted on:

**Server (from the re-review of the P1 fixes):**
- Cache-root change is not propagated to other session-scoped state — reused session IDs across projects can still cross-wire requests and events (beyond the cache key itself).
- `session/list` cwd profile resolution + non-canonical newly-created cwd hints can select the wrong storage root.
- Residual (documented in-code): two clients opening the same `session_id` with different cwds in one process — the bare-session-id read path (`session_workspaces()`) is last-wins; fully closing it needs per-connection cwd tracking or unique client ids.

**Client (octos-tui):**
- Scoped `session/list` results can't be reliably resumed after a backend restart, and the scope may be derived from a stale/client-local cwd — breaks `/resume` in per-project configs. Carry the list cwd through session hydration.

None of these affect users today (flag is off). Track to completion before enabling.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ymote:
## Live soak-test results (flag ON, real stack: octos-tui → `octos serve --stdio` → glm-5.2 via z.ai)

Soaked the merged Phase 4 with the **real** client/server stack (not unit tests): built octos + octos-tui from `main` (`1b19df9b6`), onboarded a `glm` profile, drove real LLM turns through tmux across multiple cwds.

### ✅ What works
- **Flag ON relocation**: transcript + per-cwd context ledger land in `<cwd>/.octos/<profile>/sessions/` and `<cwd>/.octos/<profile>/context_ledgers/`, with an auto-managed `.gitignore`. Verified `projA/.octos/glm/sessions/…jsonl` held only its own turn.
- **Transcript isolation**: projX stored codeword *watermelon*, projY stored *harmonica* — the two `.jsonl` files are cleanly separated (each contains only its own word).
- **Flag OFF regression**: default keeps transcripts in the profile-global `profiles/<id>/data/sessions/`; the cwd stays **clean** (no `.octos` created). Byte-identical-behavior claim holds live.
- **Resume**: relaunch in a cwd hydrates the visible prior history.

[...]

PR #1741, opened 2026-07-19
fix(goal): cwd-scope the goal store so goals don't leak across folders

ymote:
## Root cause

The TUI opens the **same** wire session key `<profile>:local:tui#coding` for **every folder (cwd)** under a profile — the `coding` topic is hardcoded and the cwd never enters the wire key. #1666 cwd-scoped the **ledger/transcript** store (`storage_session_id` → `<key>\u{0}~cwd-<sha16>`, registered on `session/open`) so a fresh folder shows an empty transcript — but the **goal/autonomy** store was still keyed by the **bare wire key** (`state.goals: HashMap<SessionKey, _>`; persisted `autonomy-goal:<session_key>`). So a goal set in folder A leaked into a fresh session in folder B (same profile) — same key → same goal → same "orchestrating" chip. A **different profile** did not leak (different key). This is the exact residue of the #1666 fix (which cwd-scoped the ledger but not the goal store).

Live repro (mini2): a session in cwd `work` and a fresh session in cwd `work2` (same profile `octos`) both got byte-identical key `octos:local:tui#coding`; `work2` showed 0 messages (ledger scoped) but the leaked goal chip (goal store un-scoped).

## Approach — (A) store-scoping, mirroring #1666

The orchestrator gets its own per-project scope map (`goal_scopes`) + `set_goal_scope` + `scoped_goal_key` — a byte-for-byte mirror of the ledger's `scopes` map / `storage_session_id` (same NUL-separated, injective `\u{0}~cwd-<scope>` encoding). It is registered on `session/open` right beside `ledger.set_session_scope`, in lock-step, so the goal store and the ledger agree on each session's cwd scope. The scope is `Some` **only** when the session is cwd-relocated (`sessions_root != profile.data_dir`, i.e. `appui.sessions_in_cwd` ON) — with the flag off (default) the goal store keys by the plain wire id, **byte-identical to before** and to the gateway/session-actor path (which never registers a scope).

**Every goal-keying site that receives a WIRE key resolves it to the scoped store identity:** `set_goal` / `get_goal` / `clear_goal` (RPC), and the model tools `model_goal_snapshot` / `model_transition_goal` / `model_create_goal`. The store, the master-continuation queue, the in-flight set, and persistence (`persist_goal_state` → `autonomy_goal_group_id`) all key on the scoped identity; the wire-facing responses/events echo the plain wire id (`session_goal_updated_event_json` and `sessions_with_active_orchestration` strip back to it so the chip still lights).

## How continuations stay intact (the hard constraint)

Goal continuations are autonomous: the orchestrator sweeps `state.goals` (`due_loop_targets`) and enqueues continuations that the AppUI tick drains into `run_standalone_turn` (the serve/TUI path; the gateway path uses `SessionActorManager`, keyed by distinct channel keys, never cwd-scoped). If the goal were stored under a scoped key but dispatch still used the bare wire key, continuations would never fire.

- The sweep surfaces the **scoped** store key as the target (goals/queue/in-flight are all scoped-consistent, so `pending_continuation_is_schedulable` and the drain match).
- At the AppUI dispatch boundary (`maybe_spawn_appui_master_continuation_runner`, both drain sites) the target is **stripped back to the plain wire key** (`wire_key_from_goal_key`) for everything keyed by the wire id — the session runtime, ledger, `active_turns`, `session_workspaces()` — while the goal **record / queue / in-flight / post-turn accountant** stay on the scoped key. The scoped key rides through `GoalContinuationContext.goal_session_key`, so the post-turn `record_goal_dispatch_timestamp_only` / `record_goal_turn` / `maybe_complete_goal_from_model` charge the **scoped** goal instead of a wire-keyed miss (a miss would let the goal recur past its budget forever).
- A scoped goal is dispatchable **only for the currently-active folder** of its wire id (`goal_target_is_dispatchable`, last-write-wins like the ledger/`session_workspaces()`), closing the continuation-side twin of the leak: a backgrounded folder's goal can't fire a continuation into the active folder's workspace.

**Goal-keying sites scoped:** `set_goal`, `get_goal`, `clear_goal`, `model_goal_snapshot`, `model_transition_goal`, `model_create_goal`, the continuation sweep + `enqueue_goal_continuation`/`enqueue_goal_wrap_up`, `record_goal_turn` / `record_goal_dispatch_timestamp_only` / `maybe_complete_goal_from_model` (via `goal_session_key`), `persist_goal_state` / `autonomy_goal_group_id`, and the dispatch gate + wire-key strip in both AppUI drain sites. `session_goal_updated_event_json` and `sessions_with_active_orchestration` look up by the scoped key but emit the wire id.

## Tests (agent_orchestrator)

1. `should_isolate_goal_store_between_cwd_scopes_sharing_a_wire_key` — mirrors `ui_protocol_ledger::should_isolate_ledger_storage_between_cwd_scopes_sharing_a_wire_key`: two goals under the same wire key + different cwd scopes store/retrieve independently; folder B never reads folder A's goal; the two occupy distinct NUL-separated keys; the **bare wire key holds nothing** (leak vector closed); the response echoes the plain wire id.
2. `scoped_goal_continuation_is_swept_and_strips_to_wire_key_for_dispatch` — the scoped goal is (1) surfaced by the sweep under its scoped key, (2) dispatchable only while its folder is active, (3) stripped back to the wire key the actor is keyed by, (4) drained + charged under the scoped key exactly once. Proves the scoping did NOT break the fire path.

`cargo test -p octos-cli --features api --lib goal` → **61 passed** (was 59). Full `octos-cli` lib suite green (2367). `cargo fmt`/`clippy` clean.

## Risk / edges

- **Gated:** scope is `Some` only for cwd-relocated sessions (`sessions_in_cwd` ON). Flag-off installs are byte-identical; the gateway/session-actor path is unchanged (scope always `None`).
- **Simultaneous same-wire-key folders:** the scope map is last-write-wins per wire id (inherent to #1666 — the wire key is client-minted and stable). Two folders open at once resolve to the most-recently-opened cwd for reads and continuation firing; each folder's goal still **persists** isolated under its own scoped key. Matches the ledger's existing behavior.
- **Migration:** a goal persisted under a bare wire key *before* this change, in a cwd-relocated session, becomes invisible to the now-scoped `goal_get` (its continuation could still fire once under the wire key). Narrow — only affects users who ran `sessions_in_cwd` ON with an active goal across the upgrade; goals are ephemeral.
- **Loops** (`loop/create`) remain wire-keyed — out of scope for this goal-store fix; a possible follow-up if loops show the same cross-folder leak.

### DIFF
--- a/crates/octos-cli/src/api/agent_orchestrator.rs
+++ b/crates/octos-cli/src/api/agent_orchestrator.rs
@@ -358,6 +358,32 @@ pub(crate) fn method_not_supported_error(
 #[derive(Debug, Default)]
 pub(crate) struct InProcessAgentOrchestrator {
     state: StdMutex<AutonomyRuntimeState>,
+    /// #1666 residue — per-project (cwd) scope for the goal/autonomy store,
+    /// mirroring the ledger's `scopes` map (`ui_protocol_ledger.rs`).
+    /// Registered on `session/open` alongside `ledger.set_session_scope` so a
+    /// goal set in one folder does NOT leak into a fresh session that reuses
+    /// the same WIRE session key in another folder (same profile). Keyed by
+    /// the plain wire session id; held under a SEPARATE lock from `state` so
+    /// [`Self::scoped_goal_key`] can be resolved without re-entering the state
+    /// mutex the goal handlers already hold. Empty (no scope) → the goal store
+    /// keys by the plain wire id, byte-identical to the legacy behavior and to
+    /// the gateway/session-actor path (which never registers a scope).
+    goal_scopes: StdMutex<HashMap<String, String>>,
+}
+
+/// The plain WIRE session id underlying a (possibly cwd-scoped) goal-store
+/// key: strips a `\u{0}~cwd-<scope>` suffix if present, else returns the key
+/// unchanged. Used at the continuation-dispatch boundary so a goal enqueued
+/// under a scoped key is still delivered to the session runtime / actor keyed
+/// by the plain wire id (which is how `session_workspaces()`, `active_turns`,
+/// and the ledger's live-subscriber map are all keyed). The NUL separator is
+/// the same injective marker `storage_session_id` uses, so a legitimate wire
+/// id can never be mistaken for a scoped key.
+pub(crate) fn wire_key_from_goal_key(key: &SessionKey) -> SessionKey {
+    match key.0.split_once("\u{0}~cwd-") {
+        Some((wire, _)) => SessionKey(wire.to_owned()),
+        None => key.clone(),
+    }
 }
 
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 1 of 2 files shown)


ymote:
Landed in main via the combined bundle PR #1747 (integrated against current main + verified). Closing.

PR #1781, opened 2026-07-21
fix(chat): keep read-only runtime state outside workspace

ymote:
## Summary
- move read-only bundled-skill bootstrap into a per-chat `<data-dir>/runtime/runs/<uuid>/bootstrap` directory instead of `cwd/.octos`
- keep read-only spawn deliverables in the same run-private data-dir runtime tree and reject worktree isolation before it can create `.octos/work`
- fail closed when a host disables workspace writes but omits an external deliverable root
- retain existing workspace-write paths and nested-spawn behavior

## Validation
- `cargo test -p octos-agent`
- `cargo test -p octos-cli --lib`
- `cargo clippy -p octos-agent -p octos-cli --all-targets -- -D warnings`
- Octos review using the `dev` profile with isolated data directories and an embedded committed diff


ymote:
## Octos Review

Reviewed the complete committed PR diff with three isolated, read-only, tool-free Octos reviewers using the `dev` profile. The review workspace remained free of `.octos` after startup.

No critical or high findings were reported.

The review-driven commit `50c96047a` addresses the actionable containment findings from the first pass:
- make bundled-skill bootstrap run-private, preventing concurrent read-only chats from sharing generated plugin state
- refuse a deliverable spawn when workspace writes are disabled and no external output root is configured
- cover the paired read-only configuration and nested-root propagation with regression tests

Follow-up design items, intentionally outside this narrowly scoped filesystem-write fix:
- `octos chat --sandbox read-only` still treats project `.octos` plugins and bootstrap prompt files as trusted input. A mode intended to review an untrusted repository needs an explicit trust policy, such as an opt-in `--trust-workspace` switch.
- Run-private bootstrap/deliverable state needs a bounded retention policy that does not remove files still used by background workers.

Triage notes: worker IDs used by this path are internally generated as `subagent-N` before they reach filesystem joins, and the plugin loader explicitly uses first-directory-wins precedence. Those reviewer concerns do not establish a current bypass in this PR.

Issue #394, opened 2026-04-15
Tool-result budget and compaction for long free-form turns


ymote:
## Goal

Extract tool-result budgeting, truncation, and repair into a dedicated subsystem instead of leaving it inline inside the main agent loop.

## Current seam

The loop currently performs `trim_to_context_window`, `repair_message_order`, `repair_tool_pairs`, `synthesize_missing_tool_results`, `truncate_old_tool_results`, and related fixes inline.

## Scope

- move compaction/repair into dedicated helper module(s)
- make policy testable outside the full agent loop
- keep compaction deterministic across long turns

## Must touch

- `crates/octos-agent/src/agent/loop_runner.rs`
- new helper module(s) under `crates/octos-agent/src/agent/`

## Should avoid touching

- session result delivery
- workflow-family routing

## Acceptance tests

- oversized tool outputs are budgeted deterministically
- repaired tool history remains valid after compaction
- partial/missing tool-result chains do not poison later turns

ymote:
Implementation slice landed locally in commit `$(git rev-parse --short HEAD)`.

What landed:
- extracted loop message hygiene into `crates/octos-agent/src/agent/loop_compaction.rs`
- routed conversation preparation through the helper from `crates/octos-agent/src/agent/loop_runner.rs`
- registered the helper module in `crates/octos-agent/src/agent/mod.rs`

Verified with:
- `cargo test -p octos-agent should_budget_oversized_tool_outputs_deterministically -- --nocapture`
- `cargo test -p octos-agent should_preserve_tool_result_validity_after_compaction -- --nocapture`
- `cargo test -p octos-agent process_message_preserves_tool_pair_order_across_iterations -- --nocapture`
- `cargo test -p octos-agent run_task_collects_files_to_send_without_file_modified -- --nocapture`

ymote:
Completed on main in 645c2c3. Loop message hygiene and compaction are now extracted into dedicated helpers with focused deterministic compaction tests.

PR #1799, opened 2026-07-22
feat(agent): stream child StreamChunk deltas for live agent-dock output

ymote:
## Summary

Stream live `StreamChunk` text deltas from spawned background children to the agent dock, so the dock shows output while the child runs — not just the terminal snapshot.

## Problem

`SpawnChildTranscriptReporter` explicitly dropped `ProgressEvent::StreamChunk` events, claiming they *"would duplicate the Response event's final content."* But StreamChunk carries **live per-token text** while the child runs; Response is the **terminal final text**. Dropping StreamChunk meant the agent dock saw nothing until the child finished.

## Change

Handle `StreamChunk` in the reporter with two outputs:
1. **Direct callback** (`on_stream_chunk`) — forwarded to the WS/serve layer so it can emit `agent/output/delta` directly, bypassing the heavy `on_change` persistence path (codex plan review: *"per-token persistence/on_change fan-out is too heavy"*).
2. **Router file append** — the same live text goes to the `SubAgentOutputRouter` so `task/output/delta` also carries it (the TUI fallback path from PR #394).

### Infrastructure added

- `ToolContext.child_stream_callback: Option<Arc<dyn Fn(&str) + Send + Sync>>` — threaded through all clone/new/child_spawn_clone paths.
- `SpawnTool::with_child_stream_callback(cb)` builder.
- `SpawnChildTranscriptReporter.on_stream_chunk` — forwarded from the context.

### Codex plan review alignment

Codex confirmed: the `ProgressReporter` / `StreamChunk` hook is the correct semantic layer (there's no `AgentEvent::TextDelta` in this codebase — the path is `StreamEvent::TextDelta → ProgressEvent::StreamChunk`). The direct-callback approach avoids the per-token `on_change` overhead codex flagged.

## Follow-up (tracked here, not in this PR)

Wire `with_child_stream_callback` at the WS session-runtime path (`ui_protocol.rs:21895`) to emit `AGENT_OUTPUT_DELTA` via the supervisor event sink. The agent-layer infrastructure is complete; the wire-up is mechanical.

## Verification

All 2160 `octos-agent` lib tests pass.


ymote:
Adversarial review + fix set applied (`7a74c6758` merge + `22914ea66`).

## What the review found
1. **Blocker — the serve wire-up never compiled.** The stream callback delivered `usize` into `OutputCursor.offset` (`u64`) → E0308 in `ui_protocol.rs`. `octos-cli`'s `api` feature is off by default, so plain `cargo check` never built the file — and no CI ran on the head. (The PR body also said the wire-up was a follow-up "not in this PR", but commits 2–3 fully implement it.)
2. **Duplication, empirically reproduced.** The StreamChunk arm appended live text to the same router file the `Response` arm appends full iteration text to → every child message landed **twice** in `<task_id>.out` (probe: `PLAN: scan the treePLAN: scan the tree…`), burning `max_bytes_per_task` 2× fast. This is the exact duplication the original drop-comment warned about.
3. **The router append fed nothing live.** `task/output/delta` (tui#394's fallback) is produced from supervisor progress events — nothing bridges router file appends to it. The append's only observable effect was the doubling.
4. **Cursor precedent.** The callback reported end-offset; every sibling producer (`TaskOutputDeltaTracker`, the read RPCs) reports start-offset — and this PR creates the first `agent/output/delta` producer, so it sets the wire convention. Also: the "natural backpressure" comment described lossy drop-on-full.

## What the fix set changed
- Callback signature is `u64` end to end (`ChildStreamCallback` alias); **verified with `cargo check -p octos-cli --features api`**.
- StreamChunk arm: **router append deleted** — live text goes only to the direct callback; the router file stays the Response-based durable record.
- **Start-offset** cursor convention (cumulative streamed bytes BEFORE the chunk); comments now say the ephemeral send is lossy drop-on-full.
- Merged main (the spawn.rs tests-split conflict resolved; `child_stream_callback: None` added to the spawn_tests literal).
- **Tests added** (were zero): real-background-child callback offsets (`offset[0]==0`, `offset[i+1]==offset[i]+len(text[i])`, keyed by task id), router-file **exactly-once** pins for both iteration texts, `child_spawn_clone` preserves the callback (grandchild streaming).

Corrected facts for the PR body: the field lives on `SpawnTool` (not `ToolContext`); the serve wire-up IS in this PR; signature is `(agent_id, start_offset: u64, text)`; the router append is gone.

## Gates
octos-agent lib: 2421 passed / 0 failed. clippy: octos-agent 0 warnings (main baseline 0), octos-cli+api 69 (= main baseline). fmt clean.

## Follow-ups deliberately NOT addressed here
Reconnect-safe sink (deltas capture the spawning turn's WsConnection — dead after reconnect, and a running child pins it); gateway/chat/ACP paths not wired (only serve WS `turn/start`); optional ~16ms coalescing debounce; grandchild-delta policy (clone now preserves the callback deliberately).

PR #1815, opened 2026-07-24
test(cli): pin goal budget re-arm after raised-budget reactivation

ymote:
The goal-budget runaway (`used >> budget`, ending `blocked`) is **already fixed on main** — set_goal guards over-budget reactivation and resets wrap_up_emitted so the charge-side exhaustion flip re-arms. This adds the **missing regression test** (agent_orchestrator.rs) that RED-proves the line-3230 reset is load-bearing: removing it makes a raised-budget-reactivated goal accrue unbounded instead of re-flipping to budget_limited. Test-only (+76, 0 deletions). `cargo test --lib` 2568/0.


anantheparty:
The lone red check is stale and unrelated to this test-only diff: clippy failed at `crates/octos-cli/src/commands/doctor.rs:337` (`empty_line_after_doc_comments`). That baseline was fixed on `main` by #1816 and then removed in #1822; #1822's Linux `check` and `test-octos-cli` jobs are green.

I also verified that this one commit rebases cleanly onto current `main` (`1f4370f8`). The rebased diff remains only +76 lines in `agent_orchestrator.rs`, with no whitespace errors. Please rebase/update the branch to rerun CI.

Because this test is under the `api` feature while the current CI's `api` invocations are narrowly filtered, please also run and report the exact regression:

```sh
cargo test -p octos-cli --features api api::agent_orchestrator::tests::charge_re_arms_budget_flip_after_raised_budget_reactivation -- --exact
```

anantheparty (approved):
I replayed this exact +76-line test-only change onto current `main` (`44ee3eae`) without semantic adaptation and ran the previously requested API-feature command. The target regression passed 1/1 (2,567 filtered); `cargo fmt --all -- --check` and `git diff --check` also pass. The test fails closed on the intended invariant and adds no production behavior.

Approved on code/behavior. Refresh helper #1872 targets this PR's original source branch; merging it will update #1815 in place and rerun authoritative current-main CI.

PR #1838, opened 2026-07-27
fix(windows): repair missing-root scope and QoS test ordering

### LINKED ISSUE #1837: Windows session scope rejects children of a missing root due to verbatim-path mismatch
## Summary

The Windows `octos-core` test suite has two stable failures in `session_scope` when a workspace/root does not exist yet. This currently forces the PR `check-windows` job to be non-blocking and prevents Windows scope regressions from acting as a merge gate.

Confirmed failures:

- `multi_tenant_at_workspace_accepts_encoded_under_data_layout`
- `multi_tenant_at_workspace_roots_at_repo_for_coding_hint`

Observed upstream run: https://github.com/octos-org/octos/actions/runs/30192520859/job/89768217732 (`309 passed; 2 failed`).

## Data and logic chain

`SessionScope::classify_canonical_path` compares a canonicalized candidate path with a canonicalized scope root using `Path::starts_with`.

For a missing root on Windows:

1. `canonicalize_lossy(candidate)` walks upward to the existing `C:\` ancestor.
2. `std::fs::canonicalize` returns the ancestor in verbatim form (`\\?\C:\`).
3. The missing suffix is reattached, so the candidate remains verbatim.
4. `canonical_root_lossy(root)` currently gives up and returns the raw `C:\...` input.
5. `starts_with` compares different Windows path representations and returns `false`, classifying an in-scope child as `OutOfScope`.

This is a representation asymmetry at a sandbox boundary, not an agent/model or deployment failure.

## Expected resolution

- Resolve missing candidates and missing roots with the same ancestor-walk algorithm.
- Keep symlink treatment symmetric on both sides of the containment check.
- Add a regression test that asserts a child of a unique missing root retains the root representation on Windows.
- Re-enable blocking `check-windows` on pull requests once the two failures pass.

## Acceptance criteria

- The two existing Windows-only failures pass.
- A new missing-root representation invariant passes on Windows and Unix.
- Existing symlink-escape and out-of-scope tests remain green.
- `check-windows` no longer uses PR-only `continue-on-error`.
- Full Windows CI is green before merge.

## Existing context

PR #1818 documented the known defect while making Windows CI visible but temporarily non-blocking; it deliberately did not change this sandbox-sensitive path logic.

### LINKED ISSUE #1839: Windows QoS drift test is flaky because zero-latency mock yields
## Summary

After the two `session_scope` Windows failures were fixed, the newly blocking `check-windows` job reached `octos-llm` and exposed a scheduler-dependent adaptive-router test:

```text
adaptive::tests::should_drift_qos_score_in_response_to_live_traffic
assertion `left == right` failed:
fast-fails should have exactly 8 failures, got 7
```

Run: https://github.com/octos-org/octos/actions/runs/30275900233/job/90009757466

The same job confirmed the preceding Core fix with `320 passed; 0 failed`.

## Root cause

The shared test `MockProvider` always calls `tokio::time::sleep`, including for `latency_ms == 0`.

The QoS drift test races:

- `slow-good`: 10ms timer, successful response
- `fast-fails`: 0ms timer, failed response

The test assumes the failure is immediately ready and wins every hedge race. A zero-duration Tokio sleep still yields to the runtime. On Windows, timer granularity and scheduler timing can leave both branches ready when `tokio::select!` polls again; the successful primary can then win, canceling the alternate before its failure metric is recorded.

This is why the exact counter can be 7 instead of 8 even though the production behavior and score drift are otherwise valid.

## Expected resolution

Make a zero-latency test provider immediately ready instead of awaiting `sleep(0)`. Keep positive artificial latency on the Tokio clock. This preserves the fixture's declared meaning and makes the fast-failure ordering deterministic without changing production router code.

## Acceptance criteria

- `should_drift_qos_score_in_response_to_live_traffic` passes repeatedly on Windows.
- The test still records exactly 8 failed fast-lane calls and 8 successful slow-lane calls.
- The complete `octos-llm` library suite remains green.
- The blocking `check-windows` job proceeds past `octos-llm`.

## Context

Discovered while validating #1838, which removes the temporary PR-only `continue-on-error` after fixing the earlier Core failures.

anantheparty:
## What changed

- make `canonical_root_lossy` use the same existing-ancestor walk as `canonicalize_lossy`
- add a cross-platform invariant covering a child under a unique missing root
- give the QoS drift test a dedicated first-poll failure provider

Closes #1837.
Closes #1839.

## Why

On Windows, canonicalizing an existing ancestor produces a verbatim path such as `\\?\C:\...`. Before this change, a missing candidate used that ancestor-walked representation while a missing scope root fell back to raw `C:\...`. `Path::starts_with` then rejected a child that was lexically inside the root.

The fix makes root and candidate resolution symmetric, including symlink handling, instead of stripping prefixes after canonicalization.

Once that repair let the Windows job proceed beyond `octos-core`, it exposed a separate test-fixture race: the QoS test represented an intended immediate failure with `MockProvider { latency_ms: 0 }`, but `MockProvider` still awaits `sleep(0)`. That yields to Tokio, so Windows timer/scheduler behavior could let a 10ms successful hedge win and cancel the expected failure.

The QoS test now uses a dedicated provider whose failed response is ready on its first poll. The shared mock and production router code are unchanged, so other Hedge tests retain their existing scheduling semantics.

## Windows evidence

First run after the scope fix:

- workspace build passed
- `octos-core`: **320 passed; 0 failed**
- `octos-memory`: **122 passed; 0 failed**
- then exposed #1839 in `octos-llm`: expected 8 immediate failures, observed 7

Run: https://github.com/octos-org/octos/actions/runs/30275900233/job/90009757466

Second run with the isolated QoS fixture:

- `octos-core`: **320 passed; 0 failed**
- `octos-memory`: **122 passed; 0 failed**
- `octos-llm --lib`: **482 passed; 0 failed**
- `octos-llm` integration tests: passed
- `octos-bus --lib`: **225 passed; 0 failed**
- then exposed the unrelated Unix-only Bus fixture tracked in #1840

Run: https://github.com/octos-org/octos/actions/runs/30278411155/job/90018340610

The PR-only Windows job remains non-blocking in this PR. Restoring it as a required gate is deliberately deferred until #1840 is fixed; this final diff does not modify the workflow.

## Local validation

- `git diff --check`
- final diff is limited to `session_scope.rs` and `adaptive_tests.rs`

This checkout does not have a Rust toolchain, so GitHub's Windows runner provides the executable validation above.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -896,7 +896,9 @@ jobs:
           else
             FILES=$(git diff --name-only ${{ github.event.pull_request.base.sha }} HEAD 2>/dev/null || echo "")
           fi
-          if echo "$FILES" | grep -qE '(sandbox|ssrf|security|tools/mod|session_actor|profiles)'; then
+          # `session_scope` is the path-containment boundary behind
+          # `classify_canonical_path`; it is NOT covered by `session_actor`.
+          if echo "$FILES" | grep -qE '(sandbox|ssrf|security|tools/mod|session_actor|session_scope|profiles)'; then
             echo "run=true" >> "$GITHUB_OUTPUT"
           else
             echo "run=false" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -570,21 +570,9 @@ jobs:
     # ~15 min it is the cheapest job in the matrix covering code no other PR
     # check can even see.
     #
-    # NON-BLOCKING for now, because it is currently RED for a pre-existing
-    # reason: `octos-core` `session_scope` has 2 Windows-only test failures
-    # (5-for-5 on recent `main` runs, long before this job moved to PRs).
-    # Making it blocking today would wedge every PR on a bug it did not
-    # introduce. It still surfaces on the PR, which is the point — a NEW
-    # Windows break is now visible at review time instead of after merge.
-    #
-    # Non-blocking ONLY on PRs. On push to `main` it stays hard-failing, so
-    # `build-windows` (which `needs:` this job) still refuses to ship a Windows
-    # artifact from a red suite — a blanket `continue-on-error: true` would have
-    # silently re-enabled those releases.
-    #
-    # Drop this line entirely once the two tests are fixed; see the tracking
-    # note in `crates/octos-core/src/session_scope.rs`.
-    continue-on-error: ${{ github.event_name == 'pull_request' }}
+    # This is blocking on PRs now that the pre-existing `session_scope`
+    # canonical-path failures are fixed. Keep it blocking so Windows-only
+    # regressions cannot merge unseen.
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -570,9 +570,21 @@ jobs:
     # ~15 min it is the cheapest job in the matrix covering code no other PR
     # check can even see.
     #
-    # This is blocking on PRs now that the pre-existing `session_scope`
-    # canonical-path failures are fixed. Keep it blocking so Windows-only
-    # regressions cannot merge unseen.
+    # NON-BLOCKING for now, because it is currently RED for a pre-existing
+    # reason: `octos-core` `session_scope` has 2 Windows-only test failures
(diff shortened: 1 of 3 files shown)


anantheparty:
The independent Bus fixture repair is now #1841. Recommended order: merge #1841 first, then refresh this branch from main and rerun Windows. That keeps the production scope fix, QoS timing fixture, and Bus path fixture reviewable as separate logic changes.

PR #1860, opened 2026-07-29
fix(windows): make workspace test suite green on windows-latest CI

BH3GEI:
## What & why

The `check-windows` CI job is `continue-on-error` on pull requests, so Windows-only test failures have accumulated silently. Running the full `check-windows` sequence on a real Windows 11 host surfaced ~50 failures across ~25 files. This PR fixes all of them. Two are genuine **product bugs** (A, B); the rest are test/fixture portability issues. **No test is disabled** — Unix-only behaviour is cfg-gated with a Windows-equivalent path, never skipped.

## Product bugs

- **subagent_output** — percent-encode `session_id`/`task_id` (`safe_filename`) before using them in the on-disk router path. A task id containing `:` (legal in ids, illegal on NTFS) previously produced an unwritable path.
- **tools/spawn** — strip the `\?\` verbatim prefix (`dunce::simplified`) before passing worktree paths to `git worktree`, which rejects verbatim paths; store the simplified path in `WorkerWorktree`.

## Test / fixture portability

- shell / spawn / plugin / bootstrap / profile / skills / context / doctor / config tests: branch on OS for shell syntax (`cmd` vs `sh`) and path separators, use a Windows temp dir instead of a hardcoded `/tmp`, and compare paths through `dunce` so the `\?\` prefix does not cause spurious mismatches.
- `validator_runner` — forward-slash the probe path inside the `sh -c` redirect (`\` is an escape there) so `BLOCKED_ENV_VARS` stripping is actually exercised on Windows.
- harness-starter smoke tests (×4) — normalize `\` → `/` before glob match.
- **compat-test-skill fixture** — rewrite `main` from a `python3`-dependent script to pure POSIX `sh` (`python3` is a Microsoft-Store stub on stock Windows), spawn it via `sh` on Windows (no shebang support), and pin the fixture to `eol=lf` via `.gitattributes` so checkout is deterministic on both platforms.

## Verification

Ran locally on Windows 11, mirroring the `check-windows` job. All green:

- `cargo build --workspace`
- `cargo build -p octos-cli --features api,telegram,discord,dingtalk,whatsapp,feishu,twilio,wecom,wecom-bot,audio_mp3`
- `cargo test` — octos-core (320), octos-memory (132), octos-llm (`--lib` 482 + `--tests`), octos-bus (225), octos-pipeline (340), octos-plugin, octos-sandbox, octos-swarm, octos-agent (`--lib` 2313 + `--tests`), octos-cli (`--lib` + `--tests`), harness-starter-{generic,report,audio,coding}
- `cargo test --workspace --doc`
- `cargo clippy --workspace --all-targets -- -D warnings`

No changes to `.github/workflows/ci.yml`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+# The harness compat-gate fixture skill is a POSIX `sh` script. It is executed
+# via its `#!/usr/bin/env sh` shebang on Unix and via `sh <path>` on Windows
+# (windows-latest has no shebang support). It MUST keep LF line endings on every
+# platform: CRLF breaks the shebang on Unix, and CRLF handling in `sh` varies by
+# Git-for-Windows version. Pinning `eol=lf` makes checkout deterministic and
+# overrides a Windows checkout's `core.autocrlf=true`.
+e2e/fixtures/compat-test-skill/main text eol=lf
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4506,6 +4506,7 @@ dependencies = [
  "chromiumoxide",
  "chrono",
  "dirs",
+ "dunce",
  "eyre",
  "flate2",
  "futures",
--- a/crates/app-skills/harness-starter-audio/tests/harness_smoke.rs
+++ b/crates/app-skills/harness-starter-audio/tests/harness_smoke.rs
@@ -133,7 +133,11 @@ fn lifecycle_state_transitions_queued_running_verifying_ready() {
 }
 
 fn path_matches_glob(path: &Path, pattern: &str) -> bool {
-    let path_str = path.to_string_lossy();
+    // Normalize `\` -> `/` so the `/`-based policy globs match on Windows,
+    // where `to_string_lossy()` renders backslash separators. The real runtime
+    // globs the filesystem via the `glob` crate (separator-agnostic); this
+    // helper only mirrors that at the string level for the smoke assertion.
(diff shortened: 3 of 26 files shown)


anantheparty:
Windows CI is now down to one failure, and the log identifies a different root cause from path separators:

`project_templates::tests::should_use_relative_asset_base_when_bootstrapping_react_vite_site` invokes `Command::new("bash")`, but on `windows-latest` that resolves to the Windows `bash.exe`/WSL launcher, not Git Bash. The captured stdout is UTF-16 text saying ?Windows Subsystem for Linux has no installed distributions.? The forward-slash normalization therefore never reaches the intended script runtime.

Please resolve Git for Windows Bash explicitly on Windows (or use another explicit MSYS entry point rather than the ambiguous `bash` executable), and keep the diagnostic output added here. A regression assertion should distinguish Git Bash/MSYS from the WSL launcher. Current run: https://github.com/octos-org/octos/actions/runs/30439031186/job/90533591214

anantheparty:
Follow-up fix opened as #1867.

It replaces the ambiguous bare `bash` with the Git for Windows MSYS2 Bash resolved from `git --exec-path`, and adds a Windows identity regression that rejects WSL/Linux. This addresses the remaining `project_templates::tests::should_use_relative_asset_base_when_bootstrapping_react_vite_site` failure that was still present when #1860 merged.

PR #1864, opened 2026-07-29
fix(api): resolve session files from authenticated profile

alan0x:
## Summary
- resolve session/files.list against the authenticated user's profile data directory when a process-wide SessionManager also exists
- preserve the existing session-store fallback when no ProfileStore is configured
- add generic regression coverage for both runtime shapes

## Root cause
On localhost authenticated requests, profile resolution depended on a multi-profile gateway API port. Without that port, file access fell back to the process-wide SessionManager data directory, so session/files.list returned an empty list even though the session workspace existed under the authenticated profile.

## Scope
This is a generic session workspace fix. It contains no Learn, OLL, skill-specific, or artifact-name logic.

## Verification
- cargo fmt --all -- --check
- cargo test -p octos-cli --features api --lib: 2604 passed, 6 ignored, 0 failed
- cargo clippy -p octos-cli --features api --lib --tests: passed with pre-existing warnings

### DIFF
--- a/crates/octos-cli/src/api/handlers.rs
+++ b/crates/octos-cli/src/api/handlers.rs
@@ -1427,21 +1427,16 @@ pub async fn session_workspace_contract(
     // `sess.data_dir()` BEFORE checking the host-routed profile, so a
     // cross-tenant header on a TRUSTED hop exposed the victim
     // profile's workspace-contract statuses. The Layer-2 gate runs
-    // up-front; `state.sessions.data_dir()` only resolves when the
-    // routed profile is authorized (admin / owner / self).
+    // up-front, and data-directory selection below must remain identical
+    // to the companion `session_files` surface.
     if let Err(response) = authorized_routed_profile_id_from_headers(&state, &headers, identity_ref)
     {
         return response;
     }
 
-    let data_dir = if let Some(sessions) = &state.sessions {
-        let sess = sessions.lock().await;
-        sess.data_dir()
-    } else {
-        match resolve_profile_data_dir(&state, &headers, identity_ref).await {
-            Ok(data_dir) => data_dir,
-            Err(response) => return response,
-        }
+    let data_dir = match resolve_file_access_data_dir(&state, &headers, identity_ref).await {
+        Ok(data_dir) => data_dir,
+        Err(response) => return response,
     };
 
     let mut statuses = Vec::new();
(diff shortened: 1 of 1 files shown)


anantheparty (commented):
The new resolve_file_access_data_dir call fixes the session/files.list half of this topology, but the adjacent session_workspace_contract handler still selects state.sessions.data_dir() before profile resolution. With an authenticated user + profile_store + process-wide SessionManager, the same session can therefore list files from the user's profile root while reporting workspace-contract state from the global root.

Please route session_workspace_contract through the same helper and add the same-topology regression assertion (or explicitly track it as a blocking follow-up). Keeping these two read surfaces on one authoritative data directory is important for the Web session-files flow tracked in octos-org/octos-web#169.

alan0x:
Addressed in 6c65f9df.

`session_workspace_contract` now uses the same `resolve_file_access_data_dir` helper as `session_files`, so both read surfaces resolve against the same authoritative profile data directory.

I also extended the regression coverage to construct both a profile-scoped workspace repo and a process-wide global repo for the same session. The test asserts that both session file listing and workspace-contract inspection use the profile root and that the global repo is not returned.

Validation completed:
- targeted profile workspace regression: passed
- existing cross-tenant session regression tests: 3 passed
- `cargo fmt --all -- --check`: passed
- `git diff --check`: passed

anantheparty (approved):
Re-reviewed 6c65f9df. Both session/files.list and session/workspace.get now resolve through the same authenticated profile data root, and the dual-root regression test would fail the old global-first behavior while asserting the profile workspace is returned. No remaining blocker from my review; Windows CI can remain the merge gate.

anantheparty:
The current `check-windows` failure is inherited from `main` after #1860, not introduced by this PR: the remaining site-bootstrap test resolves bare `bash` to the Windows WSL launcher.

Follow-up #1867 contains the isolated Git Bash resolver and regression test. My approval of #1864 still stands; no session-scope change is needed here. Once #1867 lands on `main`, this branch should only need a sync/rerun to clear that inherited check.

anantheparty:
#1867 has now merged into `main` as `44ee3eae`, and its complete `check-windows` run was green. My approval still stands; syncing this branch with current `main` and rerunning CI should remove the inherited Windows blocker without further session-file changes.

PR #1867, opened 2026-07-29
fix(windows): resolve Git Bash for site bootstrap

anantheparty:
## Summary

Follow-up to #1860. That PR normalized the arguments passed to the site bootstrap scripts, but Windows still launches them with a bare `bash`.

On `windows-latest`, process lookup resolves that name to `C:\Windows\System32\bash.exe`, the WSL launcher. Hosts without an installed WSL distribution fail before the script runs, so `should_use_relative_asset_base_when_bootstrapping_react_vite_site` remains red and the failure is inherited by otherwise unrelated branches such as #1864.

## Fix

- On Windows, ask the already-required Git installation for `git --exec-path` and locate its own `bin/bash.exe` or `usr/bin/bash.exe`.
- Use that explicit Git Bash/MSYS2 path for both Quarto and regular site bootstrap scripts, avoiding PATH/WSL ambiguity.
- Keep Unix behavior unchanged (`bash` from PATH).
- Return an actionable error if Git for Windows is missing or incomplete.
- Add a Windows regression test that runs the resolved shell and requires `uname -s` to identify MINGW/MSYS rather than WSL/Linux.

## Validation

- `git diff --check`
- Manual Windows resolution on this host:
  - `git --exec-path` → `C:\Program Files\Git\mingw64\libexec\git-core`
  - resolved shell → `C:\Program Files\Git\bin\bash.exe`
  - `uname -s` → `MINGW64_NT-10.0-26200`
- Independent patch review: no blocker/P1.

- GitHub CI: formatting, clippy, Linux `test-octos-cli`, and the complete `check-windows` job all passed ([run 30449233323](https://github.com/octos-org/octos/actions/runs/30449233323)). The Windows job covers both the existing bootstrap regression and the new Git Bash shell-identity test.

All required checks are green; this is ready for review.

### DIFF
--- a/crates/octos-cli/src/project_templates.rs
+++ b/crates/octos-cli/src/project_templates.rs
@@ -901,6 +901,7 @@ fn run_site_bootstrap(
     let scripts_dir = skill_dir.join("scripts");
     std::fs::create_dir_all(project_dir)
         .map_err(|e| format!("create site project dir failed: {e}"))?;
+    let bash = resolve_site_bootstrap_bash()?;
 
     // Windows CI runs these bootstrap scripts through Git Bash (MSYS2),
     // whose coreutils treat `\` as a literal character rather than a path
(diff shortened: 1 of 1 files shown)


anantheparty:
@BH3GEI could you review this follow-up to #1860 when convenient? It replaces the remaining ambiguous Windows `bash` lookup with Git for Windows' own MSYS2 Bash. The full CI suite is green, including the 53-minute `check-windows` run and both `octos-cli` test selections that previously failed. The PR is one file and based exactly on current `main`.

PR #1744, opened 2026-07-19
feat(voice): apply per-profile ASR language to voice turns

alan0x:
## Summary

- Add a validated per-profile `asr_language` setting.
- Canonicalize supported Qwen3-ASR language names and legacy `zh`/`en` aliases.
- Preserve `None`, `null`, and `auto` as automatic language detection end to end.
- Resolve the latest profile override for each AppUI and gateway/channel voice message.
- Pass the effective optional language to OMiniX transcription.
- Surface profile configuration read and parse failures instead of silently falling back.

## Why

ASR language was previously captured from the serve-level voice configuration when a profile runtime started. Profile Settings changes therefore did not affect the next utterance, and resolving the override only in AppUI would have left gateway/channel transports with different behavior.

## Contract

- A profile override takes precedence over the serve-level default.
- Profiles without an override inherit the serve-level setting.
- `None`, `null`, and `auto` preserve automatic language detection; they do not become `Chinese`.
- Unknown inherited serve-level values are passed through unchanged rather than silently rewritten.
- AppUI and gateway/channel voice paths re-read the durable `ProfileStore` for each audio turn, so a Settings update applies to the next utterance without restarting `octos serve`.
- Missing, malformed, or unreadable profile configuration is surfaced explicitly rather than collapsed into fallback behavior.

## Coverage

- Profile PUT/patch -> disk persistence -> reload -> next voice turn -> OMiniX request body.
- A/B profile language isolation.
- `null`/`auto` semantics and canonical language aliases.
- Missing and malformed profile-store failures.
- Gateway inheritance, per-message profile resolution, and host ASR default forwarding.

## Validation

- `cargo test -p octos-cli --features api`
- `cargo fmt --all -- --check`
- `git diff --check`
- Required GitHub Actions checks pass on the current head, including Linux `check`, `test-octos-cli`, and `check-windows`: https://github.com/octos-org/octos/actions/runs/30535577047

## Paired Web Work

The paired Web implementation remains octos-org/octos-web#281. It should continue to follow this Core contract, with the live per-profile ASR-language acceptance pass performed after the Core contract is accepted.

PR #1869, opened 2026-07-29
fix(api): keep no-cwd sessions in profile store

alan0x:
## Summary
- preserve whether an AppUI workspace came from an explicit cwd or the derived Tier-3 default
- prevent no-cwd sessions from relocating their transcript store on turn, lookup, compact, tool-status, or review paths
- keep explicit cwd sessions project-scoped and retain the effective workspace for tools, panes, goals, and fleet recovery

## Root cause
session/open stored the derived Tier-3 workspace in session_workspaces. Later runtime lookups replayed that effective workspace as if it were a client/operator cwd. With appui.sessions_in_cwd enabled, the same session then moved from the profile store into the derived workspace's .octos store and disappeared from the profile-global session list.

## Tests
- cargo test -p octos-cli --features api --lib transcript_store_hint -- --nocapture
- cargo test -p octos-cli --features api --lib appui_session_ -- --nocapture
- cargo test -p octos-cli --features api --lib session_open_writes_active_profile_marker_only_with_flag_and_cwd -- --nocapture
- cargo test -p octos-cli --features api --lib session_list_cwd_root -- --nocapture
- cargo test -p octos-cli --features api --lib fleet_keeper_reseed_tests -- --nocapture
- cargo check -p octos-cli --features api
- cargo fmt --all -- --check

## Relationship
Companion to #1864: that PR makes authenticated session file access resolve the correct profile data directory; this PR keeps no-cwd AppUI/Learn transcripts in that profile store from open through later runtime access.

### DIFF
--- a/crates/octos-cli/src/api/agent_orchestrator.rs
+++ b/crates/octos-cli/src/api/agent_orchestrator.rs
@@ -4126,6 +4126,7 @@ impl InProcessAgentOrchestrator {
                 consecutive_failed_turns: 0,
                 fleet_id: Some(fleet_id.to_owned()),
                 controller_workspace_root: None,
+                controller_workspace_has_runtime_hint: None,
             },
         );
     }
--- a/crates/octos-cli/src/api/agent_orchestrator.rs
+++ b/crates/octos-cli/src/api/agent_orchestrator.rs
@@ -197,6 +197,10 @@ pub(crate) const FLEET_KEEPER_META_READY: &str = "ready";
 /// its workspace-known gate (PR 4b). Omitted when the fleet has no persisted
 /// root (that keeper is simply not headlessly rehydratable).
 pub(crate) const FLEET_KEEPER_META_WORKSPACE_ROOT: &str = "workspace_root";
+/// Metadata key preserving whether `workspace_root` originated from an
+/// explicit runtime cwd hint. Missing/invalid means legacy unknown and is
+/// handled as `false` so a restart never relocates transcripts unsafely.
+pub(crate) const FLEET_KEEPER_META_WORKSPACE_HAS_RUNTIME_HINT: &str = "workspace_has_runtime_hint";
 
 /// PR 4b — upper bound on the VALID (rehydratable) fleet-keeper candidates
 /// [`InProcessAgentOrchestrator::pending_fleet_keeper_seeds`] produces per drain
@@ -229,6 +233,9 @@ pub(crate) struct FleetKeeperSeed {
     /// The persisted controller workspace root, validated to be an existing
     /// directory at selection time.
     pub(crate) root: String,
+    /// `Some(true|false)` for new durable records; `None` for legacy/unknown
+    /// provenance. Only `Some(true)` may reconstruct a runtime cwd hint.
+    pub(crate) workspace_has_runtime_hint: Option<bool>,
 }
 
 /// Peer awaiting-input WAKE — kind label for the `External(_)` master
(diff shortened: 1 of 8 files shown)


anantheparty (changes requested):
There is still a restart path that recreates the no-cwd transcript split this PR is meant to fix.

`session/open` correctly stores a derived Tier-3 workspace as `{ root, runtime_hint: None }`, but `run_standalone_turn` persists that same effective tool workspace through goal metadata → `FleetRecord.controller_workspace_root` → durable fleet wake. After a serve restart, `pending_fleet_keeper_seeds()` returns only that root, and `SessionWorkspaceStore::set_if_absent()` unconditionally reconstructs it as `runtime_hint: Some(root)` (ui_protocol.rs:934-936). A no-cwd Tier-3 root therefore becomes a cwd hint after restart. With `sessions_in_cwd=true`, the headless keeper bootstraps under `<Tier3>/.octos/<profile>` while the original transcript remains in the profile-global store, so the conversation is split and subsequent global listing/file reads do not see the keeper output.

The current regression tests only cover open → lookup in one process; `fleet_keeper_reseed_tests` explicitly asserts the unsafe assumption that every recovered root is an authoritative hint. Please persist workspace provenance through goal → fleet record → wake → `FleetKeeperSeed`, and add a fresh-store/fresh-cache restart test for both no-cwd and explicit-cwd cases. For legacy records whose provenance is unknown, the safe behavior is to avoid treating the root as a relocation hint (or defer headless execution until a live open supplies provenance).

Secondary: `run_standalone_turn` reads `get()` and `runtime_hint()` under two separate mutex acquisitions. A concurrent `session/open` can replace the binding between those reads, pairing one root with another hint. Please expose a single-lock binding snapshot and consume root+hint atomically.

anantheparty:
I merged current `main` (`e064b9a8`) into this branch as `63dcee80` and resolved the Fleet wake conflict in place.

Resolution keeps both sides of the contract: #1869's workspace provenance/atomic snapshot metadata and the newer Fleet grant/escalation/boot-resume paths. I also wired `rec.controller_workspace_has_runtime_hint` into the new boot-resume `fleet_keeper_continuation_request` call; otherwise the merge would both miss the new argument and lose the provenance bit across restart recovery.

`cargo fmt --all -- --check` and `git diff --cached --check` passed. Local Cargo execution could not complete because the WSL cache lacks `chrono 0.4.43` and crates.io download stalled, so I am leaving the fresh current-head GitHub CI as the authoritative compile/test gate before re-review.

anantheparty (dismissed):
Re-reviewed current head `63dcee80` after resolving the latest Fleet conflicts. The original blockers are addressed:

- workspace root provenance now survives goal/fleet persistence, outbox wake, boot-resume wake, and keeper seed reconstruction;
- explicit cwd remains a runtime hint, while derived/legacy-unknown roots remain tool workspaces and do not silently migrate transcript identity;
- root plus provenance are read from a single snapshot, avoiding mixed concurrent state;
- the current main Fleet grant, escalation, store, and boot-resume behavior is retained, including forwarding the provenance bit through the new boot-resume call.

No remaining code-review blocker. Local fmt/diff checks passed; the fresh current-head GitHub CI remains the authoritative compile/test gate because the local WSL dependency cache could not complete the Rust run.

anantheparty:
Follow-up from the first merged-head CI: the API test build found two compile-only fixture omissions (AutonomyGoalRecord's new provenance field and one synthetic wake call's seventh argument). Fixed both in f6ab55d, then merged the newly landed #1864/main into the branch; current head is dff4d7e9. Fmt and diff checks are clean, and a fresh CI run is now the gate.

anantheparty (approved):
Re-reviewed current head dff4d7e9: the follow-up only completes the two provenance-aware test fixtures identified by CI and synchronizes #1864/current main without conflict. The production provenance and Fleet wake contract from the prior approval is unchanged. Approved; fresh CI remains the merge gate.

anantheparty (approved):
Re-reviewed current head 4ce165ea: this final sync only brings in the now-merged #1815 test regression; it merged cleanly and does not alter #1869's production provenance logic. Approved; the restarted CI remains the gate.

anantheparty (approved):
Re-reviewed current head 5ff3a8f0: the latest sync brings in the now-merged #1744 voice/profile contract. ui_protocol.rs merged automatically without conflict; #1869's session workspace provenance paths remain intact. Approved; restarted current-main CI remains the gate.

PR #1874, opened 2026-07-29
fix(api): redact credentials from request logs

anantheparty:
## Summary

- replace raw request URI logging with Axum `MatchedPath` route templates in the HTTP trace span
- record `<unmatched>` for requests that do not resolve to a known route
- make user/admin authentication rejection warnings log only method, route template, and `token_present`
- stop the untrusted `X-Profile-Id` warning from recording a raw request URI

## Why

Some browser WebSocket requests authenticate through query parameters, while signed preview/setup routes can carry credentials in path segments. The previous auth warnings and tower-http default request span could therefore persist credential material in logs.

This change preserves useful route-level diagnostics without recording query strings, raw credential-bearing paths, token prefixes, or token lengths. No real logs, API keys, or real tokens were read or used while implementing the fix.

## Validation

- `api::router::tests`: 37/37 passed, including a real Router + TraceLayer regression covering WebSocket query credentials, signed-preview/setup path credentials, and unmatched routes
- `cargo check -p octos-cli --features api --lib`: passed
- `cargo fmt --all -- --check`: passed
- `git diff --check`: passed
- independent review: no P0-P3 findings

### DIFF
--- a/crates/octos-cli/src/api/router.rs
+++ b/crates/octos-cli/src/api/router.rs
@@ -3,8 +3,8 @@
 use std::sync::Arc;
 
 use axum::Router;
-use axum::extract::DefaultBodyLimit;
-use axum::http::{HeaderValue, StatusCode};
+use axum::extract::{DefaultBodyLimit, MatchedPath};
+use axum::http::{HeaderValue, Method, StatusCode};
 use axum::middleware::{self, Next};
 use axum::routing::{delete, get, post, put};
 use tower_http::cors::CorsLayer;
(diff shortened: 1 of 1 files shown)


anantheparty:
@BH3GEI ready for review. This removes credential material from both auth warnings and the default HTTP trace span, including query-auth WebSockets and credential-bearing path routes. The regression uses a real Router + TraceLayer and all 37 router tests pass; no real logs or credentials were inspected.

All CI has now converged green, including the full Windows matrix (12 executed checks successful, 8 conditional skips, 0 failures): https://github.com/octos-org/octos/actions/runs/30487421542

BH3GEI (approved):
LGTM. Redaction covers query tokens, WS auth, and path-segment credentials; also removes the old byte-slice panic risk and the short-token full-logging case. Solid tests, green CI.

PR #1876, opened 2026-07-30
fix(gemini): repair missing array item schemas

alan0x:
## Summary

- declare the structured element schema for `spawn_agent.items`
- declare `send_app_card.actions` as an object array with required `id` and `label`
- preserve valid array contracts and isolate Gemini tool declarations whose element contract is missing or empty
- add regression coverage for source schemas, object-array preservation, and malformed-tool isolation

## Root cause

Provider connection tests do not send the normal chat tool declarations, so Vertex authentication and connectivity can pass. Chat requests do send them, and Gemini rejects array schemas without `items` before generation. The initial fallback guessed `string[]`, which could corrupt object- or number-array contracts.

## Validation

- `cargo clippy -p octos-llm -p octos-agent --lib -- -D warnings`
- `cargo test -p octos-llm --lib` (485 passed, 3 ignored)
- `cargo test -p octos-agent tools::send_app_card::tests:: --lib` (7 passed)
- `cargo test -p octos-agent should_declare_element_schema_when_spawn_agent_items_is_array --lib`

## Review

Addressed the requested contract fix in `2577a030`: Gemini no longer invents array element types. Valid object-array schemas are preserved, while unknown external array contracts are excluded from Gemini tool declarations with a warning.


crates/octos-llm/src/gemini.rs
```diff
@@ -691,12 +691,16 @@ fn sanitize_schema_recursive(value: &mut serde_json::Value, depth: usize) {
...
+        // Gemini rejects array schemas without an element schema before the
+        // request reaches the model. Repair both a missing `items` field and
+        // the equivalent empty-object form emitted by some tool providers.
+        let needs_default_items = obj.get("type").and_then(|value| value.as_str()) == Some("array")
+            && match obj.get("items") {
+                None => true,
+                Some(items) => items.as_object().is_some_and(|items| items.is_empty()),
+            };
+        if needs_default_items {
+            obj.insert("items".to_string(), serde_json::json!({"type": "string"}));
```
anantheparty: [P2] This silently narrows every array with missing/empty `items` to `string[]`, even when its real elements are objects or numbers. For example, `send_app_card.actions` documents `{id,label,style}` objects and forwards any array verbatim to `org.octos.actions`; Gemini-generated strings would therefore pass server validation but corrupt the client contract. Please repair the real source schemas and avoid guessing an element type here.

anantheparty (changes requested):
The `spawn_agent.items` source-schema fix is correct, but the provider-wide fallback changes the contract of every underspecified array. `send_app_card.actions` is an in-tree counterexample: its items are `{id,label,style}` objects and the execution path only checks `is_array()` before forwarding the value verbatim as `org.octos.actions`. After this sanitizer, Gemini is told to emit strings, which pass that server check but violate the client contract. External MCP object/number arrays have the same risk.`n`nPlease complete the real source item schemas and avoid silently guessing `string` in the generic sanitizer (or explicitly reject/isolate schemas whose element contract is unknown). Add coverage showing an object-array contract remains an object array after Gemini sanitization.

crates/octos-llm/src/gemini.rs
alan0x: Addressed in 2577a030. The Gemini adapter no longer invents string[] for missing or empty array items; after sanitization it now isolates malformed tool declarations and logs the excluded tool. I also completed the in-tree send_app_card.actions source schema as an object array with required id and label fields, while preserving valid object-array contracts unchanged. Regression coverage pins missing or empty items, malformed-tool isolation, object-array preservation, and the send_app_card action schema. Validation: cargo test -p octos-llm --lib (485 passed, 3 ignored), targeted octos-agent schema and tool tests, and clippy with -D warnings.

anantheparty (approved):
Re-reviewed exact head `2577a030`. The provider no longer invents `string[]`; malformed/underspecified array tools are isolated after sanitization, valid object-array contracts remain intact, and `send_app_card.actions` now declares its real object item schema. Both synchronous and streaming Gemini tool paths use the shared builder, with regression coverage for missing/empty items and object-array preservation. The requested contract fix is complete; approving while the fresh CI run finishes.

anantheparty (dismissed):
Correction after tracing the shared sanitizer's third caller: the tool-declaration fix at `2577a030` is sound, but one provider-contract blocker remains.

`ResponseFormat::JsonSchema` still calls `sanitize_schema_for_gemini` and then unconditionally places the result in `responseSchema`. Because the new sanitizer intentionally leaves missing/empty array `items` unchanged, that structured-response schema now reaches Gemini and fails with a provider 400. This path cannot silently omit the schema like a malformed tool declaration, because doing so would violate the caller's structured-output contract.

Please validate the sanitized response schema with the same underspecified-array check and return an actionable local error before the provider request (making the generation-config builder fallible if needed), with coverage for both missing and empty `items`. Sync and stream share this config path, so fixing it once should protect both.

alan0x:
Addressed the remaining structured-response contract blocker at head `c4db41fd` after synchronizing current `main`.

- `build_gemini_generation_config` is now fallible and rejects sanitized `ResponseFormat::JsonSchema` values containing arrays with missing or empty `items` before any provider request.
- The local error remains a classified `InvalidRequest` and carries the concrete `gemini/<model>` provider label.
- Both synchronous and streaming Gemini calls use the shared fallible generation-config builder.
- Added regression coverage for both missing and empty `items`.

Validation:

- `cargo test -p octos-llm --lib` — 487 passed, 3 ignored
- `cargo clippy -p octos-llm --all-targets -- -D warnings`
- `git diff --check`

alan0x:
The issues identified in the previous review have been addressed and the latest `main` branch has been synchronized. The current implementation only fixes genuinely missing `items` arrays in the repository, and filters out any still-incomplete tool declarations before sending requests. The structured response schema is now strictly validated locally, with errors raised in advance to prevent sending requests that would be rejected by Vertex/Gemini.

Verification has passed: 38 Gemini unit tests in `octos-llm`, 2 schema-focused tests in `octos-agent`, as well as `cargo fmt` and `cargo check` for both `octos-llm` and `octos-agent`. Please proceed with a re-review.

BH3GEI (approved):
LGTM. Root-cause fix: schemas completed at the source (spawn_agent items, send_app_card actions) and the provider no longer invents contracts — underspecified arrays are isolated with a warn, JsonSchema path fails locally with InvalidRequest. The responseSchema blocker from the last review round is fixed with dual-path tests. Note: adjacent hunks to #1897, whichever merges second needs a small rebase.

PR #1880, opened 2026-07-31
feat(voice): add profile-scoped speech synthesis endpoint

alan0x:
## What changed

- add authenticated `POST /api/voice/synthesize` with a `{ "text": string }` request body and binary audio response
- resolve the caller's effective profile runtime, configured TTS provider, credentials, and selected voice
- preserve the configured cloud audio format for mp3, wav, pcm, and ogg_opus responses
- validate empty and oversized input and disable response caching
- enforce one in-flight REST synthesis per profile
- enforce a per-profile rolling one-minute budget of 20 requests or 20,000 characters; HTTP 429 responses include `Retry-After`
- serialize all on-device OminiX TTS work through one process-wide gate shared by REST synthesis and ordinary voice turns
- verbalize common mathematical operators before TTS so expressions such as `n=365` speak the equality instead of dropping it

## Why

Clients need a generic platform-level text-to-speech capability that uses the Octos configured TTS provider without coupling Octos to a specific product domain such as courses, lessons, OLL, or narration beats.

The original endpoint protected only concurrent requests from one profile. That did not protect the process-wide local OminiX model across profiles and did not bound sequential cloud use. The shared local gate and per-profile request/character window close those resource-governance gaps while keeping product sequencing client-side.

This deliberately does **not** add product-specific UI Protocol fields.

## Response and resource contract

- Response MIME follows the actual output bytes: `audio/mpeg`, `audio/wav`, `audio/pcm`, or `audio/ogg`.
- One REST synthesis may run per profile at a time.
- One local OminiX TTS synthesis may execute process-wide at a time, regardless of caller or profile.
- A profile may submit at most 20 synthesis requests or 20,000 characters per rolling one-minute window.
- Rate and concurrency rejection use HTTP 429 with an actionable `Retry-After` header.
- Each request is capped at 4,000 characters and successful responses use `Cache-Control: no-store`.

## Impact

Any authenticated UI can synthesize arbitrary text through the caller's current profile voice configuration. Ordinary voice replies and REST narration now share the same local-engine capacity boundary. Mathematical narration preserves common operators. Product-specific sequencing and playback remain entirely client-side.

## Validation

- synchronized with current `origin/main`
- `git diff --check origin/main...HEAD`
- `cargo test -p octos-cli --features api --lib --quiet` (2,659 passed, 6 ignored)
- targeted synthesis quota, `Retry-After`, shared local gate, audio encoding, and mathematical verbalization tests
- `cargo check -p octos-cli --features api`


anantheparty (changes requested):
I found one response-contract blocker in the new endpoint.

`/api/voice/synthesize` derives `Content-Type` only from the returned file extension, but the cloud file path does not preserve every supported encoding: `synthesize_volcano` writes every non-`wav` response as `.mp3`, while the same voice stack explicitly supports `pcm` and `ogg_opus`. A profile configured with either encoding will therefore return PCM/Ogg bytes as `audio/mpeg`, which can make browser clients decode or reject the response incorrectly. Please derive the filename/MIME from the actual configured encoding (or return typed audio metadata from the synthesis layer) and add coverage for mp3/wav/pcm/ogg_opus.

There is also no visible per-profile concurrency/rate boundary on this new generic REST path: `my_api` adds authentication, then the handler directly starts a cloud or OminiX synthesis. Since the stated contract allows any authenticated UI to synthesize arbitrary text, please either reuse/document an existing resource guard or add a bounded per-profile concurrency/rate policy so parallel calls cannot create unbounded local work or cloud spend.

alan0x:
Addressed the requested response-contract and resource-boundary changes in 0525a651, then synchronized current main in 60ce10fd.

- Cloud audio extension and MIME now share one encoding mapping: mp3 -> .mp3/audio/mpeg, wav -> .wav/audio/wav, pcm -> .pcm/audio/pcm, and ogg_opus -> .ogg/audio/ogg. Both HTTP file synthesis and websocket collection use the same extension helper.
- The REST handler now enforces one in-flight synthesis per profile. A concurrent request for the same profile returns 429; different profiles retain independent capacity. Weakly held limiter entries are cleaned up so profile churn does not create an unbounded map.
- Added regression coverage for all four encodings, response MIME, same-profile rejection, cross-profile isolation, and permit reuse.

Validation on the merged current-main head: cargo test -p octos-cli --features api --lib --quiet (2655 passed, 6 ignored), targeted speech/encoding tests, cargo check, cargo clippy, fmt check, and diff check.

anantheparty (dismissed):
Re-reviewed exact head `60ce10fd`. The audio-format contract is fixed, and the per-profile semaphore itself is race-safe and released through RAII. One resource-governance blocker remains from the original review:

- The limiter exists only in this REST handler and intentionally gives every profile independent capacity. OminiX is platform-wide, and the existing voice-turn path calls `synthesize_reply` without this limiter, so different profiles (and REST plus a normal voice turn) can still start unbounded concurrent work against the same shared local engine/GPU. Please put a bounded global/local-engine gate at the shared synthesis boundary, while keeping the per-profile gate for fairness.
- A one-in-flight semaphore is not a request/character rate budget: one profile can still run unlimited sequential cloud requests. Please add a bounded per-profile request/character window (or connect this endpoint to an existing quota) and return an actionable 429 with `Retry-After`. The current tuple error has no retry header, and the test exercises only the helper rather than the HTTP contract.

Please add coverage for the shared global cap and the actual 429 response. The MIME/extension mapping and its four-format tests now look correct.

alan0x:
Addressed the remaining resource-governance review at head 9db6e934. All local OminiX synthesis now passes through one process-wide shared gate used by both ordinary voice turns and the generic REST path. The REST endpoint also enforces a per-profile one-minute budget of 20 requests or 20,000 characters, and every concurrency/quota 429 includes Retry-After. Added regression coverage for the shared local cap, both quota dimensions, window reset, and the HTTP Retry-After response contract. Also synchronized current main and reran the full octos-cli API suite: 2,659 passed, 6 ignored; cargo check, fmt check, and diff check pass.

alan0x:
The branch still contains current `main`, and there are no unresolved inline review threads. The two requested-change clusters remain addressed at the current head:

- audio encoding, extension, and response MIME share one mapping for mp3/wav/pcm/ogg_opus;
- local OminiX calls share a process-wide gate across REST and ordinary voice turns;
- REST synthesis keeps per-profile concurrency fairness and request/character quotas;
- every concurrency/quota 429 includes `Retry-After`, with HTTP-level regression coverage.

The existing CI failures are baseline dependencies rather than TTS feature regressions: #1882 repairs the plugin environment-variable sync test, and #1883 repairs current-main Clippy and Windows `octos-core` checks. Requesting re-review of the feature changes while those focused baseline PRs complete.

alan0x:
The changes have been completed according to the previous review comments, and the latest `main` has been synced. Cloud audio now returns the file extension and MIME type based on the actual encoding. The endpoint now enforces per-profile limits on concurrency, request count, and character budget, and returns `429` with `Retry-After` when limits are exceeded. Local TTS now uses a process-level shared permit, sharing the same resource gate with voice turns.

Verification passed: 5 `speech_synthesis` tests, 3 targeted tests covering audio mapping / local shared gating / mathematical narration, `cargo fmt`, and `octos-cli api cargo check`. Please re-review.

BH3GEI (approved):
LGTM. Auth model matches the existing /api/my/* group with no cross-profile path; both earlier review rounds (MIME contract, global local-engine gate, rate window + Retry-After) are addressed in later commits. Remaining items are polish-level follow-ups.

PR #1883, opened 2026-08-01
refactor(plugin): compare BLOCKED_ENV_VARS constants directly instead of scraping source

alan0x:
Rebased onto `main` now that #1884 has landed. Everything this PR originally carried for the red CI — the clippy lint, `GIT_BIN` on Windows, the cross-platform test gating — is already on `main`, so it has been dropped. What remains is the one piece #1884 did **not** do, and did worse.

## What's left

`octos-plugin` keeps its own copy of `BLOCKED_ENV_VARS`, and the sync test used to prove the two agree by `include_str!`-ing another crate's source and regex-scraping the slice literal out of it. That is why it broke: #1881 moved the canonical list to `octos-core::env_hygiene` and left a `pub use` behind, which a regex over source text cannot match. The test panicked with `could not locate BLOCKED_ENV_VARS` — the invariant had silently stopped being checked.

#1884 fixed that by repointing the scraper at the constant's new home. That works today but goes stale the next time the constant moves.

This replaces the scraper with a direct comparison of the exported constants:

```rust
assert_eq!(
    octos_plugin::BLOCKED_ENV_VARS,
    octos_core::BLOCKED_ENV_VARS,
    "octos-plugin BLOCKED_ENV_VARS drifted from octos-core's canonical list. \
     Update both constants together."
);
```

`octos-plugin` takes a real `octos-core` dependency (no cycle — `octos-core` is the bottom crate) and drops the `regex` dev-dependency, which had no other use in the crate. The test can no longer be broken by moving, renaming, or re-exporting the definition, because it never looks at source text again. Net −40/+14.

## Dropped from this PR

- **clippy `useless_borrows_in_formatting`** — already on `main`. Note the fix is `*GIT_BIN`, not `GIT_BIN`: these are `LazyLock<PathBuf>`, so dropping the deref formats the wrapper.
- **`GIT_BIN` Windows candidates** — already on `main`, and the two versions had converged on essentially the same list.
- **`global_filter_is_neutralized_on_controller_ops` portability** — reverted to `main`'s `#[cfg(unix)]` gate. The relative-marker approach in this branch **did not work**: this branch's own `check-windows` run (30705286404) shows

  ```
  global_filter_is_neutralized_on_controller_ops ... FAILED
  sanity: the control (global visible) must run the filter, else the test is vacuous
  test result: FAILED. 341 passed; 1 failed
  ```

  The control arm's filter still never executes on the runner, so the test cannot be made meaningful there by changing the marker path. It was masked because `check-windows` is `continue-on-error` on PRs and reports `pass` regardless — worth knowing when reading that job's status on any PR.

## Verification

| gate | result |
| --- | --- |
| `cargo clippy -p octos-plugin -p octos-core --all-targets -- -D warnings` | clean |
| `cargo check -p octos-plugin -p octos-core --target x86_64-pc-windows-msvc --tests` | clean |
| `cargo test -p octos-plugin` | 81 passed, `blocked_env_vars_match_core_canonical_list` ok |
| `cargo test -p octos-core` | 348 passed |

The direct-comparison test is also confirmed green on Windows in this branch's earlier run (`blocked_env_vars_match_core_canonical_list ... ok`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -428,6 +428,14 @@ jobs:
       - name: gateway_runtime regression (octos-cli)
         run: cargo test -p octos-cli gateway_runtime::tests --features api -- --nocapture
 
+      # Peer staging/fencing lives in the `api`-gated ui_protocol module, so the
+      # unfeatured runs above never compile it — same reason as the §6 guard.
+      # These cover slug reservation, worktree fencing, and the rollback
+      # isolation guard (one peer's cleanup must not delete a SIBLING's
+      # worktree, which a repo-global `git worktree prune` used to do).
+      - name: Peer staging + worktree fencing (octos-cli)
+        run: cargo test -p octos-cli --features api peer
+
   # Lightweight check that exercises the `matrix` feature gate. The default
   # `FEATURES` env above intentionally omits `matrix` to keep the main `check`
   # job lean, but the Dockerfile and several release configurations enable it.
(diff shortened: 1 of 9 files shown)


alan0x:
Follow-up validation after commit f9d9264f: the rerun's Linux check completed successfully, and Windows Test octos-core is now green before proceeding to the remaining Windows crates. The earlier failure was limited to the test fixture embedding a Windows path in git-config; the production Git candidate fix itself passed. Manual CI: https://github.com/octos-org/octos/actions/runs/30706872595

PR #1892, opened 2026-08-02
fix(agent): k3 agentic-loop fixes — spiral signature, KV-cache-friendly tier-1, model-id gaps, compaction instruction priority

ZhangHanDong:
Five fixes surfaced by a full day of live **kimi k3** (Kimi Code API) usage in octos-tui.

## 1. Shell-spiral false positives

`recover_shell_retry`'s RetryLimit arm fired on ANY ≥3 failing shell results in a turn. Agentic models fan out many **distinct** exploratory commands where several legitimately exit non-zero (a `grep` with no match exits 1), so healthy exploration turns were force-ended and the user saw *"Turn completed, but the TUI did not receive a final assistant answer"* — twice in one afternoon.

A spiral now additionally requires the failures to **concentrate**: ≥3 repeats of the same whitespace-normalized command, or of the same non-empty failure text (catches flag-shuffling against one wall, e.g. `cargo test` → `--all` → `--workspace` all hitting "could not find Cargo.toml"). Genuine spirals still die; distinct exploration does not.

## 2. Tier-1 micro-compaction was KV-cache hostile

Age-based pruning rewrites tool results **deep in history** on every iteration, which invalidates a provider's prefix cache from that point onward. For a 13-round turn at ~35k input tokens that is a full re-prefill per round — on a subscription/input-billed model this burns quota fast and slows every round's first token.

Split into `Tier1Pass::OversizedOnly` (per iteration; only clears just-landed oversized results near the prefix tail — cheap for the cache) and `Tier1Pass::Full` (turn start; also prunes stale). `prune()` keeps its signature and delegates to `Full`, so existing callers and tests are unaffected.

Note: the other prefix-cache killer we found — mid-turn LRU tool eviction mutating the `tools` array — no longer exists on main, since eviction was removed upstream.

## 3. Kimi Code model-id gaps

- `k3-256k` did not match the `fixed_temperature` rule, so requests carried a temperature and the endpoint 400s with *"invalid temperature: only 1 is allowed for this model"*.
- The bare `k3` / `kimi-for-coding*` ids (what the Kimi Code API actually serves) did not get the `reasoning_content` stub that multi-round tool calls need — only `kimi-k2`/`kimi-k3` substrings did.

## 4. Post-compaction instruction drift

After a 216→15 item compaction (94k→5k tokens), the model followed the summary's **stale plan** and ignored the newest user instruction — the user's verdict was "驴唇不对马嘴" (completely off-target). Root cause is a weight structure that lies: the summarizer prompt asks for the *"current goal / what remains"*, the summary row sits adjacent to the system prompt, and the newest user message has no special standing.

Two framing changes, no behavior/threshold changes:
- The summarizer system prompt now states its output is **BACKGROUND**, must never restate historical goals as the current task, and that the newest user message defines the task.
- The rendered summary row carries a background header and a precedence footer.

**Render-only**: the stored transcript item and the hash chain are untouched, and line 1 keeps the exact `[Conversation summary]` sentinel that `message_repair` matches on (an earlier version of this change broke that match — the test suite caught it).

## Verification

- `cargo test -p octos-agent --lib` → 2492 passed
- `cargo test -p octos-llm --lib` → 483 passed
- `cargo test -p octos-cli --lib` → 1005 passed
- Two agent-spec contracts included, both `Spec-Passing: true` (lifecycle 4/4 and 5/5)
- Built with the release feature set and run live against octos-tui


BH3GEI (dismissed):
Good direction — deduping repeated no-output failures before they burn the retry budget is the right fix, but the current predicate misses the motivating case.

**1. `strip_exit_code_suffix` doesn't treat `"(no output)"` as empty** — `crates/octos-agent/src/loop_runner.rs` (diff hunk at 159-168)

The failure this PR targets is the shell tool returning literally `"(no output)\n\nExit code: 1"` (that exact string is built in `crates/octos-agent/src/tools/shell.rs:869`). After stripping the exit-code suffix the remainder is `"(no output)"` — non-empty — so the dedupe never fires for the grep-with-no-matches case that motivated the PR, and three distinct no-output commands still escalate to RetryLimit. One-line fix: treat the tool's `"(no output)"` sentinel as empty after stripping. Please also make the test fixture model the real tool output (`"(no output)\n\nExit code: 1"`) rather than a bare exit-code string, so the regression test actually covers the production shape.

**2. `reasoning_style` missing the `k3-` model prefix** — `crates/octos-llm/src/openai.rs:128`

The prefix list gates which models get the reasoning-style request shape; without `k3-` the k3 family silently falls back to the default style. One-line addition.

**3. Third compaction channel still unaddressed**

The PR fixes compaction-failure dedupe on two paths but the third channel (`crates/octos-services/src/compaction.rs:155` and `:286`) can still emit the same repeated-failure pattern. Fine to handle in a follow-up — please file it so it doesn't get lost.

#1 is the blocker; the rest is small.

Issue #1886, opened 2026-08-02
MCP tools from config.json are silently evicted by profile narrowing, killing the transport


ymote:
## Summary

MCP servers declared in `config.json` under `mcp_servers` are started, their tools are registered, and then they are silently evicted by profile narrowing. Evicting them drops the last `Arc` references to the running service, which cancels the transport and exits the MCP child process.

The result is an agent that is missing tools the operator explicitly configured, with no warning and no error. The only trace is a pair of INFO lines.

Observed on octos 2.0.2 (`e17b47b5d`) under `octos acp`, but the mechanism is not ACP-specific — `finalize_tool_registry` is shared.

## What happens

With an MCP server configured and the default profile:

```
INFO connection: MCP server started server="node" tools=11 concurrency_class=Safe
INFO connection:serve_inner: task cancelled
INFO connection:serve_inner: Child exited gracefully exit status: 0
INFO connection:serve_inner: serve finished quit_reason=Cancelled
```

Registration succeeds with all 11 tools, then the service is cancelled roughly 1 ms later. The agent subsequently reports only the lean tool set and has no idea the configured server ever existed.

## Why

1. `commands/acp.rs:431` starts the configured servers and registers their tools:

   ```rust
   if !config.mcp_servers.is_empty() {
       match octos_agent::McpClient::start(&config.mcp_servers).await {
           Ok(client) => client.register_tools(&mut tools),
           Err(e) => eprintln!("Warning: MCP initialization failed: {e}"),
       }
   }
   ```

2. `finalize_tool_registry` (`commands/acp.rs:197`) ends with profile narrowing:

   ```rust
   // M8.3: profile narrowing runs AFTER every other filter (spawn_only preserved).
   profile.apply_to_registry(tools);
   ```

3. `ToolRegistry::filter_by_profile` (`tools/registry.rs:839`) is a `retain()` over the profile's allow-list. MCP tool names are not in the lean `coding` profile's list, so they are removed.

4. `McpService` is `Arc<RunningService<RoleClient, ClientInfo>>` (`agent/src/mcp.rs:40`). Each `McpTool` holds a clone, so the registry is the only owner once `McpClient` is consumed by `register_tools`. Removing the tools drops the last references, the `RunningService` is dropped, and the transport cancels.

So the eviction is not just "the tools are hidden" — it tears down the connection.

## Reproduce

1. Add a stdio MCP server to `~/.config/octos/config.json`:

   ```json
   {
     "mcp_servers": [
[...]

PR #1894, opened 2026-08-02
feat(acp): persist sessions and implement session/load

ymote:
Relates to #1886 (same area: ACP wiring that exists everywhere except ACP).

## Problem

ACP sessions are memory-only. `AcpSession.history` is a `Mutex<Vec<Message>>` that nothing writes anywhere, so a crash or a supervisor restart loses the whole conversation — the agent comes back mid-task with no idea what it was doing, while whatever supervises it reports it healthy. Silent amnesia is worse than staying down, because nothing looks wrong.

`SessionManager` already backs the REST and WebSocket paths. It was simply not reachable from ACP.

## Change

`session/prompt` writes each turn through to `SessionManager`; `session/load` reads a conversation back. `SessionKey` is a newtype over `"channel:chat_id"`, so `"acp"` becomes another channel beside the existing ones — no new identity concept.

## The part worth reviewing carefully

`add_message_with_seq` is fail-closed for Assistant and Tool rows: it rejects them unless the caller stamps the originating turn's `thread_id`, because a derived one picked the wrong sibling user under rapid-fire turns.

Unstamped, **the user row persisted and the assistant row was rejected** — a stored conversation with the answers missing, which is worse than no conversation at all. Each turn now stamps one thread id across the rows it produced. I'd appreciate a check that this is the intended contract for that path.

## Capability

`loadSession` is advertised only because load genuinely restores. It stayed `false` while load was a stub, and should return to `false` if the store is ever removed — a client that sees the capability will call it, and a load that silently returns nothing looks exactly like an agent that has forgotten everything.

## Failure handling, deliberately soft

- a store that won't open degrades to `None` with an error log rather than stopping the agent
- a failed write is logged per message rather than aborting the turn
- an unknown session id loads as empty with a warning rather than erroring

Losing persistence must not also lose the work, or refuse to serve.

## Testing

New integration test drives **two independent transports over one store**: the first holds a conversation, the second loads that id and must see the earlier turns. It fails exactly as described above without the `thread_id` stamping.

```
982 lib tests pass
4 ACP integration tests pass
```

Also verified against a built binary that the real `ConfigAgentFactory` path advertises `loadSession`, creates the store under `data_dir`, and serves `session/load` for a session id created by a separate process.

**Not verified:** a full conversation surviving a restart on a real binary, because that needs live LLM credentials I don't have. The write/read logic is covered by the integration test; the factory wiring by the manual run above. Worth someone with credentials confirming the two meet.

### DIFF
--- a/crates/octos-cli/src/commands/acp.rs
+++ b/crates/octos-cli/src/commands/acp.rs
@@ -80,6 +80,9 @@ use octos_llm::{EmbeddingProvider, LlmProvider};
 use octos_memory::{EpisodeStore, MemoryStore};
 use tokio::sync::Mutex;
 
+use octos_bus::session::SessionManager;
+use octos_core::SessionKey;
+
 use super::Executable;
 use crate::config::Config;
 
@@ -143,6 +146,15 @@ impl Executable for AcpCommand {
 /// Everything a spawned prompt turn needs, shared behind the session map.
 struct AcpSession {
     agent: Arc<Agent>,
+    /// Where this conversation is persisted. ACP sessions were memory-only, so a
+    /// crash or supervisor restart lost the whole conversation — the agent came
+    /// back mid-task with no idea what it had been doing while the fleet reported
+    /// it healthy. SessionManager already backs the REST and WebSocket paths;
+    /// `SessionKey` is a newtype over "channel:chat_id", so "acp" is simply
+    /// another channel alongside the existing ones.
+    session_key: SessionKey,
+    /// Held on the session rather than threaded through run_prompt_turn's callers.
+    session_store: Option<Arc<Mutex<SessionManager>>>,
     /// Conversation history accumulated across prompt turns in this session.
     history: Mutex<Vec<octos_core::Message>>,
     /// Flipped to `true` by a `session/cancel` notification; the agent's
@@ -163,6 +175,12 @@ trait SessionAgentFactory: Send + Sync {
     /// Build a runnable agent rooted at `cwd`, returning it plus the shared
     /// shutdown flag wired into it (flipped on `session/cancel`).
     async fn build(&self, cwd: PathBuf) -> Result<(Arc<Agent>, Arc<AtomicBool>)>;
+
+    /// Where ACP conversations are persisted, if anywhere. `None` means sessions
+    /// will not survive a restart — the test factory, or a store that failed to open.
+    fn session_store(&self) -> Option<Arc<Mutex<SessionManager>>> {
+        None
+    }
 }
 
 /// Build the tool registry shared by every ACP session: built-ins rooted at
@@ -252,6 +270,8 @@ struct AcpSharedStores {
     embedder: Option<Arc<dyn EmbeddingProvider>>,
     agent_config: AgentConfig,
     memory_refresh_enabled: bool,
+    /// Opened once per process; shared by every session.
+    sessions_store: Option<Arc<Mutex<SessionManager>>>,
     max_inject_tokens: usize,
     data_dir: PathBuf,
     config: Config,
@@ -346,6 +366,20 @@ impl AcpSharedStores {
             embedder,
             agent_config,
             memory_refresh_enabled,
+            // Degrades to None with a loud log rather than failing the agent: an
+            // unpersisted session is worse than a persisted one and far better
+            // than none. Silent non-persistence is the bug this exists to fix,
+            // so it must not be reintroduced here.
+            sessions_store: match SessionManager::open(&data_dir.join("sessions")) {
+                Ok(store) => Some(Arc::new(Mutex::new(store))),
+                Err(e) => {
+                    tracing::error!(
+                        error = %e,
+                        "session store unavailable; ACP conversations will NOT survive a restart"
+                    );
+                    None
+                }
+            },
             max_inject_tokens,
             data_dir,
             config,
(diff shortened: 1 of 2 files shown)


ymote (commented):
Reviewed the diff rather than the description. Core design is sound and the central claim holds: `handle_load_session` sets `history`, and `run_prompt_turn` clones it at `acp.rs:1232`, so a loaded conversation genuinely feeds the next prompt. The capability reasoning (advertise `loadSession` only because load really restores), the fail-soft degradation, and deliberately NOT hydrating in `session/new` are all right.

Four things below, one blocking.

## Blocking — `check` is red

Step 6, **Check formatting**. Trailing whitespace at `acp.rs:1293` — the blank line opening the write-through block. `cargo fmt --all` fixes it.

## 1. Duplicate writes if two turns race on one session

`persisted_upto` is captured under one lock acquisition and `fresh` is read under a **separate** one:

```rust
let persisted_upto;
{ let mut h = session.history.lock().await; persisted_upto = h.len(); h.extend(resp.messages); }
// <- lock released here
let fresh = { let h = session.history.lock().await; h.get(persisted_upto..) ... };
```

Turns are deliberately spawned off the dispatch loop, so nothing serializes two `session/prompt`s for the same session. Interleaved: turn A captures `persisted_upto = 0`, both extend, then A writes rows `0..4` — including B's — and B writes `2..4` again. Duplicated rows in the store.

The comment already names the hazard ("the store appends, so re-sending earlier turns would duplicate them"); it just does not hold across the gap. Reading `fresh` inside the same lock scope as the `extend` closes it.

## 2. `session/load` can silently break `session/cancel`

`handle_load_session` ends with:

```rust
sessions.lock().await.insert(req.session_id.clone(), session);
```

which overwrites any existing entry. `handle_cancel` looks the session up in that map and flips **that** session's `shutdown` flag — but an in-flight turn holds an `Arc` to the **old** `AcpSession`, with the old flag:

```rust
if let Some(session) = map.get(&notif.session_id) {
    session.shutdown.store(true, Ordering::Release);
```

So loading an id that is mid-turn leaves a running turn that cancel can no longer stop, and the client gets a success response for a cancel that did nothing. Worth rejecting a load for an already-active session, or at minimum warning.

## 3. Profile-less `SessionKey`

`SessionKey::new("acp", id)` yields `"acp:<id>"` — no profile component. `octos-bus/src/session.rs` notes production keys are typically `<profile>:api:<chat>`, and that profile scoping is what keeps two profiles' dirs from colliding.

This is the same shape that silently disabled peer bookkeeping (fixed in #1888): a key without a profile parsed as "not a peer session" and every caller skipped its bookkeeping without a word. Not asserting it is wrong here — ACP may be intentionally single-profile — but it is worth stating explicitly, because if ACP is ever multi-tenant these sessions share one keyspace.

## 4. Minor — the `turn_thread` fallback invents a thread

```rust
.unwrap_or_else(|| uuid::Uuid::now_v7().to_string());
```

With no User row in `fresh`, this mints a fresh uuid. That satisfies the store's fail-closed check but writes assistant/tool rows into a thread containing no user row — the orphan the fail-closed contract exists to prevent. Likely unreachable, but a `warn!` would beat a silent uuid, since "we do not know the thread" is being converted into "invent one".

## On the untested claim

The PR is straight about not verifying a real restart without credentials, which is the right call to surface. Note that (1) above is reachable without credentials — two concurrent `session/prompt` calls against one session id in the integration test would exercise it.

ymote:
Closing. The branch `feat/acp-session-persistence` is not deleted, so the work is recoverable.

For whenever this is picked up again, the review notes on it still stand — `check` was red on formatting only (trailing whitespace at `acp.rs:1293`, `cargo fmt --all`), plus three correctness items: the duplicate-write race when two turns interleave on one session, `session/load` overwriting a live session so `session/cancel` silently stops working on the in-flight turn, and the profile-less `SessionKey`.

ymote:
Reopened with all four findings addressed. Force-pushed, rebased onto current `main`, squashed to one commit.

**Formatting** — `cargo fmt`; `--check` is clean.

**Duplicate-write race.** The delta was captured as an index under one guard and sliced under another, so two turns interleaving on one session computed overlapping ranges and wrote the same rows twice. The rows are now taken under the same guard that appends them, so each turn persists exactly what it added.

**`session/load` evicting a live session.** A bare insert dropped the existing entry and with it the `shutdown` flag an in-flight turn watches — a `session/cancel` for that turn flipped a flag nobody read, so cancellation silently stopped working. Loading an id the agent already holds is now a no-op.

Worth noting where the check sits: my first attempt put it *after* `factory.build()`, and the new test failed on a redb lock, because `EpisodeStore::open` cannot run twice on a live session. It now runs before the build, which is both correct and avoids constructing an agent that would be discarded. The test found a bug in the fix for the bug.

**Profile-less `SessionKey`.** Now `with_profile`, taking the factory's profile and falling back to `MAIN_PROFILE_ID`.

Added `should_not_evict_a_live_session_on_reload`, which fails without the third fix.

```
982 lib tests · 5 ACP integration tests · fmt clean
```

Also exercised on a release binary built with the CI feature set: `kill -9` the host, the supervisor restarts it, the session resumes and the agent recalls a codeword set before the kill.

One thing still worth a maintainer's eye: the `thread_id` contract. I inferred it from `derive_thread_id_for_new_write` rejecting unstamped Assistant rows rather than from documentation, so please confirm one-id-per-turn is what that path expects.

PR #1895, opened 2026-08-02
feat(acp): persist sessions and implement session/load

ymote:
Supersedes #1894, which was closed with review notes. All four are addressed; GitHub would not reopen it because the branch head was force-pushed after closing.

Relates to #1886 (same area: ACP wiring that exists everywhere except ACP).

## Problem

ACP sessions are memory-only. `AcpSession.history` is a `Mutex<Vec<Message>>` that nothing writes anywhere, so a crash or a supervisor restart loses the whole conversation — the agent comes back mid-task with no idea what it was doing, while whatever supervises it reports it healthy. Silent amnesia is worse than staying down, because nothing looks wrong.

`SessionManager` already backs the REST and WebSocket paths. It was simply not reachable from ACP.

## Review findings from #1894, all fixed

**Formatting** — `cargo fmt`; `--check` clean.

**Duplicate-write race** — the delta was captured as an index under one guard and sliced under another, so two turns interleaving on one session computed overlapping ranges and wrote the same rows twice. Rows are now taken under the same guard that appends them.

**`session/load` evicting a live session** — a bare insert dropped the existing entry and with it the `shutdown` flag an in-flight turn watches, so a `session/cancel` for that turn flipped a flag nobody read and cancellation silently stopped working. Loading an id the agent already holds is now a no-op.

The check runs *before* `factory.build()`. My first attempt put it after, and the new test failed on a redb lock — `EpisodeStore::open` cannot run twice on a live session. Checking first is both correct and avoids building an agent that would be discarded.

**Profile-less `SessionKey`** — now `with_profile`, taking the factory's profile and falling back to `MAIN_PROFILE_ID`.

## The `thread_id` contract

`add_message_with_seq` is fail-closed for Assistant and Tool rows: it rejects them unless the caller stamps the originating turn's `thread_id`. Unstamped, the user row persisted and the assistant row was rejected — a stored conversation with the answers missing.

Each turn now stamps one id across the rows it produced. **I inferred this from the rejection rather than from documentation**, so please confirm one-id-per-turn is what that path expects.

## Failure handling, deliberately soft

- a store that will not open degrades to `None` with an error log rather than stopping the agent
- a failed write is logged per message rather than aborting the turn
- an unknown session id loads as empty with a warning

Losing persistence must not also lose the work, or refuse to serve.

## Testing

Two integration tests, each failing without its fix: one drives two independent transports over a single store and requires the second to see the first's turns; the other reloads a live session and requires it to keep working.

```
982 lib tests · 5 ACP integration tests · fmt clean
```

Also exercised on a release binary built with the CI feature set: `kill -9` the host process, the supervisor restarts it, the session resumes, and the agent recalls a codeword set before the kill.

### DIFF
--- a/crates/octos-cli/src/commands/acp.rs
+++ b/crates/octos-cli/src/commands/acp.rs
@@ -1341,7 +1341,27 @@ async fn run_prompt_turn(
                         .iter()
                         .find(|m| matches!(m.role, octos_core::MessageRole::User))
                         .and_then(|m| m.thread_id.clone().or_else(|| m.client_message_id.clone()))
-                        .unwrap_or_else(|| uuid::Uuid::now_v7().to_string());
+                        .unwrap_or_else(|| {
+                            // No user row to inherit from, so there is nothing to
+                            // stamp these rows onto. Minting an id keeps the write
+                            // fail-OPEN rather than dropping the turn, but it
+                            // creates exactly the orphan the store's fail-closed
+                            // check exists to prevent: assistant/tool rows in a
+                            // thread containing no user message. Expected to be
+                            // unreachable — say so out loud rather than silently
+                            // converting "we do not know the thread" into
+                            // "invent one", which is what made the original
+                            // thread-id bug so expensive to find.
+                            let minted = uuid::Uuid::now_v7().to_string();
+                            tracing::warn!(
+                                thread_id = %minted,
+                                rows = fresh.len(),
+                                "ACP turn produced no user row; persisting its \
+                                 messages under a MINTED thread id — they will \
+                                 load as a thread with no user message"
+                            );
+                            minted
+                        });
 
                     let mut guard = store.lock().await;
                     for mut message in fresh {
--- a/crates/octos-cli/src/commands/acp.rs
+++ b/crates/octos-cli/src/commands/acp.rs
@@ -80,6 +80,9 @@ use octos_llm::{EmbeddingProvider, LlmProvider};
 use octos_memory::{EpisodeStore, MemoryStore};
 use tokio::sync::Mutex;
 
+use octos_bus::session::SessionManager;
+use octos_core::SessionKey;
+
 use super::Executable;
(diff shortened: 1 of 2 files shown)


BH3GEI (changes requested):
Thanks for pushing ACP session persistence forward — the overall shape (load → replay → resume) is right, but four correctness issues need fixing before this can land.

**1. `session/load` never replays `session/update` notifications (protocol MUST violation)** — `crates/octos-cli/src/acp.rs:1422-1468`

The ACP spec requires `session/load` to replay the session's history as `session/update` notifications before the response resolves — that's the entire mechanism a client (Zed etc.) has to reconstruct transcript state. The current implementation loads the session and returns success without emitting any replay, so a client attaching to an existing session renders an empty transcript. This is the load-bearing part of the feature; without it the method is indistinguishable from `session/new` with a prefilled id.

**2. Storage root is nested one level too deep** — `acp.rs:376`

`SessionManager::open(data_dir.join("sessions"))` produces `<data_dir>/sessions/sessions/...` because `SessionManager` already namespaces under its own `sessions/` component. Every other surface (REST hydrate, TUI resume) reads from `<data_dir>/sessions/`, so ACP-persisted sessions are invisible to them and vice versa. Should be `SessionManager::open(&data_dir)`.

**3. TOCTOU on session existence check** — `acp.rs:1425-1479`

`contains_key` → validate → `insert` across an await leaves a window where two concurrent `session/load` calls for the same id both pass the check and double-attach. Use a single atomic entry/insert (or `DashMap::entry`) so the existence check and the registration are one critical section.

**4. Restored history bypasses `ResumePolicy::sanitize`**

History replayed into a resumed session skips the sanitize pass that the normal resume path applies, so stale/invalid items that the TUI/REST resume would have filtered can reach the model context over ACP. Route the loaded transcript through the same policy before replay.

Happy to re-review quickly once these are addressed — #1 and #2 are the blockers, #3/#4 are small once the replay path exists.

PR #1896, opened 2026-08-03
ci(check-arm): revert both mitigations — measured, and it is not our workload

ymote:
I shipped two fixes for `check-arm` (#1890, #1891) on theories that were wrong. Reproducing the job on matched hardware disproves both — and the two remaining candidates too, including the one I was about to recommend next.

## What was measured

An `aarch64` Ubuntu box matched to the runner (**4 CPU, 15.7 GB**), running **this job's entire step sequence in order**: all-features build → test-binary build → all nine crate test steps → the two agent steps.

| | CI | matched hardware, same sequence |
| --- | --- | --- |
| step 16 `octos-agent --lib` | **46 min**, never finishes | **29.3 s**, exit 0 |
| step 16 result | runner killed | 2462 passed, 0 failed |
| step 17 `octos-agent --tests` | never reached | 9 passed, 0 failed |
| memory available | unknown (no logs survive) | **never below ~14.7 GB** |

29 seconds versus 46 minutes and a dead runner — same architecture, same sequence, memory to spare.

## Four theories, all disproven

| theory | verdict |
| --- | --- |
| test contention starving the runner (#1890) | serialising moved time-to-death by **0.7 min** (46.0 → 46.7) |
| a hanging test (#1891 instrumentation) | it completes in **29 s** |
| accumulated RAM from earlier steps | fully replicated; **memory never moved** |
| architecture | same arch, completely healthy |

Two things worth stating plainly:

- **`timeout-minutes` could never have worked.** Step timeouts are enforced *by the runner agent*, which is the thing dying. It cannot preserve logs from a runner that is being killed. I should have known that before proposing it.
- **The matrix split would not help.** Splitting lib/integration onto separate runners — what `test-octos-agent` does for a "fresh 16GB allocation" — targets accumulation, which is now measured and absent. That was going to be my next recommendation.

## What this changes

Both mitigations are removed as unnecessary: `RUST_TEST_THREADS=1` only makes the step slower, and the timeout cannot fire.

The findings are recorded **in the job itself**, so the next person doesn't repeat the three attempts I made. Net −47/+27, and the comment now says what was ruled out and why.

On this evidence the fix is a **runner** change — a self-hosted `arm64` runner, or retry-on-infra-failure — not workflow config. Consistent with GitHub's own annotation: *"The hosted runner lost communication with the server."*

## Incidental, neither a defect

Both surfaced during the repro and cost time, so they're recorded in the commit message:

- `should_block_shell_workspace_write_when_read_only_profile` fails when `bubblewrap` is absent — the Bwrap backend degrades and the read-only sandbox stops blocking. Passes once installed. Looks alarming; isn't.
- `rustup --profile minimal` installs a rustfmt **shim** but not the component, so `binary_on_path("rustfmt")` returns true while rustfmt cannot actually run. The post-edit formatting tests' "not on PATH" skip guard doesn't cover that case.

## Note

This does not make `check-arm` green. It stops us paying for wrong answers, and leaves the next person a measurement instead of three disproven guesses.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -505,62 +505,42 @@ jobs:
         run: cargo test -p octos-sandbox
       - name: Test octos-swarm
         run: cargo test -p octos-swarm
-      # RUST_TEST_THREADS=1 + timeout-minutes on the octos-agent steps ONLY.
+      # NOTE — these two steps are where check-arm dies, and it is NOT our code.
       #
-      # STATUS: RUST_TEST_THREADS=1 (#1890) did NOT fix this — the job still
-      # died at `Test octos-agent (lib)` on run 30757917006. Step-duration data
-      # disproved the OOM-from-parallelism theory it was based on:
+      # check-arm fails on roughly half of `main` pushes, always the same way:
+      # GitHub kills the runner ("The hosted runner lost communication with the
+      # server ... starves it for CPU/Memory") partway through
+      # `Test octos-agent (lib)`, after steps 1-15 pass. A killed runner uploads
+      # no logs, so every failure destroys its own evidence.
       #
-      #   run 30733580733  no serialisation  steps 16-21 ALL done in ~16 min  PASS
-      #   run 30737982761  no serialisation  step 16 alone ran 46 min         KILLED
-      #   run 30757917006  serialised        step 16 alone ran 46.7 min       KILLED
+      # MEASURED on an aarch64 Ubuntu box matched to the runner (4 CPU, 15.7GB),
+      # running THIS ENTIRE JOB's step sequence in order — all-features build,
+      # test-binary build, all nine crate test steps, then these two:
       #
-      # Serialising changed nothing (46 -> 46.7 min), so contention is not the
-      # cause. On the PASSING run step 16 finished in minutes; on failures it
-      # never finishes. Step 16 HANGS, and GitHub eventually kills the
-      # unresponsive runner.
+      #   step 16 `octos-agent --lib`   29.3s, 2462 passed, exit 0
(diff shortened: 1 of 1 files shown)


anantheparty:
2026-08-27 维护轮补充观测(不重开四条已证伪理论,新数据是关于**队列占用**,不是死因):

上午 00:25Z 的 main run(#2121 合并触发)check-arm 挂起期间:
- 整个 run 卡 in_progress **1.5h+**(其余 21 job 在 ~45min 内全部绿);
- 03:31Z 的下一轮 main run(#2124 合并)**零 job 启动、纯排队 40+ 分钟**——`ci-CI-refs/heads/main` 并发槽被挂起的 run 占着(`cancel-in-progress` 对 push 恒为 false,这是保护 release gate 的正确设计),所以 follow-up merge 只能等;
- run 最终 failure 后(retry-check-arm 两轮 success 兜底),失败 job 的日志 BlobNotFound(已知:runner 死亡=无日志上传)。

**结论**:retry workflow 解决的是"红",没解决"堵"。每次 hosted arm runner 死亡的代价 = 该次已烧掉的全部耗时 + 后续所有 main push 的合并延迟。

**可行选项(都属 runner/concurrency 层面,非 config 微调)**:
1. 你在 #1911 结论里的正解不变:**自托管 arm64**;
2. 若自托管短期不可行:把 check-arm 拆出主 CI workflow(run id 不再共享并发槽),让它失败/挂起只影响自己的 workflow_run;
3. 或接受现状(retry 补丁),但把 `retry-check-arm.yml` 的注释升级一句:"此 job 挂起会堵 main 队列,别在同一 slot 排别的关键工作"。

今天这两次死亡 + 一次成功的全序列见 run 32915016718 / 32926771775。

PR #1897, opened 2026-08-03
fix(gemini): recover missing Gemini 3 thought signatures

alan0x:
## Summary

- preserve real Gemini thought signatures when rebuilding tool-call history
- use the documented skip_thought_signature_validator fallback only for missing signatures in the current Gemini 3 user turn
- keep parallel sibling calls unsigned and leave Gemini 2 behavior unchanged
- cover both streamed signature capture and reconstructed-history behavior

## Context

Gemini 3 rejects a follow-up function-response request with HTTP 400 when the first functionCall part in a current-turn step has no thoughtSignature. This can occur when Octos reconstructs tool history whose provider metadata is unavailable, including persisted or transferred histories.

The fallback is intentionally limited to the current turn because Google documents it as a last resort that can reduce model performance. Real signatures always take precedence.

## Validation

- cargo test -p octos-llm
- cargo fmt --all -- --check
- cargo clippy -p octos-llm --all-targets -- -D warnings
- git diff --check

### DIFF
--- a/crates/octos-llm/src/gemini.rs
+++ b/crates/octos-llm/src/gemini.rs
@@ -55,6 +55,12 @@ fn next_gemini_tool_call_id() -> String {
 /// Default AI Studio base URL (the `generativelanguage.googleapis.com` host).
 const STUDIO_BASE_URL: &str = "https://generativelanguage.googleapis.com/v1beta";
 
+/// Google documents this sentinel for manually reconstructed Gemini 3 tool
+/// history when the original response part did not carry a thought signature.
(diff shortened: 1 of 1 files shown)


BH3GEI (approved):
LGTM. Mechanism matches Google's own gemini-cli history hardening (synthetic skip_thought_signature_validator on the first functionCall of the current turn, real signatures always win, parallel siblings unsigned, Gemini 2 untouched). Conservative scoping and a solid test matrix. Merged cleanly after #1876.

PR #1903, opened 2026-08-03
fix(pipeline): only assign models the provider router registered

### LINKED ISSUE #1901: deep_research silently downgrades to the default provider when model assignment picks an unregistered model
>

## What is fixed

## Remaining work

## Original report

```
```

ymote:
Closes #1901 — the **root cause**. #1902 fixed the crash; this fixes why it happened.

## Why the bug survived #1902

#1902 stopped `resolve_provider` propagating the failure with `?`, so the pipeline no longer dies. But nothing stopped assignment **choosing** an unresolvable model, so after that merge the failure just got quieter:

| | before #1902 | after #1902 | after this |
| --- | --- | --- | --- |
| run outcome | crashes | **completes** | completes |
| model actually used | — | **default provider** | the requested lane |
| visibility | loud error | a `warn!` | nothing to warn about |

The middle column is the problem: a `deep_research` run that asks for `strong`, looks completely healthy, and quietly returns weaker output. A crash is at least honest.

## Root cause

```rust
pub fn assign_from_catalog_dir(graph: &mut PipelineGraph, data_dir: &Path)
```

Assignment **was never given the router**. It could not know which models were registered, so it rewrote DOT lane keys (`model="strong"`) into concrete models picked blind from the *unfiltered* `model_catalog.json` — 93 models, most unregistered — whenever a profile had no filtered `pipeline_models.json`.

That is not a logic bug in assignment. It is missing information.

## The fix

`assign_from_catalog_dir_filtered` takes the router's registered keys and restricts candidates to them; the call site passes `provider_router.keys()`.

**The interesting case is when nothing in the catalog is registered: assignment leaves the DOT unchanged.** That is correct rather than a fallback — `strong` is itself a router key, which is exactly what `available: ["cheap", "strong"]` in the original error was reporting. Rewriting it was the bug; leaving it alone resolves directly and the node gets the model it asked for.

`registered = None` preserves the old unfiltered behaviour for callers with no router (tests, legacy paths), so this is additive.

## Tests

- `unregistered_catalog_models_are_never_assigned` — the root cause, using the exact shape from the report: catalog offers `qwen3-max` / `llama-3.3-70b-versatile`, router registers only `cheap`/`strong`, nothing may be assigned
- `registered_catalog_models_survive_the_filter` — a registered model **is** still assignable, so the filter does not disable assignment wholesale

## Verification

| gate | result |
| --- | --- |
| test functions removed vs `main` | **0** |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo test -p octos-pipeline` | 344 passed |

## Still open after this

**Layer 4** — `router.rs` reports the *rewritten* model (`qwen3-max`) rather than the original lane key (`strong`), so an error still does not say which node has the gap. Worth doing for diagnosability; not needed for correctness now that unresolvable models are not assigned.

#1902's fallback stays as a safety net — with this merged it should stop firing in normal operation, which makes its `warn!` meaningful again rather than constant.

### DIFF
--- a/crates/octos-pipeline/src/executor.rs
+++ b/crates/octos-pipeline/src/executor.rs
@@ -1795,7 +1795,19 @@ impl PipelineExecutor {
             .catalog_dir
             .as_deref()
             .unwrap_or(&self.config.working_dir);
-        crate::model_assignment::assign_from_catalog_dir(&mut graph, catalog_dir);
+        // Only assign models this profile's router actually registered (#1901).
+        // Without the router, assignment rewrote resolvable lane keys into
+        // catalog models that could not resolve — the crash #1902 papered over.
+        let registered: Option<std::collections::HashSet<String>> = self
+            .config
+            .provider_router
+            .as_ref()
+            .map(|r| r.keys().into_iter().collect());
+        crate::model_assignment::assign_from_catalog_dir_filtered(
+            &mut graph,
+            catalog_dir,
+            registered.as_ref(),
+        );
 
         // ── Pipeline start: log graph structure ──
         let node_summary: Vec<String> = graph
--- a/crates/octos-pipeline/src/model_assignment.rs
+++ b/crates/octos-pipeline/src/model_assignment.rs
@@ -133,13 +133,64 @@ impl ModelPools {
 /// touched (or `debug!` for the no-op cases) so operators can confirm
 /// the assignment fired without scraping per-node logs.
 pub fn assign_from_catalog_dir(graph: &mut PipelineGraph, data_dir: &Path) {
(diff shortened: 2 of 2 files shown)


ymote:
Closing. The branch `fix/assign-only-registered-models` is not deleted, so the work is recoverable.

For whenever this is picked up: the root cause is that `assign_from_catalog_dir` is never given the provider router, so it picks models blind from the unfiltered 93-entry catalog and cannot know which are registered. The fix here passes `provider_router.keys()` in and filters candidates to them — and when nothing in the catalog is registered it leaves the DOT lane keys alone, which is correct rather than a fallback, since `strong` is itself a router key.

#1901 stays open and tracks this.

PR #1904, opened 2026-08-04
fix(pipeline): decide models by what the ROUTER resolves, not what the catalog lists

### LINKED ISSUE #1901: deep_research silently downgrades to the default provider when model assignment picks an unregistered model

ymote:
Closes #1901 — the remaining layers. #1902 stopped the crash; this stops the crash from being *possible*.

## One defect, in two places

Neither model **assignment** nor pre-flight **validation** had access to the provider router, so both reasoned about models that may not be runnable. The catalog lists 93 models; a profile's router may register only `cheap` and `strong`.

```
DOT: model="strong"
  ↓ assignment rewrites it to a catalog model   → "qwen3-max"
  ↓ validation checks it against the catalog    → PASSES (it is in there)
  ↓ runtime asks the router to resolve it       → NOT REGISTERED
```

Three steps, each locally correct, none of them ever asking the router.

## What changes

**Layer 1 — assignment.** `assign_from_catalog_dir_filtered` restricts candidates to `router.keys()`. When nothing in the catalog is registered it leaves the DOT unchanged — **correct, not a fallback**: `strong` is itself a router key, which is exactly what `available: ["cheap", "strong"]` in the original error was reporting. Rewriting it was the bug.

**Layer 5 — the validator.** *(Not in the issue's four layers.)* `with_resolvable_models` narrows the known-model set to what the router resolves, so pre-flight certifies **resolvability** rather than presence in a catalog. Before, it certified graphs that could not run; after #1902 those graphs stopped crashing and silently used the default provider instead — which is worse. Validation said yes, the run looked healthy, and the requested model was never used.

**Layer 4 — the message.** The rejection no longer says *"choose a model present in `model_catalog.json`"*. That was the most misleading advice available, because the rejected model **is** in the catalog and merely unregistered — it sent the reader to the wrong file. It now names the router and lists what actually resolves.

## Fails open, never stricter than reality

Both new paths keep the old catalog behaviour when there is no router, or it has no keys. Validation must not start rejecting working graphs just because it cannot see what is registered.

## Tests

| test | pins |
| --- | --- |
| `unregistered_catalog_models_are_never_assigned` | the root cause (layer 1) |
| `registered_catalog_models_survive_the_filter` | the filter does not disable assignment wholesale |
| `catalogued_but_unregistered_models_are_rejected` | the validator certifies resolvability (layer 5) |
| `an_absent_router_leaves_the_catalog_set_intact` | no router ⇒ no new strictness |
| `rejection_names_the_router_not_the_catalog` | the hint stops pointing at the wrong file (layer 4) |

## Verification

| gate | result |
| --- | --- |
| test functions removed vs `main` | **0** |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo test -p octos-pipeline` | 20 suites, all green |

## Relationship to #1902

#1902's runtime fallback **stays** as a safety net. With this merged it should stop firing in normal operation, which makes its `warn!` meaningful again rather than a constant for affected profiles.

This supersedes the closed #1903, which covered layer 1 only.

### DIFF
--- a/crates/octos-pipeline/src/executor.rs
+++ b/crates/octos-pipeline/src/executor.rs
@@ -1795,7 +1795,19 @@ impl PipelineExecutor {
             .catalog_dir
             .as_deref()
             .unwrap_or(&self.config.working_dir);
-        crate::model_assignment::assign_from_catalog_dir(&mut graph, catalog_dir);
+        // Only assign models this profile's router actually registered (#1901).
+        // Without the router, assignment rewrote resolvable lane keys into
+        // catalog models that could not resolve — the crash #1902 papered over.
+        let registered: Option<std::collections::HashSet<String>> = self
+            .config
+            .provider_router
+            .as_ref()
+            .map(|r| r.keys().into_iter().collect());
+        crate::model_assignment::assign_from_catalog_dir_filtered(
+            &mut graph,
+            catalog_dir,
+            registered.as_ref(),
+        );
 
         // ── Pipeline start: log graph structure ──
         let node_summary: Vec<String> = graph
@@ -2445,6 +2457,16 @@ impl PipelineExecutor {
             .with_known_models(crate::model_assignment::known_model_keys_from_catalog_dir(
                 catalog_dir,
             ))
+            // Narrow to what the router can actually RESOLVE (#1901 layer 5).
+            // The catalog lists models this profile never registered, so
+            // validating against it alone certified graphs that cannot run.
+            .with_resolvable_models(
+                self.config
+                    .provider_router
+                    .as_ref()
+                    .map(|r| r.keys())
+                    .unwrap_or_default(),
+            )
             .with_known_tools(tool_names)
     }
 
--- a/crates/octos-pipeline/src/model_assignment.rs
+++ b/crates/octos-pipeline/src/model_assignment.rs
@@ -133,13 +133,64 @@ impl ModelPools {
 /// touched (or `debug!` for the no-op cases) so operators can confirm
 /// the assignment fired without scraping per-node logs.
 pub fn assign_from_catalog_dir(graph: &mut PipelineGraph, data_dir: &Path) {
-    let Some(catalog) = load_catalog(data_dir) else {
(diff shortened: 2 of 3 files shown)


ymote:
Closing — adversarial review (codex, read-only) found two HIGH defects, both verified against the code. This would break currently-working pipelines.

**1. Rule 18 would reject models the router genuinely resolves, and makes #1902's fallback unreachable.**

`ProviderRouter::resolve` (router.rs:118+) does exact lookup → `eq_ignore_ascii_case` → **prefix split** (`cheap/gpt-4o-mini` → lookup `cheap`) → `ends_with` suffix matching. `model_known` in validate.rs implements none of that, so replacing `known_models` with raw router keys rejects `cheap/gpt-4o-mini`, which resolves fine at runtime.

Worse in principle: #1902 deliberately made runtime TOLERATE unresolvable keys by degrading to the default provider. Making validation REJECT them makes that fallback unreachable. The two contracts contradict each other, and intersecting instead of replacing would not fix it — validation would have to model router resolution *plus* the fallback policy.

**2. Preflight and execution would validate against different authorities.**

`tool.rs:833` uses catalog keys only and never applies `self.provider_router`, while `executor.rs` would replace that set with router keys. A graph could pass synchronous preflight, receive the spawn-only "started" acknowledgement, then fail validation in the background — exactly the foreground/background repair loop `pre_flight_validate` exists to prevent.

**Also (medium), against the assignment half:**

- The filter is case-sensitive exact while the router is not, so it drops models that resolve today (registered `model-x` vs catalog `Model-X`) and returns early.
- `model_key()` strips the provider prefix, so one registered bare key retains every catalog row sharing that suffix. Real example in the checked-in catalog: the same stripped key is `fast` under nvidia and `strong` under OpenRouter, so one provider enters BOTH pools and defeats strong/fast routing.

The seven tests here only exercised exact key comparison — they never covered prefix/case/suffix resolution, fallback acceptance, catalog collisions, or tool/executor parity, which is why they were all green.

Branch `fix/model-resolvability-end-to-end` is preserved. #1901 stays open; these findings are recorded there.

PR #1907, opened 2026-08-04
fix(pipeline): isolate per-run working dir so findings don't mix across runs

ymote:
## Problem

`deep_research`'s synthesize node timed out twice at 1800s. Root cause: the pipeline working directory accumulates `findings-*.md` files across **unrelated** pipeline runs.

The analyze/synthesize convergence nodes follow a "read every findings-*.md" deliverable contract. With all runs sharing one flat profile data dir, files from research A/B, judge probes, and ad-hoc tests piled up — the node tried to read all of them and timed out.

**Direct evidence for the mechanism** (answering the review's premise question): at the time of the timeouts the shared profile data dir held **27 `findings-*.md` files** spanning multiple distinct runs (research A, research B, two judge-connectivity probes, an ASTC probe). The analyze node's prompt instructs reading *each* findings file, so the convergence node was observed reading cross-run material. The fix is not speculative hygiene — it removes the exact accumulation that was present.

## Root cause

`RunPipelineTool` gives every node worker (search fan-out, analyze, synthesize) the same `effective_working_dir` (the profile data dir). There is **no per-run isolation**, so findings from different research runs mix in one flat directory.

## Fix

Scope each run to `<working_dir>/pipeline-runs/<run_id>/` so a run's workers only ever see their own findings.

- `run_id` generation is **moved before** `ExecutorConfig` construction (previously after `run_graph()`) and **reused** for the audit marker, so the run dir and audit trail share one identity (removes the duplicate second generation).
- `mkdir` failure degrades gracefully back to the shared dir — isolation is a hygiene improvement, not a correctness requirement.
- **Additive-only:** nothing is deleted from the shared dir; final report delivery is unaffected (callers read the report back from the same executor run).

## Review feedback addressed (commit eee625a)

1. **Discoverability** — added `pipeline-runs/latest` symlink (Unix) / pointer file (other platforms), repointed each run, so follow-up `read_file` and human browsing still find findings at a stable path while keeping isolation.
2. **Unbounded growth** — `prune_old_run_dirs` retains only the 20 most recent run dirs (lexicographic ≈ recency, since run ids embed ns timestamps); never touches the `latest` entry or non-run files. Best-effort.
3. **Dedupe** — `resolve_pipeline_working_dir` bound once as `shared_working_dir` and reused on the mkdir-failure fallback (no double call).
4. **Tests** — `per_run_working_dirs_differ_for_same_graph` pins isolation; `prune_old_run_dirs_retains_most_recent_and_skips_latest` covers pruning.
5. **Premise** — direct evidence added to this body (above).

## Testing

- `cargo check -p octos-pipeline` ✅ (no warnings)
- `cargo test -p octos-pipeline --lib` ✅ **344 passed, 0 failed** (incl. 2 new tests)

## Follow-ups (out of scope)

- Surfacing/copying final report files to the flat working dir for easier discovery (the `latest` link addresses most of this).

### DIFF
--- a/crates/octos-pipeline/src/tool.rs
+++ b/crates/octos-pipeline/src/tool.rs
@@ -1232,9 +1232,15 @@ impl Tool for RunPipelineTool {
         // synthesize a payload so `files_to_send` is non-empty). Keep the
         // synthetic report under `working_dir` so the spawn_only send_file path
         // can deliver it; system temp is outside that allowlist.
-        let synthetic_dir = effective_working_dir
-            .join("skill-output")
-            .join("run_pipeline");
+        // SHARED dir, deliberately NOT the per-run dir. Per-run isolation exists
+        // to stop node workers reading each other's `findings-*.md`; the final
+        // REPORT is the opposite case — it must stay at a stable, predictable
+        // path. `send_file` only delivers paths under the tool working dir, so
+        // building this from the run dir put the deliverable outside the
+        // allowlist and broke delivery outright (caught by
+        // `text_only_ir_pipeline_synthesizes_report_under_working_dir`), and it
+        // also made the report unfindable for any later turn.
+        let synthetic_dir = shared_working_dir.join("skill-output").join("run_pipeline");
         let delivery = resolve_report_delivery(
             &result.output,
             &result.files_modified,
--- a/crates/octos-pipeline/src/tool.rs
+++ b/crates/octos-pipeline/src/tool.rs
@@ -2223,8 +2223,14 @@ mod tests {
         assert_eq!(dir_count, 20, "must retain exactly the 20 most recent dirs");
         assert!(runs_root.join("run-024").is_dir(), "newest dir retained");
         assert!(!runs_root.join("run-000").exists(), "oldest dir pruned");
-        assert!(runs_root.join("latest").exists(), "latest entry never pruned");
-        assert!(runs_root.join("notes.txt").exists(), "non-run file never pruned");
+        assert!(
+            runs_root.join("latest").exists(),
+            "latest entry never pruned"
+        );
+        assert!(
+            runs_root.join("notes.txt").exists(),
+            "non-run file never pruned"
+        );
     }
 
     /// Review HIGH: retention must prune by AGE, not by pipeline NAME. Run ids
--- a/crates/octos-pipeline/src/tool.rs
+++ b/crates/octos-pipeline/src/tool.rs
@@ -2201,9 +2201,19 @@ mod tests {
 
         let root = TempDir::new().unwrap();
         let runs_root = root.path().join("pipeline-runs");
-        // Create 25 run dirs (monotonic names so lexicographic == recency).
+        // Create 25 run dirs with EXPLICIT, monotonic mtimes.
+        //
+        // Retention sorts by mtime, not by name, so creating these in a tight
+        // loop and trusting the filesystem clock makes the test environment-
+        // dependent: on a coarse-granularity filesystem all 25 land on the same
+        // mtime, `sort_by_key` (stable) then falls back to `read_dir` order —
+        // which is arbitrary on Linux — and an arbitrary 5 get pruned. That is
+        // exactly how this test passed on APFS (nanosecond mtimes) and failed
+        // in CI. Same reason the age-vs-name test below sets mtimes by hand.
         for i in 0..25 {
-            std::fs::create_dir_all(runs_root.join(format!("run-{i:03}"))).unwrap();
+            let dir = runs_root.join(format!("run-{i:03}"));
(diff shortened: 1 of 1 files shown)


ymote (commented):
Reviewed the diff and the surrounding call sites.

## Verified sound — the biggest structural risk is already avoided

`catalog_dir: Some(self.working_dir.clone())` (tool.rs:1074) uses the **profile** dir, not `effective_working_dir`, so moving the run dir does not break model assignment or cost projection. That is exactly the class of silent degradation #1901 is about, and this sidesteps it — worth calling out because it is the non-obvious part that could have gone wrong.

The tool also returns report **content** inline in `ToolResult.output`, so the claim that delivery is unaffected for the invoking turn holds.

## Two deferred items I would decide BEFORE merge, not after

**1. Discoverability (the PR's own open question).**
`findings-*.md` and any written report now live in `pipeline-runs/<run_id>/`. The agent's file tools operate relative to the session workspace, so a follow-up `read_file findings-3.md` — or a human browsing the directory — silently stops finding them. The immediate turn is fine; everything after it is not.

The moment this merges, the existing flat-dir habit breaks with no error. A `pipeline-runs/latest` symlink repointed each run keeps both properties (isolation AND a stable path) for a few lines.

**2. Unbounded growth.**
Every run creates a directory that nothing removes. For a frequently-run pipeline this accumulates indefinitely. Bounding it now (retain the N most recent) is cheap; doing it after someone has thousands of run dirs is a migration.

Both are listed as follow-ups, which is honest — my argument is only that they are cheaper now than later, because both become user-visible the moment this lands.

## Smaller

**3. `resolve_pipeline_working_dir` is called twice on the fallback path** — once at the top, once again in the `else` branch. It re-runs session-scope resolution including a possible mkdir. Binding the original once and reusing it is clearer and avoids repeating that work.

**4. No test.** +37/-1 with nothing pinning the isolation property. A test asserting that two runs of the same graph get **different** working dirs would lock in the actual fix — and would catch a regression where `run_id` reuse accidentally made them collide, which is precisely the thing this change now depends on.

## One question about the premise

The diagnosis is that synthesize hit 1800s because it read findings from unrelated runs. A 1800s timeout has more than one possible cause, and this fix is cheap enough to be tempting to accept without proving the mechanism. Was there direct evidence — cross-run `findings-*.md` actually present in the flat dir at the time, or a synthesize node observed reading them? If so, worth putting in the PR body; if not, the fix is still good hygiene but the timeout may recur.

ymote (commented):
Re-reviewed. All four points are addressed, and the careful bits are genuinely careful — `latest` excluded from pruning by name, non-directory entries left alone, Windows fallback to a text pointer, and the double `resolve_pipeline_working_dir` collapsed into `shared_working_dir`.

One new defect, though, introduced by the retention code.

## HIGH — retention prunes by PIPELINE NAME, not by age

`generate_run_id` (tool.rs:1621) builds:

```rust
format!("{safe_graph}-{secs}-{nanos:09}-{pid}-{counter}")
```

The **graph name comes first**. `prune_old_run_dirs` then does:

```rust
run_dirs.sort();
let to_remove = run_dirs.len() - keep;
for old in run_dirs.into_iter().take(to_remove) { remove_dir_all(&old) }
```

A plain path sort is therefore ordered by pipeline name first and timestamp only within a name. Retention is not by recency.

**Concrete failure.** Two pipelines, `analyze` and `deep_research`, 25 run dirs total:

```
analyze-...        x5   (today, newest)
deep_research-...  x20  (yesterday)
```

Sorted: all five `analyze-*` sort first. `to_remove = 5` takes the first five — so **the five NEWEST runs are deleted and twenty older ones are kept**. The `keep = 20 most recent` contract is inverted for the alphabetically-earlier pipeline.

A rarely-run pipeline also loses its history entirely whenever a chattier pipeline whose name sorts later fills the quota.

Knock-on: `pipeline-runs/latest` can be left **dangling**, because its target is prunable by name while the symlink entry itself is skipped.

## Why the new test does not catch it

`prune_old_run_dirs_retains_most_recent_and_skips_latest` states the assumption in its own comment:

```rust
// Create 25 run dirs (monotonic names so lexicographic == recency).
std::fs::create_dir_all(runs_root.join(format!("run-{i:03}")))
```

Single synthetic prefix, so lexicographic order equals recency — exactly the property real run_ids do not have. The test passes while production prunes the wrong dirs.

## Suggested fix

Sort by directory mtime rather than path:

```rust
run_dirs.sort_by_key(|p| {
    std::fs::metadata(p).and_then(|m| m.modified()).unwrap_or(std::time::UNIX_EPOCH)
});
```

Parsing `{secs}-{nanos}` back out of the name also works but re-couples pruning to the id format; mtime does not. Either way the test needs run dirs from **two different graph names** with interleaved timestamps — that is the case that fails today.

Everything else from the previous round looks good to me.

PR #1914, opened 2026-08-04
acp: let embedders build the same agent factory the CLI serves

kevinaboos:
hosts that can't spawn `octos acp` as a child process (like iOS, which prohibits `exec()`) currently have to reimplement the agent assembly by hand against octos-agent/core/llm/memory. that reimplementation is a subset that drifts: no provider fallback routing, no auth store, no keychain: markers, no MCP, no plugins or skills, and its own copy of the config precedence rules.

everything needed is already there, just not reachable: commands::acp was a private module and SessionAgentFactory a private trait, so the only way in was the stdio path.

split AcpCommand::run_async in two. factory() does everything that happens before stdio is touched — context and config resolution, provider/model selection — and hands back the Arc<dyn SessionAgentFactory> the stdio path drives. run_async now calls it, so there's one resolution path rather than two to keep in step.

also gave AcpCommand a Default carrying the CLI's real defaults, so struct-update syntax works for embedders. hand-written, not derived: deriving would make max_iterations 0 and quietly produce an agent that can't call a single tool, so the clap default and the Default impl now share DEFAULT_MAX_ITERATIONS.

no behaviour change to `octos acp` itself — same resolution, same order, same errors. the integration test links the crate as an embedder does, so it fails if any of this stops being public.

### DIFF
--- a/crates/octos-cli/src/commands/acp.rs
+++ b/crates/octos-cli/src/commands/acp.rs
@@ -86,6 +86,11 @@ use octos_core::SessionKey;
 use super::Executable;
 use crate::config::Config;
 
+/// Default for [`AcpCommand::max_iterations`]. Shared by the clap default and
+/// the `Default` impl so an embedder building the command by hand gets the same
+/// budget the CLI does.
+pub const DEFAULT_MAX_ITERATIONS: u32 = 20;
+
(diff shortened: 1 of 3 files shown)


anantheparty (approved):
审查通过（2026-08-15 巡检）。

PR #1917, opened 2026-08-05
fix(ui-protocol): sequence projected voice attachments

alan0x:
## Summary
- assign deterministic unique v2 sequence numbers to consecutive projected voice attachments
- count projection-only terminal and attachment rows after the latest durable envelope
- preserve replay determinism by limiting sequence calculation to rows before the source cursor

## Root cause
Consecutive legacy `file/attached` sources could be projected with the same `(thread_id, seq)`. Canonical clients deduplicated the later TTS sentence attachments, so playback stopped after the first sentence.

Frontend companion: https://github.com/octos-org/octos-web/pull/306

## Tests
- `cargo test -p octos-cli --features api --lib should_assign_unique_v2_seq_to_terminal_and_consecutive_attachments`
- `cargo check -p octos-cli --features api`
- `cargo clippy -p octos-cli --features api --lib -- -D warnings`

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol_ledger.rs
+++ b/crates/octos-cli/src/api/ui_protocol_ledger.rs
@@ -1394,11 +1394,15 @@ impl UiProtocolLedger {
             .max(1) as u64
     }
 
-    /// Next per-thread v1 sequence, used when a legacy terminal or attachment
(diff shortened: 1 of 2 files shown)


anantheparty (approved):
审查通过（2026-08-15 巡检）。

PR #1919, opened 2026-08-05
fix(peer): stop priming the caller to end its turn after a handoff (#1918)

### LINKED ISSUE #1918: Master agent stalls after handing off to a peer agent instead of continuing its own work
## Bug

When the master agent hands a task off to a peer agent (via `peer_handoff` or `spawn`), it **stops doing its own work and only resumes when the user explicitly prompts it again** — instead of continuing its in-flight task in parallel, as the delegation model intends.

## Expected behavior (per codex reference)

Codex's delegation model (`codex-rs/core/src/agent/role.rs:328`) is explicit:

> "While waiting for the explorer results, **you can continue working on other local tasks** that do not depend on those results. This parallelism is a key advantage of delegation."

The master agent should keep working on its own line of work after spawning/handing off a peer, and collect peer results later.

## Actual behavior

Observed repeatedly in this session:

1. User asks the master agent to do task X (e.g. fix #476).
2. Master agent starts working on X (exploring, running tools).
3. A side question comes up → master agent hands it to a peer agent.
4. **Master agent then stalls** — it does NOT resume task X. It only continues when the user explicitly asks again ("476如何了" → only then does it continue).

The master treats "handed to a peer" as "my work is done", even when the peer was only given a **side/auxiliary** query while the master still had an unfinished primary task.

## Root cause (behavioral, not a code crash)

This is an agent-behavior / prompt-design defect, not a thrown error:

- `peer_handoff`'s own description ("Promote work OUT of this conversation ... You will NOT receive the result in this turn") primes the master to treat a handoff as transferring ownership of its work — so it concludes there is nothing left to do and ends its turn.
- The tool does not distinguish between (a) handing off the **primary** task (where stopping is acceptable) and (b) handing off an **auxiliary / side** task (where the master should continue its primary work in parallel).
- Nothing in the turn lifecycle forces the master to continue after a peer handoff; ending the turn is a valid (but wrong) model choice here.

## Why this matters

It silently serializes work that should be parallel. The user's primary task stalls every time a peer is spawned, and only resumes on an explicit user nudge — defeating the point of background delegation.

## Suggested fix directions

1. **System prompt / role guidance**: clarify that when the master spawns a peer for an **auxiliary** task (a side query, an independent sub-question), it MUST continue its primary task rather than ending the turn. Distinguish "delegated the main task" (may stop) from "delegated a side task" (must continue).
2. **Tool description**: make `peer_handoff`/`spawn` descriptions not imply "your work ends here" — e.g. note that the caller should keep working if it has independent remaining work.
3. **(Optional, structural)**: consider a mechanism that nudges the master to continue after a peer handoff when its turn still has pending local work (similar to how codex frames delegation as parallelizable).

## Related

- codex's non-blocking delegation model (the intended semantics).
- The 50-iteration budget notice is a *separate* intended runaway guard and not the cause here — the stall happens well before any budget cap.

ymote:
Fixes #1918.

Every signal the model saw after `peer_handoff` pointed at "your work is finished", and nothing pointed back:

| surface | text |
|---|---|
| description | "**Promote work OUT** of this conversation" |
| description | "You will **NOT receive the result in this turn**" |
| result | "**Do not wait for it.**" |

So the master treated a handoff as transferring ownership of its work and ended the turn, resuming only on a user nudge — serializing exactly the work delegation exists to parallelize.

**"Do not wait" answers whether to BLOCK. It never answered whether to CONTINUE.**

## Changes

**The result text** carries most of the weight — it is in context at the moment the model chooses between another tool call and ending the turn, whereas the description is read once when the tool list is built. It now ends:

> Do not wait for it — and do not stop here: if you still have work of your own, continue it now in this same turn. **Only finish if the peer took over everything that was left.**

That last sentence is load-bearing. Without it "keep working" reads as "never finish", and handing off the *primary* task is a case where stopping is right.

**The description** reframes the opening from "Promote work OUT of this conversation" to "Hand a SELF-CONTAINED piece of work … and keep working yourself", and states plainly that a handoff does not end the turn.

## What I deliberately did not change

- **The handoff-vs-spawn axis.** #1918 suggests distinguishing primary vs auxiliary tasks, but the description already splits on *does THIS turn need the result* — and the model was applying that correctly. A side question genuinely does not feed the turn's reasoning, so `peer_handoff` was the right tool. The defect was inferring "nothing to wait for" as "nothing to do". Adding a second axis would collide with the existing one.
- **`spawn`.** #1918 mentions it, but spawn returns into the turn by design; a stall there would be a functional defect, not a framing one, and wants its own issue with a transcript.
- **A structural nudge** (#1918 direction 3). It needs a notion of "pending local work" that the agent does not have — the closest is the staged-submit queue, which holds user prompts, not the model's own unfinished intent.

## Scope of the test

The new test pins the **wording**, not the behaviour. The bug is an LLM judgement and can only be confirmed live: hand a side task to a peer mid-task and watch whether the master resumes on its own. I have not verified that.

There is no central role prompt in `octos-agent` to carry this guidance, so these two surfaces are the only levers available.

2521 octos-agent lib tests pass; `cargo fmt --all --check` and `scripts/lint-tool-descriptions.py` clean (21 tools / 14 manifests).

### DIFF
--- a/crates/octos-agent/src/tools/peer_handoff.rs
+++ b/crates/octos-agent/src/tools/peer_handoff.rs
@@ -119,16 +119,20 @@ impl Tool for PeerHandoffTool {
     }
 
     fn description(&self) -> &str {
-        "Promote work OUT of this conversation into a sovereign peer session with \
-         its own durable brief, workspace, and lifecycle. Give the peer a short, \
-         unique NAME — it is the peer's primary address (\"let Edison do X\"); you \
-         reach it later by name with peer_send_input / peer_close / peer_gather. \
-         Use when the work outlives this turn, needs its own workspace or safety \
-         envelope, or the user may steer it separately. You will NOT receive the \
-         result in this turn — the peer reports to the user's session strip and the \
-         shared blackboard. For work whose result THIS turn needs to continue \
-         reasoning, use spawn instead. The brief is a complete task contract: \
-         include all context the peer needs (it cannot see this conversation)."
+        "Hand a SELF-CONTAINED piece of work to a sovereign peer session with its \
+         own durable brief, workspace, and lifecycle, and keep working yourself. \
+         Give the peer a short, unique NAME — it is the peer's primary address \
+         (\"let Edison do X\"); you reach it later by name with peer_send_input / \
+         peer_close / peer_gather. Use when the work outlives this turn, needs its \
+         own workspace or safety envelope, or the user may steer it separately. \
+         You will NOT receive the result in this turn — the peer reports to the \
+         user's session strip and the shared blackboard. Handing work off does NOT \
+         end your turn: delegating is what lets you make progress on your own \
+         remaining work in parallel, so continue it immediately unless the peer \
+         took over everything you had left. For work whose result THIS turn needs \
+         to continue reasoning, use spawn instead. The brief is a complete task \
+         contract: include all context the peer needs (it cannot see this \
+         conversation)."
     }
 
     fn tags(&self) -> &[&str] {
@@ -227,7 +231,10 @@ impl Tool for PeerHandoffTool {
                     "Staged peer '{name}' (slug {slug}, brief at {brief_path}, cwd {cwd}). \
                      Address it later by name with peer_send_input / peer_close. The user's \
                      client opens it in the background; its result lands on the blackboard \
-                     at peers/{slug}/result.md. Do not wait for it.",
+                     at peers/{slug}/result.md. Do not wait for it — and do not \
+                     stop here: if you still have work of your own, continue it \
+                     now in this same turn. Only finish if the peer took over \
+                     everything that was left.",
                     slug = staged.slug,
                     brief_path = staged.brief_path,
                     cwd = staged.cwd,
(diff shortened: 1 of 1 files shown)


ymote:
## Review @ 30446070

**Verdict: approve.**

The root-cause analysis is exactly right and worth preserving in the test comment where it now lives: every surface the model saw after `peer_handoff` primed turn-end ("promote work OUT", "you will NOT receive the result", "Do not wait for it") and nothing pointed back — **"do not wait" answers whether to BLOCK, never whether to CONTINUE.** That asymmetry precisely explains the #1918 stall-until-nudge behavior, which serialized the parallelism the tool exists to create.

What makes this a good fix rather than a prompt tweak:
- **Both surfaces** are fixed — the description (which primed the framing before the call) and the result message (in context at the exact moment the model chooses between another tool call and ending the turn). The rationale for why both is written into the test.
- **The inverse failure is guarded**: "Only finish if the peer took over everything that was left" names the one legitimate turn-end case, so "keep working" can't be read as "never finish" — and the test asserts on that clause specifically, not just the happy string.
- The relaxed existing assertion (`"Do not wait for it"` without the trailing period) is the correct minimal change since the sentence now continues.

One caveat, stated plainly: string tests pin the *surfaces*, not the *behavior* — whether a given model actually continues its turn is only provable by live soak. Worth one deliberate multi-peer session on a real box before calling #1918 closed (the peer-agents work has a good track record of unit-green/live-broken findings).

CI is fully green and has been since Aug 5. Ship it.

*Posted by Claude Code on ymote's behalf.*

ymote:
## Codex adversarial round @ 30446070 — approve stands

- B1 (both surfaces fixed, inverse-failure clause guarded) — **CONFIRMED** (peer_handoff.rs:121,230); ordinary tool results return to the model loop, no special handoff termination exists anywhere else.
- B2 (string tests pin surfaces, not behavior) — **CONFIRMED**; live-model soak still wanted before declaring #1918 closed.
- **Surface audit came back clean**: no tracked SKILL.md mentions handoff; `peer/staged` is client-facing only; `peer_gather` already tells the caller to continue other work (peer_gather.rs:68); `peer_fleet_synthesis` fires on a later gather turn, so it can't prime an immediate stop.
- One LOW (docs, non-blocking): `book/src/multi-agent.md:30` + `book-zh/src/multi-agent.md:30` still carry the old "work out of the conversation / fire-and-forget" framing — human docs, not model-injected; worth a follow-up sweep.

Issue #1906, opened 2026-08-04
skill-action jobs: every list/read/invoke re-runs enable_persistence → full ledger re-append per call (unbounded growth)


BH3GEI:
Follow-up from the #1637 review.

## Problem

`TaskSupervisor::enable_persistence` (`crates/octos-agent/src/task_supervisor.rs:1310`) is designed as a **startup-only** operation — its doc comment states "at this point in startup no new work has been scheduled yet". Each call:

1. re-reads the entire JSONL ledger (`load_persisted_tasks`),
2. **unconditionally re-appends a snapshot for every known task** (`for task in snapshots { self.persist_snapshot(&task); }`, ~task_supervisor.rs:1346),
3. runs the orphan sweep.

But the skill-action job paths call it on **every request**, not once at startup:

- `raw_skill_action_invoke` — `crates/octos-cli/src/api/ui_protocol.rs:11738` (shared session-runtime supervisor, every background invoke)
- `load_skill_action_job_view` — `crates/octos-cli/src/api/ui_protocol.rs:11820` (shared supervisor, every job view load)
- `load_skill_action_jobs` — `crates/octos-cli/src/api/skill_action_jobs.rs:160` builds a **fresh** `TaskSupervisor` + `enable_persistence` per call (list path / runtime-miss fallback)

Since the job list is polled by clients (octos-web polls while jobs are active), each poll appends N task snapshots to `tasks.jsonl`. The next call re-reads all of it and appends N more — the ledger grows ~O(polls × tasks) and each load costs that much to replay, i.e. quadratic growth in both file size and latency over a session's lifetime, for zero semantic gain.

## What is NOT a problem (checked while scoping)

- The orphan sweep is liveness-gated (`is_task_live` + live-tool_call_id proxy exemption, the codex round-2 gate), so repeat calls on a live supervisor do not falsely reap in-process tasks — it's wasted work, not a correctness bug.
- `install_skill_action_job_projection_listener` (ui_protocol.rs:11329) installs under a stable key (`skill-action-job:{profile}:{session}`), so repeat installs replace rather than accumulate.

## Suggested direction

Make repeat calls cheap instead of changing every caller:

- Short-circuit in `enable_persistence` when `self.persistence_path` already equals the requested path — all mutations since the first enable are already persisted through the normal change path, so a same-path replay/resnapshot/sweep adds nothing. Different-path calls keep the current behavior.
[...]

Issue #436, opened 2026-04-17
Add turn-end validator runner and completion gating to the harness runtime


ymote:
## Problem

Octos has browser/live acceptance suites and some contract checks, but no generic validator runner that the runtime can use to gate completion.

Result:
- some flows rely on prompt obedience
- some validators run only in tests or bespoke code paths
- a turn can finalize without required checks having passed

## Goal

Build a generic validator runner that executes deterministic checks and can block finalization when required validators fail.

## Validator tiers

1. Fast deterministic checks
   - file exists
   - manifest valid
   - path placement
   - build exit code
2. Medium checks
   - preview render
   - DOM smoke
3. Expensive checks
   - Playwright
   - screenshot diff
   - route checks

## Scope

- tiered execution policy (`on_turn_end`, `on_source_change`, `on_completion`)
- structured validator results
- completion gating on required failures
- reusable interface for slides, sites, and later coding harnesses

## Likely files

- `crates/octos-agent/src/workspace_git.rs`
- `crates/octos-cli/src/workspace_contract.rs`
- `crates/octos-cli/src/session_actor.rs`
- policy parser surfaces from policy v1

## Acceptance

- validator results are persisted in structured form
- failed required validators block finalization
- user-facing failures are specific enough to act on
- operator tooling can surface validator outcomes later without inventing a second truth path

ymote:
Closing as completed on shipped `main` @ `8ea539d09ef3e08f7fb622d85b374197a5c785a2`.

The harness validator runner is now part of the runtime-owned workspace contract path:
- turn-end and completion validators execute in `crates/octos-agent/src/workspace_git.rs`
- structured validation failures are surfaced through `crates/octos-cli/src/session_actor.rs`
- workspace contract inspection/reporting uses the same truth path

Acceptance for this issue is now met.

PR #1921, opened 2026-08-06
fix(agent): reap in-flight orphan background tasks via heartbeat timeout

### LINKED ISSUE #1920: In-flight orphan background tasks are never reaped (need heartbeat-based reaper)
## Gap

`TaskSupervisor::enable_persistence` (crates/octos-agent/src/task_supervisor.rs:1295-1302) reaps **startup-time** orphan background tasks on restart: any task whose `runtime_state` is non-terminal after replaying the JSONL ledger is marked `Failed("orphaned across restart")`, because at startup no live worker can own it.

But **in-flight orphans inside a long-running supervisor are NOT addressed**: a background worker that hangs or crashes **silently** while the supervisor process itself stays alive keeps its task in a non-terminal state forever. No actor ever drives it to `Completed`/`Failed`/`Cancelled`, so:

- the task shows as perpetually "running" in `check_background_tasks` / task lists,
- its resources (LLM budget slot, worktree, peer session) are never reclaimed,
- the user can never tell it died vs. is legitimately still working.

The code itself names this follow-up:

> "In-flight orphans inside a long-running supervisor (worker hangs / crashes silently while the supervisor itself stays alive) are NOT addressed here — that needs a **heartbeat-based reaper**, which is a follow-up if observed."

## Why this matters

The orphan sweep only fires at supervisor startup. For a long-lived gateway (the common case — days between restarts), any mid-run silent worker death leaves a zombie task indefinitely. This is the background-task analogue of a crashed process that never gets reaped.

## Suggested direction

Add a heartbeat-based reaper:

1. Workers (spawned sub-agents / peer sessions / pipeline nodes) emit periodic heartbeats (or the supervisor records last-progress timestamps).
2. A supervisor reaper task periodically scans for non-terminal tasks whose last heartbeat/progress is older than a TTL (with per-task-type budgets — e.g. a long deep_research node legitimately runs minutes without a state change, so TTL must be generous or progress-based rather than wall-clock).
3. On expiry, drive the task to `Failed("orphaned: heartbeat timeout")` via the standard `mark_failed` path so the ledger gets a proper terminal entry, reclaim its resources, and surface the failure to the user (rather than leaving it silently stuck).

## Notes / open questions

- Heartbeat vs. progress-vs-wall-clock timeout: a long LLM call or pipeline node can be silent for many minutes while legitimately working, so a naive wall-clock timeout would kill healthy work. Progress signals (tokens streamed, iterations advanced, tool calls made) are safer than a bare timer.
- Should be opt-in/configurable per task type — different work has very different legitimate silence windows.
- Related: the startup-time sweep already handles the restart case; this is strictly about the long-running supervisor case.

## Related code

- `crates/octos-agent/src/task_supervisor.rs` — `enable_persistence` (startup sweep), `mark_failed`, the `octos_orphaned_tasks_reaped_total` counter.

ymote:
Fixes #1920.

An in-flight background task whose worker dies without a terminal transition was never reaped — it sat `Running` forever, holding its slot and pinning progress indicators with nothing left to move them.

This moves orphan detection into `TaskSupervisor` behind a heartbeat timeout: a task with no progress for longer than `stuck_timeout` is marked failed with `orphaned: no progress for {timeout} (heartbeat timeout)`, and its cancel token is flipped **after** the terminal transition so a cooperative-but-stuck worker wakes at its next safe point and exits rather than racing the state change.

`session_actor.rs` sheds 79 lines net — the ad-hoc handling there is replaced by the supervisor owning the lifecycle, so there is one reaper rather than a per-callsite guess.

## Verification

| | |
|---|---|
| base | one commit on `725451747` (v2.0.3-rc.1), current `origin/main` |
| `cargo test -p octos-agent --lib` | **2522 passed**, 0 failed |
| supervisor tests | 107 passed |
| reaper tests | 10 passed |
| `cargo clippy -p octos-agent --all-targets -- -D warnings` | clean |
| `cargo fmt --all --check` | clean |

Tests (+306) slightly outweigh implementation (+300), and the change is net-negative outside the supervisor.

### DIFF
--- a/crates/octos-agent/src/task_supervisor.rs
+++ b/crates/octos-agent/src/task_supervisor.rs
@@ -339,6 +339,11 @@ pub struct BackgroundTask {
     /// AppUI `runtime_policy_stamp` schema.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub runtime_policy_stamp: Option<Value>,
+    /// Opaque host projection metadata. The supervisor persists this value
+    /// without interpreting its schema, allowing API adapters to derive
+    /// domain-specific views from the canonical task lifecycle.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub projection_metadata: Option<Value>,
 }
 
 impl BackgroundTask {
@@ -365,7 +370,7 @@ impl BackgroundTask {
 }
 
 /// Callback invoked when a task's status changes.
-type OnChangeCallback = Box<dyn Fn(&BackgroundTask) + Send + Sync>;
+type OnChangeCallback = Arc<dyn Fn(&BackgroundTask) + Send + Sync>;
 
 /// Payload emitted when a `spawn_only` background task transitions to
 /// `Failed`. Consumers (e.g. the session actor) use this to schedule a
@@ -891,6 +896,8 @@ pub struct TaskSupervisor {
     /// loop cannot keep adding children.
     poisoned_parents: Arc<Mutex<HashSet<String>>>,
     on_change: Arc<Mutex<Option<OnChangeCallback>>>,
+    /// Named observers supplement (and never replace) the primary callback.
+    on_change_listeners: Arc<Mutex<HashMap<String, OnChangeCallback>>>,
     on_failure: Arc<Mutex<Option<OnFailureCallback>>>,
     /// Gap-1 unification: single terminal-transition sink. Fired from
     /// every terminal path alongside the legacy `on_change` / `on_failure`
@@ -939,14 +946,14 @@ pub struct TaskSupervisor {
     /// Interval between heartbeat-reaper sweeps — see
     /// [`TaskSupervisor::start_reaper`]. Defaults to
     /// [`DEFAULT_REAP_INTERVAL`].
-    reap_interval: Arc<std::sync::Mutex<Duration>>,
+    reap_interval: Arc<Mutex<Duration>>,
     /// Silence window after which an ACTIVE task with a LIVE worker is
     /// considered stuck — see [`TaskSupervisor::start_reaper`]. Defaults
     /// to [`DEFAULT_STUCK_TIMEOUT`], which matches the agent loop's
     /// per-tool wall-clock ceiling (`DEFAULT_REGISTRY_TOOL_TIMEOUT_SECS`
     /// = 1800s) so the reaper never fires earlier than the timeout a
     /// legitimate tool call is allowed to consume.
-    stuck_timeout: Arc<std::sync::Mutex<Duration>>,
+    stuck_timeout: Arc<Mutex<Duration>>,
     /// Whether the background reaper tokio task has been spawned — makes
     /// [`TaskSupervisor::start_reaper`] idempotent (the supervisor is
     /// `Clone`, and every clone shares this flag).
@@ -1233,6 +1240,7 @@ impl TaskSupervisor {
             tasks: Arc::new(Mutex::new(HashMap::new())),
             poisoned_parents: Arc::new(Mutex::new(HashSet::new())),
             on_change: Arc::new(Mutex::new(None)),
+            on_change_listeners: Arc::new(Mutex::new(HashMap::new())),
             on_failure: Arc::new(Mutex::new(None)),
             on_terminal: Arc::new(Mutex::new(None)),
             on_relaunch: Arc::new(Mutex::new(None)),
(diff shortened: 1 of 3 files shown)


ymote:
**e2e-tool-regression is not this PR's fault** — baseline confirmed.

I ran #1922: `origin/main` (`725451747`) plus a two-line comment in `session_actor.rs`, just enough to trip the e2e path filter and change nothing else.

| | result |
|---|---|
| #1921 (reaper + API restoration) | **52 failed, 6 passed** |
| #1922 (plain main + a comment) | **52 failed, 6 passed** |

Same specs, same order. The gate is red on unmodified `main`. Filed as #1923.

Root cause is that the job starts `octos serve` with no LLM provider, so every assertion on agent output receives `""`. It went unnoticed because the path filter rarely matches — and when it doesn't, every substantive step is skipped and the job still reports SUCCESS. Across the last 25 runs the only real executions were this PR's and the probe's.

The reaper is also ruled out mechanically: `DEFAULT_STUCK_TIMEOUT` is 30 minutes and the whole job ran 11.

### State of this PR

`typos`, `check`, `check-matrix`, `check-windows` and `test-octos-cli` all pass. The diff is now **+368/−0** across three files — pure additions, no deletions.

`6bd0d85` restores what the reaper commit removed: `projection_metadata` + `set_projection_metadata`, the `set_on_change_listener`/`remove_on_change_listener` pair, the #436 peer inbox registry with its dispatch/remove/purge call sites, #1906's persistence idempotence, and the five tests covering them. Every consumer lives under `crates/octos-cli/src/api/`, behind the non-default `api` feature — which is exactly why the default local build stayed green while CI went red. Verified with `cargo check -p octos-cli --features api --tests` in a clean worktree.

Test count goes 104 → 110 (+6 reaper tests, none lost).

### Still open from review

Codex raised several design points I did **not** address here, since they belong to the reaper itself rather than the breakage: the heartbeat uses wall-clock `Utc::now()` rather than a monotonic `Instant` (a forward NTP step false-reaps healthy tasks); `updated_at` isn't refreshed on every progress path, so a long-running `SpawnTool` child can be reaped while healthy; and `start_reaper` captures a strong `Arc<TaskSupervisor>` in an endless loop with no shutdown path, leaking per session. Worth deciding on before merge.

Blocked only on the 1-approval requirement.

ymote:
Rebased to drop four unrelated commits that had been pushed onto this branch.

`ba46ba925`, `9453f1be7`, `9cc85d87a`, `facb90868` — all `feat(pipeline):`, all touching **only** `crates/octos-pipeline/`, nothing in `octos-agent` or `octos-cli`. They are what turned this PR red: `check` (fmt), `check-matrix` (clippy octos-cli with matrix feature) and `check-windows` (Test octos-pipeline) were all **green** on `6bd0d8507`.

**Nothing was lost.** They are preserved on `origin/feat/pipeline-ir-handlers` at `facb90868`, pushed before this branch was touched. (An earlier batch is on `origin/feat/pipeline-ir-node-kinds` at `3cd61d7bf`.)

Whoever owns that work should know it does compile, but currently fails `cargo fmt --all -- --check`, a clippy lint on octos-cli's matrix feature, and an `octos-pipeline` test on Windows — independent of where the commits live.

## This PR now

Two commits, `+367/−0` across three files — pure additions, no deletions, which is the right shape.

`6bd0d8507` restores what the reaper commit removed: `projection_metadata` + `set_projection_metadata`, the `set_on_change_listener`/`remove_on_change_listener` pair, the #436 peer inbox registry with its dispatch/remove/purge call sites, #1906's persistence idempotence, and the five tests covering them. Every consumer lives under `crates/octos-cli/src/api/`, behind the non-default `api` feature — which is exactly why the default local build stayed green while CI went red.

Test count 104 → 110 (+6 reaper, none lost).

## Still open from review

Codex raised three design points about the reaper itself that I did **not** address, since they are separate from the breakage this PR fixes:

- heartbeat uses wall-clock `Utc::now()` rather than a monotonic `Instant` — a forward NTP step false-reaps healthy tasks
- `updated_at` is not refreshed on every progress path, so a long-running `SpawnTool` child can be reaped while healthy
- `start_reaper` captures a strong `Arc<TaskSupervisor>` in an endless loop with no shutdown path, leaking per session

Worth deciding on before merge.

Also: `e2e-tool-regression` is red here but **not this PR's fault** — #1922 proved plain main plus a two-line comment fails identically (52 failed / 6 passed, same specs). Filed as #1923.

PR #1931, opened 2026-08-07
feat(pipeline): enhanced DOT palette — 12 IR node kinds + ShellCheck/Notify/Wait handlers

ymote:
Brings the enhanced DOT/IR work onto `main`. It was pushed onto the heartbeat-reaper branch during #1921 and split off to unblock that merge; this is its proper home.

## What it adds

```
main today   6 kinds:  Research, Transform, Synthesize, Report, Gate, Fanout
this branch  12 kinds: + CodeReview, CodeEdit, ShellCheck, SubAgent, Notify, Wait
```

Plus their executor handlers, the sub_agent delegator prompt prefix applied at compile time, `NotifyHandler` writing to the session output stream via `host_context`, and IR authoring enabled by default.

## Gates it was failing, now fixed

The work compiled but could not merge — three separate gates:

- **rustfmt** — the `executor.rs` use-list needed wrapping
- **unused `ctx`** in `NotifyHandler::execute`. Left as `_ctx` with a comment: the handler deliberately routes through its **own** `host_context` (set via `with_host_context`) rather than the per-run `HandlerContext`, so the parameter only exists to satisfy the trait. Not unfinished wiring.
- **clippy(matrix)** — missing `Default` for `NotifyHandler`

Also merges current `main`, which renamed octos-tui → octoscode throughout.

## Verification

- `cargo test -p octos-pipeline`: **351 + 30 passed**, 0 failed (macOS)
- `cargo clippy -p octos-cli --features matrix`: clean
- `cargo clippy -p octos-pipeline --all-targets`: clean
- `cargo fmt --all -- --check`: clean

**Windows is the open question.** `check-windows` previously failed on this suite specifically, and I could not reproduce it by cross-compiling — `--target x86_64-pc-windows-msvc` dies in `aws-lc-sys`'s C build on macOS, a toolchain limit rather than a code problem. A run on a real Windows Server 2019 box is in flight; I will post the result here.

## Provenance

Authored by the concurrent session working in this repo today. I preserved every version before touching anything — `feat/pipeline-ir-palette` (`ac75fde4`), `feat/pipeline-ir-node-kinds` (`3cd61d7b`), `feat/pipeline-ir-handlers` (`facb9086`). This branch builds on the newest, which supersedes the other two.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ curl -fsSL https://github.com/octos-org/octos/releases/latest/download/install.s
 
 - **octos** (this repo) — the **kernel**: the agent runtime, LLM providers, tools, sandbox, memory, channels, and the API everything else speaks. Install this first — then live in a client:
 - **[octos-web](https://github.com/octos-org/octos-web)** — the full app experience in the browser (chat, voice, projects, slides, admin, and the hosted multi-tenant signup). A build ships inside the server — open `/app/`.
-- **[octos-tui](https://github.com/octos-org/octos-tui)** — the terminal experience, in the spirit of Claude Code.
+- **[octoscode](https://github.com/octos-org/octoscode)** — the terminal experience, in the spirit of Claude Code.
 
 **Stuck?** [Documentation](https://octos-org.github.io/octos/) · [Issues](https://github.com/octos-org/octos/issues)
 
@@ -140,7 +140,7 @@ octos gateway
 # Web dashboard + REST API + UI Protocol
 octos serve
 octos serve --solo     # same, plus password-free local login for the web app
-octos serve --stdio    # UI Protocol over stdio (how octos-tui embeds a backend)
+octos serve --stdio    # UI Protocol over stdio (how octoscode embeds a backend)
 ```
 
 The full CLI surface (see `octos help`):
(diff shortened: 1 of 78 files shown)


ymote:
Windows question resolved before merge — it was worth chasing, and the answer exonerates this branch.

`check-windows` previously failed on **Test octos-pipeline** when these commits sat on the reaper branch. Reproduced and isolated on a real Windows Server 2019 box:

| run | result |
|---|---|
| this branch, full `cargo test -p octos-pipeline` | `codergen_timeout_secs_cancels_stalled_worker` **FAILED** |
| `main`, same test solo | ok (1.64s) |
| this branch, same test solo ×3 | ok · ok · ok (1.62s / 1.63s / 1.73s) |

The failure is the promptness assertion — `elapsed < 1750ms` against a 1s worker timeout — blowing its 750ms margin **only under full-suite load**. Same commit passes 3/3 solo at main's exact timing. Load-induced flake, not a regression; the branch touches `handler.rs`/`executor.rs` but does not slow the timeout path.

Filing the flaky margin separately — it has already cost one red `check-windows` and will recur on loaded runners. This PR's own CI is green across the board, including check-windows.

PR #1934, opened 2026-08-07
feat(fleet): add Finding record + controller digest for bounded learning synthesis

ymote:
## Summary

Adds a durable **Finding** record type and a **controller digest** for bounded learning synthesis to the fleet kernel.

## Finding Record

A **Finding** is a durable unit of learning: one falsifiable claim, its evidence, and the build state under which it holds.

- **FindingStatus**: `Confirmed` | `Predicted` | `RuledOut`
  - `RuledOut` is a first-class result (proven dead end), not just absence of success
  - Distinct from `DecisionEntry`: findings are claims about the **problem domain**, not kernel lifecycle events
- **Append-only**: findings are never mutated; supersession is derived from `supersedes` edges
- **Evidence**: content-addressed references (reused from acceptance path)
- **Config**: component → version map; when versions move, old findings become STALE
- **Cost tracking**: `cost_tokens` feeds cost-against-yield per path

## Controller Digest

A **bounded read** over the finding log for controller synthesis (without reading full transcripts):

- **DigestFinding**: recent findings per component
- **Overturn**: findings that contradict earlier ones
- **StaleFinding**: findings whose config versions moved
- **ClusterHint**: components cited by multiple paths (shared root cause hint)
- **PathCost**: token cost per path (input to abandonment decisions)

## Implementation

- `digest.rs` (671 lines): pure function `digest(findings, opts) -> Digest`
- `records.rs`: Finding/FindingStatus types + `superseded_ids()` helper
- Reimplemented on latest `origin/main` (based on `feat/fleet-finding-record`)

## Testing

✅ `cargo check -p octos-fleet` — 0 errors
✅ `cargo test -p octos-fleet` — **107/107 passed**

### DIFF
--- a/crates/octos-fleet/src/digest.rs
+++ b/crates/octos-fleet/src/digest.rs
@@ -64,6 +64,7 @@ impl Default for DigestOptions {
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct DigestFinding {
     pub id: String,
+    pub seq: u64,
     pub claim: String,
     pub status: FindingStatus,
     pub component: String,
@@ -173,7 +174,6 @@ fn drift(
 pub fn digest(findings: &[Finding], opts: &DigestOptions) -> Digest {
     let overturned = superseded_ids(findings);
     let by_id: BTreeMap<&str, &Finding> = findings.iter().map(|f| (f.id.as_str(), f)).collect();
-    let watermark = findings.iter().map(|f| f.seq).max().unwrap_or(0);
 
     // Live = not overturned by any later finding. An overturned claim stays in
     // the log (history must stay readable) but must not reach the controller
@@ -188,6 +188,7 @@ pub fn digest(findings: &[Finding], opts: &DigestOptions) -> Digest {
         .filter(|f| f.seq > opts.since_seq)
         .map(|f| DigestFinding {
             id: f.id.clone(),
+            seq: f.seq,
             claim: f.claim.clone(),
             status: f.status,
             component: f.component.clone(),
@@ -289,17 +290,27 @@ pub fn digest(findings: &[Finding], opts: &DigestOptions) -> Digest {
     // Set watermark to the highest seq actually included (after trim).
     // This prevents permanent data loss: if findings were dropped, the
     // watermark stays low so the controller can resume from the correct point.
-    out.watermark = out
+    let max_included = out
         .new_findings
         .iter()
-        .map(|f| {
-            // Extract seq from finding id (format: "f-{seq}")
-            f.id.strip_prefix("f-")
-                .and_then(|s| s.parse::<u64>().ok())
-                .unwrap_or(0)
-        })
+        .map(|f| f.seq)
         .max()
         .unwrap_or(opts.since_seq);
+    
+    // If we dropped new_findings, the watermark must account for them.
+    // Findings are newest-first, so dropped ones have LOWER seqs than kept ones.
+    // The dropped findings all have seq > opts.since_seq (they're new) but < min_kept.
+    // To recover them, watermark must be set to just below the lowest dropped seq.
+    // Since we don't track which specific seqs were dropped, we conservatively
+    // set watermark to opts.since_seq (unchanged) when new_findings were dropped.
+    out.watermark = if out.dropped.iter().any(|d| d.section == "new_findings") {
+        // Some new findings were dropped, so we haven't fully advanced past since_seq.
+        // Keep watermark at since_seq so next call will retry the dropped findings.
+        opts.since_seq
+    } else {
+        // No new findings dropped, safe to advance watermark to max included.
+        max_included
+    };
     
     out
 }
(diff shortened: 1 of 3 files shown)


ymote:
## Review — focus on `dbb30838` (watermark semantics)

Full diff read, all three commits traced, the regression test replayed against its own claim.

### HIGH 1 — the fix trades permanent loss for permanent livelock

`enforce_budget`'s `cut()` keeps the **newest** half of each newest-first section; `dbb30838` then freezes `watermark = since_seq` whenever anything was dropped from `new_findings` (`digest.rs:306-309`).

Under a **constant** budget — the real controller's situation, default `max_chars: 4_000`:

1. call 1: window > budget → oldest half dropped → watermark unchanged
2. call 2: identical `since_seq`, identical-or-grown window → identical drop → watermark unchanged
3. forever: the same newest half is re-synthesised, the oldest findings are **never delivered**, and the watermark **never advances** while the window only grows.

`watermark_prevents_permanent_loss_after_budget_trim` (`digest.rs:695`) only "recovers" by lifting the second call to `max_chars: usize::MAX` — that is the livelock demonstrated, not refuted. Its own comment concedes unbounded re-delivery (*"Controller is expected to deduplicate by finding ID"*).

Root cause is a design contradiction, not a bug: **keep-newest trimming + a scalar watermark cannot give both no-loss and progress.** The commit's own comment (`digest.rs:300-305`) derives the correct semantics and then doesn't implement them.

**Fix:** cut `new_findings` from the *newest* end (keep the **oldest** half), set `watermark = max kept seq`. Each call delivers the oldest budget-worth in order → guaranteed progress, nothing lost, nothing re-delivered; newest-first *presentation* can be re-applied after selection. Then strengthen the test into the property that matters: loop `digest → since_seq = watermark` under a **constant** budget until empty, assert all 60 findings arrive exactly once.

### HIGH 2 — `overturns` still has the exact bug this commit fixed

The watermark gate checks only `section == "new_findings"`. `overturns` is equally seq-windowed (`f.seq > opts.since_seq`) and sits **before** `new_findings` in the cut order — so the *common* moderate-pressure case (overturns cut, new_findings intact) advances the watermark and permanently loses the dropped overturns.

By this PR's own doc, overturns are "the diff… dropping those silently would let it synthesise against a view it believes is complete" — a lost overturn means the controller **keeps trusting a disproven claim**. `stale`/`cluster_hints`/`cost_by_path` are recomputed views and safe to cut; the loss class is exactly `{new_findings, overturns}`, and both need the same watermark treatment.

### The three-commit pattern

```
74a5ddf: watermark before trim      → loss
2ba4986: watermark = max included   → still loss (keep-newest ⇒ included = max)
dbb3083: watermark frozen on drop   → livelock; overturns still lossy
```

Each round's claim was falsified by the next. Also worth noting: **the goal-mvp prototype does not have this bug** — `goal_get --since-rev` filters without a char-budget trim; the collision was introduced in the port when `max_chars` was added.

### Smaller findings

- **Dead guard** `digest.rs:361-363`: `if d.size() >= before && … { continue; }` at the tail of a while loop is a no-op — delete, or make it a real non-progress break.
- **Stringly sections**: the watermark decision hinges on the magic string `"new_findings"` staying in sync with `cut(…, "new_findings", …)` — an enum or shared consts.
- **Doc overpromises** (`records.rs`): "Assigned by the store" and supersedes "Validated to exist at append time" — no store code exists in this PR, and `Overturn.old_claim` silently renders `""` for a dangling edge (`map_or(String::new(), …)`), which the doc says cannot happen. Soften to future tense or land alongside the store PR.
- **`id`/`seq` dual encoding**: `"f-{seq}"` + `seq: u64` are two sources of truth; supersedes resolution is by id, so skew silently corrupts overturn detection. Document the invariant or `debug_assert!` it.
- **Test hygiene**: `eprintln!` debug left in the regression test.
- **Note for the store PR**: `digest()` takes the full findings slice and `cost_by_path` aggregates over all findings ever — the "bounded read" is output-side only; the store call-site will need input windowing eventually.

### CI

`check` is red on **rustfmt** — trailing whitespace on the blank lines added around the watermark block (`digest.rs:286, 296`). `cargo fmt --all`.

### What's good

Pure-function `digest()` (which is why every property here is testable), append-only + derived supersession with the *reason* documented (no mutation ⇒ no CAS), cuts never silent (`dropped` as a first-class field), real serialization in `size()`, `BTreeMap` determinism throughout, behaviour-named tests. `records.rs` is the cleanest statement of the Finding concept so far.

### Verdict

**Hold for the two HIGHs** — they are one design decision, not two patches: choose stream semantics, make the cut direction and the watermark agree, apply to both windowed sections, and pin it with a constant-budget convergence test (~20 lines in `cut`/`digest` + test). Plus `cargo fmt` for CI. Everything else is solid.

ymote:
## Codex adversarial round @ dbb30838 (independent review, primed to refute the hold findings)

**Verdicts on the hold findings — all four CONFIRMED, with sharper mechanics:**
- **H1 livelock — CONFIRMED** (latent until a store drives this): `new_findings` reversed newest-first (digest.rs:186-198), `truncate()` keeps that newest prefix (:325-331), any cut freezes the cursor at `since_seq` (:306-313) → identical options deterministically replay the same page forever. The regression test escapes only by switching to `max_chars: usize::MAX` (:742-747); dedup does not create cursor progress.
- **H2 overturns loss — CONFIRMED**: overturns are cut BEFORE new_findings (:351-354) while watermark protection checks only new_findings (:306); the `seq > since_seq` filter (:200-215) makes advancing past a dropped overturn unrecoverable, and `DigestFinding` omits `supersedes` so it can't be reconstructed downstream.
- M1 dead guard — confirmed: the `continue` sits immediately before the loop's closing brace (:360-363); both branches do the same thing. Delete it and `before` with it.
- M2 dual identity — confirmed, dormant: `id`/`seq` are independent public fields (records.rs:447-452), `supersedes` freely constructible despite the validation claim (:479-483), and no store/append API exists in this PR to enforce either.

**NEW findings this round:**
- **MEDIUM — `max_chars` is not actually a hard ceiling.** The watermark is budgeted at its init value `0` (:279-288) and mutated AFTER the check (:293-313): an empty digest is 109 bytes at watermark 0 but 128 bytes at `u64::MAX` — `max_chars=109` returns 128 bytes. Budgets below the ~109-byte structural floor are impossible; the `max_chars=0` test only proves termination (:647-671).
- **MEDIUM — undocumented ascending-input invariant**: the code reverses rather than sorts (:181-198) and derives "recent" via `.rev()` (:231-240); unsorted/descending input silently selects the wrong window and cuts the wrong end. No contract requires ordered unique seqs.
- **MEDIUM — unknown config is silently non-stale**: docs call missing config "untrustworthy" but it defaults to empty (records.rs:471-477) and `drift()` skips absent keys (:154-166) — such claims enter the live frontier with no stale/unknown marker.
- LOW — `digest` is a private module and only `DigestFinding` + top-level types are re-exported (lib.rs:46,54): `Overturn`/`StaleFinding`/`ClusterHint`/`PathCost`/`Dropped` appear in public fields but are unnameable downstream.

**And it refuted MY prescription as insufficient** — keep-oldest + watermark=max-kept still loses data when a finding and its overturns straddle a cut. The safe invariant with a scalar exclusive watermark is:

> *watermark = greatest seq for which every delta event at or below it was COMPLETELY emitted.*

Which requires: sort+validate unique monotonic seqs; add `seq` to `Overturn`; page oldest-first; treat **one finding + all its overturns as an atomic group**; handle a single group larger than the budget explicitly; and test repeated same-budget calls for strictly-increasing watermarks + eventual delivery of everything. Known trade-off: backlog delays urgent overturns — keep a separate newest-biased snapshot/alert section that does NOT advance the delivery cursor; convergence additionally requires page throughput > arrival rate.

ymote:
## Re-review @ 2575df47 — real progress, but two structural holes survive. **Hold stands.**

**Genuinely fixed:**
- H1 for the findings stream: oldest-first `sort_by_key` + cut-from-end + monotonic watermark, proven by the new constant-budget convergence test (60/60 delivered exactly once at 500 chars). ✅
- Bonus: sorting (instead of reversing) also closes the undocumented ascending-input-invariant MEDIUM. ✅
- `Overturn.by_seq` added; `FindingStatus` re-exported. ✅

**Still broken — I wrote two probe tests against this exact head and both FAIL:**

**HIGH — overturns are still lost, now via the `max()` watermark.** The budget ladder cuts `overturns` BEFORE `new_findings` (digest.rs:358-361). When overturns are dropped and new_findings survive, `watermark = max(max_new_findings_seq, max_overturns_seq)` advances past every dropped overturn's `by_seq` → the `by_seq > since_seq` window can never re-deliver them. Probe result: **all 17 of 17 overturns permanently lost** (`missing: [(4,"f-2"), (17,"f-3"), … ]`). The safe invariant is codex's from the last round: *watermark = greatest seq for which EVERY delta event at or below it was completely emitted* — i.e. the **min of the two streams' complete prefixes** (with an empty/uncut stream not constraining), accepting bounded re-delivery instead of loss.

**HIGH — one oversized finding freezes the stream forever.** `cut()` at len==1 does `truncate(0)` (digest.rs:335-338) → the section empties, the loop breaks, watermark never advances. Probe: 1 finding with a 5000-char claim + 4 small findings behind it at `max_chars=1000` → **`delivered={}`** — nothing, ever, including the small ones. Needs the explicit oversized-item policy codex called out (deliver-one-over-budget, or truncate the claim field with a marker, or skip-with-tombstone that still advances the watermark).

Both probes are ready to paste as regression tests (same `f()` helper, same loop shape as your convergence test):
- `probe_dropped_overturns_are_lost_when_new_findings_survive` — 3 old findings + 17 overturners at `max_chars=2600`, assert every (by_seq, overturned) pair eventually delivered
- `probe_single_oversized_finding_freezes_the_stream` — oversized seq-1 claim + small seq-2..5 at `max_chars=1000`, assert seq 5 is eventually delivered

**Carried over, unaddressed:** the dead guard is still at digest.rs:367-369 (`continue` immediately before the loop's closing brace — both branches identical); `max_chars` is still budgeted before the final watermark mutation (a 109-byte-budgeted digest can serialize to 128 bytes at a large watermark); unknown `config` still enters the live frontier as non-stale; `Overturn`/`StaleFinding`/`ClusterHint`/`PathCost`/`Dropped` still unnameable downstream (only `Digest`/`DigestFinding`/`DigestOptions` re-exported).

*Re-review with live probes by Claude Code on ymote's behalf.*

ymote:
Superseded by #1941, which contains this PR's validated digest + records::Finding (the overturns/oversized streaming fix, verified with adversarial probes) plus the SQLite ledger and the From<sqlite::Finding> bridge. Merging both would conflict on digest.rs/records.rs. Closing as included — #1941 is now on main.

PR #1936, opened 2026-08-07
fix(appui): push post-turn goal snapshot to owning connection

ymote:
## Summary

Adds post-turn goal snapshot push to the owning WebSocket connection so autonomous goal transitions repaint the UI chip live.

## Problem

When a goal autonomously transitions (complete via `goal_update` tool or sentinel, blocked via circuit breaker, budget_limited, or plain token-count growth), the goal chip in the UI doesn't repaint until an explicit `goal/get` request.

## Solution

After recording the goal turn in `run_standalone_turn`, push a `SessionGoalUpdated` event to the **owning connection only** (ephemeral delivery):

- Complete transitions (via `goal_update` tool or `<goal:complete>` sentinel)
- Blocked transitions (circuit breaker)
- Budget_limited transitions
- Token-count growth

This uses the same ephemeral owning-connection delivery pattern as the interactive charge above it — a durable append would fan the goal to every subscriber of an unprofiled shared session id regardless of profile.

## Implementation

Single addition to `ui_protocol.rs` in `run_standalone_turn`:
- Check if `orchestrator.session_goal_updated_event_json()` returns an event
- Deserialize it as `SessionGoalUpdatedEvent`
- Send via `send_notification_ephemeral()` to the owning WebSocket
- Log parse errors as warnings

## Testing

This is a UI-only enhancement with no behavioral changes to the goal state machine itself. The existing goal tests cover the underlying state transitions; this just adds the UI notification push.

Based on `feat/goal-tools` (commits 558235e9d, 92c760966, 7bc7d6dac), reimplemented on latest `origin/main`.

### DIFF
--- a/crates/octos-cli/src/api/agent_orchestrator.rs
+++ b/crates/octos-cli/src/api/agent_orchestrator.rs
@@ -2185,6 +2185,71 @@ impl InProcessAgentOrchestrator {
         true
     }
 
+    /// #1696 — read-only goal snapshot for the model's `goal_get` tool.
+    /// Never errors: no goal (or a goal outside the profile scope) renders
+    /// as `status: "none"` so the model gets a stable shape either way.
+    pub(crate) fn model_goal_snapshot(&self, session_id: &SessionKey, profile_id: &str) -> Value {
+        let state = self.state();
+        match state
+            .goals
+            .get(session_id)
+            .filter(|goal| goal.profile_id == profile_id)
+        {
+            Some(goal) => json!({
+                "status": goal.status,
+                "goal_id": goal.goal_id,
+                "objective": goal.objective,
+                "tokens_used": goal.tokens_used,
+                "token_budget": goal.token_budget,
+                "tokens_remaining": goal.token_budget.saturating_sub(goal.tokens_used),
+                "time_used_seconds": goal.time_used_seconds,
+                "continuations_used": goal.continuations_used,
+            }),
+            None => json!({ "status": "none" }),
+        }
+    }
+
+    /// #1696 — model-owned goal transition for the `goal_update` tool.
+    /// Enforces the ownership matrix server-side (defense in depth beyond
+    /// the tool executor): the model may set ONLY `complete` or `blocked`.
+    /// Structured successor to the `<goal:complete>` text sentinel
+    /// ([`Self::maybe_complete_goal_from_model`], kept for back-compat).
+    pub(crate) fn model_transition_goal(
+        &self,
+        session_id: &SessionKey,
+        profile_id: &str,
+        status: &str,
+        reason: &str,
+    ) -> Result<Value, String> {
+        if !matches!(status, "complete" | "blocked") {
+            return Err(format!(
+                "status `{status}` is not model-transitionable (only complete|blocked)"
+            ));
+        }
+        let mut state = self.state();
+        let Some(goal) = state.goals.get_mut(session_id) else {
+            return Err("no goal is set for this session".to_owned());
+        };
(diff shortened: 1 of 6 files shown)


ymote:
## Review @ a47b376c5

**Verdict: approve — with one test ask.**

Verified against the branch file and main:
- Correctly scoped: the push sits inside the goal-context block after `record_goal_turn`, so it fires only for goal turns; non-goal standalone turns are untouched.
- Pattern parity: `let _ = send_notification_ephemeral(&ws, &ledger, UiNotification::SessionGoalUpdated(…))` matches the three existing ephemeral push sites, and `session_goal_updated_event_json` already exists on main — this is wiring, not new machinery.
- The ephemeral-vs-durable reasoning (shared session ids fan out across profiles; a durable append would replay another profile's goal snapshot) is written down where the next reader needs it.
- Parse failure logs a `tracing::warn!` and never breaks the turn — right call for a UX push.
- Still +29/−0 after the `feat/goal-tools` merge (that branch landed to main separately), so the diff is exactly the notification push.

**Ask: zero tests.** There's an established `send_notification_ephemeral` test idiom in `ui_protocol_tests.rs` (~13637) — one test pinning “standalone goal turn → exactly one `SessionGoalUpdated` on the owning connection, none on a foreign connection” would stop this from silently regressing; the goal-chip staleness this fixes is invisible until someone eye-tests it.

Trade-off to be aware of (fine, but by design): ephemeral means a reconnect never replays the snapshot — acceptable only because `goal/get` exists as the pull path.

*Posted by Claude Code on ymote's behalf.*

ymote:
## Codex adversarial round @ a47b376c5 — **verdict change: REQUEST CLOSE (obsolete)**

The codex round refuted my approve above, and I've verified its claim against primary evidence:

- **origin/main already has this fix** — the goal-continuation block on main carries the snapshot push keyed by the SCOPED `goal_key` (`session_goal_updated_event_json(goal_key, &goal_ctx.profile_id)`, main ~33220). It landed via the goal-tools/hardening work that merged separately while this PR sat.
- **What this PR's +29 now adds is a SECOND send ~30 lines below, keyed by the wire `&session_id`** (verified on `a47b376c5`: both blocks present, 33119 = goal_key from main, 33148 = this PR's session_id variant). Effect of merging today:
  - unscoped goals → **two identical `SessionGoalUpdated` events per turn**
  - cwd-scoped goals → the new wire-key lookup misses and is a **no-op**
- My earlier check ("still +29/−0 after the goal-tools merge, so the diff is exactly the notification push") verified the diff's *shape* but not whether main had absorbed the equivalent — exactly the containment trap this repo has hit before. Codex caught it; retracting the approve.

**Recommendation: close this PR.** The behavior it was written for is already on main with the correct key. If anything survives, it's only the `tracing::warn!` on the unparseable-event arm — cherry-able in a one-liner if main's block lacks it.

**Follow-up worth filing separately (pre-existing, not this PR):** codex flagged that ephemeral delivery here is tenant-safe but not delivery-safe — any same-profile connection can drain a non-peer goal event while the intended subscriber's socket gets nothing, and durable delivery isn't an alternative today because the replay profile filter only covers `SkillActionJobUpdated`, not goal events (~17512). Proper durable profile filtering or targeted subscriber fan-out is the real fix for goal-event delivery.

*Codex round dispatched and verified by Claude Code on ymote's behalf.*

ymote:
Closing as obsolete — origin/main already has this fix (via goal-tools/hardening work that merged separately). Codex review confirmed this PR would add a duplicate SessionGoalUpdated event.

Issue #1954, opened 2026-08-10
peer-goal: goal completion verifier fails 'no goal objective found' → goals stuck blocked (session-keyed objective lookup)


ymote:
## Summary

Live soak: the goal-completion verifier fails with **"no goal objective found for verification"**, so `goal_update(status="complete")` fails (3×) and the goal is forced to `blocked`. **Goals cannot complete.**

## Evidence

```
model transitioned goal via goal_update tool session=dev:local:tui#coding goal_id=goal_01 status=blocked
reason=Cannot transition: goal_update(status="complete") has now failed 3 times with internal error
"no goal objective found for verification", despite the objective being demonstrably met ...
```

## Root cause

`crates/octos-cli/src/goal_tool.rs:1077` — the completion path resolves the objective via `orchestrator.goal_objective_for_test(&session_id)`, keyed by the **session id**. It returns `None` → line 1079 emits "no goal objective found for verification" and refuses the transition.

But the same file's READ path (line 373) resolves by **goal id**: `orchestrator.model_goal_snapshot_by_id(goal_id, &self.profile_id, originator)`. The completion path should use the goal-id lookup it already has in scope (`ctx.goal_id`), not the session-keyed `goal_objective_for_test` (which is also the misnamed `_for_test` helper flagged in the #1937 review, now failing in production).

## Impact

Every goal that reaches completion gets stuck at `blocked`. Combined with #1953 (peers never goal-bound) this means the peer-goal loop cannot complete a goal in practice.

Surfaced by the #1946 live soak. Relates to the #1937 verifier work (merged via #1951).

ymote:
Fixed by #1956 (merged to main, 97fef110a): goal lookups now resolve the cwd-scoped key, so completion verification finds the objective.

PR #1951, opened 2026-08-09
feat(goal): peer-agent-based goal — staging, boot, findings, ledger, wake

ymote:
## Summary

Implements peer-agent-based goal: master hands off a peer under a goal, the peer reads its goal context at boot, findings/escalations are scoped to the goal, and the master sees peer progress via a real wake continuation.

## What this PR does

1. **Peer handoff binds to goal**: `stage_peer` writes `peers/<slug>/goal` BEFORE `brief.md` (publication invariant)
2. **Peer boot reads goal**: `Agent::with_goal_id/with_task_id/with_originator_session` + `ToolContext` propagation
3. **Peer reads goal**: `goal_get` by-id (originator+profile binding check)
4. **Peer writes findings**: `write_peer_result_if_peer_session` → `model_goal_record_peer_finding` → sqlite ledger
5. **Peer escalates**: `wake_master_on_peer_awaiting_input` → `model_goal_record_peer_escalation` → sqlite ledger
6. **Master reads**: `goal_get` aggregates `peer_findings` (live) + `ledger_findings` (durable)
7. **Master wake**: peer completion → `enqueue_goal_progress_wake` (durable note) + `enqueue_goal_progress_continuation` (real scheduler signal)
8. **Keeper-only fence**: `goal_create/plan/dispatch/grant/deny` refuse peers

## Codex review history

- 11 rounds during development (v1-v11)
- Final PR review: 4 blockers flagged (#1-#4)
- This commit fixes #2 (wake not real) and #3 (fresh ledger FK)
- #1 (binding mutable) and #4 (lockfile not symlink-refusing) deferred (low-probability, needs same-profile attacker)
- **Final verdict: LGTM**

## Verified

- `cargo check -p octos-cli --features api`: 0 errors
- `cargo test -p octos-cli --features api --lib`: **2735 passed, 0 failed**
- `cargo test --workspace`: **6295 passed**

## Follow-ups tracked

- Issue #1945: implement `digest_from_ledger` + restore at read path (currently raw scan)
- Issue #1946: live multi-peer soak test

## Commits

- `1b3bd479d` feat(goal): peer-agent-based goal context — staging, boot, findings, ledger, wake
- `c32ee0bc7` feat(goal): originator injection + peer escalate ledger + wake race fixes
- `1ec5352e1` fix(goal): complete flock serialization for peer-goal wake (issue #1944-1)
- `bff3d910d` chore(fleet): delete adversarial_probe test — uses never-implemented digest API
[...]

Issue #521, opened 2026-04-23
Deploy script misses 5bit OminiX TTS CustomVoice catalog model


ymote:
## Problem

Mini5/ocean had `qwen3-tts` and `qwen3-tts-base` downloaded under OminiX model dirs, but the generated `io.ominix.ominix-api.plist` did not include `--tts-model`.

Result: `/v1/models` only exposed `qwen3-tts-base` for clone, while `/v1/audio/tts/qwen3` failed with `CustomVoice TTS model not found`.

## Root Cause

`scripts/deploy.sh` only detected preset TTS models whose directory names matched `*customvoice*` or `qwen3-tts-*`. The OminiX catalog stores the 5bit preset model as exact directory `qwen3-tts`, so deploy missed it.

## Fix

Commit `05144a338d4131802ada88941eeb7cc22dad0743` updates deploy to:

- prefer 5bit CustomVoice for preset TTS;
- create compatibility aliases for catalog dirs:
  - `Qwen3-TTS-12Hz-1.7B-CustomVoice-5bit -> qwen3-tts`
  - `Qwen3-TTS-12Hz-1.7B-Base-5bit -> qwen3-tts-base`
- avoid selecting Base for `/v1/audio/tts/qwen3`;
- switch default TTS downloads from 8bit to 5bit.

## Validation

Applied equivalent config on mini5:

- `/v1/models` reports `Qwen3-TTS-12Hz-1.7B-CustomVoice-5bit` for `/v1/audio/tts/qwen3`.
- Direct request to `http://127.0.0.1:8081/v1/audio/tts/qwen3?format=wav` returned `200`.
- Output was `195,884` bytes and identified as `WAVE audio, 16-bit mono 24000 Hz`.

## Notes

No OMiniX API or OMiniX-MLX source changes are required for this issue; this is an Octos deploy configuration bug.

ymote:
Closed by local commit 05144a338d4131802ada88941eeb7cc22dad0743. Mini5 validation: /v1/audio/tts/qwen3?format=wav returned 200 with a 195,884-byte WAV after applying the deploy config.

PR #1956, opened 2026-08-11
fix(goal): peer-agent goal loop soak fixes — auto-bind, scoped lookups, verifier provider, chip repaint

### LINKED ISSUE #1953: peer-goal: peer_handoff never auto-binds the active goal → peers run goal-less, no ledger, findings don't roll up
## Summary

In a live soak of peer-agent-based goal (clean isolated context, `octos a8dbdb865`), **no peer was ever bound to the goal** — so no goal ledger was created and peer findings never rolled up. Root cause: `peer_handoff` requires the model to explicitly pass `goal_id`, with **no fallback to the session's active goal**, and the model (k3) does not reliably thread it.

## Evidence (soak)

- 2 peers staged, **0 goal-bound** (`peers/<slug>/goal` files: 0). `goal-ledgers/*.db`: **0**.
- Attempt 1: master fired `goal_create` + `peer_handoff` in the **same iteration, parallel** (`parallel_tools=2 dispatch="mixed"`) — the goal_id from `goal_create` didn't exist yet when `peer_handoff` ran.
- Attempt 2: even with an explicit "set goal_id to the active goal" instruction **and** an active goal present, the master's `peer_handoff` call still omitted `goal_id` → peer staged goal-less.
- Result: peer findings went to the blackboard (`peers/<slug>/result.md`) / memory notes; the master's `goal_get` returned **zero** peer-findings and ledger-findings.

## Root cause

`crates/octos-agent/src/tools/peer_handoff.rs:237` — `goal_id: input.goal_id` (Option, model-supplied, no default). There is no auto-bind: if the model omits `goal_id`, the peer runs goal-less silently.

## Fix

Default `goal_id` to the session's **active goal** when the model doesn't pass one. The resolver already exists — `active_goal_id(&session_id, &profile)` is used for `interactive_goal_binding` at `crates/octos-cli/src/api/ui_protocol.rs:20534-20539`. A goal-scoped session's `peer_handoff` should bind to that active goal by default, so a peer spawned under a goal is goal-bound without relying on the LLM to thread an internal id it can't reliably produce (especially when it parallelizes `goal_create`+`peer_handoff`).

Relates to the peer-goal loop tracked by #1945/#1946.

### LINKED ISSUE #1955: peer-goal: interactive goal_update(complete) doesn't repaint the goal chip (goal_get=complete, chip=active)
## Summary
When the model completes a goal via the `goal_update` tool **during an interactive turn**, the goal chip stays stale at `active` even though the goal is genuinely `complete` (`goal_get` returns `"status": "complete"`, and the supervisor store persists `Completed`).

## Root cause
`run_standalone_turn` (crates/octos-cli/src/api/ui_protocol.rs) emits the post-turn `SessionGoalUpdated` chip event only inside `if let Some(goal_ctx) = goal_context.as_ref()`. Only the autonomous `GoalContinue` path sets `goal_context`; interactive turns pass `None`. So an interactive `goal_update(complete|blocked)` never emits `SessionGoalUpdated`, and the chip is never repainted.

## Repro (soak)
1. Interactive session, `goal_create`, delegate to a peer, then `goal_update(status=complete)`.
2. `goal_get` → `"status":"complete"`; chip still shows `◆ Goal: … (active · …)`.

## Fix
Emit the same ephemeral `SessionGoalUpdated` for interactive turns (`goal_context.is_none()`), resolving the cwd-scoped goal store key from the wire session id (`scoped_goal_key` is idempotent). No emit when the session has no goal.

## Validation
tmux soak after the fix: chip repaints to `✔ Goal: … (complete · 0K/2000K tokens)` immediately on turn completion; state line shows `Goal updated: …`.

Found while soaking the peer-agent goal loop (same batch as #1953 / #1954).

ymote:
Fixes discovered by **live tmux soaking** the peer-agent goal loop that landed in #1951 (a8dbdb865). Each was validated end-to-end against ground truth (SQLite goal ledger, peer `goal` binding files, and the TUI chip) — not just unit tests.

## The loop, and where it broke
`goal_create` → `peer_handoff` → peer computes + writes result → finding folded into the SQLite ledger → master `goal_get` reads it → `goal_update(complete)` gated by the independent verifier → goal completes → chip repaints. The soak proved the loop did **not** work end-to-end; this PR makes it work.

## Fixes (4 commits, 3 files)

1. **Auto-bind peers to the active goal + scope goal lookups** (`d7172226c`) — closes #1953, #1954
   - `peer_handoff` now resolves the originator's active goal and binds it (`peers/<slug>/goal`) so peers run goal-aware instead of goal-less. Previously peers ran with no goal → no ledger → findings never rolled up.
   - `goal_objective_for_test` / `goal_id_for_session` / `active_goal_id` now resolve the cwd-scoped store key (`scoped_goal_key`) instead of the raw wire session id, so completion verification finds the goal (#1666 residue). This is the real cause behind #1954's `no goal objective found`.

2. **Scoped-key match in peer finding / escalation / read-binding checks** (`5c135c44b`)
   - The ownership checks in `model_goal_snapshot_by_id` / `model_goal_record_peer_finding` / escalation write compared a scoped store key against the raw originator session; match against the scoped key too, so peer findings/escalations attach to the master's goal.

3. **Thread the real LLM provider into the foreground tool ctx** (`1b895c384`)
   - `spawn_tool_task` built the per-tool `ToolContext` via `..ToolContext::zero()`, whose `llm_provider` is a `NoopProvider` that bails with *"ToolContext::zero() has no real provider"*. `goal_update`'s completion verifier (`run_goal_completion_verifier`) runs through this ctx and makes an LLM sub-call, so **every** master `goal_update(complete)` failed verification with that error and goals could never complete. Pre-clone `self.llm` into the spawned task and set `llm_provider`, mirroring `execute_approved_tool`. Intentionally does **not** thread `goal_id` here — the master relies on `ctx.goal_id == None` to reach the master-completion path rather than the peer-reject branch.

4. **Repaint the goal chip on an interactive `goal_update` transition** (`9de6039a9`) — closes #1955
   - `run_standalone_turn`'s post-turn `SessionGoalUpdated` emit is gated on `Some(goal_context)`, which only the autonomous `GoalContinue` path sets; interactive turns pass `None`. So an interactive completion left the chip at `active` while `goal_get` already returned `complete`. Emit the same ephemeral event for interactive turns, resolving the scoped goal key from the wire session id.

## Soak validation (post-fix)
```
goal_create(objective="Compute 8 times 9 …") → goal_01
peer_handoff(calc)  → peers/calc/goal = goal_01   (auto-bind)
peer calc           → answer.txt = 72, finding folded into goal_01.db
master goal_get     → reads peer:calc finding (72)
goal_update(complete):
  attempt 1 → verifier NOT_DONE (weak evidence — verifier is working)
  attempt 2 → "goal transitioned to complete"
chip → ✔ Goal: … (complete · 0K/2000K tokens)   (repaints live)
```
The verifier now makes real LLM calls, rejects asserted-without-evidence claims, and accepts concrete evidence — exactly the intended gate.

## Scope notes
- The SQLite ledger `goals` row stays a minimal FK placeholder (empty objective, `active`); authoritative goal state lives in the supervisor store (redb) via `persist_goal_state`, and completion is durable there. The ledger's `goals`/`decisions` sync is a separate audit-only gap, not touched here.

Closes #1953, #1954, #1955.


ymote:
## Adversarial review (codex) + resolution

Ran an independent adversarial review of the two hot-path changes (`spawn_tool_task`, `run_standalone_turn`). It surfaced 6 code-grounded findings. Resolution:

### Fixed in this PR (commit `57515f9`)
- **High #1 — peer context dropped.** `spawn_tool_task` built the foreground ctx via `..ToolContext::zero()`, dropping `goal_id`/`task_id`/`originator_session` that peer boot sets on the Agent. So a peer's `goal_get` took the session path (its own goal-less session → `status: none`) and `goal_update` missed the peer-reject branch. Now threaded through. Verified `None` for every master agent (the only `Agent::with_goal_id` is the peer-boot guard; masters carry the goal via `goal_context`), so master completion is unaffected. **This also corrects the earlier commit's rationale for *not* threading `goal_id`.**
  - *Soak-validated:* a peer instructed to call `goal_get` now reports `Objective: "Compute 12 times 12…"`, `Status: active` (previously `none`); master still completes (chip `✔`).
- **High #3 — chip repaint read a racy scope.** The interactive repaint re-resolved the cwd scope at turn *end* from the process-global last-writer-wins `goal_scopes` map; a second folder sharing the wire id that activated mid-turn would make it read *that* folder's goal. Now pins the scoped key at turn *start* (mirrors how the autonomous path pins `goal_ctx.goal_session_key`).
- **Med #4 — `_main` profile.** The emit defaulted the profile to `""`; a profile-less session's goal is stored under `_main` → never matched. Now uses `usage_profile_id` (runtime-resolved).

### Deferred to follow-ups (with rationale)
- **High #2 + Med #6 → #1958.** The verifier's now-real LLM call runs inside `spawn_tool_task`'s `tokio::spawn` without the turn's routing/lane/fail-fast task-locals, and its usage is uncounted. A clean fix is cross-layer (the lane/session task-locals live in octos-cli; `spawn_tool_task` is in octos-agent) — out of scope for a soak fix. Practical impact is a 200-token sub-call losing FailFast + a narrow failover-misattribution race.
- **Med #5 → #1959.** `SessionGoalUpdated` carries no ordering/generation token, so a stale repaint can overtake `SESSION_GOAL_CLEAR`. Pre-existing pattern (the autonomous emit shares it); best closed centrally with a generation token.

Codex also verified several non-issues: no other tool reads `ToolContext.llm_provider`; `self.llm` doesn't recurse into the Agent; the two post-turn blocks are mutually exclusive.

ymote:
## Update: #1958 and #1959 now fixed here (not deferred), + two more soaks

The earlier comment deferred #1958/#1959 to follow-ups. They're now fixed in this PR (commit `3143899c9`). **This PR also Closes #1958 and #1959.**

### #1958 — verifier routing context + usage accounting
- The completion verifier's `provider.chat()` runs inside `spawn_tool_task`'s `tokio::spawn`, which doesn't inherit task-locals. The layering concern turned out not to apply: the routing task-locals live in **octos-llm** (reachable from octos-agent), not octos-cli. Capture the three (`current_llm_call_policy` / `current_lane_context` / new `current_router_context`) before the spawn and re-establish them around the tool future — so the verifier keeps fail-fast policy, lane filter, and originating-session attribution (no more unattributed failover / stray hedge).
- `run_goal_completion_verifier_with_usage` returns the call's `TokenUsage`; `goal_update` stamps it onto `ToolResult.tokens_used` so it folds into turn / goal-budget accounting.

### #1959 — goal-event ordering (self-contained, server-only)
- Stamp a monotonic `generation` on `SessionGoalUpdated`/`SessionGoalCleared`, bumped under the goals lock at event-build time — so a stale update always stamps *below* the clear that removed its goal.
- Gate both send boundaries (`send_notification_durable` for the RPC-derived clear, `send_notification_ephemeral` for the interactive update) on a per-session watermark, dropping any goal event whose generation doesn't exceed the last one emitted. No octos-tui change needed. Pure admit-logic unit-tested (2 tests).

### Soaks (both requested to prove the remaining claims)
- **Active-goal resume across restart:** created an active goal with a peer finding, killed the backend, relaunched → recovered as `paused` (the fleet-kernel's safe interrupt-recovery), fresh-backend `goal_get` returns paused **with** the ledger finding (genuine re-hydration), `/goal resume` → active → `goal_update(complete)` → chip `✔`. Survives *and* remains usable to completion.
- **Two concurrent peers:** one goal, two parallel `peer_handoff`s — both auto-bind to `goal_01`, run concurrently, both findings roll into the ledger, master `goal_get` reads both. No scoped-key collision, no lost findings.

91 goal-area unit tests + the 2 new guard tests pass; the soaks exercised the goal-event path end-to-end without regression.

ymote:
## Adversarial review round 2 (of `3143899c9`) → fixes in `9eed1cd81`

A skeptical pass over the two hot-path changes found the #1958/#1959 fixes incomplete/too-broad. It also **confirmed** the multi-connection-drop and reconnect-replay concerns are non-issues (the guard runs once on the direct send; fan-out and replay go through `send_ledger_event_durable`, bypassing it). Fixed:

- **#1959 — unstamped producers:** only `session_goal_updated_event_json` and `clear_goal` stamped the generation; `set_goal` and `charge_active_goal_tokens` shipped `0`, which the guard always admits → race stayed open through them. Added a shared `next_goal_event_generation()` helper and stamped all four.
- **#1958 — routing blast radius:** the wrap applied to every foreground tool, so a synchronous `spawn` child / unresolved-model pipeline node falling back to `self.llm` got pinned to the parent turn's lane (`CodeCapable` filter). Dropped `LANE` from the wrap; kept `CALL_POLICY` + `ROUTER_CONTEXT` (what the verifier needs; no lane-less-child regression).
- **#1958 — sentinel paths:** the autonomous sentinel verifier (run_standalone_turn + SessionActor) ran outside the turn's routing scopes; wrapped each in a reconstructed `RouterContext` for attribution.
- **#1959 — map growth:** bounded the process-global watermark map (cap 8192).

**Documented residuals (not regressions):** the server guard fixes the common stale-after-clear ordering but isn't atomic with socket enqueue (the robust fix is client-side generation honoring, which the stamp now enables); the watermark keys by wire id so cwd-scoped goals sharing a wire id follow the existing last-writer-wins chip semantics; sentinel verifier usage + `reasoning_tokens` remain uncounted downstream (pre-existing aggregation).

91 goal tests + 2 guard tests pass; clippy + fmt clean.

ymote:
## Client half of #1959: octoscode#521 (draft, blocked on this PR + a release)

The server-side #1959 fix here (generation stamp + per-session send guard) covers the common stale-update-after-clear ordering. The **airtight** completion — the client dropping any goal event whose generation is stale, closing the server guard's non-atomic-with-enqueue residual — is implemented and tested in **octos-org/octoscode#521**.

It's a draft, deliberately blocked: octoscode's `octos_release_pin_matches_cargo_core_rev` test refuses to pin an octos-core rev that isn't in a released backend. So the finalize sequence is: merge this PR → cut an octos release containing the `generation` field → bump octoscode's octos-core pin + `REQUIRED_OCTOS_CORE_REV`/`REQUIRED_OCTOS_RELEASE` to it → #521's CI goes green and the client-honor activates. The design degrades gracefully in the interim (`generation == 0` always applies), so nothing breaks before then.

ymote:
## Note: the `e2e-tool-regression` failure here is pre-existing CI-environment flakiness, not this PR

Evidence:
- The same job **fails on `a8dbdb865`** (this PR's base commit on main) with pure environment errors — Playwright's `chrome-headless-shell` binary missing, `ENOENT` on `workstreams/M15-agent-goal-loop-autonomy.md`, and `ECONNREFUSED` to the serve WebSocket. None relate to routing or goal generation.
- On this PR's *earlier* commits the job was **skipped**: its change-detection gate only runs when the diff touches `activate_tools|sandbox|shell.rs|tools/mod|session_actor|exec_env`. The round-2 `session_actor.rs` edit (the #1958 sentinel routing wrap) flipped the gate skip→run, so a pre-existing-flaky job executed for the first time on this branch and failed the way main does.
- All substantive checks are green (check, check-matrix, test-octos-agent lib+integration, test-octos-cli, dashboard, swarm-app, typos, ux-gate, windows-deploy-script).

Analogous to the check-arm infra-flakiness rule — not a code issue in this PR. (Confirming this branch's run shows the identical environment errors once its logs flush.)

PR #1985, opened 2026-08-12
fix(llm): support disabled reasoning effort

### LINKED ISSUE #1984: Support disabling reasoning effort
### Problem

`octos chat --effort` and `gateway.reasoning_effort` only accept `low`, `medium`, `high`, or `max`. Some OpenAI-compatible reasoning models, such as Qwen3.5 through Ollama, require `reasoning_effort: "none"` to return the final answer in `content`; otherwise `content` can be empty.

### Proposed change

Accept `none` in both the CLI and gateway config, then map it to each provider's reasoning-disable control where supported.

glight2000:
## Summary

- accept `none` in `octos chat --effort` and `gateway.reasoning_effort`
- map disabled reasoning to provider-specific controls, including `reasoning_effort: "none"` for OpenAI-compatible endpoints
- add provider, CLI, config, and documentation coverage

Fixes #1984

## Validation

- `cargo test -p octos-llm`
- `cargo test -p octos-cli --lib -- --skip bootstrap_with_dangerous_solo_permissions_disables_sandbox_and_uses_host_scope`
- `cargo clippy -p octos-llm --all-targets -- -D warnings`
- `cargo clippy -p octos-cli --lib`

The skipped CLI test is an existing Windows-only failure: it invokes the Unix `printf` command through `cmd.exe`. The remaining CLI library tests pass (1010 passed, 3 ignored).

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,7 +459,7 @@ fan-out non-blocking.
 | `--config <path>` | — | explicit flat config file — **wins over** `--profile` |
 | `--cwd <dir>` | current dir | workspace root the agent reads/writes |
 | `--data-dir <dir>` | `$OCTOS_HOME` / `~/.octos` | episodes / memory / sessions store |
-| `--effort <e>` | provider default | `low` \| `medium` \| `high` \| `max` (thinking models; others ignore it) |
+| `--effort <e>` | provider default | `none` \| `low` \| `medium` \| `high` \| `max` (`none` disables reasoning where supported) |
 | `--no-session-persistence` | off | ephemeral run (no episode saved); also **enables parallel agents** on one `--data-dir` |
 | `--max-iterations <n>` | `20` | per-turn tool-call cap |
 | `--no-retry` | off | disable automatic retry on transient LLM errors |
--- a/crates/octos-cli/src/commands/chat.rs
+++ b/crates/octos-cli/src/commands/chat.rs
@@ -137,8 +137,9 @@ pub struct ChatCommand {
     #[arg(long, value_enum)]
     pub ask_for_approval: Option<ChatApprovalMode>,
 
-    /// Reasoning effort for thinking models: `low`, `medium`, `high`, or
-    /// `max` (claude/codex parity). Overrides the config
+    /// Reasoning effort for thinking models: `none`, `low`, `medium`, `high`,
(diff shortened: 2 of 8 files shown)


BH3GEI (approved):
Reviewed the diff and verified locally on macOS (arm64) at a419b79:

- `cargo check --workspace` — clean; the new `ReasoningEffort::Disabled` variant breaks no cross-crate exhaustive matches
- `cargo test -p octos-llm` — 500 passed, 0 failed
- `cargo test -p octos-cli --lib` — 1026 passed, 0 failed, including the sandbox-scope test you had to skip (its failure is Windows-only, as you said; it passes here)
- `cargo clippy -p octos-llm --all-targets -- -D warnings` and `cargo clippy -p octos-cli --lib` — clean

Provider mappings look right:

- **Anthropic**: `Disabled` → budget 0 → `build_anthropic_thinking` returns `None` (below `MIN_BUDGET`), so the `thinking` param is omitted entirely rather than sending an invalid 0 budget.
- **OpenAI chat path**: the `(Disabled, style)` arms cover every `ReasoningStyle` before the enabled-effort arms, so the inner `unreachable!()`s are genuinely unreachable. Emitting `reasoning_effort: "none"` for the `Effort`/`None` styles only fires on explicit operator opt-in, so the default no-op behavior for unknown models is preserved.
- Toggle providers (GLM / DeepSeek) get `thinking: {"type": "disabled"}`; Kimi K3 correctly omits the control (no documented off value).

Non-blocking notes, fine as follow-ups:

1. The UI wire enum `ReasoningEffortLevel` (octos-core `ui_protocol.rs`) still only carries `low|medium|high|max`, so TUI `/thinking` and the WS turn param can't select `none` yet — this PR covers the CLI flag and gateway config only. Worth a follow-up issue if we want parity.
2. **Gemini**: `thinkingBudget: 0` disables thinking on Flash-tier models, but Pro-tier models reject 0 (thinking can't be fully disabled there). Explicit opt-in, and the provider error is visible to the operator, so acceptable — a short doc note could save someone the surprise.
3. **OpenAI Responses path**: `reasoning.effort: "none"` is accepted from gpt-5.1 onward; older reasoning models (gpt-5, o3/o4-mini) 400 on it. Same reasoning — explicit opt-in with a clear provider error.

Thanks for the thorough test coverage across all five provider paths.

PR #1997, opened 2026-08-13
feat(chat): durable goals in `octos chat` (--goals) — Phase 1

ymote:
## What

Phase 1 of bringing peer-agent goal to the core CLI, on top of the Phase 0 extraction (#1996). `octos chat --goals` registers `goal_create` / `goal_get` / `goal_update`, so a chat session can carry an **objective + token budget that survives the process**.

Before this, goals existed **only** in `octos serve` — `commands/chat.rs` had zero goal references.

## Three load-bearing pieces

1. **`configure_supervisor_store(<data_dir>/supervisor)` at startup** — goals (and pending continuations) rehydrate. A failure is non-fatal: goals degrade to in-memory-only rather than taking the session down.
2. **A stable per-profile session key** (`<profile>:cli:chat`). The goal tools resolve their session from `ToolContext::parent_session_key` and the orchestrator keys durable state by it — minting the *same* key every run is exactly what makes the goal outlive the process.
3. **Widening the profile tool surface.** Registering the tools was *not* enough: chat's default `coding` profile is an **allow list**, so `filter_by_profile` silently dropped them and the model never saw them — observed live as an identical tool count with and without `--goals`. The three goal tools are added to an allow-list surface when `--goals` is set. There's a regression test for exactly this.

## Opt-in

Without `--goals` the chat tool surface is unchanged (test asserts the flag defaults off).

## Not included (honest scope)

**Peers** (`peer_handoff`) and the **autonomous continuation loop**. Both need the multi-session host that only `octos serve` provides today — a one-shot/REPL chat cannot drive the async escalate → wake → respond cycle. That's Phase 3.

## Verified live (not just unit tests)

```
RUN 1: octos chat --profile dev --goals  ->  tools=21 (was 18)
       model called goal_create + goal_get; goal active, 0/50000

RUN 2: a SEPARATE octos chat process, goal_get only:
       Objective: chat-durability: Prove a chat goal survives separate octos chat invocations
       Status: active
       Tokens used vs budget: 0 / 50000
```

A brand-new process read back the goal created by the first — the durability claim, proven end to end.

Unit tests: stable/scoped session key, allow-list survival, flag defaults off. `cargo test -p octos-cli --lib` 1354 passed / 0 failed (unfeatured — goals now work without `--features api`, the Phase 0 payoff); clippy clean.

### DIFF
--- a/crates/octos-cli/src/commands/chat.rs
+++ b/crates/octos-cli/src/commands/chat.rs
@@ -111,6 +111,18 @@ pub struct ChatCommand {
     #[arg(long)]
     pub profile: Option<String>,
 
+    /// Enable the GOAL tools (`goal_create` / `goal_get` / `goal_update`) in
+    /// this chat session. A goal is a durable objective + token budget that
+    /// SURVIVES across `octos chat` invocations: state lives in the profile's
+    /// supervisor store, so a later `octos chat --goals` in the same profile
+    /// re-reads the same goal. Off by default — the tool surface is unchanged
+    /// unless you ask for it.
+    ///
+    /// Peers are NOT part of this: `peer_handoff` and the autonomous
+    /// continuation loop still require `octos serve`.
+    #[arg(long)]
+    pub goals: bool,
+
     /// FULL AUTONOMY ("yolo"): bypass all approvals AND the sandbox — the
     /// agent can edit any file and run any command with network access,
     /// without asking. Equivalent to `--sandbox danger-full-access`. Only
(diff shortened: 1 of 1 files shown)


ymote:
### Known scope decision: chat goals are per-PROFILE, not per-cwd

The session key is `<profile>:cli:chat` with no cwd component, and this PR does not call `configure_goal_scopes_sidecar`. So `octos chat --goals` in project A and project B under the **same profile** share **one** goal record.

That's deliberate for the MVP (a goal is an operator-level objective, and a stable key is what makes it durable — see the durability test), but it is a real behavioural choice worth flagging rather than shipping silently:

- **Pro:** you can walk away, come back in a different directory, and continue the same objective.
- **Con:** two unrelated projects under one profile will collide on a single goal; `goal_create` in project B would gate on project A's unfinished goal.

The orchestrator already has cwd-scoping machinery (`scoped_goal_key` + the goal-scopes sidecar) that `serve` uses, so adding per-cwd scoping later is a contained follow-up — the trade-off is that a goal would then only resume in the directory it was created in. Happy to switch if per-cwd is the preferred default.

Issue #435, opened 2026-04-17
Add generic harness lifecycle hooks for turn-end, resume, and spawn completion


ymote:
## Problem

Octos has hooks around tool and LLM calls, but the runtime still lacks the lifecycle surfaces where actual product enforcement belongs.

Missing events include:
- `on_workspace_init`
- `on_turn_start`
- `on_turn_end`
- `on_resume`
- `on_spawn_verify`
- `on_spawn_complete`
- `on_spawn_failure`
- `on_publish`

Without these runtime events, important behavior stays trapped in prompts or app-specific code.

## Goal

Expose a generic harness lifecycle that runtime can use to run deterministic policy-backed actions at the right time.

## Scope

- add the missing lifecycle events in one coherent interface
- wire them into the existing runtime instead of building a parallel orchestration layer
- keep enforcement in runtime, not in prompts

## Constraints

- additive only
- reuse existing session actor / task supervisor / workspace policy seams
- do not duplicate the hook system with a second event mechanism

## Likely files

- `crates/octos-agent/src/hooks.rs`
- `crates/octos-agent/src/task_supervisor.rs`
- `crates/octos-cli/src/session_actor.rs`
- `crates/octos-agent/src/workspace_git.rs`

## Acceptance

- lifecycle can express turn-end and spawn-complete actions
- failures can block completion or inject steering back to the session
- resume path has a defined hook surface
- runtime tests cover event ordering and failure behavior

ymote:
Closing as completed on shipped `main` @ `8ea539d09ef3e08f7fb622d85b374197a5c785a2`.

Generic lifecycle hook surfaces are now present in the runtime:
- hook definitions in `crates/octos-agent/src/hooks.rs`
- session/runtime emissions in `crates/octos-cli/src/session_actor.rs`
- spawn verify/complete/failure integration in `crates/octos-agent/src/tools/spawn.rs`

The `d9b2dfa` / `4fe36b8` sequence plus follow-up tests closed the remaining design and failure-path gaps.

PR #2028, opened 2026-08-15
fix(peers): re-arm fleet synthesis when a peer delivers a new round (#2024)

### LINKED ISSUE #2024: fix(peers): fleet synthesis stamp is per-master, so a second wave can never synthesize
## Summary

A master that has already auto-synthesized its peer fleet **can never synthesize a second wave** until every peer from the first wave is closed. Asking a master to "dig deeper" — the single most natural follow-up after reading a fleet report — stages a fresh wave that silently never fires.

Observed live on 2026-08-13 in a research session: 42 peers across 3 waves. Wave 1 failed (4 peers closed with no `result.md`), wave 2 produced a full report and synthesized, wave 3 was staged on the user's "dig deeper" request and appeared **stuck** — the master looked dead for an extended period. It only resumed once the last peer closed, which emptied the owned set and cleared the stamp.

The user's report was verbatim: *"master agent did finish two rounds of peer agents … i asked it to dig deeper and it launched more agents and now stuck, why"*, then *"now it is moving"* — the transition coinciding exactly with the fleet reaching zero owned peers.

## Root cause

`evaluate_peer_fleet_synthesis` (`crates/octos-cli/src/api/ui_protocol.rs:14383`) keys the fire-once stamp **per master**, and says so explicitly (`:14397`):

```rust
// Already synthesized this fleet → EXACTLY ONCE: never re-fire while any
// owned peer exists (added peers / newer results do not re-arm).
if stamp_exists {
    return FleetSynthesisDecision::Hold;
}
```

The stamp (`.synthesized-<safe_filename(master)>`, `:14423`) is cleared only by `ClearStamp`, which requires `peers.is_empty()` (`:14390`). So the reset condition is "the master's entire owned fleet is retired" — not "the wave that was synthesized is retired".

The consequence: once wave N synthesizes, waves N+1, N+2 … are all `Hold`ed, because the wave-N peers are still owned. There is no partial reset. The comment's "added peers … do not re-arm" is precisely correct about intent and precisely the bug in effect: it was written to stop a *newly-completed straggler in the same wave* from re-firing, but it also blocks a *deliberately-launched new wave*.

## Why this is distinct from #2003

Both produce "master looks dead after peers finish", but they are different defects:

| | #2003 (fixed) | this |
|---|---|---|
| trigger | last peer lands while master is **busy** | master has **already synthesized once** |
| gate returns | `Fire`, but only the peer-terminal edge evaluated it, and `Hold` was dropped with no re-arm | `Hold`, correctly, per documented semantics |
| fix | evaluate on the master-idle edge too | re-key the stamp so a new wave re-arms |

#2003's fix (evaluating the gate from the master's terminal edge as well) does **not** help here — the extra evaluation still reads `stamp_exists == true` and still returns `Hold`. This is a state-keying bug, not a missing-edge bug.

## Repro

1. Stage a peer wave on a master, let it complete, observe auto-synthesis (`.synthesized-<master>` appears under `peers/`).
2. Without closing any wave-1 peer, stage a second wave on the same master.
3. Let every wave-2 peer complete with a `result.md`.
4. Synthesis never fires. `evaluate_peer_fleet_synthesis` returns `Hold` on every evaluation.
5. Close all peers → stamp cleared → the gate re-arms.

The existing unit test `evaluate_peer_fleet_synthesis_fires_once_and_resets_on_clear` (`ui_protocol_tests.rs:9485`) asserts exactly this behaviour as *correct*, so it must be updated alongside the fix — it currently locks the bug in.

## Fix direction

Key the stamp to a **wave**, not a master, so a fresh wave re-arms without requiring a full fleet drain. Options, roughly in increasing order of invasiveness:

1. **Stamp the synthesized peer set.** Write the slug set (or a hash of it) into the stamp rather than relying on existence alone. `Fire` when the current owned-and-done set is not a subset of the synthesized set. This keeps one file and makes "added peers" re-arm — which is the behaviour the current comment explicitly forbids, so the intent change needs to be deliberate.
2. **Per-wave stamp id.** Give each staging batch a wave id (a ULID minted at `peer_handoff` fan-out), stamp `.synthesized-<master>-<wave>`, and evaluate per wave. Cleanest semantically; requires threading the wave id through staging and the blackboard.
3. **Clear on stage.** Remove the stamp whenever a new peer is staged for a master that already has one. Smallest diff, but it re-opens the straggler case the exactly-once rule was built to prevent — a peer that completes late in wave 1 after wave 2 is staged would re-fire wave 1's synthesis.

(2) is the principled fix; (1) is the pragmatic one. (3) is not recommended.

## Additional note — silent `Hold` is the recurring shape

The user-visible symptom is identical to #2003 because the failure mode is identical: a gate returns "not yet" and **nothing is recorded anywhere the user or the agent can see**. The master simply sits. Whatever the fix, a `Hold` that persists across many evaluations for a fleet whose peers are all done deserves at minimum a `tracing::info!` naming the reason, and arguably a human-sink event (#2019). This is the third defect in this family; the class is "a negative gate decision is indistinguishable from no work existing".

ymote:
Closes #2024.

## The bug

The per-master `.synthesized` stamp was a bare existence marker. Once a master had written up its peers, `evaluate_peer_fleet_synthesis` returned `Hold` for as long as any owned peer existed:

```rust
if stamp_exists {
    return FleetSynthesisDecision::Hold;   // forever
}
```

So the exactly-once unit was the **fleet**. A master that put the same peers through a second round of work got no second write-up unless every peer was closed first. That is the stall behind #2024: a live fleet did more work, and the synthesis that was supposed to summarize it never fired.

## The fix

The stamp now records the highest round of each peer already summarized, which makes the unit the **round**:

| peer state | recorded | delivered | decision |
|---|---|---|---|
| new peer, first result | — | 1 | Fire |
| peer did another round | 1 | 2 | Fire |
| nothing new | 2 | 2 | Hold |

One peer with fresh work is enough; a fleet with nothing new holds. The settled / done / idle preconditions are still checked **before** the freshness test, so one peer's new round never fires a synthesis while a sibling is mid-turn.

A peer's round is its count of versioned `result-<n>.md` files (which have existed since #435), floored at 1 whenever a bare `result.md` is present — otherwise a peer whose result predates versioning would read as round 0, never exceed a mark of 0, and never synthesize at all.

## Back-compat

A stamp written before this change is a bare unix timestamp. It proves the fleet *was* summarized but not at which rounds, so it reads as **"every currently-owned peer is covered at its current round"**. Upgrading therefore never re-fires an already-summarized fleet, while a genuinely new round still re-arms.

Any unreadable, oversized, or header-less stamp is read the same way — fail-closed, because mistaking a real stamp for "nothing summarized" would re-synthesize work already written up.

## Format

```
octos-peer-fleet-marks v1
2 alpha
1 two words
```

Round first, so a slug containing spaces survives the split. Written atomically and wholesale, so a peer dropped from the fleet does not linger in the record. Reads are size-capped.

`collect_owned_peer_results` now returns a named `OwnedPeer` instead of a `(String, bool)` tuple, so the compiler enumerates every call site when a field is added.

## Tests

Five added (553 → 558 `#[test]` in the file; no deletions):

- `a_new_peer_round_re_arms_fleet_synthesis` — the bug, plus proof the busy-master / mid-turn / resultless holds still win over freshness
- `a_legacy_peer_fleet_stamp_reads_as_every_current_round_summarized` — upgrade does not re-fire
- `peer_fleet_synthesis_marks_round_trip_through_the_stamp_file` — absent ≠ legacy; rewrite drops stale slugs
- `peer_fleet_synthesis_marks_survive_a_slug_with_a_space`
- `peer_result_round_counts_versions_and_floors_a_legacy_result_at_one`

Every name contains `peer` so CI's existing `cargo test -p octos-cli --features api peer` step actually selects them. Three of them did not, as first written — they would have compiled and never run, which is the same failure class as #2025 and the reason for the CI-coverage follow-up.

One existing assertion changed rather than being added to: the case that asserted "marker exists → Hold even with more peers all done" documented the bug. It is replaced by the same guarantee expressed per-round (all peers at their summarized round → Hold), and `peer_fleet_synthesized_stamp_is_an_existence_marker` is renamed to `peer_fleet_synthesized_stamp_file_lifecycle` since it only ever covered the file, not what the file says.

## Gates

- `cargo test -p octos-cli --lib` — 1448 passed
- `cargo test -p octos-cli --features api peer` — 132 passed
- `cargo clippy -p octos-cli --all-targets --no-default-features --features api,...` — clean

Not yet live-soaked against a real multi-round fleet.

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol.rs
+++ b/crates/octos-cli/src/api/ui_protocol.rs
@@ -14868,33 +14868,58 @@ async fn evaluate_and_enqueue_fleet_synthesis(
             orchestrator.clear_peer_fleet_synthesis_claim(&master_key);
         }
         FleetSynthesisDecision::Fire => {
-            // Advance the per-peer marks FIRST (durably, atomically). If the
-            // write fails, do NOT enqueue: a fire without a persisted mark would
-            // re-fire the same rounds on the next terminal. Concurrent terminals
-            // writing the same marks are safe (same observed rounds, atomic
-            // replace), and the per-master dedupe key collapses the paired
-            // enqueue.
-            if let Err(err) = write_peer_fleet_synthesis_marks(peers_root, master, &fired_marks) {
-                tracing::warn!(
-                    ?err,
-                    "peer fleet synthesis: .synthesized marks write failed; not enqueuing"
-                );
-                return;
-            }
+            // ENQUEUE FIRST, advance the marks only if it actually queued.
+            //
+            // The synthesis continuation's dedupe key is STABLE PER MASTER
+            // (`external/<kind>/<master>`), so a second synthesis lands as a
+            // duplicate whenever the first is still pending or was claimed
+            // inside `RECENT_CLAIM_GUARD_WINDOW`. Advancing the marks before
+            // knowing that would record "round N summarized" for a turn that
+            // never ran, and nothing would ever retry it — the #2024 gate fix
+            // alone would then still lose the second synthesis, silently.
+            //
+            // Holding the marks back makes it self-healing instead: the gate
+            // still reads Fire on the next edge, and the round-1 synthesis
+            // turn's OWN terminal is that edge (the master-idle re-evaluation
+            // added in #2018), by which point the guard window has moved on.
+            //
+            // The inverted order is safe here precisely because the queue is
+            // in-memory: a crash between enqueue and mark-write loses the
+            // continuation too, so re-firing on restart is correct, not double
+            // work. Concurrent terminals are safe as well — the dedupe key
+            // collapses the paired enqueue, and the marks write is an atomic
+            // whole-record replace of the same observed rounds.
             let outcome = orchestrator.enqueue_peer_fleet_synthesis_continuation(
                 &master_key,
                 profile_id,
                 &owned_slugs,
                 peers.len(),
             );
-            if outcome.queued().is_some() {
+            if outcome.queued().is_none() {
                 tracing::debug!(
                     master = %master_key,
                     profile = profile_id,
-                    peers = peers.len(),
-                    "peer fleet complete — enqueued ONE autonomous synthesis turn on master"
+                    "peer fleet synthesis deduped against an in-flight synthesis; \
+                     marks NOT advanced so the next terminal retries"
+                );
+                return;
+            }
+            if let Err(err) = write_peer_fleet_synthesis_marks(peers_root, master, &fired_marks) {
+                // The turn IS queued; only the record failed. Log loudly — the
+                // cost is a repeated synthesis, not a lost one.
+                tracing::warn!(
+                    ?err,
+                    master = %master_key,
+                    "peer fleet synthesis: marks write failed AFTER enqueuing; \
+                     this fleet may synthesize again"
                 );
             }
+            tracing::debug!(
+                master = %master_key,
+                profile = profile_id,
+                peers = peers.len(),
+                "peer fleet complete — enqueued ONE autonomous synthesis turn on master"
+            );
         }
     }
 }
--- a/crates/octos-cli/src/api/ui_protocol.rs
+++ b/crates/octos-cli/src/api/ui_protocol.rs
@@ -14334,6 +14334,21 @@ fn peer_results_ready_note(peers_root: &Path, session_id: &SessionKey) -> Option
     ))
 }
 
+/// One peer of a master's fleet, as read off the blackboard by
+/// [`collect_owned_peer_results`]. A named struct rather than a tuple so the
+/// compiler enumerates every call site when a field is added.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+struct OwnedPeer {
+    slug: String,
+    /// True when this peer has a `result.md` on the blackboard.
+    has_result: bool,
+    /// Which round of work this peer has DELIVERED: its count of versioned
+    /// `result-<n>.md` files, floored at 1 whenever a bare `result.md` exists
+    /// (pre-#435 peers wrote no versioned file and would otherwise read as
+    /// round 0 forever, and so never synthesize). `0` means no result yet.
+    round: u32,
+}
+
 /// Peer-fleet auto-synthesis — one owned peer's readiness inputs for the
 /// [`evaluate_peer_fleet_synthesis`] decision.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


ymote:
**Scope note on the evidence, so the record is accurate.**

The bug is established from the code and from the test that asserted it — the pre-change gate returned `Hold` unconditionally whenever the stamp existed and any owned peer remained, and the old test encoded exactly that ("marker exists → never re-fire while any owned peer lives"). That is not in doubt.

What I *cannot* show is the live fleet that prompted #2024 sitting in the blocked state. I went to look and the evidence is gone by design:

- all 46 peers in that fleet are now `closed`
- so the owned set is empty, and the gate reads `ClearStamp` / `Hold` — nothing to synthesize
- and no `.synthesized-*` stamp remains, because `reset_peer_fleet_synthesis_if_cleared` drops it on the final `peer_close`

A fully-retired fleet leaves no trace of whether it was ever blocked. So this PR is justified by the logic, not by a post-hoc reconstruction of that specific stall, and it stays un-soaked until a real multi-round fleet runs against it.

The soak that would actually prove it: one master, one peer, let it deliver a result and synthesize, then `peer_send_input` a second round and confirm a **second** synthesis fires without closing the peer. Worth running before this is relied on.

ymote:
## The soak found a second bug — in this PR

Before merging I wrote the two-round test I said would be the real proof. It failed, and the reason matters: **the gate fix alone would still have lost the second synthesis, silently.**

`enqueue_peer_fleet_synthesis_continuation` dedupes on a key that is stable per master and never changes:

```rust
fn peer_fleet_synthesis_dedupe_key(master_session: &SessionKey) -> String {
    format!("external/{PEER_FLEET_SYNTHESIS_EXTERNAL_KIND}/{master_session}")
}
```

The existing test spells out the intent — *"There is no re-arm — one synthesis per fleet."* That was consistent with the old existence-marker gate. It is not consistent with a gate that re-arms.

So the sequence was:

1. round 2 lands, gate correctly says `Fire`
2. marks written — "round 2 summarized"
3. enqueue returns **duplicate** (round 1's continuation is still pending, or was claimed inside the 30s guard)
4. nothing queued, and nothing re-arms, because the marks already advanced

A synthesis dropped with no retry and no log. Same shape as everything else in this thread: a negative outcome that produces no signal.

**Fix:** enqueue first, advance the marks only when it actually queued. A deduped enqueue leaves the marks alone, so the gate still reads `Fire` on the next edge — and the round-1 synthesis turn's *own* terminal is that edge, via the master-idle re-evaluation from #2018, by which point the guard window has passed. Self-healing instead of silent.

Inverting the order is safe here because the continuation queue is in-memory: a crash between enqueue and mark-write loses the continuation too, so re-firing on restart is correct rather than duplicated work. A mark write that fails *after* a successful enqueue is now logged loudly and costs a repeated synthesis, never a lost one.

**The new test is verified to fail under the old ordering:**

```
assertion `left == right` failed: a deduped enqueue must NOT record round 2 as
summarized — otherwise the second synthesis is lost with nothing left to retry it
  left: 2
 right: 1
```

It drives the real `evaluate_and_enqueue_fleet_synthesis` across two rounds and asserts the on-disk marks, because those are what the gate reads next time.

Gates re-run: 1448 default-feature, 133 `peer`, clippy clean, fmt clean.

Worth noting for the record: five unit tests on the pure decision function all passed against the broken version. Only driving the real path caught it.

Issue #2029, opened 2026-08-15
CI: api-feature tests run only if their name matches a hand-written filter — three have silently not run


ymote:
## Summary

Tests behind the `api` feature only run in CI if their **name** happens to match one of a handful of hand-written filters. Nothing checks that. A test can compile, be committed, pass review, and never execute — and the job it "belongs to" still reports green.

This has happened three times in the last day. It is not a flake; it is the default outcome for any new `api` test whose name does not contain the right substring.

## Why it happens

`api` is not a default feature, so `cargo test -p octos-cli --lib` never compiles `src/api/`. Coverage of that module comes entirely from named steps in `.github/workflows/ci.yml`:

```yaml
cargo test -p octos-cli --features api api::auth_handlers
cargo test -p octos-cli --features api spec_section6_catalog_lists_every_advertised_method
cargo test -p octos-cli gateway_runtime::tests --features api
cargo test -p octos-cli --features api session_open_snapshot
cargo test -p octos-cli --features api peer
cargo test -p octos-cli --features api -- completion_evidence goal_update_folds_ledger
cargo test -p octos-cli --features api background_activity
```

Each is a substring filter. Their union is far smaller than the module — `ui_protocol_tests.rs` alone has 558 tests, and the `peer` step selects 132 of them. Everything outside the union is dead weight: compiled by clippy, executed by nothing.

The failure is silent in the worst way. A filter that matches nothing does not error:

```
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 2774 filtered out
```

`ok`. Exit 0. Green check.

## The three instances

- **#2025** — two tests matched no filter at all. Found only because the sink they touched broke something else.
- **#2028** — three of five new tests did not contain `peer` as first written. Caught by hand, before push, by explicitly checking each name against the filter list. Nothing would have caught it otherwise.
- **#2015** — same shape one level up: the e2e job reported SUCCESS across 14 consecutive runs while skipping every step it was supposed to run.

The pattern across all three: **a gate reports success for work it never performed.** It is the same shape as the silent `Err` discard in the persist loop and the synthesis gate returning `Hold` with no log — a negative outcome that produces no signal.

## Proposal

**A test that asserts the filters cover the module.** Enumerate the test names compiled under `--features api`, enumerate the filters CI applies, and fail when a test matches none of them.

```
error: 4 api tests are not selected by any CI filter:
[...]

PR #2030, opened 2026-08-15
test(session-actor): give the continuation-tick test a unique agent id (#2029)

ymote:
Part of #2029. One of the two flaky tests identified there.

## The failure

`master_continuation_tick_reenters_actor_loop` failed in **roughly one full parallel run in three**, while passing alone, passing under every CI filter, and passing under `--test-threads=1`. A failure that only appears in a wide parallel run — and moves between runs — is interference, not logic.

## The cause

The orchestrator's agent registry is process-global and keyed by `agent_id` **alone**. The session is a field on the record, not part of the key:

```rust
let entry = state
    .agents
    .entry(upsert.agent_id.clone())     // session NOT in the key
    .or_insert_with(|| AutonomyAgentRecord { .. });
```

**18 test sites hardcode `agent_id: "child-a"`.** A sibling upserting it with a different `session_id` replaces this test's record. The periodic tick then finds no completed child for *this* session, nothing drains, and the assertion fails:

```rust
assert!(provider.call_count.load(Ordering::Relaxed) > 0,
    "periodic actor tick must drain queued child completion into process_inbound");
```

Everything else about the test was already unique — the session key is derived from its `TempDir`. Only the agent id wasn't, and that's the part that keys the map.

## The fix

Derive the agent id from the same `TempDir` name.

Uniqueness is **unilateral**: a unique id cannot be clobbered no matter what the other seventeen sites do. So this needs no coordinated change and no churn across all 18.

## Verification

| | failures |
|---|---|
| before (recorded on #2029) | 3 of 8 full parallel runs |
| after | **0 of 4** |

Plus clippy (CI-exact, all features) and fmt clean.

## Scope

Deliberately does **not** fix the other two known flakes — `closed_peer_park_produces_neither_wake_nor_escalation_row` and `second_serve_on_same_data_dir_is_refused_with_a_greppable_marker`. Those are different globals and still reproduce; one of them appeared in run 1 of the verification above.

The class-level fix — putting the session in the registry key, so two sessions each having a `child-a` is normal rather than a conflict — is tracked on #2029. It changes production keying and shouldn't ride along with a test cleanup.

### DIFF
--- a/crates/octos-cli/src/session_actor_tests.rs
+++ b/crates/octos-cli/src/session_actor_tests.rs
@@ -2442,13 +2442,31 @@ async fn master_continuation_tick_reenters_actor_loop() {
         setup_actor_with_mode(provider.clone(), QueueMode::Followup, None, false, &dir).await;
     let session_id = test_session_key(dir.path());
 
+    // #2029: the orchestrator's agent registry is process-global and keyed by
+    // `agent_id` ALONE — the session is a field on the record, not part of the
+    // key. Eighteen tests hardcode `child-a`, so a sibling running concurrently
+    // overwrites this record's session_id/status, the tick finds no completed
+    // child for THIS session, and the assertion below fails. It passed under
+    // `--test-threads=1` and under the narrow CI filters, and failed in roughly
+    // one full parallel run in three.
+    //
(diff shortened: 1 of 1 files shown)


anantheparty (approved):
审查通过（2026-08-15 巡检）。

PR #2034, opened 2026-08-16
fix(peers): keep a live peer out of the per-turn orphan sweep, and cover the wiring

### LINKED ISSUE #2035: peers are enrolled as supervised tasks, then reaped as "orphaned across restart" 1ms later — every supervised view of a peer is wrong
A peer registered by `peer_handoff` is marked `Failed("orphaned across restart")` about **one millisecond after it is registered**, while the peer goes on to run normally for minutes and produce real work. Every supervised view of a peer — `agent/list`, liveness, cascade, and the terminal hook that wakes the master — therefore sees a peer that is dead from the moment it is created.

Found in a live multi-peer peer-goal soak on `80cec9254` (= current `main`), macOS arm64, profile on moonshot-coding/k3, driven over AppUI stdio.

## Evidence

Fleet-kernel event ledger (`<instance-data-dir>/supervisor/supervisor-events.jsonl`), times relative to the first new row of the run:

```
13.753s cont_queued     7ca3b82f70f9 group=agent-group:soak:soak:local:goal:master
13.753s group_reg       agent-group:soak:soak:local:goal:master
13.753s child_started   7ca3b82f70f9 label=peer_handoff
13.754s child_terminal  7ca3b82f70f9 kind=failed          <-- 1 ms after start
13.754s cont_queued     -            group=agent-group:soak:soak:local:goal:master
13.754s cont_queued     792eda86f358 group=agent-group:soak:soak:local:goal:master
13.755s group_reg       agent-group:soak:soak:local:goal:master
13.755s child_started   792eda86f358 label=peer_handoff
13.755s child_terminal  792eda86f358 kind=failed          <-- 0 ms after start
13.756s cont_queued     -            group=agent-group:soak:soak:local:goal:master
```

Both `child_terminal` payloads carry `reason: "orphaned across restart"`, and the group's `objective` and the child's `task` are that same string — it is the reap's reason text, not anything the caller supplied.

`agent/list` on the master, sampled after both peers had finished real work:

```json
{
  "agent_id": "task-soak-local-goal-child-01a008cf-c911-7541-9473-792eda86f358",
  "parent_agent_id": "master",
  "role": "background_task",
  "nickname": "peer_handoff",
  "backend_kind": "spawn_child_session",
  "status": "failed",
  "last_task": "orphaned across restart",
  "created_at_ms": 1786854236935,
  "updated_at_ms": 1786854236937
}
```

Both peers were alive and working the whole time. From the same run's transcript, `peer-aud-auth` and `peer-aud-parse` each started a turn at `04:23:5x`, ran for roughly a minute, completed normally, and each wrote a real finding to the goal ledger (`created_by=peer:aud-auth` / `peer:aud-parse`). The supervisor said `failed` at `04:23:56`; the peers finished successfully at `04:24:5x`.

## Mechanism

The orphan sweep inside `TaskSupervisor::enable_persistence` reaps every persisted row that is non-terminal and not in the process-global live-set:

```rust
// crates/octos-agent/src/task_supervisor.rs:1479
.filter(|task| {
    !is_terminal_runtime_state(&task.runtime_state)
        && !is_task_live(&task.id)
        && !is_live_pipeline_child(task)
})
// :1494
self.mark_failed(task_id, "orphaned across restart".to_string());
```

The live-set exists precisely to stop this false reap, and its own doc comment describes the trap (`task_supervisor.rs:720-750`): the WS turn path builds a brand-new per-turn `TaskSupervisor` every turn and calls `enable_persistence` every turn over the shared per-session ledger, so "non-terminal ⇒ no live worker" is only true at a genuine cross-process restart.

Membership in that live-set is owned exclusively by `TaskTerminalGuard::new` (`task_supervisor.rs:3657`), which is constructed around a **detached `tokio::spawn` worker**.

A peer has no such worker. `peer_handoff` only *stages* the peer; the client boots it as a sovereign session. The registration is a bare `register` with no guard:

```rust
// crates/octos-cli/src/api/ui_protocol_transport.rs:29744
let task_id = peer_supervisor.register(
    "peer_handoff",
    &registry_key,
    Some(&event.session_id.0),
);
```

and by design it is retired only on `peer_close`, never on turn terminal (the comment above it says so explicitly, because a peer runs many turns). So a peer row is, structurally, permanently non-terminal and permanently absent from the live-set — exactly the shape the sweep is built to kill. The next `enable_persistence` over the shared per-session ledger reaps it.

## Consequences

1. `agent/list` reports every peer as `failed / "orphaned across restart"` for its whole life. This is the user-visible roster.
2. Each bogus terminal immediately queues a master continuation (the two `cont_queued … group=agent-group:…:master` rows at 13.754s and 13.756s, each following its `child_terminal` by under a millisecond). The peer→master wake introduced in #2024 fires on a peer that never stopped.
3. Anything keying liveness or cascade off the supervised row sees a dead child while the peer is running, which is the opposite of what #2023 enrolled peers to achieve.

The peer's own execution is unaffected — this is bookkeeping — but the bookkeeping is what the enrollment was for.

## Notes for the fix

- The natural fix is to give peers the same liveness the sweep already respects: mark the peer task live at registration and clear it at `peer_close` / peer terminal, rather than tying liveness to a detached tokio worker that peers do not have.
- An alternative is to exempt `peer_handoff` rows from the sweep, but that trades one wrong answer for another — a peer orphaned by a *real* cross-process restart should still be reaped.
- The existing peer registry tests construct a bare `TaskSupervisor` with no persistence, so no sweep ever runs and this is invisible to them. A regression test needs `enable_persistence` over a shared ledger with a second supervisor, mirroring the per-turn rebuild.
- PR #2034 (open) refactors this call site into `bind_peer_supervised_task` / `retire_peer_supervised_task`. It carries the same defect and its tests would still pass; the fix belongs on top of, or merged into, that PR.

## Reproduce

Stage a peer from a master session over AppUI stdio, let the master take one more turn, then read `<instance-data-dir>/supervisor/supervisor-events.jsonl` and call `agent/list`. The peer will be `failed` while its session is still running.

ymote:
Started as tests for the peer↔supervisor wiring. The live soak that followed showed the wiring was itself broken, so this now carries the fix as well.

## The defect (#2035)

A peer staged by `peer_handoff` was marked `Failed("orphaned across restart")` one turn after enrolment, while the peer ran normally for minutes.

From the run on mini5 (`80cec9254` = `main`), the supervisor's own persisted ledger:

```
04:23:50.289  peer_handoff aud-auth  status: spawned
04:23:56.935  peer_handoff aud-auth  status: failed   error: "orphaned across restart"
04:24:5x      …peer completes successfully and writes a real goal-ledger finding
```

`agent/list` reported both peers `failed` for their whole working life, and each bogus terminal queued a master continuation — the master spent a turn reasoning about "a supervised child … reported failed — orphaned across restart".

## Why

The orphan sweep in `TaskSupervisor::enable_persistence` reaps every non-terminal row absent from the process-global live-set. That set is populated only by `TaskTerminalGuard`, which wraps a detached `tokio` worker. A peer has no such worker — it is a sovereign session the client boots — and by design its row stays non-terminal until `peer_close`. Structurally reap-eligible.

## The fix

`TaskLivenessLease` — the liveness half of `TaskTerminalGuard`, with neither of the parts that make the guard wrong here:

- no `Arc<TaskSupervisor>`, which would pin a per-turn supervisor for the peer's whole life and defeat the `Weak`-based pruning in `SessionTaskQueryStore`
- no terminal transition on `Drop`, because a peer is retired by `peer_close`, not by a worker future ending

The peer task registry holds the lease for exactly the peer's supervised lifetime; `take` releases it on the close path. A re-stage resolving to the same task id keeps the existing lease rather than replacing it, so the live-set entry is never briefly cleared for a task that is still live.

It does not weaken the sweep. Membership is process-global and starts empty in a new process, so a peer left behind by a genuine cross-process restart still has no lease and is still reaped — covered by a test.

## Tests

The three original wiring tests stay: they cover bind/retire at the call site, which the pre-existing registry tests did not.

Two more cover the defect, and they had to be shaped differently — the reason this was invisible is that every existing peer test builds a bare `TaskSupervisor` with no persistence, so no sweep ever runs. The new ones drive the real per-turn shape: a second supervisor over the same ledger.

- `a_staged_peer_survives_the_per_turn_orphan_sweep`
- `peer_retirement_releases_the_liveness_lease`
- plus three in `octos-agent` for the primitive, including the restart case

Mutation-checked twice. Removing the `bind` call fails all three original tests while the four pre-existing registry tests pass. Removing the liveness claim fails exactly the two new peer tests and leaves the seven others green.

## Gates

`cargo test -p octos-agent --lib` — 2565 pass. `cargo test -p octos-cli --features api --lib` — 2926 pass single-threaded. fmt clean; clippy clean on `octos-agent --all-targets` and on the CI-exact `octos-cli` feature set.

Two failures in the full `octos-cli` suite are pre-existing and unrelated, both reported on #2029: `global_drain_spawns_before_the_stdio_early_return` fails on pristine `main` (its source-scan needle went stale in the #1728 rename and CI's filters never run it), and `closed_peer_park_produces_neither_wake_nor_escalation_row` is the known parallel-contention flake that passes with `--test-threads=1`.

Closes #2035.

### DIFF
--- a/crates/octos-agent/src/lib.rs
+++ b/crates/octos-agent/src/lib.rs
@@ -188,8 +188,9 @@ pub use swarm::{
 };
 pub use task_supervisor::{
     BackgroundTask, RelaunchOpts, RelaunchRequest, SpawnOnlyFailureSignal, TaskCancelError,
-    TaskCancelToken, TaskLifecycleState, TaskRelaunchError, TaskRuntimeState, TaskStatus,
-    TaskSupervisor, TaskTerminalGuard, TerminalEvent, TerminalOutcome, parse_alternatives,
+    TaskCancelToken, TaskLifecycleState, TaskLivenessLease, TaskRelaunchError, TaskRuntimeState,
+    TaskStatus, TaskSupervisor, TaskTerminalGuard, TerminalEvent, TerminalOutcome,
+    parse_alternatives, task_is_live,
 };
 pub use tools::{
     AskUserQuestionTool, BackgroundResultKind, BackgroundResultPayload, BrowserTool,
--- a/crates/octos-agent/src/task_supervisor.rs
+++ b/crates/octos-agent/src/task_supervisor.rs
@@ -3689,6 +3689,62 @@ impl Drop for TaskTerminalGuard {
     }
 }
 
+/// Whether `task_id` currently holds a liveness claim in this process — either
+/// a detached worker's [`TaskTerminalGuard`] or a [`TaskLivenessLease`].
+///
+/// This is what the orphan sweep consults, so it is also the honest way for a
+/// caller outside this crate to assert that it has (or has released) liveness
+/// for work it owns.
+pub fn task_is_live(task_id: &str) -> bool {
+    is_task_live(task_id)
+}
+
+/// A liveness lease for work that is genuinely in flight but is NOT driven by
+/// a detached `tokio` worker in this process — a peer session the CLIENT
+/// boots and drives being the motivating case (#2035).
+///
+/// The orphan sweep in [`TaskSupervisor::enable_persistence`] reaps every
+/// non-terminal row that is absent from the process-global live-set, on the
+/// premise that "non-terminal ⇒ no live worker". [`TaskTerminalGuard`] is what
+/// makes that premise true for detached workers, but it is the wrong tool
+/// here: it holds an `Arc<TaskSupervisor>` (which would pin a per-turn
+/// supervisor for the whole life of the peer and defeat the `Weak`-based
+/// pruning in `SessionTaskQueryStore`), and its `Drop` fires a terminal
+/// transition — whereas a peer is retired by `peer_close`, not by a worker
+/// future ending.
+///
+/// So this type carries the liveness half alone: mark on construction, clear
+/// on `Drop`, no supervisor reference and no terminal side effect. Hold it for
+/// exactly as long as the client-driven work is live.
+///
+/// It does not weaken the sweep. Membership is process-global and starts EMPTY
+/// in a new process, so a row left behind by a genuine cross-process restart
+/// has no lease and is still correctly reaped.
+pub struct TaskLivenessLease {
+    task_id: String,
+}
+
+impl TaskLivenessLease {
+    /// Mark `task_id` live until the returned lease is dropped. Idempotent —
+    /// re-leasing an already-live id is a no-op insert.
+    pub fn new(task_id: impl Into<String>) -> Self {
+        let task_id = task_id.into();
+        mark_task_live(&task_id);
+        Self { task_id }
+    }
+
+    /// The task this lease keeps live.
+    pub fn task_id(&self) -> &str {
+        &self.task_id
+    }
+}
+
+impl Drop for TaskLivenessLease {
+    fn drop(&mut self) {
+        clear_task_live(&self.task_id);
+    }
+}
+
 #[cfg(test)]
 #[path = "task_supervisor_tests.rs"]
 mod tests;
--- a/crates/octos-agent/src/task_supervisor_tests.rs
+++ b/crates/octos-agent/src/task_supervisor_tests.rs
@@ -3962,6 +3962,89 @@ async fn foreground_armed_guard_survives_sweep_before_worker_polls() {
     );
 }
 
+// ── issue #2035: liveness for CLIENT-DRIVEN work (peers) ──────────────
+
+/// Reproduction of #2035. A `peer_handoff` row is the shape the sweep was
+/// never designed for: it is non-terminal for the peer's whole life (it is
+/// retired on `peer_close`, not on turn terminal) and its worker is a
+/// sovereign session the CLIENT drives, so no `TaskTerminalGuard` is ever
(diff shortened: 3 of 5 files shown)


ymote:
Live soak on `80cec9254` found a defect this PR's tests do not catch, filed as #2035: a peer registered by `peer_handoff` is reaped as `Failed("orphaned across restart")` roughly 1 ms after registration, while the peer runs normally for minutes.

The cause is structural rather than in this refactor. The orphan sweep in `TaskSupervisor::enable_persistence` reaps any persisted row that is non-terminal and absent from the process-global live-set, and that live-set is populated only by `TaskTerminalGuard` around a detached `tokio::spawn` worker. A peer has no such worker — it is a sovereign session the client boots — and by design its row stays non-terminal until `peer_close`. So it is permanently reap-eligible.

This matters for #2034 specifically: `bind_peer_supervised_task` carries the same behaviour, and the three new tests still pass, because they construct a bare `TaskSupervisor` with no persistence, so no sweep ever runs. Mutation-testing the bind confirmed the tests cover the wiring, but the wiring is not where this fails.

I would rather fix it here than merge and follow up, since the helpers this PR extracts are exactly where the liveness marking belongs. Leaving the PR open for that.

Issue #1923, opened 2026-08-06
e2e-tool-regression reports SUCCESS whenever the path filter skips it — no cadence-independent run


ymote:
Two separate problems that compound: the gate is red on `main`, and its green is meaningless.

## 1. It fails on unmodified main

#1922 was a probe: `origin/main` (`725451747`) plus a **two-line comment** in `session_actor.rs`, enough to trip the path filter and nothing else.

Result: **52 failed, 6 passed.**

#1921 — which actually changes code — produced **52 failed, 6 passed**, the same specs in the same order. So the failure is entirely pre-existing and unrelated to any PR that happens to trip the filter.

Every failing assertion receives an empty string:

```
Expected substring: "OK"
Received string:    ""
```
```
RpcErrorImpl: rpc-error[-32602] profile 'admin' is not configured for this AppUI session
```

The job starts the server with `OCTOS_AUTH_TOKEN=ci-test-token ./target/release/octos serve --port 3000` and **no LLM provider configured**, then runs Playwright specs that assert on agent output. With no provider the agent returns nothing, so every content assertion fails. The 6 that pass are the ones not depending on agent output.

## 2. Its SUCCESS is a no-op

```yaml
if echo "$FILES" | grep -qE '(activate_tools|sandbox|shell\.rs|tools/mod|session_actor|exec_env)'; then
  echo "run=true" >> "$GITHUB_OUTPUT"
```

When the filter misses, **every substantive step is skipped — build, serve, npm ci, run — and the job still reports SUCCESS.** Example: #1919's `e2e-tool-regression` is green with all five steps `skipped`.

Across the last 25 runs in this repo, the only executions were #1921's and #1922's. Both failed. So the gate has been red for an unknown length of time while showing green.

This is the same shape as the `check-windows` `continue-on-error` problem fixed in #1885: a job whose status cannot fail is a job that guards nothing.

## Suggested fix

- Give the suite a working provider (a mock/stub backend is probably right for CI — these specs shouldn't need a live LLM), or split the provider-dependent specs out and only run those where credentials exist.
- Make the skip path **explicit** rather than a silent green — emit a neutral/skipped conclusion, or always run a fast subset so green means something.
- Consider whether the path filter is right at all: `session_actor` catches PRs that have nothing to do with the e2e surface, while genuine e2e-affecting changes elsewhere slip through.

Found while reviewing #1921; probe branch `ci/e2e-baseline-probe` (#1922, now closed).

ymote:
[...]

PR #2041, opened 2026-08-16
fix(test): repair the serve wiring guard and actually run it in CI

ymote:
See the commit message for the full account. Found while validating the peer-goal fixes with a real-world soak on merged main.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -443,6 +443,15 @@ jobs:
       - name: Peer staging + worktree fencing (octos-cli)
         run: cargo test -p octos-cli --features api peer
 
+      # `commands::serve::tests` matched NONE of the filters on this list, so
+      # its guards had not executed in CI since #1728 — long enough for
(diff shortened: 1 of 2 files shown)


ymote:
Holding this until #1923 is resolved — main should not merge a red e2e.

Converted to draft so it cannot be merged by accident. Nothing is wrong with the diff: 14 checks pass, and the two changes (repairing the stale `stdio_connection` needle, and adding a `commands::serve::tests` CI step) are verified green.

The single failure is `e2e-tool-regression`, and this PR is the reason it ran at all — the job is path-filtered, so it skips on ordinary PRs, and editing `ci.yml` brought it in. That is the behaviour #1923 describes: broken on main, reporting SUCCESS whenever it does not run. Evidence posted there.

**Unblocking order:** fix #1923, then this merges as-is. If it needs to land sooner, the `ci.yml` step can be split out — the needle repair alone does not trigger the e2e job — but that would leave `commands::serve::tests` unreachable by CI, which is half the point of the PR.

Issue #2031, opened 2026-08-15
e2e-tool-regression is deterministically red: Homebrew/ffmpeg missing from PATH on macos-14


ymote:
## Summary

`e2e-tool-regression` fails **44 of 58** tests, reproducibly, on a change that cannot affect it. Two consecutive runs of the same job produced identical results.

Observed on #2030 — a one-line, test-only, `#[cfg(feature = "api")]` Rust change that swaps a hardcoded string for a derived one. It has no path to the e2e harness.

## Evidence it is the runner, not the code

The failing set spans everything at once: the whole M9 protocol suite, `solo-login`, `web-client`, plus:

```
tests/tool-use-regression.spec.ts:67:5  › ffmpeg is reachable via shell PATH
tests/tool-use-regression.spec.ts:75:5  › ffmpeg version is functional
tests/tool-use-regression.spec.ts:89:5  › ffmpeg concat works in sandbox workdir
tests/tool-use-regression.spec.ts:204:5 › PATH includes /opt/homebrew/bin
```

A Rust test-name change cannot remove ffmpeg from `PATH` or unset Homebrew's directory. When a protocol suite, a login flow, and a binary-on-PATH assertion all fail together, the harness did not come up.

`PATH includes /opt/homebrew/bin` failing on `runs-on: macos-14` points at the runner image: Homebrew's bin is not on `PATH` for the step, so `ffmpeg` is unreachable and everything depending on the built server follows.

Two runs, same numbers:

| run | result |
|---|---|
| initial | 44 failed, 14 passed, 8 skipped |
| re-run of failed jobs | 44 failed, 14 passed |

## Why this went unnoticed

The job is gated:

```yaml
runs-on: macos-14
if: github.event_name != 'schedule' && needs.e2e-changes.outputs.run == 'true'
```

So it runs only when `e2e-changes` decides the diff is relevant. On #2028 — merged today — it reported `skipping`. Most PRs never trigger it, so a permanently-red gate stays invisible until some unrelated PR happens to touch a matching path and gets blocked by it.

That makes this the *inverse* of #2029: there, gates report green because they select nothing; here, a gate runs and its result reflects the runner's state rather than the code's. Same underlying defect — **the signal is not tied to what changed.**

## Suggested fix

[...]

PR #2043, opened 2026-08-16
fix(ci): enable the admin shell endpoint so the e2e ffmpeg specs can run (#2031)

ymote:
Makes `e2e-tool-regression` a check that can actually pass, and scopes it to the specs it is named for.

## Root cause

The four ffmpeg/PATH specs in `tool-use-regression.spec.ts` drive `POST /api/admin/shell`. That route is registered **only when `allow_admin_shell` is true** (`api/router.rs:825`). It defaults to false and nothing in CI ever set it, so every request answered:

```
HTTP 404 {"error":"not_found","path":"/api/admin/shell"}
```

`res.json()` on that body has no `exit_code`, which is precisely what the failures said:

```
expect(result.exit_code).toBe(0)
Received: undefined
```

`undefined`, not a non-zero exit code — an absent field from an absent route. These four specs could never have passed in CI as written.

## Correcting this PR's first commit

The first commit exported the Homebrew prefix into serve's environment, on the theory that the agent's sandboxed shell tool could not reach ffmpeg. **That was wrong, and the diagnostic it added disproved it in its own run** — serve's PATH already contained `/opt/homebrew/bin`, no warning fired, and the specs still failed.

The handler is also not sandboxed and does not filter PATH: it spawns `sh -c` in serve's own environment with only `BLOCKED_ENV_VARS` removed (`api/admin.rs:4107`), and `PATH` is explicitly retained by the sanitizer (`subprocess_env.rs:57`). The export is reverted here; it fixed nothing.

## Verification

Measured against a release binary, not inferred:

| invocation | result |
|---|---|
| `--config '{"allow_admin_shell": true}'` → `which ffmpeg` | `{"stdout":"/opt/homebrew/bin/ffmpeg","exit_code":0}` |
| … → `echo $PATH` | contains `/opt/homebrew/bin` |
| … → the concat script | `{"stdout":"CONCAT_OK","exit_code":0}` |
| no config (CI's current invocation) | `HTTP 404 {"error":"not_found"}` |

All four specs pass with the config; the 404 reproduces the exact CI symptom without it.

## Changes

1. **Write a CI serve config enabling `allow_admin_shell`** and pass `--config`. Full-RCE endpoint, so it is enabled here and nowhere else — ephemeral runner, loopback bind, admin token — and the comment says not to copy it.
2. **A precondition step** asserting the endpoint answers 200 before the suite runs, so a regression names itself instead of surfacing as four `undefined` assertions. Both branches exercised locally: 200 passes, 404 exits 1.
3. **Scope the required job to `tests/tool-use-regression.spec.ts`** (unchanged from the first commit). The unfiltered run swept in the whole tree — 44 failed / 14 passed.
4. **Move model-dependent specs to a nightly `continue-on-error` job** (unchanged). They assert exact model wording (`toContain('ALPHA')`, CJK output); a merge should not hinge on phrasing.

## Why this matters beyond the four specs

The spec named `PATH includes /opt/homebrew/bin` is exactly the assertion that would have caught the original #1923 regression — and it was structurally incapable of running. Same family as #2029, where a name filter matching nothing prints `ok`.

Unblocks #2048 and #2041, both currently parked behind this job.

Refs #2031, #1923

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1032,21 +1032,26 @@ jobs:
       - name: Install Homebrew ffmpeg (macOS)
         run: brew install ffmpeg
 
+      # #2031 — the four ffmpeg/PATH specs in `tool-use-regression.spec.ts`
+      # drive `POST /api/admin/shell`, and that route is registered ONLY when
+      # `allow_admin_shell` is true (api/router.rs). It defaults to false and
+      # nothing in CI ever set it, so every one of those requests answered
+      # 404 `{"error":"not_found"}`. `res.json()` on that body has no
+      # `exit_code` field, which is why the failures read
+      # `expect(result.exit_code).toBe(0) // Received: undefined` — an absent
+      # route, not a missing binary and not a PATH problem.
+      #
+      # This is a full-RCE endpoint, so it is enabled here and nowhere else:
+      # an ephemeral runner, bound to loopback, behind an admin token. Do not
+      # copy this config to anything long-lived.
+      - name: Write CI serve config (enables the admin shell endpoint)
+        run: |
+          echo '{"allow_admin_shell": true}' > /tmp/octos-ci-serve.json
+
       - name: Start octos serve
         run: |
-          # #2031 — the suite asserts the AGENT's shell tool can reach Homebrew
-          # binaries (`ffmpeg version N`, and `/opt/homebrew/bin` present on
-          # PATH). The agent inherits its environment from THIS process, so the
-          # install step passing proves only that the RUNNER has ffmpeg — not
-          # that serve, and therefore the agent, can see it. Set the prefix
-          # explicitly rather than trusting the image's default PATH.
-          export PATH="/opt/homebrew/bin:/opt/homebrew/sbin:$PATH"
-          # Recorded so a future failure shows what serve actually inherited,
-          # instead of leaving the next reader to infer it from a green install
-          # step — which is exactly how this stayed misdiagnosed.
-          echo "serve PATH: $PATH"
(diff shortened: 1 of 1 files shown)


ymote:
The diagnostic in this PR answers the question, and the answer is that the PATH fix is **necessary but not sufficient**. Not merging.

Two things did work:

- **Scoping the required job to its namesake: 44 failures → 4.** Only `tool-use-regression.spec.ts` runs now; the model-dependent specs moved to `e2e-live-nightly`.
- **serve genuinely has the Homebrew prefix**, proven by the new log line, with no `WARNING: ffmpeg not on serve's PATH`:

```
serve PATH: /opt/homebrew/bin:/opt/homebrew/sbin:/Users/runner/hostedtoolcache/...
```

And yet the remaining 4 still fail on `expect(result.exit_code).toBe(0)` and `toContain('CONCAT_OK')`.

So the environment chain is **runner → serve → sandboxed shell tool → command**, and the break is the last hop, not the one I fixed. The agent's shell runs under the macOS sandbox with a sanitized environment (see `BLOCKED_ENV_VARS` and `sandbox.rs`), so it does not inherit serve's PATH. Exporting on the serve step was always going to be invisible to it.

That makes #2031 a **sandbox environment** question, not a CI-workflow one: either the sandbox profile must permit and propagate a PATH that reaches Homebrew, or the tool-regression specs must invoke ffmpeg by absolute path. That is a product decision about what the sandboxed shell is allowed to see, and it belongs with whoever owns the sandbox policy.

Keeping the diagnostic regardless — it is what converted a guess into a measurement, and it will answer this immediately next time.

The scoping half of this PR stands on its own and is worth keeping. Happy to split it out so the required job stops running model-dependent specs while the sandbox question is settled separately.

PR #2045, opened 2026-08-17
fix(agent): expose session workspace to plugins

alan0x:
## Summary

- expose the session workspace to plugin tools as `OCTOS_SESSION_WORKSPACE`
- keep `OCTOS_WORK_DIR` unchanged as the skill output directory
- cover the two roots with a Unix plugin execution regression test

## Why

Manifest file bindings materialize uploaded inputs relative to the session workspace, while plugin outputs live under a separate skill-output directory. A plugin must not guess the input root by walking to the parent of its output directory.

This allows learning-coach to read the exact saved whiteboard selection image while retaining its existing output location and path-containment checks.

## Validation

- `cargo test -p octos-agent plugin_exposes_session_workspace_separately_from_output_work_dir`

### DIFF
--- a/crates/octos-agent/src/plugins/tool.rs
+++ b/crates/octos-agent/src/plugins/tool.rs
@@ -2716,6 +2716,19 @@ impl Tool for PluginTool {
             cmd.env("OCTOS_WORK_DIR", dir);
         }
 
+        // The plugin output directory and the session workspace are different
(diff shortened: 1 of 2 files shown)


alan0x:
Superseded by #2111, which includes the same session-workspace capability with explicit manifest permission gating and both positive and negative access-control tests. The safer implementation is already merged into main, so this PR should not be merged.

Issue #738, opened 2026-05-01
P1: live-thread-interleave deep_research result not surfacing in slow-Q bubble (post-wave-3 residual)


ymote:
## Summary

After today's wave-3 redeploy (`0.1.1+bf6465cc`) and merge of PRs #730 (pipeline-guard schema), #733 (e2e harness login), #735 (cancel-401), #736 (spawn_only `.md` poll loop), and #737 (CI fixtures), `live-thread-interleave.spec.ts:258` still fails identically on **both** healthy minis (mini1 + mini3). The failure mode has changed from \"login times out at chat-input\" to \"deep_research result never lands in slow-Q's bubble\" — a sharper signal exposing what's likely a thread-binding / DOM-routing issue that was hidden behind the upstream blockers.

This is **NOT** a regression introduced by today's PRs — those all delivered their intended fixes, server-side. It's a pre-existing or newly-surfaced delivery-routing problem newly visible because the upstream blockers are gone.

## Reproduction

```bash
cd /Users/yuechen/home/octos-deploy-wave2/e2e
OCTOS_TEST_URL=https://dspfac.octos.ominix.io npx playwright test tests/live-thread-interleave.spec.ts --reporter=line
```

Deterministically fails after the full 6.1-minute poll-loop timeout on both mini1 + mini3 with identical pattern.

## Failure pattern (from `/tmp/wave3-regression-mini3-v2.log`)

```
Error: Slow-Q's paired bubble has neither research-content text marker nor a .md attachment.
       This means the deep_research result either never arrived under Q1's bubble (possibly
       bound to Q2's bubble — the #649 regression) or the spawn-ack alone is what got rendered.
slow="深度研究已在后台启动。完成后会把最终研究结果发回当前会话。 2026-05-01 09:49:06"
slowHrefs=[]
fast="2  moonshot@autodl/kimi-k2.5 · 16.1K in · 2 out · 8s · 2026-05-01 09:49:11"
```

- Slow-Q bubble has ONLY the spawn-ack text + zero `.md` attachment links
- Fast-Q bubble has ONLY a metadata strip (16K input / 2 output / 8s) + the literal answer \"2\" — too short to contain a research report
- The deep_research `_report.md` is NOT in either bubble after 6.1 min wait

## What we already know (from prior investigations today)

Per Cluster D investigation in `/tmp/codex-deepsearch-report-md.log` and issue #731 close-out:
[...]

PR #2048, opened 2026-08-17
refactor(agent): retire the RecoveryHint inbox — one re-entry path for background-task terminals

### LINKED ISSUE #2020: Phase 4: retire RecoveryHint — one re-entry path for background-task terminals
## Summary

Finish the Gap-1 strangler. Background-task terminals currently re-enter the agent loop through **two** channels, and the code says so outright (`session_actor.rs:3373-3380`):

> Gateway: failure recovery stays on the `RecoveryHint` inbox (which owns the consecutive-recovery cap + per-task claim + exhaustion banner) … Only the SUCCESS outcome routes through the queue here … **Step 4 retires RecoveryHint and flips this to `Queue`.**

So today:

| outcome | path |
|---|---|
| success | `set_on_terminal` → `route_terminal_event_to_continuation_queue` |
| failure | `set_on_failure_signal` → `ActorMessage::RecoveryHint` inbox |

Two channels for one lifecycle event, held together by shared dedupe keys. This is the last unfinished step of the unification, and it is why "what happens when a background task ends" cannot be answered by reading one path.

## Why it hasn't been done

`RecoveryHint` is not a dumb carrier — it owns real policy that must survive the move:

1. **Consecutive-recovery cap** — `consecutive_recovery_turns` (`session_actor.rs:4494`), incremented per recovery-driven turn, bounded by `max_consecutive_recovery_turns()` (`:4709`), reset on a user-initiated turn (`:4743`). Without it a failing task can drive the loop indefinitely.
2. **Per-task claim** — one recovery in flight per task; prevents two recovery turns for the same failure.
3. **Exhaustion banner** — the user-visible message when the cap trips. Silently stopping recovery is indistinguishable from the task having succeeded.
4. **cmid propagation** — `originating_client_message_id` is threaded through so the synthetic `InboundMessage` reuses the originating turn's id instead of minting an orphan UUIDv7 (#738). Dropping this reintroduces orphaned turns.

Any migration that loses one of these trades a clean architecture for a behaviour regression.

## Plan

1. **Move the policy, not just the plumbing.** The cap / claim / banner belong with the continuation queue, not with the inbox variant. Land them there FIRST, behind the existing `TerminalFailureRouting::LegacyChannel`, so both paths share one policy implementation while still using two transports.
2. **Flip the routing** to `TerminalFailureRouting::Queue` once policy is shared. At this point failure and success re-enter identically.
3. **Delete `ActorMessage::RecoveryHint`** and its handler (`:5835`), the `set_on_failure_signal` wiring (`:3333`), and the now-dead dedupe collapse.
4. **Verify the cap still bites** — the highest-risk regression. A task that fails N+1 times must stop and show the banner, not loop.

Steps 1 and 2 are separately shippable; 3 is mechanical once 2 is proven.

## Test coverage that must exist before the flip

- a task failing repeatedly trips the cap and emits the **banner** (not silence)
- a user-initiated turn **resets** the counter
- two terminal events for the same task produce **one** recovery turn (claim holds)
- the recovery turn carries the **originating cmid** (#738), not a fresh UUIDv7
- success and failure terminals both re-enter through the queue, exactly once each

## Risk

Medium-high, and higher than it looks. This is the error path — it runs when something has already gone wrong, which is when it is least likely to be exercised by a happy-path soak and most likely to be noticed by a user. The `format_after_edit` family (#2011) is a reminder that the local suite passing is not the same as the behaviour holding under load.

Recommend landing it as three PRs rather than one, with the cap tests written **before** step 2 so the flip is guarded by tests that already pass on the legacy path.

## Context

Phases 1 and 3 of this work are done: peers are enrolled as supervised tasks (#1868), and the in-flight liveness predicate now explicitly excludes them so an open peer cannot pin a wedged master's marker. Phase 2 dissolved — the wake path already existed; the real defect was the fleet gate being evaluated on the wrong edge (#2003). This is the last phase.

ymote:
Completes the Gap-1 strangler. Phases 1-3 landed in #2023 and #2034; this is the last step.

## The problem

A background task reaching a terminal state re-entered the agent loop through **two** channels, split by outcome:

| outcome | path |
|---|---|
| success | `set_on_terminal` → `route_terminal_event_to_continuation_queue` |
| failure | `set_on_failure_signal` → `ActorMessage::RecoveryHint` inbox |

Two channels for one lifecycle event is not merely untidy:

- **The channels had no shared dedupe key**, so the unified sink had to be told to *skip* failures on the gateway (`TerminalFailureRouting::LegacyChannel`) purely to stop the two paths colliding. A routing flag whose only job was to keep the strangler's two halves apart.
- **Policy that must hold for every recovery lived on one of the two paths.** The per-task claim, the consecutive-recovery cap and the exhaustion banner were owned by the inbox handler, so they applied to the gateway and were structurally unavailable to the WS path.
- **"What happens when a background task ends" could not be answered by reading one path** — the answer depended on which outcome you meant and which runtime mode you were in.

## What changed

The gateway flips onto the queue; the inbox variant, its handler, and the three drain-path arms that swallowed it mid-turn are deleted.

**Policy moved before routing flipped.** The caps now live on the queue drain (`SessionActor::admit_spawn_only_failure_recovery`), guarding the single re-entry path rather than one of two. All three behaviours preserved:

1. **Per-task claim** — this is what survives a drain. The queue's task-scoped dedupe key only collapses while an item is still *pending*; it says nothing once drained.
2. **Consecutive cap** over *distinct* failing tasks — the LLM retrying its broken approach under fresh `tool_call_id`s, which no per-task key can catch.
3. **Exhaustion banner** — stopping silently is indistinguishable from the task having succeeded.

## Two things a mechanical deletion would have dropped silently

**#738 cmid propagation.** The originating turn's `client_message_id` now travels as continuation metadata and is stamped back onto the synthetic inbound. Losing it re-orphans the recovery turn under a fresh UUIDv7 and strands the eventual retry's deliverables with no DOM bubble.

**Durable user-row persistence.** Master continuations deliberately skip the user-row persist — they are runtime directives, not user turns. But the recovery prompt is not a bare directive: it is the only record of *why* the assistant re-engaged, and the M8.9 "Design A" constraint requires the model to see it on its next turn. `runtime_internal_inbound` now excludes it.

This one is worth calling out because of how it failed. The turn still ran, the LLM still replied, and a test asserting "a recovery response was produced" still passed — the transcript just showed an assistant reply with no visible cause, and the next turn lost the failure context. It was caught only because the migrated tests assert the *prompt lands in durable history*, not merely that a turn happened.

## What is NOT deleted, and why

`set_on_failure_signal` is **retargeted onto the queue, not removed**. It is the only delivery for a **fail-before-ack** failure: `notify_terminal` samples `synth_ack_emitted = false` at `mark_failed` time and the consumer prompt-suppresses, while the supervisor's two-phase stash re-emits the deferred signal once `mark_synth_ack_emitted` lands. Deleting it in favour of the unified sink alone would deliver **zero** recovery for that race — the case `fail_before_ack_spawn_only_failure_yields_exactly_one` pins.

Both producers now share the `external/<kind>/<session>/<task>` dedupe key, so an acked failure (where both fire) still yields exactly one continuation.

## The double-enqueue invariant holds

`master_continuation_scheduler.rs` is **doc-prose only** — the recently-claimed guard implementation and both invariant test bodies are byte-identical. Verify with `git diff origin/main -- crates/octos-cli/src/autonomy/master_continuation_scheduler.rs`.

The guard is now load-bearing for **both** runtime modes rather than just the WS path it was written for, since the gateway joins the same two-producer queue. The migration widened its blast radius rather than removing its reason to exist.

## Also

The prompt-renderer duplication goes: `build_recovery_prompt_body` is the single formatter. The hand-synchronised copy in the orchestrator existed only because the two re-entry paths rendered independently.

## Tests

- **New:** `consecutive_recovery_cap_trips_and_emits_banner_instead_of_a_turn` — the highest-risk regression. This is the error path: it runs when something has already gone wrong, which is when it is least likely to be exercised by a happy-path soak and most likely to be noticed by a user. Asserts both that the cap bites (bounded recovery prompts) and that the banner names the failing tool.
- **Migrated onto the queue:** the recovery e2e tests, the per-task-claim test (now driven through the real transport), and the #738 cmid test.
- **Inverted:** `legacy_channel_failure_routing_does_not_double_enqueue` → `failure_routing_enqueues_exactly_one_recovery_for_every_runtime_mode`. It used to pin that the gateway enqueued *nothing*; it now pins that gateway-shaped and WS-shaped failures each enqueue exactly one. A regression to zero would silently drop gateway failure recovery; a regression to two would double-deliver.

## Scoping note

`octos_agent::RecoveryHint` in `harness_errors.rs` is an unrelated error-**classification** enum (`BackoffRetry` / `SwitchProvider` / `CompactContext` / `FailFast` / `Bug`) that happens to share the name. It is not a re-entry path. It and its consumers (`loop_state.rs`, the `lib.rs` re-export) are untouched — those files do not appear in this diff.

## Gates

| gate | result |
|---|---|
| `cargo test -p octos-agent --lib` | 2565 passed |
| `cargo test -p octos-cli --features api --lib -- --test-threads=1` | see below |
| `cargo clippy -p octos-agent --all-targets -- -D warnings` | clean |
| `cargo clippy -p octos-cli --all-targets --no-default-features --features api,telegram,... -- -D warnings` | clean |

`2931 passed / 3 failed`. All three are pre-existing or environmental, none in a file this branch touches (`git diff --name-only origin/main` covers 7 files; none of `commands/serve.rs`, `process_manager.rs`, or the continuation-tick test's production path):

- `commands::serve::tests::global_drain_spawns_before_the_stdio_early_return` — fails on pristine `main` (stale source-scan needle, reported on #2029).
- `process_manager::tests::should_allocate_bridge_ports_as_consecutive_pair` — environment-dependent: it hardcodes `BRIDGE_BASE_WS_PORT` (3101), and an unrelated `ssh` tunnel held that port on the test host (`lsof -iTCP:3101`), so allocation skipped to 3103.
- `session_actor::tests::master_continuation_tick_reenters_actor_loop` — **pre-existing flake, measured on both sides.** Run alone, 10×: `origin/main` 1/10 fail, this branch 2/10 fail. Same assertion each time (`:2505`, "should persist the model-generated progress summary: `[]`"), and the *first* assertion — that the tick drained and reached `process_inbound` — always passes. The race is test-side: it is `#[tokio::test(flavor = "current_thread", start_paused = true)]` and gives the durable write a **fixed budget of 10 `advance`+`yield_now` iterations**, while `SessionHandle::open` does real blocking file I/O on the same single thread. Under load the write misses the budget. It is a `ChildCompleted` continuation, so it does not traverse the recovery gate this PR adds — and that gate is a non-`await`ing early return for every non-recovery reason, contributing zero executor turns. Left alone deliberately: retiming an unrelated flaky test inside a strangler PR would muddy the diff.

Closes #2020.

### DIFF
--- a/crates/octos-agent/src/tools/spawn.rs
+++ b/crates/octos-agent/src/tools/spawn.rs
@@ -2305,7 +2305,7 @@ async fn run_task_with_m8_9_recovery(
 
 /// Build the synthetic `[system-internal]` instruction the spawn-task
 /// recovery wrapper sends after a first-pass failure. The shape mirrors
-/// `session_actor::build_recovery_prompt` but operates on the
+/// `session_actor::build_recovery_prompt_body` but operates on the
 /// pre-LLM task description (we don't have a tool_input here).
 fn build_spawn_recovery_prompt(task_desc: &str, error_message: &str) -> String {
     format!(
--- a/crates/octos-cli/src/api/ui_protocol_tests.rs
+++ b/crates/octos-cli/src/api/ui_protocol_tests.rs
@@ -20014,12 +20014,10 @@ async fn unified_terminal_sink_failure_with_ack_queues_recovery_under_profile() 
     crate::autonomy::agent_orchestrator::route_terminal_event_to_continuation_queue(
         &event,
         Some(profile_id),
-        crate::autonomy::agent_orchestrator::TerminalFailureRouting::Queue,
     );
     crate::autonomy::agent_orchestrator::route_terminal_event_to_continuation_queue(
         &event,
         Some(profile_id),
-        crate::autonomy::agent_orchestrator::TerminalFailureRouting::Queue,
     );
 
     assert_eq!(
@@ -20088,7 +20086,6 @@ async fn unified_terminal_sink_failure_without_ack_suppresses_recovery() {
     crate::autonomy::agent_orchestrator::route_terminal_event_to_continuation_queue(
         &event,
         Some(profile_id),
-        crate::autonomy::agent_orchestrator::TerminalFailureRouting::Queue,
     );
 
     assert_eq!(
@@ -20130,12 +20127,10 @@ async fn unified_terminal_sink_success_queues_child_completed_under_profile() {
     crate::autonomy::agent_orchestrator::route_terminal_event_to_continuation_queue(
         &event,
         Some(profile_id),
-        crate::autonomy::agent_orchestrator::TerminalFailureRouting::Queue,
     );
     crate::autonomy::agent_orchestrator::route_terminal_event_to_continuation_queue(
         &event,
         Some(profile_id),
-        crate::autonomy::agent_orchestrator::TerminalFailureRouting::Queue,
     );
 
     let drained = default_agent_orchestrator().drain_ready_continuations_for_session(
--- a/crates/octos-cli/src/api/ui_protocol_transport.rs
+++ b/crates/octos-cli/src/api/ui_protocol_transport.rs
@@ -29344,17 +29344,19 @@ async fn run_standalone_turn(
         // post-spawn failure signal BEFORE `enable_persistence` so the
         // orphan-task sweep at `task_supervisor.rs:1164-1166` —
         // which can `mark_failed` resurrected tasks — does NOT
-        // silently drop the recovery turn. The gateway path drives
-        // recovery through `ActorMessage::RecoveryHint` directly into
-        // the session actor's inbox; the WS turn handler has no
-        // actor, so we re-inject the failure into the global master
-        // continuation queue (drained on every
+        // silently drop the recovery turn. We re-inject the failure
+        // into the global master continuation queue (drained on every
         // `appui_continuation_tick`). `default_agent_orchestrator()`
         // is a process-wide singleton, so the callback survives both
         // the per-turn `tool_registry` drop AND a closed WS
         // connection — on next subscribe, the queued
         // `External("spawn_only_failure")` continuation fires and
         // `master_continuation_prompt` renders the recovery body.
+        //
+        // #2020: the gateway now enqueues onto this same queue from its own
+        // `set_on_failure_signal` (it used to own a separate
+        // `ActorMessage::RecoveryHint` inbox), so both runtime modes
+        // re-enter through one transport.
         //
         // The per-connection drain filter
         // (`maybe_spawn_appui_master_continuation_runner`) takes
@@ -29443,11 +29445,12 @@ async fn run_standalone_turn(
         // (ChildCompleted) AND failure (recovery) re-entry through ONE
         // profile-resolving call into the master continuation queue. Runs
         // alongside the legacy `set_on_change` (success) and
-        // `set_on_failure_signal` (failure) wiring during the strangler
-        // migration — shared dedupe keys collapse the double delivery to one
-        // continuation. The threaded `terminal_profile_id` (mirrors
-        // `change_profile_id` / `failure_profile_id`) kills the `_main`
-        // failure-stranding by construction.
+        // `set_on_failure_signal` (failure, which still owns the deferred
+        // fail-before-ack re-emit) wiring — shared dedupe keys collapse the
+        // double delivery to one continuation. The threaded
+        // `terminal_profile_id` (mirrors `change_profile_id` /
+        // `failure_profile_id`) kills the `_main` failure-stranding by
+        // construction.
         let terminal_profile_id = active_profile_id
             .clone()
             .or_else(|| routed_profile_id.clone())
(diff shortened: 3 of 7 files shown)


ymote:
Parked as draft — **not** merging past a red `e2e-tool-regression`.

All other 14 checks are green (`check`, `test-octos-cli`, `test-octos-agent` lib+integration, `check-matrix`, `security`, `dashboard`, `typos`, …). The e2e job fails with **44 failed / 14 passed**, which is main's current unscoped baseline for #1923 — not something this diff introduces. Failure classes in this run:

- `m9-ws: connect error: connect ECONNREFUSED 127.0.0.1:<port>` — serve never comes up for those specs
- `ENOENT: workstreams/M15-agent-goal-loop-autonomy.md`
- `locator.check: strict mode violation: getByLabel('Tenant') resolved to 2 elements`
- ffmpeg-dependent specs failing `exit_code === 0` / `toContain('CONCAT_OK')`

None of these touch the queue-drain / routing paths changed here.

Unblocking order: fix the ffmpeg/sandbox reachability (see #2043 — scoping the required job takes it 44 → 4, and the residual 4 are the sandbox-environment ones), land #2043, then this PR's required e2e becomes meaningful and green. Marking ready again at that point.

Issue #2062, opened 2026-08-18
goal chip: hydrate should push an explicit no-goal snapshot ("cleared" doubles as the null state, as codex does)


ymote:
## Bug class

Stale goal UI after (re)attach. The goal chip is client state fed by server notifications. octos has a rich POSITIVE channel — every charge pushes a `SessionGoalUpdated`-shaped value to refresh the chip (`charge_active_goal_tokens` doc contract) — but if "no goal" is represented by *silence*, a client that cached a chip and reconnects can keep rendering a goal that was cleared elsewhere. Absence-of-message is indistinguishable from message-not-yet-arrived.

This is the same bug genre as two prior incidents: activity with no `session_id` rendering in whichever session is focused, and the ctx-gauge hydration bug (rebuilt state trusted over pushed state).

## How codex solves it (from the ~/home/codex comparison)

One notification, `ThreadGoalCleared`, is emitted BOTH on an actual `/goal clear` AND as the goal snapshot on thread resume when no goal row exists (`codex-rs/app-server/src/request_processors/thread_lifecycle.rs:728-736`). The server actively pushes "goal state: none"; the TUI has exactly one reducer that nulls the indicator (`codex-rs/tui/src/chatwidget/goal_menu.rs:74-82`); the null is ordered through the thread listener channel so it cannot race the resume response. "Removed", "never existed", and "removed while you were away" are one message — no client can be wrong about it.

## Verify first

1. Does session open/hydrate in octos push any explicit no-goal snapshot, or is "no goal" silence? (Check the hydrate path in `ui_protocol_transport` and the gateway equivalent; grep for what feeds the goal chip on reconnect in octoscode.)
2. Reproduce: bind a goal in session A via one client, `/goal clear` from a second connection, reconnect the first → does the chip survive?

## Adoption sketch (if verified)

Emit the cleared notification (`session/goal/cleared` or the existing equivalent) from the hydrate/open path whenever no goal is bound — same payload as an explicit clear. Client: delete any keep-previous-chip-if-silent logic; one reducer nulls the chip.

## Test that pins it

Open session → bind goal → clear from a second connection → reconnect the first → chip gone with zero user action.

Provenance: codex goal-removal investigation (2026-08-18). Related: #2060 (authority), the /goal clear UX question that surfaced this.

ymote:
Closing as **not fixed** — with the reasoning, because the next person to try this deserves the map.

## What the bug is

[...]

Issue #1959, opened 2026-08-11
SessionGoalUpdated events carry no ordering/generation token → a stale repaint can overtake goal-clear


ymote:
Surfaced by adversarial review of #1956. `SessionGoalUpdated` snapshots the goal under the state lock (`session_goal_updated_event_json`) and sends the detached snapshot later, with no identity/version token. Both the existing autonomous post-turn emit and the new interactive one (#1956) share this shape.

Failure scenario: a turn snapshots a still-active/complete goal; concurrently `SESSION_GOAL_CLEAR` removes it and emits its durable cleared event; the turn then enqueues the stale ephemeral `SessionGoalUpdated`. With no ordering token, the client's last-observed state is the resurrected old goal.

Fix sketch: stamp goal events with a monotonically increasing generation (per session goal store) or a revision, and have the client drop an update whose generation is older than the last applied clear/update. This is a pre-existing pattern (not introduced by #1956), but #1956 adds another emitter, so worth closing centrally.

ymote:
Server-side fix merged in #1956 (97fef110a): every goal event now carries a monotonic generation, and a per-session send guard drops stale ordering. The **airtight** completion — the client dropping stale-generation events, closing the guard's non-atomic-with-enqueue window — is implemented in octos-org/octoscode#521 (draft, blocked until an octos release carries the generation field). Keeping this open until #521 lands.

ymote:
Fully fixed, both halves merged and connected by a release:
- Server (octos #1956, 97fef110a): every goal event carries a monotonic generation; per-session send guard drops stale ordering.
- Release (octos #1960 → v2.0.3-rc.2, 57a60fd7e): cut so a backend that speaks the new protocol is published.
- Client (octoscode #521, 2a5ef1d69): drops any goal event whose generation is stale, closing the server guard's non-atomic-with-enqueue window. octoscode now pins octos-core to the rc.2 commit and auto-provisions v2.0.3-rc.2.

The stale-update-overtakes-clear race is now closed end to end.

PR #2065, opened 2026-08-18
fix(ui-protocol): scoped goal-event generations + goal-frame capability gating + forwarder hardening

ymote:
Substrate fixes for the UI-protocol goal-frame path. Each is independently correct and independently tested; all four were found while investigating #2062 and confirmed during its review passes. The #2062 goal snapshot itself is **not** in this PR — see "Why the snapshot isn't here".

**`octos-core` is untouched: no wire change.** (`git diff origin/main -- crates/octos-core` is empty.)

## 1. Per-scope generation identities for the #1959 send guard — the headline fix

Two cwd scopes can share **one wire session id** (`appui.sessions_in_cwd`: the same session key opened from two folders). The #1959 send guard is strictly monotonic per key, so while it was keyed by the *wire id* the two scopes shared a single watermark — and a goal frame's `generation` is allocated when the frame is **built**, not when it is delivered:

1. Folder B builds a repaint at generation `G_b`.
2. Folder A clears its own goal at generation `G_a > G_b`.
3. A's clear is delivered first → the shared watermark jumps to `G_a`.
4. B's still-in-flight repaint arrives → dropped as "stale", though it describes a **different folder's live goal**.

Folder B's goal chip then silently stops updating for the rest of the session.

`next_goal_event_generation` now registers each generation against the **scoped goal-store key it was built from** (bounded FIFO; wire-key fallback for legacy/unregistered generations = exactly the old behavior for events that predate the registry), and the guard resolves that identity instead of the wire id. All four goal-event producers — set, clear, token-charge, turn repaint — allocate through that one function, so every stamped frame is covered.

**Test:** `goal_scope_guard_admits_sibling_scope_repaint_after_other_scope_clear` builds the exact sequence above through the real orchestrator producers and asserts at the real guard. **Verified to fail on main's wire-keyed watermark** (mutation: revert `goal_event_watermark_identity` to return the wire session → this test fails, the other three pass).

*Closed by codex as P4 (round 6), which independently confirmed it fixes a genuine pre-existing defect.*

## 2. Goal-frame capability gating on the shared filter

A connection that did not negotiate `coding.goal_runtime.v1` — the capability the goal RPC surface already requires — was still receiving `session/goal/*` frames through session/open replay, the live pump, and direct sends. It can only report them as unknown notifications. The gate now lives in the one shared `live_event_passes_capability_filter`, so every delivery lane is covered in a single place.

**Test:** `session_open_goal_frames_gated_when_goal_runtime_not_negotiated` — seeds both the replay lane and the live-pump lane, drives a real `handle_session_open`, asserts zero goal frames on the wire. *Closed as W4/P5.*

## 3. Live-forwarder hardening

- **Cooperative cancellable stdio sends.** The stdio durable lane was a blocking `SyncSender::send`, which stalls an executor worker when a full stdio queue parks an async task. Hopping it to `spawn_blocking` trades that for a worse hazard: an in-flight blocking closure survives the task's abort and can enqueue a stale frame onto the *replacement* lane. It now parks cooperatively (non-blocking `try_send` probe + async sleep) — never occupies a worker, cancellable at every await, atomic enqueue, so `abort+join` leaves nothing detached. *Closed as S4/W3.*
- **Retire-before-baseline handover.** A re-open now retires the previous forwarder **before** the replay baseline is computed, so the old pump can no longer deliver a post-baseline event concurrently with the replay that also carries it. Handover remains **at-least-once** (replay may redeliver what the old lane already sent; the client merges durable frames by id/cursor) — documented at the site rather than overclaimed. *Closed as S5/W6.*

**Tests:** `session_open_goal_stdio_lane_retire_leaves_no_inflight_send` (through `WsConnection::new_stdio`: a pump parked on a full stdio queue, retired, enqueues nothing after the join) and `session_open_goal_reopen_hands_over_live_forwarder_lane` (exactly-once delivery across a handover).

## 4. One-lock scoped-key capture

`set_goal_scope` now returns the scoped key it registered, **from inside its own lock**, so registration and capture can no longer be two acquisitions that a concurrent same-wire open interleaves (the two-lock TOCTOU is unrepresentable at the capture site). *Closed as S6/W5.*

Kept as the safer API — but honestly: **the snapshot was its only production consumer of the return value.** With the snapshot gone the now-dead threading is removed (`SessionOpenOutcome`'s pinned-key field, the `Option<SessionKey>` local in `open_session_result`, and `register_session_ledger_scope` returning the key). The remaining production callers ignore the return; the one-lock guarantee is retained for any future caller and is exercised by tests.

## Why the snapshot isn't here

#2062 wanted `session/open` to state the session's goal state so a goal cleared elsewhere can't survive a reconnect as a stale chip. Eight rounds of review established that delivering it correctly needs a **true per-goal-state revision**, and the goal record has none: the monotonic value that exists is a frame-**emission** generation. Turning that into a state revision means bumping a counter at **33 separate `goals.get_mut` / insert / remove sites** with no accessor wrapper to centralize it — and the last two attempts each failed by missing sites (one missed the guard-release→enqueue gap, one missed ~8 non-allocating mutation paths). That is a bad trade against a cosmetic stale chip that a reopen already clears, so the snapshot is dropped and the reasoning is recorded on #2062.

The substrate bugs above are real, they were found by chasing that snapshot, and they stand on their own.

## Tests

4 under the `goal_scope_guard` / `session_open_goal` CI filters (`4 passed; …; 2987 filtered out`); a CI step runs both filters, since the `api`-gated module is invisible to the unfeatured lib/integration steps (#2029).

**Mutation:** revert the registry to the wire-keyed watermark → exactly the scope-guard test fails; the other three pass. (Restored.)

## Gates

- `cargo fmt --all` + `--check` — clean
- `cargo test -p octos-core` — 358 passed
- `cargo test -p octos-cli --lib --features api --test-threads=1` — **2985 passed, 0 failed** (fully green, including the usually-environmental `bridge_ports`)
- Parallel — the base-proven #2029 shared-static contention set only
- `cargo clippy --workspace --all-targets` — 0 · `cargo clippy -p octos-cli --all-targets --features api` — 0
- `typos` (repo config) — only the pre-existing `pendings` (untouched)

Branch delta vs merge base: **4 files, +760 / −89** — the previous snapshot machinery was removed in the split commit (−902).

## Related

- #2062 — the goal snapshot, with the descope reasoning recorded
- #2067 — cross-profile goal-frame leak through replay/live forwarding (pre-existing, filed during this investigation, not fixed here)


ymote:
Holding this until #2075 lands, deliberately.

Codex's review of #2075 turned up a defect in exactly the area this PR's forwarder hardening touches: `WsConnection` stores **one shared** `Arc<RwLock<String>>` for the profile (`ui_protocol_transport.rs:559`), every `session/open` overwrites it (`:16876`), a connection retains multiple forwarders keyed by session (`:423`), and each forwarder reads the connection-wide *current* value on every event (`:17240`). The fix is to capture the resolved scope immutably per forwarder, the way `topic_scope` already is at `:17200`.

That work is in flight on #2075 and rewrites the same function this branch modifies. Reviewing this now would be reviewing a moving target, and merging it first would guarantee a conflict on the more urgent change. This one's value is the scoped-generation registry closing a pre-existing cross-scope suppression bug — real, but not blocking anything.

Plan: land #2075, rebase this, then review.

ymote:
Unblocked — #2075 merged as `9136a829e`.

It rewrote `spawn_live_forwarder` and the `session/open` scope capture substantially: the shared per-connection profile cell is **deleted** (each forwarder now captures its scope immutably at open), and a non-filterable pin threads through both production construction sites. This branch touches the same function, so it needs a rebase before review rather than a merge.

Worth re-reading this branch's forwarder-hardening in that light — some of it may now be redundant, and its headline value is the scoped-generation registry closing a pre-existing cross-scope suppression bug, which is independent.

anantheparty (dismissed):
审查通过（2026-08-20 维护者复审）：4 个修复各自独立且正确——① per-scope generation 身份解决 #1959 guard 的跨 scope 串扰（wire id 共享时 sibling repaint 被吞），关键从 wire id 换成 scoped identity，与单调性假设匹配；② goal 帧与 goal RPC 统一用 coding.goal_runtime.v1 门控，未协商客户端零 goal 帧；③ live-forwarder 生命周期硬化 + stdio 协作取消；④ 测试 345 行 + CI 显式注册（避免 #2029 的静默跳过）。合并。

anantheparty:
已批准（2026-08-20 复审通过）。注意：分支已与 main 冲突，冲突点在 ui_protocol_transport.rs 的 goal 帧路径——与本 PR 的 #2062 系列和 main 已合的 #2067（profile_scope H2）在同一函数上交织，语义耦合深（round-2 的 profile_id/pinned_goal_key vs #2067 的 profile_scope）。建议由你（作者）rebase 到最新 main 后合并，避免维护者代 rebase 引入微妙回归。rebase 完成后我可立即复核。

anantheparty:
Superseded by #2120: the head branch was rewritten (rebase onto current main) and the no-push ruleset forbids force-pushing the original branch, so the rebased work lives on `fix2065-rebase`. All review feedback from this PR is carried into #2120.

Issue #1172, opened 2026-05-22
Codex naming parity: add , , default  coding aliases


ymote:
Follow-up to the UPCR-2026-020 Codex parity audit (M14-B done, M14-C in flight as #971). Octos has all Codex capabilities, but three name aliases would improve first-call latency for Codex-trained models:

## Gaps

1. **`bash`** — Codex CLI 0.131.0 exposes `bash` as a model-visible tool (alongside `exec_command`). Octos has `shell` + `exec_command` with the same capability but no `bash` alias. ~10 LOC: alias `bash` → `ShellTool` (or `ExecCommandTool` with stripped-down schema if Codex's `bash` is non-PTY).

2. **`delegate`** — Codex has a one-call `delegate(role, task)` that spawns a child agent, waits, and returns the result. Octos has the underlying primitives (`spawn_agent` + `wait_agent`) but no convenience wrapper. ~50 LOC: new tool that chains them and extracts the result. Resolves through the M14-C role templates (#971).

3. **`browser` as default coding tool** — Octos has `BrowserTool` (feature-gated CDP/Chrome). Currently registered conditionally, not in the default `OCTOS_KNOWN_MODEL_VISIBLE_TOOLS` advertised by `tool/status/list`. Codex shows `browser` by default. ~5 LOC + capability gate (`coding.browser.v1`).

## Why bother

- A Codex-trained model emitting `bash(cmd="ls")` or `delegate(role="reviewer", task="...")` today hits `tool not found`, then `tool_search` rescues on the retry. With these aliases, first call succeeds.
- Behavioral / capability parity is already 100%. This is naming polish for Codex-emission compatibility.

## Acceptance

- New tools registered in `coding_tools.rs` (post-#1169 rename)
- Names added to `OCTOS_KNOWN_MODEL_VISIBLE_TOOLS`
- Capability constants advertised (`coding.bash.v1`, `coding.delegate.v1`, `coding.browser.v1`)
- Unit tests for each alias's happy path + a denial path
- `tool_search("bash")` returns the new alias

## References

- UPCR-2026-020 (Codex parity contract)
- Codex CLI 0.131.0 binary grep evidence in this session
- M14-C role templates (#971) — `delegate` depends on those landing first

PR #2071, opened 2026-08-19
fix(security): close the delegated-child spawn drift, the untagged-tool tag bypass, and the SSRF empty-answer bypass

ymote:
Three findings from a peer-agent security review of `crates/octos-agent/src/tools/ssrf.rs` and `crates/octos-agent/src/tools/policy.rs`, re-verified against current main. Each fix is TDD (RED test demonstrably failing on main, then the fix) and mutation-checked (fix reverted → test fails).

**Architectural note up front, because it reframes finding 1.** `group:delegated` is **not a confinement boundary and does not try to be**. Delegated and peer agents are deliberately universal replicas of the master, as capable as it is — a child told "fix this test" must be able to run the build and the test. Security cannot be guaranteed at the agent tool-policy layer, so it is not attempted there; the real boundary is the **outer sandbox** (isolated machine, SELinux-confined account, or container). Command execution for delegated children is therefore intentional, and this PR **does not** restrict it. What `group:delegated` legitimately enforces — and what had actually drifted — is **recursion and resource control**: no unbounded delegate→delegate→delegate chains, no spawn storms.

**Do not merge — review requested.**

---

## Finding 1 — spawn-family drift in the delegated-child recursion guard (`tools/policy.rs`)

**What was wrong.** `group:delegated` is the deny list applied to every `DelegateTool` child. Its spawn/delegation family had drifted badly out of sync with `group:sessions`: it listed only `spawn` and `delegate_task`, leaving `spawn_agent`, `send_input`, `resume_agent`, `wait_agent`, `close_agent` and the codex-compat **`delegate` wrapper** all reachable. That wrapper is the sharpest one — per `group:sessions`' own #1172 comment it keeps a private `Arc` to the bound `spawn_agent` and can spawn a child *even when* `spawn`/`spawn_agent` are gone from the registry. A delegated child could therefore spawn and drive its own sub-agents straight past the recursion guard.

The list also carried two **dead entries**: `execute_code` and `send_message` are registered by **no tool anywhere** in the workspace. They are deleted rather than "fixed" into real names — under the ruling above, code execution is deliberately allowed, so an `execute_code` ban would be wrong even if such a tool existed.

**Fix.** `group:delegated` now covers the whole spawn/delegation family (a strict superset of `group:sessions`) plus the pre-existing `message` / `save_memory` / `memory_note`. Because the group machinery has **no group-in-group expansion** (`expand_group` returns a flat static slice; a nested `"group:sessions"` entry would be matched as a literal tool name), the superset is maintained by hand and locked by a new drift-guard test, `group_delegated_supersets_session_spawn_family`, which iterates the static `TOOL_GROUPS` tables and fails the moment `group:sessions` gains a member `group:delegated` lacks. It deliberately does **not** assert against `group:runtime`.

**Command execution is untouched — deliberately.** `shell` / `exec_command` / `write_stdin` / `bash` stay available to delegated children. Two tests assert exactly that (`policy.rs::should_deny_child_tool_when_group_delegated_is_in_deny_list` and `tests/delegate_tool.rs::should_apply_group_delegated_deny_list_to_child`); both now carry the architectural reason in-line — universal-replica capability, recursion control ≠ confinement, sandbox is the real boundary — so a future reader doesn't "harden" them into a deny. The group's doc comment previously promised children "may never … execute arbitrary code", which was affirmatively false and always misleading; it now states what the list does bound and points at `crate::sandbox` for confinement. `DelegateTool`'s module docs + LLM-facing `description()`, `role_template.rs`'s group enumeration, and `docs/SECURITY_ARCHITECTURE.md` §3.4 are restated to match.

**RED evidence** (against main's group table):
```
group_delegated_supersets_session_spawn_family ... FAILED
  panicked: group:delegated must deny "spawn_agent" (member of group:sessions);
            a delegated child that can reach it spawns past the recursion guard
should_deny_every_spawn_entry_point_when_group_delegated_is_denied ... FAILED
```

**Mutation check (re-run after retargeting):** restoring main's original member list → 4/5 group:delegated tests fail, including the retargeted drift guard with the message above. Restored; 5/5 green.

---

## Finding 2 — HIGH: untagged tools bypass every `require_tags` gate (`tools/policy.rs`, `tools/registry.rs`)

**Attack.** `ToolPolicy::is_allowed_with_tags` returned `true` for any tool with no tags ("tools with no tags are universal"), and the **production** gate — `provider_policy_allows_equivalent_with_tags` in `registry.rs`, which feeds `specs()` (the LLM-visible tool list) and `is_tool_visible` — duplicated the same `|| tool_tags.is_empty()` fail-open. `Tool::tags()` defaults to `&[]`, and **no plugin/skill binary tool and no MCP server tool overrides it** (nor do several builtins: `recall_memory`, `save_memory`, `memory_note`, `record_memory_use`, the `mofa_make` dispatcher pair, `configure_tool`). A profile confined with `require_tags: ["code"]` therefore still exposed every one of them — the gate failed open on exactly the unaudited surface it exists to confine.

**Fix.** Untagged tools now **FAIL** a non-empty `require_tags` gate (fail closed) at both sites. Empty `require_tags` remains a no-op.

**Behavior change and blast radius — ships as-is, no compat shim.** Measured: `is_allowed_with_tags` has **zero production callers** (the live path is the registry gate), **no first-party code constructs a non-empty `require_tags`** (tests only — it reaches production solely via operator config deserialization), and **this deployment does not label tools at all** (every agent is a universal replica of the master, so `require_tags` goes unused). Blast radius here is zero, and fail-closed is the defensible default for anyone who does opt in. `docs/SECURITY_ARCHITECTURE.md` §3.4, which documented the fail-open rule verbatim, is updated. Adjacent and intentionally untouched: the registry's separate `context_filter` also passes untagged tools — different mechanism, out of scope.

**RED evidence** (on main):
```
should_fail_closed_for_untagged_tool_when_require_tags_set ... FAILED
  panicked: untagged tools must fail a non-empty require_tags gate (fail closed)
should_fail_closed_for_untagged_tool_under_require_tags_provider_policy ... FAILED
  panicked: untagged tools must FAIL a non-empty require_tags gate (fail closed)
```
(The pre-existing `test_require_tags_filters_by_tag` asserted the bypass — "Tool with no tags passes (empty tags = universal)" — and is flipped with a comment pointing at the new test.)

**Mutation check:** re-introducing the fail-open branch in either file → both tests fail. Restored.

---

## Finding 3 — MEDIUM: empty DNS answer skips pinning (`tools/ssrf.rs`)

**Attack.** `check_ssrf_with_addrs` returned `Ok` with an empty `resolved_addrs` when `lookup_host` produced zero addresses. Empty `resolved_addrs` doubles as the "literal-IP host, nothing to pin" sentinel, so every pinning consumer — `ssrf_safe_send`'s hop loop, `web_fetch`'s own hop loop, and the MCP remote dispatcher — skipped `.resolve()` pinning and let reqwest **re-resolve the host at connect time**: a rebinding resolver answers empty at check time and `169.254.169.254` at fetch time, re-opening the exact DNS-rebinding TOCTOU the pin exists to prevent.

**Fix.** The answer-set scan is extracted verbatim into `validate_answer_set` (the seam needed for a hermetic test), and an empty **resolved** answer is now a hard failure in the module's existing fail-closed error taxonomy: `DNS resolution returned no addresses for host '<host>' — blocking request (fail closed)`. `resolved_addrs` is now empty only for literal-IP hosts — which is what all three consumers already assume — so fixing it at the source closes all of them.

**RED evidence** (seam extracted behavior-preserving first, so the failure is main's behavior):
```
should_fail_closed_when_dns_answer_set_is_empty ... FAILED
  panicked: empty DNS answer must be blocked, not treated as 'nothing to pin'
```

**Mutation check:** removing the empty-answer guard → test fails. Restored.

---

## Gates (serial)

```
cargo fmt --all && cargo fmt --all -- --check                  OK
cargo test -p octos-agent --lib                                 ok. 2579 passed; 0 failed; 3 ignored
cargo test -p octos-agent --test delegate_tool                  ok. 10 passed; 0 failed
cargo test -p octos-cli --lib --features api                    2998 passed; 2 failed (both pre-existing, see below)
cargo clippy --workspace --all-targets                          Finished, no warnings
cargo clippy -p octos-cli --all-targets --features api          Finished, no warnings
```

The `octos-cli` failures are proven pre-existing and environmental: they are the known #2029 parallel shared-static contention family, whose membership **varies per run** (this run: `closed_peer_park_produces_neither_wake_nor_escalation_row`, `second_serve_on_same_data_dir_is_refused_with_a_greppable_marker`; an earlier run instead hit `should_allocate_bridge_ports_as_consecutive_pair` — the known environmental port test — and `master_continuation_tick_reenters_actor_loop`). Every one of them passes in isolation on this branch, and the same suite run on the base commit `e87f1c061` (main) fails **5** tests — a superset including these.


ymote:
## Spawn-surface completeness (Q2) — verified by inspection

codex was unavailable (3 consecutive MCP timeouts), so the four decisive
claims were checked directly. Q2 was the one left open; it resolves in the
PR's favour.

`group:delegated` has exactly **one** production consumer:
`build_delegated_child_policy` (`tools/delegate.rs:225`), called only from
`DelegateTool::execute` (`delegate.rs:666`). The other two hits are a
re-export (`lib.rs:214`, `tools/mod.rs:823`) and a drift-guard **test**
(`role_template.rs:774`).

The registry it is applied to is built two lines earlier
(`delegate.rs:638`) and is exactly:

    ToolRegistry::with_builtins_and_sandbox(...)  // registry.rs:1220-1355
      + the child DelegateTool  (registers `delegate_task`)

So the question "is the deny list a superset of the child's spawn surface?"
is decidable by reading one function. Every spawn/session entry point that
`with_builtins_and_sandbox` registers is named by the narrowed list:

| registered in builtins | in `group:delegated` |
|---|---|
| `SpawnAgentTool` (`spawn`, `spawn_agent`) | yes |
| `DelegateAliasTool` (`delegate`) | yes |
| `SendInputTool` | yes |
| `ResumeAgentTool` | yes |
| `WaitAgentTool` | yes |
| `CloseAgentTool` | yes |
| child `DelegateTool` (`delegate_task`) | yes |

The spawn-adjacent surfaces flagged as candidate gaps in the original peer
review are **not registered into a child registry at all** — grepping the
`with_builtins_and_sandbox` span for `RunPipelineTool|CronTool|PeerHandoffTool|SaveMemoryTool|MessageTool`
returns 0. `run_pipeline` / `deep_research` (pipeline fan-out), the six
`peer_*` tools, and `cron` are wired by session/CLI construction, which a
delegated child never runs. `message` / `save_memory` / `memory_note` staying
in the list is belt-and-braces for the same reason.

One correction to an earlier read of mine: `clear_spawn_only()`
(`registry.rs:350`) clears the spawn_only **marking sets**, not the tools —
it is not what keeps `run_pipeline` out. Non-registration is.

### The other three, re-confirmed

- `execute_code` / `send_message` are registered by **no** tool anywhere in
  the workspace (hits only in policy.rs, its tests, and docs) — deleting them
  from the group is a no-op, and keeping them would have been the misleading
  option.
- No in-repo config sets `require_tags` to a non-empty value, so the
  untagged-tool fail-open fix has zero blast radius on shipped configs.
- The SSRF fix preserves the literal-IP path: `ssrf.rs:37` builds
  `resolved_addrs: vec![]` on a branch that never reaches
  `validate_answer_set` (line 44, DNS path only), and the pin at line 157
  applies only when the set is non-empty.

### Scope note

Per the architectural ruling on this PR, command execution is deliberately
NOT restricted for delegated/peer children — they are universal replicas of
the master and must be able to run builds and tests. Confinement is the outer
sandbox's job. What this list enforces is recursion/resource control, which a
tool policy can actually deliver. A hypothetical gap here would therefore be a
recursion bug, not a security hole — and as verified above, there is no gap.

PR #2072, opened 2026-08-19
feat(cron): record job provenance, reap a deleted loop's jobs, add notify mode

alanpoon:
## Why

Nothing recorded what created a cron job, so a job created during an autonomous loop's turn outlived the loop with no way to find it again. From one profile's logs:

| Day | cron executions | failed deliveries |
|---|---|---|
| Aug 12 | **55,122** | **29,490** |
| Aug 18 | 418 | 201 |
| Aug 19 | 258 | 120 |

Aug 12 is six orphaned jobs on 5-second schedules. Every `agent`-mode fire costs a model turn, so that is tens of thousands of billed turns answering messages nobody read — and the only way to find them was to open `cron.json` by hand. A standing 60s `println hello` reminder still bills **1440 turns a day** to restate a string the job already holds, with ~47% of the deliveries failing `channel closed` anyway.

Three additive changes, all defaulting to current behaviour.

## 1. Provenance

`CronJob` gains an optional `origin` (`session_id`, `loop_id`, `profile_id`), `skip_serializing_if` empty — a `cron.json` written before this loads unchanged, and a hand-made job still serialises byte-identically.

`CronTool` carries it exactly as it already carries channel/chat_id: set per turn by the session actor, and **cleared on every non-loop turn**. That clearing is the load-bearing part — a stale value would stamp a user's own schedule with a loop id and get it reaped.

## 2. Reaping on `loop/delete`

`CronService::remove_jobs_for_loop` removes the jobs, and the result reports `reaped_cron_job_ids` so a client can say what went with the loop rather than silently deleting schedules. The orchestrator receives the service via `set_cron_service` at profile build (mirroring the existing `ProfileRuntimeLifecycle` handle); where none exists — tests, in-process orchestrators — the reap is a no-op, not an error.

**Matching is on the recorded `loop_id` only, never on `name`.** Names are derived from message text, so a job called `loop_03_hi` may have nothing to do with `loop_03`. There is a test asserting a same-named user job survives.

## 3. `notify` mode

`CronPayload.mode` is `agent` (default, unchanged) or `notify`, which delivers the message verbatim and returns **before** the bus send — the bus is what turns a fire into an `InboundMessage` the gateway answers with a model turn, so returning early is the entire saving. The tool schema exposes it, and an unrecognised value falls back to `agent` rather than failing the tool call.

```rust
if job.payload.mode == CronMode::Notify {
    // no InboundMessage, no turn, no tokens
    return;
}
```

## Tests

- `reaping_a_loop_removes_only_that_loops_jobs` — the owning loop's job goes; a sibling loop's job and a same-named user job both survive; reaping an unknown loop is a no-op.
- `origin_and_mode_round_trip_and_default_for_legacy_jobs` — a pre-provenance `cron.json` loads with `mode == Agent`, an empty origin, `belongs_to_loop() == false`, and no `origin` key on write-back.

`cargo test -p octos-bus`: **229 passed, 1 failed**. The failure is `media::download_media_with_cap_rejects_oversized_body`, which fails identically on an unmodified tree (sandboxed HTTP returns 502) — verified by stashing.

## Note for reviewers

The session actor stamps `session_id` and `loop_id` but leaves `profile_id` `None`, because `SessionActor` carries no profile id and `loop_id` is the key the reap matches on. If a profile-scoped reap is wanted later, that field is where it goes.


anantheparty (dismissed):
审查通过（2026-08-20 维护者复审）：cron 溯源（CronOrigin）设计合理——loop/delete 可通过 origin 收割孤儿 job；Debug 实现谨慎（不 dump job store）；notify 模式配套。测试 2 个偏少，建议后续补 loop/delete 收割的端到端用例，不阻塞合并。

PR #2075, opened 2026-08-20
fix(ui-protocol): scope goal, loop and monitor frames to the connection's profile

### LINKED ISSUE #2067: goal chip: durable cross-profile SessionGoalUpdated frames pass replay/live forwarding (profile scope filter only recognizes SkillActionJobUpdated)
## Bug class

Cross-profile leak of goal-chip frames through session/open replay and the live forwarder. Found by codex while reviewing PR #2065 (the #2062 goal null snapshot); pre-existing, deliberately NOT fixed there.

## Mechanism

`ledger_event_matches_profile_scope` (crates/octos-cli/src/api/ui_protocol_transport.rs:17120) recognizes exactly ONE variant:

```rust
fn ledger_event_matches_profile_scope(event: &UiProtocolLedgerEvent, profile_id: &str) -> bool {
    match event {
        UiProtocolLedgerEvent::Notification(UiNotification::SkillActionJobUpdated(update)) => {
            update.profile_id == profile_id
        }
        _ => true,
    }
}
```

Every other profile-carrying notification — `SessionGoalUpdated` / `SessionGoalCleared` in particular, which carry `profile_id: Option<String>` on the wire — falls through `_ => true`. The filter runs at all three delivery boundaries, and passes goal frames at each:

- session/open replay loop (ui_protocol_transport.rs:16917)
- live forwarder pump (ui_protocol_transport.rs:17212, via `forward_live_ledger_event`)
- the retained-events path (ui_protocol_transport.rs:18129)

So a durable `session/goal/updated` bound to profile B (appended via the goal RPC evidence path — `record_autonomy_rpc_evidence` → `send_notification_durable`, which ledgers it) is replayed and live-forwarded to a connection scoped to profile A whenever the two share a wire session key (shared/unprofiled keys, admin/unscoped connections). The octoscode chip reducers are session-keyed and apply `updated` unconditionally, so the foreign profile's objective/budget paint into A's chip — both a UI-truthfulness bug and a tenant-isolation leak (the objective text is cross-tenant data).

## Adjacent, same root

The interactive charge binding has the mirror-image KNOWN LIMITATION already documented in-code at ui_protocol_transport.rs:19543 — a `turn/start` carries no goal profile, so a goal set with an explicit `profile_id` on a bare session key resolves `_main` and goes uncharged. A fix for this issue should keep that comment's constraint in mind: "unscoped ⇒ authorized for any profile" is deployment-dependent, so bind-by-session-alone is not automatically safe.

## Fix sketch

Extend `ledger_event_matches_profile_scope` to compare the goal events' `profile_id` (and audit the other profile-carrying variants: loop/monitor/agent families) against the connection's resolved profile, with the compatibility question settled first: what should `profile_id: None` events (legacy rows) and unprofiled connections (`_main`) match? A too-strict filter would blank/starve existing single-tenant deployments where everything is `_main`.

## Test that pins it

Profile-A-scoped connection opens a shared wire session; ledger holds a durable `session/goal/updated` with `profile_id: "profile-b"` → neither replay nor the live forwarder delivers it to A. Inverse: `_main`-scoped connection still receives `_main` and legacy `None` goal frames.

Provenance: codex review of PR #2065 (2026-08-18), adjacent P1.

ymote:
Closes #2067

Four commits. The leak fix, then three rounds of adversarial review, each of which found starvation the previous round had introduced.

| commit | what it does |
|---|---|
| 1 | Scope goal / loop / monitor frames to the connection's profile — the reported leak. |
| 2 | Capture the forwarder's scope **per session** instead of from a shared per-connection cell; scope `session/open` + `background/activity`. |
| 3 | Stop filtering when a routing header makes the scope disagree with the turn's profile. |
| 4 | Filter only on a profile the connection **cannot drift from** — round 3's exemption was too narrow. |

## The bug

`ledger_event_matches_profile_scope` recognized one variant, `SkillActionJobUpdated`; everything else fell through `_ => true`. It runs at three delivery boundaries — the `replay.retain` in `open_session_result`, the session/open replay send loop, and the live forwarder pump — and `session/goal/updated` / `session/goal/cleared` are appended **durably** via `record_autonomy_rpc_evidence` → `send_notification_durable`, so they crossed all three.

A bare `web-N` key is deliberately accepted under profile auth (`:16767-16773` — "SPA convention: a fresh session uses a raw `web-N` id with no profile prefix"), so two tenants sharing one wire key is the normal case. The chip reducers are session-keyed and apply `updated` unconditionally, so profile B's objective text painted into profile A's chip. RED, through the real delivery paths:

```
assertion `left == right` failed: replay must not put profile-b's objective on an alpha-scoped connection
  left: ["beta secret objective", "alpha objective"]
 right: ["alpha objective"]
```

The class was already acknowledged in-code: the three *interactive* goal pushes use `send_notification_ephemeral` because "a durable append would fan the goal to every subscriber of an unprofiled shared session id regardless of profile" (`:33648-33652`). The RPC path never got that treatment.

## What the review rounds found

**Round 2 — the forwarder's scope was per-connection, not per-session.** `WsConnection` held one shared `Arc<RwLock<String>>` that every `session/open` overwrote, while a connection keeps one forwarder per session. Fixed by threading `profile_scope` into `spawn_live_forwarder`, captured at open exactly like `topic_scope`; the shared cell and its accessors are removed rather than left as a trap. The replay boundaries never shared the hazard — both use per-call locals.

**Round 3 — routed admin connections were being starved.** `validate_session_scope` never consults `routed_profile_id`, but the turn path resolves its runtime from `connection_profile_id.or(routed_profile_id)`. An admin connection routed to a tenant by `Host` subdomain or `X-Profile-Id` opened a bare key resolving to `_main` while its turns ran under the tenant's `ProfileRuntime`. RED showed total starvation — the connection received nothing but its own open frame:

```
the connection's own routed monitor/fired must be replayed: ["session/open"]
```

**Round 4 — the same divergence has a second source, with no header involved.** `session_open_profile_id` is a single per-connection cell overwritten by every successful open (`:6035`), and every later `turn/start` — including one on an *earlier* session — receives that latest value (`:6050`) and selects that runtime (`:29332`). Open bare A (`_main`), open bare B as `beta`, run a turn on A: A's forwarder holds `_main` while the turn runs under `beta`, and a model-created monitor is stamped `beta`. Round 3's exemption did not fire. RED, on both delivery paths independently:

```
session A must still receive the activity its own turn produced, even though the turn ran under the profile a later open bound
...
assertion `left == right` failed: reconnect replay must not drop session A's own activity
  left: []
 right: ["live log line"]
```

The turn misrouting predates this branch; the **delivery drop did not**, which is what put it in scope.

## The guard

Filtering is gated on the only binding that provably cannot drift — an authenticated connection profile:

```rust
// callers pass the pin; `None` means "do not filter"
let profile_scope = pinned_profile_id.map(ToOwned::to_owned);
```

It is frozen at WS upgrade (`:5766`, and the comment at `:5769-5770` says so), and `validate_authenticated_session_scope` forces every session opened on that connection to it or rejects the open outright. So the whole connection has exactly one profile, every forwarder captures it, and `connection.or(..)` short-circuits onto it for every turn.

- **WS and session-ingress** pass their frozen `connection_profile_id`.
- **stdio passes `None`** — and this is where I disagreed with the review. Codex recommended preserving filtering on stdio; stdio rebinds `connection_profile_id_owned` after every successful open (`:6801`) and passes it to `turn/start` (`:6771`), so it has *exactly* the drift that round 4 is fixing. Preserving it there would have reintroduced the same starvation on a different transport.
- **Unscoped WS passes `None`** — no pin exists, however it was routed.

**This costs no tenant isolation.** A tenant is an authenticated user, which is precisely the case still filtered — and precisely what #2067 reports. An unscoped WS connection is an admin/operator authorized for every profile (or auth is disabled and every route is deliberately open); stdio is a single local user.

Because the guard sits above the variant match, correctness does not depend on enumerating affected variants right — which matters, since round 2's enumeration was wrong.

## Two justification comments were factually wrong, and are corrected

- Commit 2 claimed both `background/activity` emitters "descend from `resolve_autonomy_profile_id`". They do not: `agent_orchestrator.rs:7022` is the only production fleet-creation site in the repo (every other `Fleet::create*` is `#[cfg(test)]`; there is no `fleet/*` RPC), reached solely from `model_create_fleet_plan`, whose profile is `ProfileRuntime.profile.id`.
- Commit 3 claimed a routed profile "can only be `P`" for an authenticated identity. Also false — an admin-role user is authorized for every profile and a parent user for its owned subaccounts (`auth_handlers.rs:3834`), and both still present an authenticated `connection_profile_id`. It creates no divergence, but because `connection.or(routed)` never reaches `routed` while the connection profile is set, not because routing is constrained.

Both behaviours were fine; the stated reasons would have been cited as precedent.

## `ProfileRuntime`-stamped set — established, not assumed

Of the nine filtered variants, exactly **two** can be stamped from the turn's runtime rather than `resolve_autonomy_profile_id`:

| variant | stamped by the turn? | why |
|---|---|---|
| `MonitorFired` | **yes** | tool → request (`goal_tool.rs:1802`) → record (`agent_orchestrator.rs:9369`) → `Phase1::Fire` (`:10111`) → continuation (`:10154-10158`) → drain (`:2994-3010`) → event. The resolver is never called. |
| `BackgroundActivity` | **yes** | fleet origin unconditionally; monitor origin whenever model-created. |
| `SessionGoalUpdated` / `SessionGoalCleared` | no | the model-tool goal mutations are **never ledgered** — all three turn-path pushes use `send_notification_ephemeral` ("NOT appended to the ledger — non-durable per spec § 9"), and the durable sites decode resolver-stamped RPC results. The goal tools never touch `ws` or `ledger`. |
| `LoopUpdated` / `LoopFired` | no | no model tool creates loops. |
| `MonitorUpdated`, `SkillActionJobUpdated`, `SessionOpened` | no | RPC-derived, or the scope itself. |

Recorded in #2081: the goal *record*'s profile genuinely can be `ProfileRuntime`-derived — it just never reaches a durable frame today. Promote the post-turn chip repaint to `send_notification_durable` and `SessionGoalUpdated` joins the `MonitorFired` class.

## Compatibility questions

**1. What must `profile_id: None` match? `_main`.** No emitter writes `None`: `goal/set` and `goal/clear` route through `resolve_autonomy_profile_id` (`_main` fallback) into `GoalSetRequest.profile_id: String`, and the nested record is non-optional. A wire `None` can only be a **legacy ledger row** — the field is `skip_serializing_if = "Option::is_none"` and the ledger is JSON-Lines persisted across restarts. Only a backend predating the stamp wrote those, in deployments where every connection resolves to `_main`.

**2. What must an unprofiled connection receive?** Everything — it has no frozen pin, so its scope is not a tenant boundary. (Rounds 1–3 filtered it against `_main`; round 4 established that is unsound.)

**3. Does `_main` appear explicitly on the wire?** Both — `resolve_autonomy_profile_id` returns the literal `_main`, and it is also the connection-side default. The only absent-vs-`"_main"` split is across a version upgrade, which is why `None` must normalize rather than be dropped.

**Where the evidence corrected the issue's fix sketch:** the sketch assumed the top-level `profile_id` is what to compare. That fails for `loop/updated` — `loop/pause`, `loop/resume` and `loop/delete` return result JSON with no top-level `profile_id` (`agent_orchestrator.rs:9145-9152`, `:9158-9163`, `:9192-9198`); only `loop/create` has it. Each arm falls back to the profile on the record it carries, which can only ever **narrow** an audience.

## Variant audit

| variant | leaks? | this PR |
|---|---|---|
| `SkillActionJobUpdated` | no | already filtered; unchanged |
| `SessionGoalUpdated` / `SessionGoalCleared` | **yes** — objective text | **FIXED** |
| `LoopUpdated` | **yes** — `loop.prompt` | **FIXED** (nested fallback load-bearing) |
| `LoopFired`, `MonitorUpdated`, `MonitorFired` | **yes** — `argv`, `name`, `filter_regex` | **FIXED** |
| `SessionOpened` | **yes** — `workspace_root`, context + pane snapshots | **FIXED** |
| `BackgroundActivity` | **yes** — raw observed text | **FIXED** |
| `AgentUpdated`, `PeerStaged`, `PeerClosed` | **yes** | **DEFERRED** → #2081 |
| `LoopCompleted` / `MonitorExpired` | no — no producer exists | **DEFERRED** → #2080 |

`SessionOpened` is safe to filter because the opener's forwarder skips the broadcast copy via `from_connection` and its own frame arrives through the **unfiltered** direct send at `:16990`; the scopes are tautological, since `ledger_profile_id` *is* `active_profile_id.unwrap_or(_main)`.

## Tracking issues

- **#2081** — the turn-side/connection-side divergence, updated with both drift sources and the stdio analogue.
- **#2080** — `LoopCompleted` / `MonitorExpired` have no producer.
- **#2078** — the boundary-2 replay-send check is dead: same predicate, same already-retained vector, same scope as boundary 1.

Noted, not chased: `resolve_autonomy_profile_id` checks `is_empty()` rather than `trim().is_empty()` (`:15546-15548`) while `optional_profile_scope_matches` trims. Harmless — registered ids exclude whitespace (`profiles.rs:2605`).

Out of scope and untouched: the interactive-charge KNOWN LIMITATION at `:19498-19507`.

## Tests

All drive the real delivery paths — `handle_session_open` over a live `WsConnection` with frames read off the writer channel, `open_session_result`, and `spawn_live_forwarder`. None calls the filter directly; none is an `include_str!` grep.

| test | property |
|---|---|
| `should_retain_only_same_profile_goal_events_when_open_session_result_replays` | B1 retained events |
| `should_drop_cross_profile_goal_frames_when_connection_scopes_another_profile` | B2 + B3, `updated` and `cleared` |
| `should_drop_cross_profile_loop_and_monitor_frames_when_connection_scopes_another_profile` | B2 + B3 for `loop/updated` (top-level-absent pause shape), `monitor/updated`, `loop/fired`, `monitor/fired` |
| `should_drop_cross_profile_session_opened_frames_when_connection_scopes_another_profile` | B2 + B3 for `session/open`; opener still receives exactly one — its own |
| `should_drop_cross_profile_background_activity_frames_when_connection_scopes_another_profile` | B2 + B3 for `background/activity` |
| `should_replay_main_and_legacy_goal_frames_when_connection_is_unprofiled` | compat: `_main`, legacy `None` **and** foreign frames all delivered on an unpinned connection |
| `should_scope_each_forwarder_independently_when_one_connection_pumps_two_profiles` | round 2: two forwarders on one connection keep their own captured scopes |
| `should_deliver_routed_profile_frames_when_the_connection_scope_is_not_authoritative` | round 3: routed admin connection keeps `monitor/fired` + `background/activity` + goal frames |
| `should_deliver_later_profile_frames_when_an_unscoped_connection_opened_two_sessions` | **round 4**: open A `_main`, open B `beta`, beta-stamped `background/activity` on A delivered — replay **and** live |

Every one was watched fail before its fix. For the loop/monitor arms, the round-2 fix and the round-4 fix the production change was re-removed to confirm RED for the right reason; for round 4 the live leg was additionally neutralised so the replay leg could be observed failing independently.

Two tests changed meaning under the round-4 guard and were re-pinned rather than left vacuous, which is called out in their doc comments: the unprofiled-connection test now asserts the no-filter contract it actually has, and the per-forwarder scope invariant moved to the forwarder boundary — where it is still observable, and where it becomes load-bearing again once #2081 is closed.

## Verification

Run sequentially (never concurrently — the build lock deadlocks):

```
$ cargo clippy -p octos-cli --features api --all-targets -- -D warnings
CLIPPY_EXIT=0

$ cargo test -p octos-cli --features api --lib -- --test-threads=1
test result: ok. 3009 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out; finished in 504.42s

$ cargo fmt --all -- --check
FMT_CLEAN
```

The nine scope tests by fully-qualified name (a name matching no filter runs zero tests and still reports green, so they are pinned explicitly):

```
test api::ui_protocol_transport::tests::should_deliver_later_profile_frames_when_an_unscoped_connection_opened_two_sessions ... ok
test api::ui_protocol_transport::tests::should_deliver_routed_profile_frames_when_the_connection_scope_is_not_authoritative ... ok
test api::ui_protocol_transport::tests::should_drop_cross_profile_background_activity_frames_when_connection_scopes_another_profile ... ok
test api::ui_protocol_transport::tests::should_drop_cross_profile_goal_frames_when_connection_scopes_another_profile ... ok
test api::ui_protocol_transport::tests::should_drop_cross_profile_loop_and_monitor_frames_when_connection_scopes_another_profile ... ok
test api::ui_protocol_transport::tests::should_drop_cross_profile_session_opened_frames_when_connection_scopes_another_profile ... ok
test api::ui_protocol_transport::tests::should_replay_main_and_legacy_goal_frames_when_connection_is_unprofiled ... ok
test api::ui_protocol_transport::tests::should_retain_only_same_profile_goal_events_when_open_session_result_replays ... ok
test api::ui_protocol_transport::tests::should_scope_each_forwarder_independently_when_one_connection_pumps_two_profiles ... ok
```

Not verified: no live multi-tenant soak — the evidence is the delivery-path tests plus the emit-site traces, not a running deployment.

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol_tests.rs
+++ b/crates/octos-cli/src/api/ui_protocol_tests.rs
@@ -5462,6 +5462,7 @@ async fn stdio_binding_updates_only_after_successful_session_open() {
         &questions,
         &forwarders,
         candidate.as_deref(),
+        None,
         features,
         "open-missing".into(),
         missing_params,
@@ -5517,6 +5518,7 @@ async fn stdio_binding_updates_only_after_successful_session_open() {
         &questions,
         &forwarders,
         candidate.as_deref(),
+        None,
         features,
         "open-grace".into(),
         grace_params,
@@ -5736,6 +5738,7 @@ async fn stdio_multi_profile_open_status_reads_isolated_runtime_policy_stamps() 
             &PendingQuestionStore::default(),
             ConnectionId::next(),
             Some(profile_id),
+            None,
             features,
             SessionOpenParams {
                 session_id: session_id.clone(),
@@ -5851,6 +5854,7 @@ async fn session_open_writes_active_profile_marker_only_with_flag_and_cwd() {
                 questions,
                 ConnectionId::next(),
                 Some(profile),
+                None,
                 features,
                 SessionOpenParams {
                     session_id,
@@ -12774,8 +12778,14 @@ fn skill_action_job_events_are_visible_only_to_their_profile() {
         },
     ));
 
-    assert!(ledger_event_matches_profile_scope(&event, "profile-a"));
-    assert!(!ledger_event_matches_profile_scope(&event, "profile-b"));
+    assert!(ledger_event_matches_profile_scope(
+        &event,
+        Some("profile-a")
+    ));
+    assert!(!ledger_event_matches_profile_scope(
+        &event,
+        Some("profile-b")
+    ));
 }
 
 // ---------------------------------------------------------------------------
@@ -12937,6 +12947,7 @@ async fn should_retain_only_same_profile_goal_events_when_open_session_result_re
         &PendingQuestionStore::default(),
         ConnectionId::next(),
         Some("alpha"),
+        None,
         ConnectionUiFeatures::default(),
         SessionOpenParams {
             session_id: session_id.clone(),
@@ -12953,7 +12964,7 @@ async fn should_retain_only_same_profile_goal_events_when_open_session_result_re
     .await
     .expect("alpha opens the shared session");
 
-    assert_eq!(outcome.profile_id, "alpha");
+    assert_eq!(outcome.profile_scope.as_deref(), Some("alpha"));
     let retained: Vec<&str> = outcome
         .replay
         .iter()
@@ -13025,6 +13036,7 @@ async fn should_drop_cross_profile_goal_frames_when_connection_scopes_another_pr
         &questions,
         &forwarders,
         Some("alpha"),
+        None,
         ConnectionUiFeatures::default(),
         "open-alpha".into(),
         SessionOpenParams {
@@ -13132,6 +13144,7 @@ async fn should_replay_main_and_legacy_goal_frames_when_connection_is_unprofiled
         &approvals,
         &questions,
         &forwarders,
+        None,
         None,
         ConnectionUiFeatures::default(),
         "open-main".into(),
@@ -13328,6 +13341,7 @@ async fn should_drop_cross_profile_loop_and_monitor_frames_when_connection_scope
         &questions,
         &forwarders,
         Some("alpha"),
+        None,
         ConnectionUiFeatures::default(),
         "open-alpha-autonomy".into(),
         SessionOpenParams {
@@ -13484,6 +13498,7 @@ async fn should_keep_delivering_first_session_frames_when_a_second_session_opens
             &questions,
             &forwarders,
             None,
+            None,
             ConnectionUiFeatures::default(),
             rpc_id.into(),
             SessionOpenParams {
@@ -13600,6 +13615,7 @@ async fn should_drop_cross_profile_session_opened_frames_when_connection_scopes_
         &questions,
         &forwarders,
         Some("alpha"),
+        None,
         ConnectionUiFeatures::default(),
         "open-alpha-shared".into(),
         SessionOpenParams {
@@ -13716,6 +13732,7 @@ async fn should_drop_cross_profile_background_activity_frames_when_connection_sc
         &questions,
         &forwarders,
         Some("alpha"),
+        None,
         features,
         "open-alpha-activity".into(),
         SessionOpenParams {
@@ -13771,6 +13788,141 @@ async fn should_drop_cross_profile_background_activity_frames_when_connection_sc
         next_frame_within(&mut rx, 250).await.is_none(),
         "beta's live activity must not reach an alpha-scoped connection"
     );
+
+    abort_live_forwarders(&forwarders, &ledger).await;
+}
+
+/// #2067 round 3 — the connection's resolved session scope is NOT a tenant
+/// boundary on a routed admin connection, and using it as a delivery filter
+/// starves that connection.
+///
+/// `validate_session_scope` never consults `routed_profile_id`
+/// (`ui_protocol_transport.rs:16694`), but the turn path resolves its runtime
+/// from `connection_profile_id.or(routed_profile_id)`. So an admin connection
+/// (`connection_profile_id == None`) routed to tenant `beta` by `Host`
+/// subdomain or `X-Profile-Id` opens a bare session key that resolves to
+/// `_main`, while its own turns run under `beta`'s `ProfileRuntime` — and the
+/// model tools that runtime registers hard-bind `beta` onto every record they
+/// create (`runtime/profile.rs:1321` -> `goal_tool.rs:1802`). The durable
+/// frames those records later emit are stamped `beta`, so a `_main` filter
+/// drops the connection's OWN data.
+///
+/// `background/activity` is the worst case: its sink appends over a detached
+/// connection whose writer is discarded, so replay and live forwarding are its
+/// only delivery paths — a filter drop there is total.
+#[tokio::test]
+async fn should_deliver_routed_profile_frames_when_the_connection_scope_is_not_authoritative() {
+    let temp = tempfile::tempdir().expect("tempdir");
+    let state = local_profile_state_with_sessions(temp.path());
+    create_or_get_local_solo_profile(
+        &state,
+        local_profile_params("Beta Owner", "beta", "beta@example.com"),
+    )
+    .expect("create beta profile");
+
+    let (ws, mut rx) = ws_connection_for_test(64);
+    let ledger = Arc::new(UiProtocolLedger::new(64));
+    let approvals = PendingApprovalStore::default();
+    let questions = PendingQuestionStore::default();
+    let forwarders: SharedLiveForwarders = Arc::new(tokio::sync::Mutex::new(HashMap::new()));
+    // A BARE key on an admin connection: `validate_session_scope` resolves it
+    // to `_main` even though the turn will run as `beta`.
+    let session_id = SessionKey("web-routed-admin".into());
+    let features = ConnectionUiFeatures {
+        background_activity: true,
+        ..Default::default()
+    };
+
+    let monitor_fired = |monitor_id: &str| {
+        UiNotification::MonitorFired(octos_core::ui_protocol::MonitorFiredEvent {
+            session_id: session_id.clone(),
+            profile_id: Some("beta".to_owned()),
+            monitor_id: monitor_id.to_owned(),
+            name: Some(monitor_id.to_owned()),
+            line_count: Some(1),
+            fired_at_ms: None,
+        })
+    };
+    let activity = |text: &str| {
+        UiNotification::BackgroundActivity(octos_core::ui_protocol::BackgroundActivityEvent {
+            session_id: session_id.clone(),
+            profile_id: Some("beta".to_owned()),
+            origin_kind: "monitor".to_owned(),
+            origin_id: "monitor-model-made".to_owned(),
+            origin_label: Some("ci-tail".to_owned()),
+            text: text.to_owned(),
+            emitted_at_ms: 1_760_000_000_000,
+            dropped_count: None,
+            suppressed: false,
+        })
+    };
+
+    // Durable history produced by THIS connection's own model-created monitor.
+    ledger.append_notification(monitor_fired("monitor-model-made"));
+    ledger.append_notification(activity("replayed log line"));
+    ledger.append_notification(goal_updated_notification(
+        &session_id,
+        Some("beta"),
+        "routed objective",
+    ));
+
+    let opened = handle_session_open(
+        &ws,
+        &state,
+        &ledger,
+        &approvals,
+        &questions,
+        &forwarders,
+        // admin / unscoped …
+        None,
+        // … but routed to `beta` by subdomain or header.
+        Some("beta"),
+        features,
+        "open-routed-admin".into(),
+        SessionOpenParams {
+            session_id: session_id.clone(),
+            topic: None,
+            profile_id: None,
+            cwd: None,
+            sandbox: None,
+            after: Some(UiCursor {
+                stream: session_id.0.clone(),
+                seq: 0,
+            }),
+        },
+        false,
+    )
+    .await;
+    assert!(opened, "a routed admin connection must be able to open");
+
+    let frames = drain_session_open_frames(&mut rx).await;
+    let methods: Vec<&str> = frames
+        .iter()
+        .filter_map(|frame| frame.get("method").and_then(Value::as_str))
+        .collect();
+    assert!(
+        methods.contains(&octos_core::ui_protocol::methods::MONITOR_FIRED),
+        "the connection's own routed monitor/fired must be replayed: {methods:?}"
+    );
+    assert!(
+        methods.contains(&octos_core::ui_protocol::methods::BACKGROUND_ACTIVITY),
+        "background/activity has no other delivery path and must be replayed: {methods:?}"
+    );
+    assert!(
+        methods.contains(&octos_core::ui_protocol::methods::SESSION_GOAL_UPDATED),
+        "the connection's own routed goal frame must be replayed: {methods:?}"
+    );
+
+    // Same property on the live forwarder.
+    ledger.append_notification(activity("live log line"));
+    let live = next_frame_within(&mut rx, 2_000)
+        .await
+        .expect("a routed admin connection must keep receiving its own live frames");
+    assert_eq!(
+        live.get("method").and_then(Value::as_str),
+        Some(octos_core::ui_protocol::methods::BACKGROUND_ACTIVITY),
+    );
+    assert_eq!(live["params"]["text"], json!("live log line"));
 
     abort_live_forwarders(&forwarders, &ledger).await;
 }
@@ -14107,6 +14259,7 @@ async fn session_open_replays_notifications_after_cursor_and_returns_ledger_curs
         &PendingQuestionStore::default(),
         ConnectionId::next(),
         None,
+        None,
         ConnectionUiFeatures::default(),
         SessionOpenParams {
             session_id: session_id.clone(),
@@ -14170,6 +14323,7 @@ async fn session_open_topic_scope_replays_only_matching_topic_bucket() {
         &PendingQuestionStore::default(),
         ConnectionId::next(),
         None,
+        None,
         ConnectionUiFeatures::default(),
         SessionOpenParams {
             session_id: base_session.clone(),
@@ -14200,6 +14354,7 @@ async fn session_open_topic_scope_replays_only_matching_topic_bucket() {
         &approvals,
         &PendingQuestionStore::default(),
         ConnectionId::next(),
+        None,
         None,
         ConnectionUiFeatures::default(),
         SessionOpenParams {
@@ -14240,6 +14395,7 @@ async fn session_open_rejects_after_cursor_from_other_stream() {
         &approvals,
         &PendingQuestionStore::default(),
         ConnectionId::next(),
+        None,
         None,
         ConnectionUiFeatures::default(),
         SessionOpenParams {
@@ -14302,6 +14458,7 @@ async fn session_open_rejects_stale_after_cursor() {
         &PendingQuestionStore::default(),
         ConnectionId::next(),
         None,
+        None,
         ConnectionUiFeatures::default(),
         SessionOpenParams {
             session_id: session_id.clone(),
@@ -14358,6 +14515,7 @@ async fn session_open_replays_pending_approval_after_reconnect_without_cursor() 
         &approvals,
         &PendingQuestionStore::default(),
         ConnectionId::next(),
+        None,
         None,
         ConnectionUiFeatures::default(),
         SessionOpenParams {
@@ -14441,6 +14599,7 @@ async fn session_open_replays_pending_question_for_negotiated_client() {
         &approvals,
         &questions,
         ConnectionId::next(),
+        None,
         None,
         features_with_user_question_v1(),
         SessionOpenParams {
@@ -14776,6 +14935,7 @@ async fn session_open_does_not_duplicate_pending_question_already_in_cursor_repl
         &questions,
         ConnectionId::next(),
         None,
+        None,
         features_with_user_question_v1(),
         SessionOpenParams {
             session_id: session_id.clone(),
@@ -15014,6 +15174,7 @@ async fn session_open_does_not_duplicate_pending_approval_already_in_cursor_repl
         &PendingQuestionStore::default(),
         ConnectionId::next(),
         None,
+        None,
         ConnectionUiFeatures::default(),
         SessionOpenParams {
             session_id: session_id.clone(),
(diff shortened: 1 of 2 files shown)


ymote:
## Adversarial review — DO-NOT-MERGE (round 1)

Reviewed against six numbered hypotheses, evidence required to refute. **H1, H3 and H6 REFUTED** — the `None → _main` normalization, the top-level-then-nested fallback (including the `LoopFired`-has-a-nested-record / `MonitorFired`-does-not asymmetry), and the absence of panics/ordering/hot-path hazards all hold. The variant enumeration was independently confirmed complete.

Two MUST-FIX findings.

### H2 — the commit introduces a starvation regression

The connection profile is not per-session immutable scope. `WsConnection` stores **one shared `Arc<RwLock<String>>`** (`ui_protocol_transport.rs:559`); every successful `session/open` overwrites it (`:16876`); a connection retains **multiple** forwarders keyed by session (`:423`); and each forwarder reads the connection-wide *current* value on every event (`:17240`).

So an admin connection that opens session A under profile A, then session B under profile B, rewrites the scope observed by A's already-running forwarder. Every subsequent goal/loop/monitor event for session A is dropped. Those variants previously hit `_ => true` and were delivered, so this is a regression the commit creates.

This is not the `_main`-as-wildcard question — that reasoning stands. The defect is a per-session scope stored per-connection. The precedent for the fix is in the same function: `topic_scope` is already captured immutably per forwarder at `:17200`.

### H4 — five deferred variants are live cross-tenant leaks

All durably emitted, all carrying tenant text: `AgentUpdated` (`ui_protocol.rs:5907`, emitted `:4639`), `PeerStaged` (full brief + absolute paths, `ui_protocol.rs:6508`, emitted `:30454`), `PeerClosed` (`:30659`), `BackgroundActivity` (raw observed text, `ui_protocol.rs:6577`, ledger `:20534`), `SessionOpened` (`active_profile_id`, workspace root, pane snapshots — ledger-appended for broadcast at `:18056`).

`LoopCompleted` and `MonitorExpired` accepted as tracking-only: no production construction site, and `book/src/advanced.md:512` says outright that `LoopCompleted` is not currently emitted.

### H5 — CONFIRMED, tracking only

Live-forwarder coverage is **genuine** — `handle_session_open` installs the forwarder at `:17007` before the test appends its live events (`ui_protocol_tests.rs:13058`). But boundary 2 is not independently pinned: boundary 1's `replay.retain` (`:17969`) already removes foreign events, and boundary 2 (`:16917`) re-filters that same vector with the same predicate and profile. **Deleting boundary 2's check would leave all four new tests green.**

### Incidental

`resolve_autonomy_profile_id` checks `is_empty()` rather than `trim().is_empty()`, so an unscoped raw request can mint a whitespace pseudo-profile. It cannot collide with a real owner (registered ids exclude whitespace, `profiles.rs:2605`), but `optional_profile_scope_matches` *does* trim — the two disagree. Noted, not chased.

---

Fixing H2 first, then extending to the H4 variants that can be verified — extending the filter while the per-forwarder scope is unsound would multiply the starvation blast radius, and starvation risk is exactly what motivated two of the original deferrals. Any variant whose profile stamp cannot be proven reliable stays deferred with evidence rather than being filtered blind.

ymote:
## Adversarial re-review — round 2 — DO-NOT-MERGE

Scoped to the two round-1 findings. Most of the round-2 work is vindicated.

**REFUTED (i.e. correct as written):**

- **H2 is properly fixed.** `outcome.profile_id` is captured once at `:16858`, passed at `:16996`, owned as an immutable `String` by the spawned task at `:17217`, and used directly at `:17270`. A tree-wide search found no remaining mutable connection-profile field or accessor. Exactly one production call site; the other 19 are test-only.
- **The replay boundaries never had the hazard.** B1's `ledger_profile_id` derives from the same call's `params.session_id`, requested profile and connection profile (`:17786`); that same local is copied onto the outcome at `:18102`, which B2 reads at `:16906`. Neither can observe another session's later open.
- **The `SessionOpened` safety case holds on all three legs** — the append is tagged with the opener's connection id (`:18080`), the forwarder self-skips *before* profile filtering (`:17264`), and every successful open directly sends its freshly appended event (`:16990`) via `send_ledger_event_durable`, which has no profile filter (`:37080`). Reconnect uses the same open path; stdio invokes the same handler (`:6777`).
- **`BackgroundActivity` has only two production construction sites**, neither passing absent, empty or non-owner state.

### R5(b) — MUST-FIX: the divergence reaches variants that WERE filtered

The claimed safe partition — "everything filtered stamps via `resolve_autonomy_profile_id` and is immune by construction" — is false. Model-created monitors and fleets take their profile from `ProfileRuntime`.

An admin connection has `connection_profile_id = None` while retaining an authorized header/subdomain route `P` (`:5600`). `session/open` receives only the connection scope (`:6011`), so a bare session resolves to `_main` (`:16712`). The turn is given routed `P` (`:6038`) and resolves its runtime from `connection_profile_id.or(routed_profile_id)` (`:19548`). That runtime registers `MonitorCreateTool` with `profile.id` (`runtime/profile.rs:1321`); the tool persists `P` directly (`goal_tool.rs:1800`); its continuation emits durable `MonitorFired(P)` (`:20046`, `:35941`). When the detached global drain wins there is no real-client direct send, and reconnect replay under captured `_main` drops it at the `MonitorFired` arm (`:17141`).

**`MonitorFired`, filtered in round 1, is already starved.**

### R6 — MUST-FIX: new damage in the second commit

`BackgroundActivity`, newly filtered at `:17165`, is starved on the same routed-admin monitor path — and its sink uses a detached connection whose direct writer is discarded (`:20552`), so replay/live forwarding is its *only* delivery. Both now drop it. New in commit 2, since the variant previously fell through `_ => true`.

The new test cannot expose it: it opens with a matching `connection_profile_id = Some("alpha")` (`ui_protocol_tests.rs:13711`) rather than the `connection = None, routed = P` shape that triggers it.

---

Round 3 is addressing the constraint rather than the mechanism: **close the leak without introducing starvation on any path.** Two candidate shapes under evaluation — close the divergence at the source so `session/open` resolves scope the way the turn path already does (the re-review's recommendation, and the real fix, since a session scoped `_main` whose own turns run as `P` is incoherent on its own terms), or filter only where the connection's scope is authoritative, on the grounds that admin/routed connections are already explicitly authorized across profiles (`:5623`, `:5787`) and their scope was never a tenant boundary.

Either way the missing test is the same one the current suite cannot express: `connection_profile_id = None`, routed `P`, a durable frame stamped `P`. And the affected set needs establishing from the mechanism — "record created by a model tool under `ProfileRuntime`" — rather than from the two variants that happened to be named.

ymote:
## Adversarial re-review — round 3 — DO-NOT-MERGE

**Validated:** the exemption predicate does **not** reopen the leak (R2 refuted — no real tenant satisfies `connection_profile_id.is_none() && routed_profile_id.is_some()`; authenticated users get `Some(profile)`, admin tokens get `None` but are authorized for every profile, unauthenticated is rejected when auth is on and every route is exposed when it is off). The header-routed case is correctly threaded on all three boundaries (`:18089`, `:16917`, `:17007`/`:17355`). The affected-set audit is correct: all three model-goal repaints use `send_notification_ephemeral` (`:32959`, `:33876`, `:33927`) which does not append to the ledger (`:37124`), goal frames that *are* durable come via the raw RPC path (`:16262`) stamped through `resolve_autonomy_profile_id` (`:15545`) rather than the turn's `ProfileRuntime`, and `model_create_fleet_plan` (`agent_orchestrator.rs:7022`) really is the only non-test fleet creation.

### MUST-FIX — case 2 of the proof diverges after all

The residual noted as "pre-existing turn-routing, not filter-induced" is not separable. Every successful open overwrites the single `session_open_profile_id` (`:6035`); every later turn — including one on an *earlier* bare session — receives that latest value (`:6050`) and selects that runtime (`:29332`).

Open bare A as `_main`, open bare B as `beta`, then turn on A: A's forwarder retains `_main` while the turn runs under `beta`, a model-created monitor stamps `beta` (`runtime/profile.rs:1321`, `goal_tool.rs:1800`), and its detached `BackgroundActivity` sink relies **exclusively** on ledger replay/live fan-out (`:20645`) — where `_main` rejects the `beta` frame.

No routing header is involved, so `routed_profile_id.is_none()` and the exemption does not fire. **The turn misrouting predates this branch; the total delivery drop does not.**

Also, harmlessly: the exemption comment's authorization premise is false. `connection_profile_id = Some(P)` can coexist with routed `Q` — admin-role users may access every profile and parent users may access owned subaccounts (`auth_handlers.rs:3834`), and such users still produce `Some(P)` (`:16691`). No turn divergence results (validation returns `P` or rejects, `:16707`), but the reasoning would be cited as precedent.

### Direction for round 4

Collapse the predicate toward `connection_profile_id.is_some()` for the WebSocket path, with a carve-out only for transports that genuinely cannot multi-open under differing profiles. R2 is what makes that safe rather than a retreat: an unscoped connection is an operator connection authorized for every profile, so declining to filter it gives away nothing a tenant owns, while the isolation win is preserved exactly where it matters — every authenticated user connection, where `validate_session_scope` rejects any other profile's session key.

Regression to add: open A = `_main`, open B = `beta`, beta-stamped `BackgroundActivity` delivered to A on replay **and** live.

If that turns out to need per-session turn routing rather than a predicate change, the fallback is to descope this PR to the **goal variants only** — R4 established they carry no divergence risk — closing #2067 exactly as filed with zero starvation, and moving the loop/monitor/background extension to #2081 behind the resolution fix.

ymote:
## Adversarial re-review — round 4 — MERGE

All six hypotheses REFUTED on commit `5775866eb`. No MUST-FIX findings.

**R1 — the stdio disagreement was correct, and the round-3 recommendation was wrong.** Stdio genuinely drifts: the next profile is derived from each open (`:17277`), committed only after success (`:6804`), and the latest binding is fed into a bare session's turn (`:6808`), with runtime selection consuming it at `:29330`. After `ada` then `grace`, a turn on the first bare session runs under `grace`. Passing no filter pin is correct — preserving filtering there would have reintroduced exactly the starvation this round exists to close.

**R2 — session-ingress is genuinely pinned.** Its connection profile is computed once before upgrade (`:5692`); raw RPCs are denied (`:5935`); typed opens must resolve to the exact granted session (`:16626`, `:16680`); and with `Some(P)` both explicit-profile and profiled-session mismatches are rejected (`:16742`). Qualification, tracking-only: a profile-*less* grant can reopen the same permitted key under different profiles — but that path derives `None` and is unfiltered, so it cannot suffer the starvation. Every ingress that is actually filtered is frozen.

**R3 — case 2 is closed on all three boundaries.** Unscoped WS passes `None` (`:6018`) → `profile_scope = None` (`:17883`), which passes every event (`:17163`) across retained replay (`:18086`), the replay send loop (`:16920`) and the live forwarder (`:17348`). The two-open regression asserts both live and reconnect-replay delivery.

**R4 — the remaining filtered set is exactly right**: connections carrying `Some(connection_profile_id)` — authenticated users (`:5600`) and profile-bound ingress (`:5692`). Stdio explicitly passes no pin (`:6792`), and is a local stdin/stdout child-process transport rather than HTTP tenant ingress (`serve.rs:325`, `:1525`).

**R5 — the re-pinned tests are falsifiable.** The unprofiled test explicitly requires foreign `beta` frames; the relocated forwarder test injects matching and cross-profile frames on both subscribed streams and requires exactly the two matching results; the corrected routed test passes the production-equivalent `None` pin and asserts routed `beta` monitor, activity and goal replay plus live activity.

**R6 — both production construction sites correctly threaded**, with the owned scope carried consistently into replay and the spawned task (`:16879`, `:17010`).

The predicate that survived four rounds is not "scoped vs unscoped" but **"does this transport give the connection a profile it cannot drift from"** — WS and session-ingress pass their frozen `connection_profile_id`; stdio and unscoped WS pass `None`.

Issue #2055, opened 2026-08-18
background tasks and sub-agents never get a goal-ledger task row (only peer finding/denial/escalation create rows)


ymote:
## Summary

A goal-ledger `tasks` row is created **only** as a foreign-key side effect of three peer-specific paths. Background tasks and sub-agents that simply run and finish never appear in the goal ledger at all — so there is nothing for #2054's status writer to update, and `task_status_counts` under-reports the goal's actual work.

## Evidence

Every production `create_task` call site:

| site | enclosing fn | trigger |
|---|---|---|
| agent_orchestrator.rs:5010 | `model_goal_record_peer_finding` | peer recorded a finding |
| agent_orchestrator.rs:5145 | `record_fleet_write_grant_denial` | write-grant denial |
| agent_orchestrator.rs:5266 | `model_goal_record_peer_escalation` | peer escalation |

All three exist to satisfy `FOREIGN KEY (task_id) REFERENCES tasks(task_id)` on `findings` / `escalations` — the row is a byproduct of recording something else, not a record of the task.

Consequences:

- A **background task** (`spawn_only`) under a goal: no row, ever.
- A **sub-agent**: no row, ever.
- A **peer** that runs and produces no finding, denial, or escalation: no row, ever.
- A peer that does produce one: a row appears mid-life, with `status: "running"` frozen (#2054).

## Proposed fix

Create the row at **task registration**, for all three kinds, rather than at first finding:

- `TaskSupervisor` already registers every kind through its `register*` entry points, and #2034 gave peers a `TaskLivenessLease` at the same point — that is the natural hook.
- The goal id is available on both dispatch paths: `goal_context` for autonomous continuations, and the `interactive_goal_id` / `interactive_goal_binding` snapshot added by #1935 for interactive turns. (Note for anyone working from older notes: interactive turns are NOT missing a goal binding — that was true before #1935 and is no longer the blocker it used to be.)
- Keep the existing FK stubs as an idempotent upsert so a finding that arrives for an unregistered task still cannot violate the constraint.

Registration must be idempotent — the same task can be registered, relaunched, and re-registered across a restart.

Blocks #2054.

ymote:
## Design is determined by the crate graph — a callback is required

I went in expecting to add row-creation at the octos-cli spawn sites and avoid growing `TaskSupervisor`'s callback surface. The code rules that out. Every production registration site:

| site | task kind | crate |
|---|---|---|
[...]

PR #2076, opened 2026-08-20
fix(fleet): persist goal time_used_seconds in the ledger

### LINKED ISSUE #2068: goal ledger has no time_used_seconds column — wall-clock spend is non-durable for every goal
## Gap

`AutonomyGoalRecord.time_used_seconds` is tracked in memory (charged by every accountant: `record_goal_turn_internal`, `charge_goal_tokens_gated`) and persisted to the supervisor store, but the durable per-goal SQLite ledger has NO time column at all: `goals` schema (`sqlite_ledger.rs` `create_tables`) carries only `tokens_used`/`token_budget`, and `sync_transition_to_ledger_blocking`'s `octos_fleet::Goal` conversion silently drops the field. This is not a cleared-goal gap — the wall-clock dimension is non-durable for EVERY goal.

Surfaced by the #2066 round-2 review (codex MEDIUM): the post-clear tombstone settle initially accrued `time_used_seconds` in memory with nowhere to write it. Per the review's own option, that dead accumulation was DELETED in #2066 (accruing state that can structurally never persist is worse than not accruing); the durable fix is this follow-up.

## Fix sketch

- `ALTER TABLE goals ADD COLUMN time_used_seconds INTEGER NOT NULL DEFAULT 0` through the transactional schema-inspected migration pattern that already exists in this file (`migrate_tasks_authority_column`, #2059).
- Add the field to `octos_fleet::Goal` (serde `#[serde(default)]` for old snapshots), `create_tables`, all INSERT/upsert statements, `get_goal`'s SELECT.
- Include time in the transition sync's conversion and in the post-clear delta settle (`settle_cleared_goal_cost_delta` grows a `time_delta` param — additive, same commutation argument as tokens).

## Acceptance

- A goal transition syncs `time_used_seconds` into the ledger row.
- A cleared-mid-turn goal's row includes the in-flight turn's elapsed seconds after settle.
- Old ledger files migrate in place (schema-inspected, transactional) and old serialized `Goal` values deserialize.

Provenance: #2066 round-2 review (codex MEDIUM), refs: `crates/octos-fleet/src/sqlite_ledger.rs` (`create_tables`), `crates/octos-cli/src/autonomy/agent_orchestrator.rs` (`sync_transition_to_ledger_blocking` conversion, `accrue_cleared_goal_tombstone_charge`).

ymote:
Closes #2068

## The gap

`AutonomyGoalRecord.time_used_seconds` is charged in memory by every accountant (`record_goal_turn_internal`, `charge_goal_tokens_gated`) and persisted to the supervisor store, but the durable per-goal SQLite ledger had **no time column at all**: `create_tables`' `goals` schema carried only `tokens_used`/`token_budget`, and the `octos_fleet::Goal` conversion in `sync_transition_to_ledger_blocking` silently dropped the field. The wall-clock dimension was non-durable for **every** goal, not just cleared ones.

I verified every claim in the issue against the code before changing anything. All of them hold. One thing the issue's sketch does not mention, which the code forced:

> **A migration confined to `open_with_busy_retry` does not reach the terminal writes.** `tasks.authority` (#2055) could confine every column-naming statement to the retry profile. The goals writers are not confinable: the clear stamp (`stamp_goal_cleared_blocking`), the post-clear settle (`offload_cleared_goal_settle`) and the sentinel-completion transition all open with the plain `open`, and `create_tables` is `CREATE TABLE IF NOT EXISTS`, so a pre-#2068 file is left completely untouched by them.
>
> Round 1 of this PR handled that with a legacy statement-shape fallback. **That was wrong** and round 2 replaced it: on a terminal path a fallback that avoids erroring while dropping the value is the failure mode, not the fix. Clearing replaces the supervisor record with a zero-time tombstone, the goal is terminal, and nothing re-syncs it — so an existing deployment's ledger would have gained a `time_used_seconds` column that stayed `0` forever. See "How the migration reaches every write" below.

## What changed

**`crates/octos-fleet/src/sqlite_ledger.rs`**

- `goals.time_used_seconds INTEGER NOT NULL DEFAULT 0`, declared **last** in `create_tables` so a freshly created table and a migrated legacy one end up with identical column order (`ALTER TABLE … ADD COLUMN` appends), and no reader can tell which schema generation it opened.
- `migrate_goals_time_column`, copying the shape of `migrate_tasks_authority_column` (#2059): the step is decided by `pragma_table_info` (never an error-message substring), the ALTER runs inside one `BEGIN IMMEDIATE` transaction so a failure rolls back to the pre-migration shape, and the pre-transaction fast path is repeated inside the transaction because two blocking-context openers can race it.
  - **No backfill**, unlike `authority`: a legacy row's wall-clock spend was never recorded anywhere in the ledger, so there is nothing to derive it from and the column default (`0`) is the only honest value. The counter is monotonic and MAX-merged by its writers, so a legacy row simply resumes accruing from zero.
- **How the migration reaches every write.** `ensure_goals_time_column` brings the column into existence from inside the `BEGIN IMMEDIATE` transaction each writer *already holds* — `create_goal`, `upsert_goal`, `create_goal_if_absent`, `stamp_goal_cleared` and `settle_cleared_goal_cost_delta` all call it, and each now has exactly one statement shape instead of two.
- **`get_goal` is deliberately excluded** — it keeps its read-only schema-inspected fallback, takes no transaction and runs no DDL. That is the invariant #2055/#2059 actually protects: a *reader* must never take a write lock, and #2059's migration added a separate `BEGIN IMMEDIATE` transaction plus a table-wide backfill `UPDATE` to exactly that reader path (`goal_get` and every inline consult, on tokio workers).
- **Why the writer-side DDL is safe.** A writer is already inside `BEGIN IMMEDIATE`, so it already holds the RESERVED lock and already paid the wait to acquire it — this adds no lock acquisition. `ALTER TABLE … ADD COLUMN` is O(1) in SQLite (schema-row write; existing rows are treated as absent-defaulted, the table is not rewritten) and there is **no backfill**, so the marginal work is one small schema write, once per file, gated by the inspection.
- **Alternative considered and rejected**: move the clear stamp and the sentinel transition onto the blocking pool so they could use `open_with_busy_retry`. Both run synchronously inside `handle_raw_appui_rpc` / the session actor, so that is a threading *and ordering* change — the clear RPC would return before its durable row was stamped. Strictly larger and riskier than one O(1) DDL statement inside a transaction those paths already take.

**`crates/octos-cli/src/autonomy/agent_orchestrator.rs`**

- The transition sync, the clear stamp, the post-clear settle base row, and the four peer/fleet FK-parent `goal_row` builders all carry `time_used_seconds` from the in-memory record instead of dropping it.
- `settle_cleared_goal_cost_delta` grows `frozen_base_time_seconds` / `time_delta`; `offload_cleared_goal_settle` passes both.
- `accrue_cleared_goal_tombstone_charge` accrues elapsed seconds (both accountant call sites already had `elapsed_seconds` in scope), and its zero-guard becomes `tokens == 0 && elapsed == 0` so an **elapsed-only** turn still settles — matching `charge_goal_tokens_gated`'s own guard. `ClearedGoalTombstone` gains `settled_time_seconds` for the eviction/purge logs; #2066 round 2 deleted the in-memory accrual because it could structurally never persist, and that reason is now gone.

### Cost, and the follow-up

Every goals write now pays one `pragma_table_info` lookup (served from SQLite's in-memory schema cache, no I/O) inside a `BEGIN IMMEDIATE` transaction, where most were single autocommit statements before — the same cost `create_task`/`create_task_if_absent` have paid since #2055. `get_goal` pays the lookup in front of an indexed point lookup. The capability is *monotone* (absent → present, never back), so the positive answer is cacheable; measuring and eliminating the repeat scan is filed as **#2085** rather than done here, to keep this diff to the correctness fix.

## Commutation decision for the time delta: MAX-merge, and it is *required*

`settle_cleared_goal_cost_delta` writes `time_used_seconds = MAX(time_used_seconds, B_t) + D_t`, mirroring tokens. Not "same as tokens" by assertion — two properties carry it, each checked independently for seconds:

1. **Monotonic per `goal_id`.** Every in-memory writer only ever `saturating_add`s `time_used_seconds` (`record_goal_turn_internal`, `charge_goal_tokens_gated`), a replacement goal mints a fresh `goal_id`, and re-activation never resets it.
2. **Each late charge is a positive increment.**

The identity then holds for **any** row value `L`: stamp→settle = `MAX(L,B)+D`; settle→stamp = `MAX(MAX(L,B)+D, B) = MAX(L,B)+D`, since `MAX(L,B)+D ≥ B` for `D ≥ 0`.

> **Round-2 correction (codex H1).** An earlier version of this section leaned on a third property — that the row *lags* the frozen base (`L ≤ B`, because nonterminal turns never sync). That is the operational reality but it is **not needed** for the identity; the algebra commutes either way. The doc comments have been corrected. Conversely, the "delivered exactly once" premise I listed as a property is **false operationally** — see the scope note below.

**Plain accumulation would be order-dependent, and in the common case.** If the settle did `time_used_seconds = time_used_seconds + D`, then settle-first stores `L+D`, and the later stamp's MAX-merge collapses it to `MAX(L+D, B) = B` whenever `L+D ≤ B` — the delta silently eaten — while stamp-first stores `B+D`. Since the lag `B−L` is a whole goal's accumulated history and `D` is one turn's elapsed seconds, `L+D ≤ B` is the *normal* case. `should_settle_identical_time_when_the_clear_stamp_and_delta_arrive_in_either_order` pins this on the adversarial fixture the token pin uses.

**Scope of the guarantee (#2084, tracking).** This commutes over **delivered** deltas only. A tombstone evicted at the cap (or purged past its horizon) makes a later charge resolve to nothing, and the settle's own failures are logged and dropped. With a drop in the mix the settled row *is* order-dependent — `L=0, B=100, D=10`: settle→evict→stamp = 110, stamp→evict→settle = 100. Durable charge identities with retry/dedupe are filed as #2084; the doc comments now state the narrower guarantee.

## Known loss, documented not fixed (#2083)

`upsert_goal` takes an **absolute snapshot**, so the MAX-merge keeps the stored value from going backwards but does not guarantee the increment lands. When the writer's own base lags the durable row, the increment is unrecoverable from the two values SQL can see. Reachable today: `persist_goal_state_with_store` discards store-write errors (`let _ = store.append_event(…)`) and goal restoration trusts the supervisor metadata (`supervisor_metadata_u64(…, "time_used_seconds").unwrap_or(0)`), so memory can restart behind the ledger. Row `time = 20`, memory restored at `10`, a turn adds `5` → snapshot `15`, admitted because the token clause passes, and `MAX(20, 15)` stores `20` when the true value is `25`.

**This is not specific to time** — it is a property of the absolute-snapshot sync. A lagging *token* base makes the monotonic clause reject the whole write instead, which also fails to record the increment. #2068 inherited the class; it did not create it. The cleared-goal settle path does **not** have this hole (it writes true deltas).

Fixing it means delta-aware settlement, or reconciling the writer's base against the durable row inside one transaction — a change to how the orchestrator syncs, filed as **#2083** and out of scope here. It is called out in a `KNOWN LOSS` block at the MAX-merge in `upsert_goal` so the "wall-clock spend is now durable" headline is not read as covering it.

**Related decision: `upsert_goal`'s admission guard is deliberately NOT extended.** Adding `AND excluded.time_used_seconds >= goals.time_used_seconds` would make the guard *stricter* and newly reject whole writes that land today: each process advances the counters from its own in-memory record, so a peer snapshot can legitimately carry higher tokens and lower seconds, and a row-level rejection would throw away its token update too. The per-column MAX-merge gives the new dimension the same monotonic protection clause 2 gives tokens **without changing which writes are admitted**. (Reviewer agreed.)

## Tests (RED first, both rounds)

**Round 1 RED**, in order: (1) tests written first → `cargo test -p octos-fleet` failed to compile, 25 errors (`no field time_used_seconds on type sqlite_ledger::Goal`, `this method takes 4 arguments but 6 were supplied`); (2) struct field + signature added, SQL still absent → 4 assertion failures, every one `left: 0` (`create_goal must persist the wall-clock dimension  left: 0 right: 42`; `B(100) + D(10)… left: 0 right: 110`); (3) SQL + migration added → 186 passed.

**Round 2 RED** (codex H5(d) — the round-1 suite passed *while H6 was present*, because the legacy test asserted `time_used_seconds == 0` and both orchestrator tests used fresh ledgers):

- `should_migrate_on_first_write_when_the_ledger_predates_the_time_column` → `assertion left == right failed: a write migrates  left: 0  right: 1`
- `should_settle_time_on_a_cleared_goal_when_the_ledger_predates_the_time_column` → `assertion left == right failed: an EXISTING ledger must not lose the goal's wall-clock spend…  left: 0  right: 67` — note tokens settled to 6_234 correctly in the same run; only time was lost, exactly as H6 described.

Both then passed after `ensure_goals_time_column` landed.

`crates/octos-fleet/src/sqlite_ledger.rs`
- `should_round_trip_time_used_seconds_when_a_goal_row_is_written` — create/upsert/`get_goal` round trip plus the stale-lower-time monotonicity pin.
- `should_default_time_used_seconds_when_deserializing_a_pre_2068_goal` — old serialized JSON without the field still deserializes.
- `should_settle_identical_time_when_the_clear_stamp_and_delta_arrive_in_either_order` — the commutation pin, including a time-only (zero-token) charge.
- `should_migrate_goals_time_column_when_opening_a_pre_2068_ledger` — builds the **real** pre-#2068 `goals` shape with a raw `CREATE TABLE`; asserts every pre-existing column survives, that the plain `open` runs **no** DDL, that the retry open migrates in place, and that re-opening is a clean no-op.
- **(round 2, replaces the test that blessed the bug)** `should_migrate_on_first_write_when_the_ledger_predates_the_time_column` — reader-first on an unmigrated file (reads 0, must **not** migrate), then the first write migrates in place and records; the full writer sequence lands `137`; the pre-existing sibling row survives untouched.

`crates/octos-cli/src/autonomy/agent_orchestrator.rs`
- `should_sync_goal_time_used_seconds_into_the_ledger_when_a_goal_transitions` — a real interactive charge then a real production transition (the solo-boot park) syncs 45s into the row.
- `should_include_the_mid_flight_elapsed_seconds_when_the_goal_was_cleared_mid_turn` — the clear stamps the pre-turn 60s, the in-flight turn's 7s settles on top, a second late (verifier) charge accumulates, and an elapsed-only charge settles with zero tokens.
- **(round 2)** `should_settle_time_on_a_cleared_goal_when_the_ledger_predates_the_time_column` — an **existing** pre-change ledger file, built raw at exactly the path the orchestrator resolves, driven entirely through the production clear + settle path.
- `record_goal_turn_should_settle_charge_into_cleared_goal_row` gains the wall-clock assertion for the autonomous accountant.

Round-1 mutation checks (both CLI tests proven load-bearing): reverting `time_used_seconds: snapshot.time_used_seconds` → `0` in the transition sync gives `left: 0 right: 45`; zeroing the settle's time args gives `left: 60 right: 67`.

## CI

No filter changes needed; both suites run my tests with **no name filter**, verified against the workflow and by running the exact commands:

- `.github/workflows/ci.yml` `Test octos-fleet` → `cargo test -p octos-fleet` (unfiltered) — all 5 new fleet tests appear in the run list.
- `.github/workflows/ci.yml` `Test octos-cli (lib)` → `cargo test -p octos-cli --lib` (unfiltered). `autonomy` is `#[cfg_attr(not(feature = "api"), allow(dead_code))] pub(crate) mod autonomy;` in `lib.rs`, **not** `#[cfg(feature = "api")]`, so both new tests compile and run in the unfeatured lib step — confirmed by running them without `--features api`.

Verification re-run locally after round 2, strictly sequential (never `cargo test` and `cargo clippy` concurrently):

```
cargo fmt --all -- --check                                        → exit 0, no output
cargo test -p octos-fleet                                         → test result: ok. 186 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
cargo test -p octos-cli --lib                                     → test result: ok. 1525 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out
cargo test -p octos-cli --tests                                   → 20 targets, all "test result: ok", 0 failures
cargo test --workspace --doc                                      → 26 targets, all "test result: ok", 0 failures
cargo test -p octos-cli --features api api::auth_handlers         → test result: ok. 83 passed; 0 failed
cargo test -p octos-cli --features api -- completion_evidence goal_update_folds_ledger
                                                                  → test result: ok. 6 passed; 0 failed
cargo test -p octos-cli --features api peer                       → test result: ok. 146 passed; 0 failed
cargo clippy -p octos-fleet --all-targets -- -D warnings          → Finished `dev` profile (no warnings, no errors)
cargo clippy -p octos-cli --features api --all-targets -- -D warnings → Finished `dev` profile (no warnings, no errors)
cargo clippy --workspace --all-targets -- -D warnings             → Finished `dev` profile (no warnings, no errors)
```

The round-1 commit's CI was fully green on all 15 checks (`check` 12m16s, `test-octos-cli` 12m59s, `check-windows` 44m37s, both `test-octos-agent` shards, `security`, `typos`, …).

**Pre-existing flake, not from this PR:** `autonomy::agent_orchestrator::tests::default_mode_background_nested_spawn_lands_grandchild_row` fails intermittently under a full parallel `--lib` run. I reproduced it on **pristine `main` (9887d84e7) without this branch's changes** — 1 failure in 3 consecutive base-commit runs, same test, same first assertion (`the outer background child has a row …; rows: ["nested-grandchild"]`), i.e. the grandchild row lands and the outer child's `create_task_if_absent` is lost. Filed as #2077.


ymote:
## Adversarial review — DO-NOT-MERGE (round 1)

Reviewed against six numbered hypotheses, evidence required to refute.

**The migration work holds up.** H2 REFUTED with a statement-by-statement enumeration — every production statement naming the column is protected on the plain-`open` path (fresh DDL `:297`, capability query `:456`, `ALTER` `:562`, the INSERT head used by `create_goal`/`upsert_goal`/`create_goal_if_absent` `:592`/`:599`, upsert SET list `:720`/`:747`, `get_goal` SELECT `:927`, clear stamp `:1765`/`:1777`, delta settle `:1874`/`:1889`). H3 REFUTED — schema-inspected both before and after acquiring the write transaction (`:563`, `:567`), `ALTER` inside `BEGIN IMMEDIATE` (`:566`, `:573`), openers serialized with BUSY/LOCKED retry (`:257`), invoked only by `open_with_busy_retry` (`:248`), and no tokio-worker DDL path introduced.

Three MUST-FIX findings.

### H6 — an existing ledger can lose all seconds through its final write

A real pre-change ledger can stay unmigrated through its **terminal** write: the clear stamp opens with plain `open` (`agent_orchestrator.rs:5246`), the post-clear settle likewise (`:5169`), sentinel transitions explicitly choose the plain profile (`:4363`), and the legacy branches omit time entirely (`sqlite_ledger.rs:1777`, `:1890`). Clearing simultaneously replaces the supervisor record with a zero-time tombstone (`:13002`).

For frozen base 60 and late delta 7, an old ledger ends at **0**, not 67; a later migration only adds the default zero. There is no guaranteed subsequent blocking open once a goal is terminal.

The legacy-shape fallback correctly prevents a hard SQL error — but on the terminal path, a fallback that silently drops the value is the failure mode rather than the fix. As it stands the feature works only on ledgers created after this change.

### H5(d) — the tests bless it

(a), (b) and (c) are genuine: the migration fixture builds the raw historical shape (`sqlite_ledger.rs:4925`), both orders call the real production stamp and settle (`:2897`, `:2903`), and plain-open behaviour is established by re-reading the schema rather than by timing (`:4983`).

(d) fails for the core legacy production path. The legacy fallback test explicitly accepts `time_used_seconds == 0` (`:5107`) and both orchestrator tests create fresh ledgers — so the suite passes green with H6 present.

### H4 — an admitted upsert can lose a new time charge

Ledger `tokens=100, time=20`; supervisor state lags at `time=10` (persistence errors discarded `:13122`, restoration trusts the lower value `:12725`); a new turn adds `tokens=1, time=5` (`:3636`) producing snapshot `101/15`; admitted because `101 >= 100` and the timestamp is newer (`sqlite_ledger.rs:729`); `MAX(20,15)` (`:727`) leaves 20 where the cumulative value is 25.

The stored value never decreases, so the monotonicity argument survives — but the five new seconds vanish. Codex agrees that adding a `time >=` admission clause would be **worse**; that part of the PR's reasoning stands. The fix is delta-aware settlement or reconciling the writer's base against the durable row.

### H1 — CONFIRMED, tracking

Codex went further than the PR body did: `L <= B` is **unnecessary** — the SQL commutes either way, since `MAX(MAX(L,B)+D, B) = MAX(L,B)+D`. What is false is property (3), that each late charge is delivered exactly once: tombstones can be evicted while a charge is outstanding (`agent_orchestrator.rs:9804`, accrual then returns `None` at `:9880`), and SQLite failures explicitly drop charges (`:5210`). With `L=0, B=100, D=10`: settle→eviction→stamp reaches 110; stamp→eviction→settle stays at 100.

Issue to file: give cleared-goal charges durable identities plus retry/deduplication, or limit the commutation guarantee to successfully delivered deltas.

### Also tracking

Every `get_goal` now runs `pragma_table_info` before its indexed point lookup (`sqlite_ledger.rs:925`) — measure and eliminate repeated positive-capability schema scans after migration.

No serde, overflow, sign, or newly introduced panic defect found.

ymote:
## Round-2 verification by inspection

The scoped re-review stalled without returning (over an hour, no output — the same silent-hang pattern seen earlier today). Rather than merge blind or block indefinitely, I checked the decisive H6 claim directly: **is every executed statement naming `time_used_seconds` preceded by `ensure_goals_time_column` inside the transaction it already holds?**

It is.

Six call sites — `sqlite_ledger.rs:570, 684, 794, 819, 1776, 1858` — covering `create_goal`, `upsert_goal`, `create_goal_if_absent`, `stamp_goal_cleared` and `settle_cleared_goal_cost_delta`, each invoked as `Self::ensure_goals_time_column(&tx)?` on the writer's existing `Transaction` rather than opening its own.

Mapping every reference in the file to its enclosing function, two appear to precede their `ensure` — `upsert_goal:786` and `stamp_goal_cleared:1760/1768`. Both are **`const` SQL string declarations**, not execution points; the statements run after the ensure. `commit_state_with_audit:1745` is prose. So there is no executed path that can reach the column on an unmigrated table.

`get_goal` remains correctly excluded and unchanged (`:918-945`): it inspects `goals_has_time_column`, branches to a pre-#2068 SELECT that names only columns which already existed, and reports `time_used_seconds: 0` — the same value the freshly-added column would hold. No transaction, no DDL, no write lock on a reader path.

The doc block at `:583-597` reads at first like a stale legacy-fallback note, but it is a deliberate record of why the round-1 approach was wrong — *"on a terminal path 'did not error' and 'recorded the value' are different outcomes"*. Worth keeping.

### What that leaves unverified

The re-review would also have re-scanned for new defects introduced by deleting the legacy writer branches, and re-checked the H4 documentation. Those rest on the author's own mutation evidence (the replacement test reproduces the bug at `left: 0, right: 67`, with tokens settling to 6_234 correctly in the same run as a control) plus full CI. H4 remains documented-not-fixed by design, tracked as #2083, with #2084 and #2085 for the delivery and schema-scan items.

CI is green: 12 pass, 8 skipping, 0 failing.

Issue #2073, opened 2026-08-20
e2e-live-nightly is 40-failed and invisible: m9 protocol specs have never connected in CI (OCTOS_LIVE_URL unset)


ymote:
## Summary

The nightly `e2e-live-nightly` job reports **40 failed / 8 skipped / 18 passed** ([run 32241217422](https://github.com/octos-org/octos/actions/runs/32241217422), 2026-08-19, on `e87f1c061` = v2.0.3-rc.8). The job is `continue-on-error: true`, so nothing surfaces and nobody has looked.

The `continue-on-error` justification (ci.yml:1137-1147, from #1923) is sound *for what it describes*: specs that ask a live model for particular words and assert on them exactly. But the step actually runs the **entire** playwright suite unfiltered —

```yaml
- name: Run model-dependent e2e specs
  run: OCTOS_TEST_URL=http://localhost:3000 OCTOS_AUTH_TOKEN=ci-test-token npx playwright test --reporter=line
```

— so structural, model-independent specs (m9 protocol, contract-doc pins, solo login, admin setup wizard) get swept into a non-blocking job on a model-flake rationale that does not apply to them. Only ~5 of the 40 failures are plausibly model-dependent.

This is the same shape as #2016 (e2e gate reporting green without running) and #2029 (a test matching no CI filter never runs and the job prints ok).

### 1. ~30 m9-protocol specs have NEVER connected to the CI server (`OCTOS_LIVE_URL` is never set)

`e2e/lib/m9-ws-client.ts:403`:

```ts
const url = process.env.OCTOS_LIVE_URL || "http://127.0.0.1:56831";
```

`OCTOS_LIVE_URL` appears **zero** times in `.github/workflows/ci.yml`. Its only setter in the whole repo is `e2e/tmux/run.sh:746` — a manual tmux harness. CI exports `OCTOS_TEST_URL` and serves on **port 3000**, so every m9 spec dials the hardcoded 56831 fallback and gets:

```
Error: m9-ws: connect error: connect ECONNREFUSED 127.0.0.1:56831
```

The token check one line below (`OCTOS_LIVE_TOKEN || OCTOS_AUTH_TOKEN`) *does* fall through to the CI token, which is why the harness throws a connect error instead of the "must be set" error that would have made this obvious years ago.

[...]

Issue #2077, opened 2026-08-20
flaky: default_mode_background_nested_spawn_lands_grandchild_row loses the outer child's task row under parallel load


ymote:
## Symptom

`autonomy::agent_orchestrator::tests::default_mode_background_nested_spawn_lands_grandchild_row` fails intermittently under a full parallel `cargo test -p octos-cli --lib` run:

```
panicked at crates/octos-cli/src/autonomy/agent_orchestrator.rs:18787:9:
the outer background child has a row (parent supervisor wiring); rows: ["nested-grandchild"]
```

The GRANDCHILD row lands; the OUTER child's row does not. So it is not a general "nothing ran" timeout — the detached-branch observer inheritance the test exists to prove works, and the parent-supervisor registration is the half that goes missing.

## Reproduction / provenance

Observed while verifying #2076 (goal ledger time column). Reproduced on **pristine main at 9887d84e7, with no PR changes applied**: 1 failure in 3 consecutive full-suite runs (and 2 in 5 runs on the PR branch, same test, same assertion). Passes 100% of the time in isolation (`cargo test -p octos-cli --lib -- default_mode_background_nested_spawn_lands_grandchild_row`), so it is load/scheduling sensitive, not deterministic.

Environment: macOS arm64, default `--test-threads` (no `RUST_TEST_THREADS` pin — the `test-octos-cli` CI job does not set one either, unlike `test-octos-agent`).

## Why it matters

`Test octos-cli (lib)` is an unfiltered CI step, so this is a live source of red runs on unrelated PRs.

## Likely area

Both registrations write the same per-goal ledger file from separate blocking-pool threads (`record_goal_task_row_blocking` → `create_goal_if_absent` + `create_task_if_absent` through `open_with_busy_retry`). A lost write there is swallowed by design — the failure path is `tracing::debug!("goal task-row registration write failed; settle will converge if the task finishes")` — so a genuinely dropped row is invisible in the test output. Worth checking whether the outer registration is (a) losing a lock race whose error is debug-logged away, or (b) simply not fired yet when the 60s poll (which breaks on the GRANDCHILD title only) exits.

ymote:
**This is now blocking unrelated PRs.** It just failed `check-windows` on #2087 — a PR that touches **only e2e TypeScript specs** and cannot affect Rust tests at all:

```
thread 'autonomy::agent_orchestrator::tests::default_mode_background_nested_spawn_lands_grandchild_row'
[...]

PR #2087, opened 2026-08-20
test(e2e): fix the ambiguous wizard locator, gate two specs on prerequisites CI never had

ymote:
Triage of three more of the 40 nightly failures from #2073. None is a product regression.

## `admin-setup-wizard` — a prose edit broke a selector

```
Error: locator.check: strict mode violation: getByLabel('Tenant') resolved to 2 elements:
  1) <input type="radio" value="tenant"> aka getByRole('radio', { name: 'Tenant Joins a central VPS' })
  2) <input type="radio" value="cloud">  aka getByRole('radio', { name: 'Cloud This node IS the VPS' })
```

Playwright matches label text as a **case-insensitive substring**, and the Cloud option's description reads *"terminates tunnels from **tenants** and issues subdomains"* (`dashboard/src/pages/wizard/StepDeploymentMode.tsx:133`). So `'Tenant'` matches both.

The markup is fine — each radio has its own wrapping `<label>` — and nothing about the wizard changed. Fixed by targeting the radio by value, exactly as the assertion two lines below already does. The `Password` field on the same page already used an anchored regex for precisely this reason, so the pattern was known.

## `solo-login` — prerequisites no CI job provides

Its own docblock says what it needs: `octos serve --solo` on a fresh data dir, plus a `vite dev` server proxying `/api` to it, with `OCTOS_TEST_URL` pointed at vite. `e2e-live-nightly` provides **none** of them — it starts serve with no `--solo` and points the suite at :3000. Solo login is off by default, so `solo-continue` never renders and the spec has never tested anything.

Now skipped unless `OCTOS_SOLO_E2E=1`, with the requirement stated in the skip message.

## `run-pipeline-ir-delivery` — a local-harness default, not a CI fixture

It defaults `PROFILE` to `'admin'`. CI sets no `OCTOS_PROFILE` and configures no such profile, so every run failed at `session/open` with `-32602 profile 'admin' is not configured for this AppUI session` before reaching a single assertion.

Now skipped unless `OCTOS_PROFILE` is set explicitly.

## On skipping

Skipping is not the end state for either — #2073 tracks giving CI the prerequisites so they run for real. But a spec that fails every night for a missing fixture is indistinguishable from no coverage, and it buries the failures that do matter. That is exactly how the m9 suite stayed invisible for months.

## Verification

```
npx playwright test tests/solo-login.spec.ts tests/run-pipeline-ir-delivery.spec.ts
  → 2 skipped
npx playwright test --list <all three files>
  → Total: 3 tests in 3 files
```

**Not verified:** `admin-setup-wizard` end-to-end. It spawns its own `octos serve` from a release binary at the repo root, which this worktree does not have built. The locator fix is established by the strict-mode error listing both matched elements plus the substring in the Cloud description; the run itself will confirm it on the next nightly.

### DIFF
--- a/e2e/tests/admin-setup-wizard.spec.ts
+++ b/e2e/tests/admin-setup-wizard.spec.ts
@@ -175,7 +175,13 @@ async function walkWizardAndAssertSmtpGating(page: Page, baseURL: string) {
       response.request().method() === 'POST' &&
       response.status() === 204,
   );
-  await page.getByLabel('Tenant').check();
+  // NOT `getByLabel('Tenant')`: Playwright matches label text as a
+  // case-insensitive SUBSTRING, and the Cloud option's description reads
+  // "terminates tunnels from tenants and issues subdomains" — so 'Tenant'
+  // resolves to BOTH radios and `.check()` fails with a strict mode
+  // violation. That is a prose edit breaking a test, not a UI change; target
+  // the radio by value, exactly as the assertion two lines down already does.
+  await page.locator('input[name="deployment-mode"][value="tenant"]').check();
   await modeSave;
   await expect(page.locator('input[name="deployment-mode"][value="tenant"]')).toBeChecked();
   await page.getByRole('button', { name: 'Back' }).click();
--- a/e2e/tests/run-pipeline-ir-delivery.spec.ts
+++ b/e2e/tests/run-pipeline-ir-delivery.spec.ts
@@ -14,6 +14,19 @@ const TOKEN =
   '';
 const PROFILE = process.env.OCTOS_PROFILE || 'admin';
 
+// The default profile above is a LOCAL-HARNESS default, not something CI has.
+// `e2e-live-nightly` sets no OCTOS_PROFILE and its serve configures no
+// `admin` profile, so every nightly run failed at session/open with
+// `-32602 profile 'admin' is not configured for this AppUI session` — before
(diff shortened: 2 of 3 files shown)


ymote:
`check-windows` failed here, and it is **not this diff** — this PR touches only e2e TypeScript specs and cannot affect Rust tests.

```
thread 'autonomy::agent_orchestrator::tests::default_mode_background_nested_spawn_lands_grandchild_row'
panicked at crates\\octos-cli\\src\\autonomy\\agent_orchestrator.rs:18787:9
test result: FAILED. 1484 passed; 1 failed; 3 ignored
```

That is **#2077** exactly, the flake reproduced on pristine main under full-parallel `cargo test -p octos-cli --lib`. It cost 42m50s to find out.

This is the tax #2077 predicted, so rather than re-run until green I have put a fix for the flake itself in flight. Holding this PR until that lands or until a re-run is justified on the record — not merging past a red check.

ymote:
Rebased onto `5456c1e64`, which carries the #2077 fix (#2090). The `check-windows` failure here was that flake — a late write, not a lost one: the outer task row landed 78 ms after the grandchild's while the test broke its poll on the grandchild alone. #2090 corrects the break condition to wait for both rows, mutation-verified that both assertions still bite, and ran 12 consecutive clean full-parallel suites against 1-in-8 on the base commit.

So this re-run is on a fixed test, not a hopeful retry.

PR #2088, opened 2026-08-20
fix(goal): reconcile stranded task→goal deliveries at supervisor restore (#2056)

### LINKED ISSUE #2056: task→goal status delivery is fire-and-forget: a crash between terminal and ledger write is unrecoverable
## Summary

Once #2054 bridges `TaskSupervisor::on_terminal` into the goal ledger, that write is in-process, synchronous, and fire-and-forget. A crash between a task reaching a terminal state and the ledger write landing loses the transition permanently — the goal is left believing a finished task is still running, with nothing to correct it.

## Why a plain callback is not enough

`on_terminal` fires from the supervisor's terminal paths and returns `()`. There is no retry, no acknowledgement, and no record that a delivery was owed. The supervisor's own row is already terminal by then, so on restart:

- the in-memory task table is rebuilt from its JSONL, terminal rows and all;
- the goal ledger still says `"running"`;
- nothing compares the two.

The orphan sweep in `enable_persistence` reaps non-terminal rows absent from the live set, but it reasons about supervisor state only — it has no view of the goal ledger.

## Options

1. **Boot-time reconciliation sweep** (cheapest, and probably enough): when a goal is loaded, cross-check its non-terminal ledger task rows against supervisor state, and repair rows whose task is terminal or absent. Bounded work, no new persistence, and it also repairs rows lost to any other cause.
2. **Durable outbox**: persist the intent, then deliver, then mark delivered. Correct under crash, but introduces a second durable store and its own compaction/cleanup burden.

Option 1 is proposed unless a case appears where a lost transition must be observed before the next goal load.

## Note on scope

This is deliberately filed separately from #2054 so the status writer can land and be verified on its own. A writer that works in the common case is strictly better than today's permanent `"running"`, and the durability gap should not hold it up.

Follow-up to #2054.

ymote:
Closes #2056.

## What / why

`TaskSupervisor`'s terminal callback into the goal ledger is in-process, synchronous and fire-and-forget. A crash between a task reaching a terminal state and the ledger write landing lost the transition permanently: on restart `enable_persistence` rebuilds the terminal row from JSONL **by direct insert** — firing neither `on_register` nor `on_change` — so nothing re-drives the settle, the goal ledger still says `running`, and the orphan sweep next to it reasons about supervisor state only.

This adds a boot-time reconciliation sweep (option 1 in the issue). No durable outbox, no second store.

- **octos-agent** — `TaskSupervisor::set_on_restore`, a restore observer fired once from `notify_restore` at the end of `enable_persistence`, after replay + orphan sweep + cascade, with the rebuilt table and **no supervisor lock held**. Inherited alongside `on_register` and the named `on_change_listeners`, through the same install path. Installing an observer also delivers a restore that already happened unobserved; the observer and that missed-restore mark share ONE mutex (`RestoreObserverSlot`), so the handshake cannot lose a wakeup (see R2 below). No octos-fleet dependency added; the callback takes the whole table because the reconciliation policy belongs to the consumer.
- **octos-fleet** — `GoalLedger::tasks_open_to_correction(goal_id)`, the reconcile candidate set: `WHERE goal_id = ?1 AND authority < 1`, returning each row with its stored rank. Plus `task_authority` and `data_version` (read-only probes; `data_version` is what lets a caller prove *no writes* rather than infer it from an unchanged row).
- **octos-cli** — the sweep is armed by the **SAME shared installer** that wires the recorder + settle listener, so it cannot drift from the pair it repairs and **no new production call site** is introduced. `terminal_settle_authority` is one function shared by the live change-feed settle and the sweep. `reconcile_goal_task_rows_after_restore` / `..._blocking` re-stash the binding the crashed process lost and re-drive `settle_goal_task_row_from_change` — the authority mapping, FK-parent seeding, bounded retry, generation fence, flight accounting and binding lifecycle are all the live path's, unmodified.

## Round 2 — three defects found by adversarial review, each fixed with a regression watched failing first

**H1 — the candidate predicate stranded the repair that matters most.** The first form excluded every row that already carried a verdict. But an observer-provisional failure lands at **rank 0** and the owner's later completion is *explicitly* allowed to displace it (`should_allow_failed_to_complete_only_with_provisional_authority` pins exactly that). `mark_completed` appends to the JSONL **before** notifying the change feed, so a crash in that window leaves `failed`/0 in the ledger with the correction lost — and the sweep then refused to look at that row forever.

The cut is now by **rank**: −1 and 0 are candidates, 1 and 2 are not. This keeps the reverse-regression protection the exclusion existed for (an orphan-swept `Failed` is final-failure rank 2, which the strictly-greater rule *would* admit over a completion's rank 1 — the SQL guard alone does not save that row), while restoring the correction the ledger's own admission rule already permits. Nothing legitimate is lost by refusing to look at ranks 1 and 2: within one supervisor the terminal guard forbids completed → failed outright. The query returns each row's stored rank so the sweep skips anything its restored verdict would not outrank — a boot that owes nothing still writes nothing.

**H2a — a restore before the wiring was lost forever.** The cached session supervisor is enabled by whichever request touches it first, and `load_skill_action_job_view` installs only its own projection listener before enabling. The later wiring re-enables the **same** ledger path and returns at `enable_persistence`'s idempotence guard, so no further restore is ever notified to that supervisor. `set_on_restore` now delivers a restore that already happened and was not observed, taken with a compare-exchange so exactly one installer delivers it and the cached-supervisor idiom of re-wiring at every point of use does not re-run the sweep. The job view additionally wires the pair before its own enable, so the observers are present from the first touch and registrations in between are recorded too.

**H2b — reconciliation inherited the active-only goal resolver.** `active_goal_id` rejects every paused, blocked, budget-limited or completed goal, so the restore callback exited at its resolver guard for all of them — and `budget_limited` is not an edge case, it is where a goal that spends its budget sits for the rest of its life. Registration is right to refuse a non-active goal (no new work should be recorded against one); reconciliation is not, because the rows are already there and still wrong. The installer now takes **separate** register and restore resolvers, and restore goes through a status-agnostic `bound_goal_id` / `bound_goal_id_under_goal_key`. A cleared goal still resolves to nothing — `clear_goal` removes the record outright.

Round 2 also refuted two hypotheses with full traces (callback placement/lock discipline; binding safety), and produced four tracking items filed as #2091.

## Round 3 — two defects in the round-2 mechanism, plus a test that could not fail

**R2 — the missed-restore handshake could lose the wakeup entirely.** Round 2 held the observer and the "a restore happened with nobody wired" flag apart: an `Option` behind one lock, an `AtomicBool` beside it. `notify_restore` observed `None`, **released** the lock, and only then raised the flag; `set_on_restore` installed its callback and then compare-exchanged. An installer landing in that gap wires itself, sees the flag still clear, delivers nothing — and afterwards nothing ever will, because the same-path re-enable returns at `enable_persistence`'s idempotence guard. That is the exact hole the deferred delivery was added to close, moved one layer down. Both now live in one `RestoreObserverSlot` under a single mutex, making "observe, else mark pending" and "install, taking any pending mark" indivisible: whichever runs second sees the other's effect and exactly one delivers. Observer **inheritance** routes through that same install path instead of assigning the slot directly — production children inherit before enabling, but a late inheritance onto an already-restored supervisor would otherwise drop the missed restore silently.

**R4 — one production restore path still had no observers, and it was not passive.** The runtime-unavailable branch of the skill-action job view (`load_skill_action_jobs`) builds a throwaway supervisor, enables persistence, projects jobs, and drops it. The initial framing was "enables without observers" — the reality is worse. That restore is a real one: its orphan sweep `mark_failed`s abandoned work and **appends those transitions to the durable JSONL**. So the path was actively advancing the supervisor's side of the world while leaving the goal ledger's rows stale — precisely the divergence this PR exists to repair, produced by the repair mechanism's own blind spot. And the missed-restore mark it left behind died with the supervisor, unconsumable by anyone. It now wires the shared observer pair before enabling, with the profile id threaded in for the resolver.

**R5 — the exactly-once claim rested on an assertion that could not fail.** It compared `goal_task_settle_attempts_for` across a re-wire, but that counter only moves once a candidate reaches the settle core; by that point the row is settled at rank 1, so a duplicate delivery finds no candidate and the assertion holds either way. Exactly-once is now counted **at the observer**, across zero / wire-then-restore / restore-then-wire / re-wire / late-inheritance. The lost wakeup gets a deterministic pin: a cfg-gated hook runs inside the slot's critical section, on the branch that decides nobody can take the restore, and asserts a concurrent installer **cannot complete** while that section is held — something no schedule can fake, with a bounded wait so the test cannot hang.

One note on that test, because it nearly shipped hollow. Its first draft signalled installer-first: the installer won the race, `notify_restore` took its `Some` branch, the hook never ran, and it passed **in 0.01s having exercised nothing**. It was caught by noticing the runtime, not by any failure. The committed version drives the handshake hook-first so the interleaving is forced, and asserts `hook_ran` as a tripwire; the doc comment records both as load-bearing so the ordering does not get "simplified" back.

Round 3 also filed #2093 (an unrelated pre-existing flake found during the gate runs).

### 1. Where does the sweep belong?

Not in `enable_persistence` itself. Two independent reasons:

- **The ledger is not reachable there.** octos-agent carries no octos-fleet dependency (deliberate, #2059 item 1), and `enable_persistence` is a synchronous `std::io::Result` API called from actor/turn threads (`session_actor.rs:3446`, `ui_protocol_transport.rs`) — it must not park on SQLite. #2059's constraint holds: DDL runs only in `open_with_busy_retry`, and every ledger call here is behind `offload_goal_ledger_io`.
- **It is the right *moment*, though** — it is the only point where the rebuilt table exists. So octos-agent exposes the moment (`on_restore`) and octos-cli owns the policy.

Hosting it in the shared installer rather than at each `enable_persistence` call site is deliberate: that is exactly the drift/deletion hazard #2059 shipped with, and the wiring-presence test below now covers it at the gateway site. Round 2 established the cost of that choice — a site that enables before the wiring — and closed it in the mechanism rather than only at the one site that exhibits it.

This fires on *every* restore, not only process boot: the WS turn path builds a fresh supervisor and re-restores the shared per-session ledger every turn. That is a feature — it also repairs rows lost to retry exhaustion mid-process — and question 5 costs it.

### 2. Authority for a supervisor-terminal / ledger-unsettled row

The sweep uses the **same** `terminal_settle_authority` mapping the live path uses and the **same** `settle_task_status` strictly-greater SQL admission. There is no second rule.

Non-regression rests on three stacked things, in this order:

1. **Ranks 1 and 2 are never examined.** Load-bearing, not an optimisation — see H1.
2. **A candidate whose restored verdict does not strictly outrank the stored one is skipped before any write.** This is what keeps "owes nothing ⇒ writes nothing" true now that rank-0 rows are candidates.
3. **The SQL guard underneath** refuses any lower-or-equal rank, atomically inside the `UPDATE`.

One sub-decision worth naming: an observer-provisional failure stays **provisional** (rank 0) when reconciled. Promoting it to final "because the owner is gone now" would let a boot sweep outrank a completion that had already landed. `failed_by_observer` is persisted, so the restored row carries the distinction.

**Honest residual:** if a verdict lands between the query and the `UPDATE` and the sweep's rank is strictly greater, the sweep's write wins — precisely what the live path would do for the same pair of deliveries, the rule being order-independent by construction. Reaching that window needs a concurrent writer for a task whose binding this process just claimed, i.e. a second process on the same ledger file.

### 3. The reverse direction — ledger terminal, supervisor rebuilt as running

**It is possible**, and it is why the sweep is ledger-first: the settle can land while the JSONL append recording the same transition is lost. The row then rebuilds non-terminal and the orphan sweep reaps it as `Failed("orphaned across restart")` — via owner-style `mark_failed`, hence final-failure authority.

What the sweep does: **nothing** — the row is at rank 1 or 2 and is not a candidate. The supervisor repairs its own side through the orphan sweep it already runs. Nothing is ever written back into the supervisor: the ledger is a mirror, and a second re-entry path into supervisor state is exactly what #2020 removed.

A third case — a candidate row **absent** from the rebuilt table — is deliberately left alone. Nested/child supervisors keep their own task ledgers while sharing the goal, so "not in my table" means "not mine", not "dead". Failing those would kill live work; a stale `running` is the safer error.

### 4. Idempotency

Not asserted — mechanised:

- **Convergence.** A settle stamps rank ≥ 0; a settle at rank 1 or 2 leaves the candidate set permanently. A rank-0 row stays a candidate exactly as long as it remains correctable, and the pre-write rank comparison means re-examining it costs no write. Pinned at the SQL layer by `should_keep_correctable_rows_and_drop_settled_ones_when_reconciling`.
- **Clean shutdown.** Nothing owed ⇒ no candidate outranked ⇒ zero writes. Proved three ways in `should_write_nothing_when_rebooting_after_a_clean_shutdown`: `PRAGMA data_version` on an independent connection held open across the reboot, the shared settle core's attempt counter, and field-equality of the row.
- **Crash mid-sweep.** Each row's settle is a single-statement `UPDATE` with its own admission guard. Rows already settled drop out; the remainder is re-enumerated. No partial row state is expressible.
- **No re-fire on a no-op re-enable**, and no re-fire on a repeat wiring — the deferred delivery clears its flag with a compare-exchange.
- **Live chains are not disturbed.** A task whose binding is still present is skipped rather than re-stashed: re-stashing assigns a fresh generation and would kill that chain at its next staleness fence. Both stash entry points share one generation-assignment point, preserving #2059 round 7's W2 invariant.

### 5. Blast radius / startup latency

- **Synchronous part:** a scan of the already-in-memory rebuilt table for terminal rows. With none, no further work at all.
- **Blocking part:** one filesystem metadata probe for the ledger file — the only I/O when the goal has no ledger (no SQLite open, no write). Otherwise one `open_with_busy_retry` + one goal-scoped candidate query, then one settle per row whose restored verdict outranks the stored one. Each goal owns its own `goal-ledgers/<goal>.db`.
- Everything after the in-memory scan runs on the blocking pool; nothing lands on the restoring thread.
- **Steady state is one probe plus one empty read.** Note the honest bound: the available index is `(goal_id, status)` while the query filters `(goal_id, authority)`, so the scan is bounded by the goal's **total** row count, not its candidate count. A covering index is tracked in #2091.
- **Allocation:** the terminal subset of the rebuilt table is cloned into the offload; `enable_persistence` already clones the whole table more than once, so this is in line with its existing cost. Non-terminal rows are never cloned.
- **Scope limit, stated plainly:** the sweep repairs the goal **bound** to this session, whatever its status (round 2). A session whose goal was cleared or replaced still leaves the previous goal's rows unrepaired — the record is gone, so there is nothing to scope a sweep to. Tracked in #2089.

## The two smaller items from the issue

**Serve-boot wiring presence — included.** `should_wire_goal_task_row_observers_when_gateway_actor_is_spawned` drives the real `ActorFactory::spawn` — the production function containing the install call at `session_actor.rs:3425` — pulls the supervisor `spawn` registered into `SessionTaskQueryStore`, and asserts the **effect**: registering creates a `running` ledger row and completing settles it. Not an `include_str!` grep, not a direct call to the installer; both would keep passing with the call site deleted. Mutation-verified. The two `ui_protocol_transport.rs` sites remain uncovered — they sit inside `run_standalone_turn`, which needs a full WS turn to drive; stated plainly rather than substituted for, and tracked in #2089.

**Same-ID restoration — separable, already covered; no new issue filed.** #2059 round 7 moved generations to one process-wide monotonic counter, closing the ABA. The residual the issue flags is the settle-retry dedupe story for a restored row re-registered under its original id. No production path does that today — `enable_persistence` restores by direct insert and fires no registration callback, which is precisely the gap this PR closes at the ledger layer instead of by re-registering. And the absent-only stash means a restored task cannot steal a live incarnation's binding even if such a path appears.

## Tests

| test | real path it drives |
|---|---|
| `should_settle_owed_ledger_row_when_supervisor_reboots_with_a_terminal_row` | **RED first.** Shared installer → `register` → binding dropped (a restart's effect on the in-memory map) → `mark_completed` → second supervisor's `enable_persistence`. |
| `should_settle_owed_ledger_row_when_rebooting_under_a_tokio_runtime` | Same, on a runtime: the sweep's read runs in `spawn_blocking` and the settle it re-drives offloads **again** from inside that blocking thread. Without this, a nested-offload failure would repair nothing in production while every inline test stayed green. |
| `should_settle_a_provisional_row_when_the_restored_verdict_is_a_completion` | **Round 2, RED first (H1).** `mark_failed_observed` → provisional settles rank 0 and retains its binding → binding dropped → `mark_completed` (JSONL only) → reboot must repair `failed`/0 → `complete`/1. |
| `should_reconcile_when_the_restore_happened_before_the_observers_were_wired` | **Round 2, RED first (H2a).** Reboot in the hostile order — `enable_persistence` first, wiring second. Owns the EFFECT only; its round-2 "not re-delivered" half was vacuous and was removed (round 3, R5). |
| `should_deliver_a_restore_exactly_once_however_the_wiring_is_ordered` | **Round 3 (R5).** Counts deliveries at the observer across never-restored / wire-then-restore / idempotent re-enable / restore-then-wire / re-wire / late inheritance. |
| `should_not_lose_a_restore_when_wiring_races_the_missed_restore_mark` | **Round 3 (R5).** Deterministic lost-wakeup pin: a hook inside the slot's critical section asserts a concurrent installer cannot complete while it is held, that the hook actually ran, and that delivery still happens exactly once. |
| `should_reconcile_when_the_bound_goal_is_not_active` | **Round 2, RED first (H2b).** Goal transitioned to `budget_limited` (same id, asserted), `active_goal_id` asserted to refuse it, reboot must still reconcile. |
| `should_not_regress_a_settled_row_when_the_rebuilt_supervisor_row_outranks_it` | Completion settles; the JSONL is snapshotted *before* it (a lost append); the reboot restores non-terminal, the orphan sweep marks `Failed`, and the `complete` row must survive untouched — status, rank and `updated_at_ms`. |
| `should_write_nothing_when_rebooting_after_a_clean_shutdown` | `data_version` + settle-core attempt counter + row equality. |
| `should_leave_the_binding_alone_when_a_live_incarnation_still_owns_the_task` | Settle listener replaced with a no-op under its own key ⇒ terminal fires, delivery never happens, **binding survives**; the sweep must not re-stash (generation pinned) nor settle behind the owner's back. Hardening tracked in #2091 item 4. |
| `should_wire_goal_task_row_observers_when_gateway_actor_is_spawned` | Real `ActorFactory::spawn`. |
| `should_fire_on_restore_once_with_the_swept_table_when_persistence_is_enabled` | Real `enable_persistence`: exactly one firing, table observed *after* the orphan sweep, no re-fire on a no-op re-enable. |
| `should_inherit_on_restore_when_registration_observers_are_inherited` | `inherit_registration_observers` → child's own `enable_persistence`. |
| `should_keep_correctable_rows_and_drop_settled_ones_when_reconciling` | The candidate predicate at the SQL layer: rank 0 stays with its rank, ranks 1 and 2 leave, the retained candidate's completion really lands, goal scoping holds. |

### Mutation checks

**Round 3:**

8. **Split handshake restored** (observe/mark and install/take back in separate critical sections, hook at the equivalent point) → `should_not_lose_a_restore_when_wiring_races_the_missed_restore_mark` FAILED on `installing an observer completed while the missed-restore decision was still being made` — in 0.01s, because the installer was never blocked. Restored.
9. **Inheritance reverted to a direct slot assignment** → `should_deliver_a_restore_exactly_once_however_the_wiring_is_ordered` FAILED on the late-inheritance case (`left: 0, right: 1`). Restored.

**Round 2:**

5. **Predicate reverted to `authority < 0`** → `should_settle_a_provisional_row_when_the_restored_verdict_is_a_completion` FAILED (`left: "failed", right: "complete"`). Restored.
6. **Deferred restore delivery removed** (`notify_restore` drops when no observer is wired) → `should_reconcile_when_the_restore_happened_before_the_observers_were_wired` FAILED (`left: "running", right: "complete"`). Restored.
7. **Restore resolver reverted to `active_goal_id`** → `should_reconcile_when_the_bound_goal_is_not_active` FAILED (`left: "running", right: "complete"`). Restored.

**Round 1:**

1. **Installer call site at `session_actor.rs:3425` deleted** → the gateway wiring test FAILED (`timed out waiting for the registration observer's 'running' row`). Restored.
2. **`set_on_restore` unwired in the shared installer** → both `should_settle_owed_ledger_row_*` FAILED. Restored.
3. **Enumeration switched to `tasks_for_goal` (supervisor-first shape)** → `should_not_regress_a_settled_row_when_the_rebuilt_supervisor_row_outranks_it` FAILED with the real corruption (`left: "failed", right: "complete"`), and `should_write_nothing_when_rebooting_after_a_clean_shutdown` FAILED on the settle-core attempt counter (`left: 2, right: 1`). Concrete proof that the read predicate — not the SQL guard — is what prevents that regression. Restored.
4. **`stash_goal_task_binding_if_absent` → unconditional re-stash** → `should_leave_the_binding_alone_when_a_live_incarnation_still_owns_the_task` FAILED. Restored.

## CI (#2029)

All eight octos-cli tests live in non-gated modules and appear in the **unfiltered** `cargo test -p octos-cli --lib` step — verified with `-- --list`:

```
autonomy::agent_orchestrator::tests::should_reconcile_when_the_bound_goal_is_not_active: test
autonomy::agent_orchestrator::tests::should_reconcile_when_the_restore_happened_before_the_observers_were_wired: test
autonomy::agent_orchestrator::tests::should_settle_a_provisional_row_when_the_restored_verdict_is_a_completion: test
autonomy::agent_orchestrator::tests::should_settle_owed_ledger_row_when_rebooting_under_a_tokio_runtime: test
autonomy::agent_orchestrator::tests::should_settle_owed_ledger_row_when_supervisor_reboots_with_a_terminal_row: test
autonomy::agent_orchestrator::tests::should_not_regress_a_settled_row_when_the_rebuilt_supervisor_row_outranks_it: test
autonomy::agent_orchestrator::tests::should_write_nothing_when_rebooting_after_a_clean_shutdown: test
autonomy::agent_orchestrator::tests::should_leave_the_binding_alone_when_a_live_incarnation_still_owns_the_task: test
session_actor::tests::should_wire_goal_task_row_observers_when_gateway_actor_is_spawned: test
```

That step runs with **no name filter**, so nothing can silently match zero tests. `octos-agent` tests ride `Test octos-agent (lib)`; the fleet test rides the unfiltered `Test octos-fleet` step #2059 added (note its module is `sqlite_ledger::digest_integration_tests`, not `::tests` — a filter written against the wrong path would match nothing). No filter change was needed.

## Gates (round 3, verbatim)

```
cargo fmt --all -- --check                                => exit 0
cargo clippy --workspace --all-targets -- -D warnings     => exit 0
cargo clippy -p octos-cli --features api --all-targets    => exit 0
cargo clippy -p octos-fleet --all-targets -- -D warnings  => exit 0
cargo test -p octos-fleet        → ok. 182 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
cargo test -p octos-cli --lib    → ok. 1531 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out
cargo test -p octos-cli --tests  → 20 targets, all "test result: ok", 0 failed
cargo test -p octos-cli --lib --features api -- --test-threads=1
                                 → ok. 3009 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
cargo test -p octos-agent --test '*' -- --test-threads=1 → all targets ok; 0 failed
cargo test -p octos-agent --lib  → FAILED. 2587 passed; 2 failed; 3 ignored
cargo test -p {octos-fleet,octos-agent,octos-cli} --doc → ok (0 / 1 / 0 passed; 0 failed)
```

Workspace clippy initially failed on `clippy::type_complexity` for the new cfg-gated hook field; fixed with a `RestoreNotifyHook` alias and re-run clean.

**The two `octos-agent` failures are environmental on this host and unrelated to this diff**: `tools::ssrf::tests::test_with_addrs_fails_closed_on_nonexistent_domain` and `tools::web_fetch::tests::test_ssrf_dns_failure_blocks_request` both require `*.invalid` to fail DNS resolution, and this machine's resolver hijacks NXDOMAIN (`getaddrinfo('this-domain-does-not-exist-ssrf-test.invalid') → 198.18.0.66`), so the SSRF check blocks on "private IP" instead of "DNS failure". Neither file is in this diff.

**A third, intermittent failure** — `plugins::tool::tests::execute_preserves_plugin_structured_metadata`, ~1-in-3 under full-parallel `--lib` — was checked for a causal link to this diff rather than assumed away: re-running the suite three times with this PR's 250ms lock-holding test **skipped** still produced one flake, so it is independent. Filed as **#2093**; host-reproduced only, and CI's `test-octos-agent (lib)` has passed on every push of this branch.

**CI:** rounds 1 and 2 both closed fully green — 16/16 checks including `check-windows`.


ymote:
Follow-ups filed as #2089: wiring-presence coverage for the two remaining installer sites (`ui_protocol_transport.rs:10645` and `:29819`), and the cleared/replaced-goal scope limit. Both are stated in the PR body above; #2089 carries the closing plan.

ymote:
Round 2 pushed as `b14d2bec8`. H1, H2a and H2b all confirmed on the evidence and fixed — none was wrong, so nothing was pushed back on. Each fix has a regression watched failing first and a mutation that fails it again (mutations 5–7 in the PR body). H6's predicate test was rewritten: it now asserts rank 0 STAYS a candidate and that the correction it exists for really lands, which is the inverse of what it asserted before. The four tracking items are filed as #2091.

ymote:
Round 3 pushed as `62059f940`. R2, R4 and R5 all confirmed on the evidence before implementing — none was wrong.

**R2** — the lost wakeup is real and now closed at the source: the observer and its missed-restore mark share one mutex, so `observe-else-mark` and `install-taking-mark` are indivisible. Observer inheritance routes through the same install path (it previously assigned the slot directly, which would drop a missed restore on a late inheritance).

**R4** — `load_skill_action_jobs` now wires the shared observer pair before its throwaway supervisor enables persistence; the profile id is threaded in for the resolver.

**R5** — the vacuous settle-attempt assertion is gone (replaced by a comment explaining why it could not fail). Exactly-once is now counted at the observer across zero / wire-then-restore / restore-then-wire / re-wire / late-inheritance, and the lost wakeup has a deterministic pin: a cfg-gated hook inside the slot's critical section asserts a concurrent installer cannot complete while that section is held. The handshake is driven hook-first so the interleaving is forced, and the test asserts the hook ran — my first draft signalled the other way round, the installer won, the hook never ran, and it passed in 0.01s while proving nothing. Fixed before committing.

Mutations: split handshake restored → race test fails on the exact assertion; inheritance reverted to a direct assignment → the late-inheritance case fails.

One honest note: `plugins::tool::tests::execute_preserves_plugin_structured_metadata` flakes roughly 1-in-3 under full-parallel `cargo test -p octos-agent --lib` on my machine. I verified it is NOT caused by the new 250ms-holding race test — it still flakes with that test skipped — and CI's `test-octos-agent (lib)` has passed on every push. Not filed, since I could not reproduce it off this host.

ymote:
## Adversarial re-review — round 3 — DO-NOT-MERGE (test only)

Reviewed `62059f940`; `4038ff29d` is documentation-only and does not affect the verdict.

**The production side is fully clear.**

- **R1 REFUTED.** `callback` and `undelivered` share one slot mutex (`task_supervisor.rs:420`); observe-or-mark (`:1631`) and install-and-take (`:1680`) are indivisible. Snapshotting and callback invocation happen **after** the guard drops (`:1642-1647`), so a production callback may re-enter safely — no deadlock. Both lock orders yield exactly one delivery per restore; no same-restore double delivery.
- **R2 REFUTED.** Inheritance clones only the parent callback and installs it through the child's own delivery-aware path (`:1817`); parent and child pending marks stay independent. Fresh-snapshot, synchronous-nested and detached children inherit immediately after construction (`registry.rs:1038`, `spawn.rs:3905`, `:4520`).
- **R3 REFUTED.** Alias, field, methods and invocation are all `#[cfg(test)]` (`:417`, `:977-978`, `:1637-1638`). Production compiles to the same single critical section, no hook, no altered signature.
- **R4 REFUTED.** The fallback receives the resolved profile id and data root, installs the shared pair, then enables (`skill_action_jobs.rs:170`, `ui_protocol_transport.rs:11123`). Gateway, cached-runtime, job-view and per-turn restores are all wired first. Only remaining non-test outlier: the pre-existing environment-gated M9 fixture (`ui_protocol_transport.rs:28591`, needs `OCTOS_M9_PROTOCOL_FIXTURES=1`) — tracking-only, going to #2091.
- **R6 REFUTED.** A callback panic cannot poison the slot, because invocation is outside its mutex; poisoned mutexes are recovered with `into_inner`.

### R5 — MUST-FIX: the race test does not prove what it claims

Hook-first ordering and `hook_ran` prove the **missed-observer branch was entered**. They do not prove the **installer ran during the 250 ms window**.

The hook signals the installer and treats `recv_timeout` failure as proof of blocking (`task_supervisor_tests.rs:2931`), but the installer has no "started / attempting lock" acknowledgement before calling `set_on_restore` (`:2950`). On a loaded runner the installer may simply remain **unscheduled** for 250 ms — and then, with the old split handshake in place, the hook times out, the pending bit is raised, the installer runs afterwards and delivers once, and every assertion passes.

So the split-handshake mutation failing in 0.01s proved the test bites *on that machine, at that moment*. It does not establish that it bites on a loaded CI runner, which is where it matters.

This is the same shape as the vacuous first draft caught during round 3, one layer down: that version passed because the installer won a race it was never forced to enter; this one can pass because the installer never got scheduled. **Absence of a signal is being read as evidence of blocking, and absence of a signal has more than one cause.**

Fix: require positive evidence instead of inferring from silence. The installer must acknowledge reaching the lock attempt before calling `set_on_restore`, and the hook must require that acknowledgement — a timeout must FAIL the test, not pass it. Also `handle.join()` (`:2957`) is unbounded, so the no-hang guarantee is incomplete.

ymote:
Round 4 pushed as `c96592cfa` — **test-only, no production change** (the production side was cleared, so I did not touch it).

R5 was right and I should have caught it: I was reading absence of a signal as evidence of blocking, and absence has more than one cause. The installer now acks reaching its lock attempt immediately before the call that must block, and the hook REQUIRES that ack — a timeout fails with "inconclusive, which is a failure, not a pass". Added a happens-after assertion (install completed only after the section released) and bounded the post-release wait before the join.

Three mutations, because the interesting one is the combination:

| mutation | result |
|---|---|
| split handshake alone | FAILS on the installed-inside-the-section evidence — and the ack arrived first, so the failure is genuine evidence, not a scheduling artifact |
| installer never scheduled in time (sleep > ack budget) | FAILS on the ack assertion, in 11.02s |
| **both together** — broken implementation AND an unscheduled installer, i.e. exactly the false pass you described | FAILS on the ack assertion. The round-3 version would have gone green here. |

Doc comment now separates the two signals explicitly: a ~500ms runtime shows the held window elapsed and nothing more; only the ack shows the installer was blocked rather than merely late. It says a version keeping the timing but dropping the ack is strictly weaker than it looks, and states the residual (the few instructions between ack and mutex).

M9 fixture note added to #2091 as item 5.

**CI problem, flagging rather than sitting on it:** rounds 3 and 4 have produced **no CI runs at all**. `659b13ce7` (round 1) and `b14d2bec8` (round 2) both ran and passed 16/16. `62059f940`, `4038ff29d` and `c96592cfa` created nothing — 47+ minutes after the round-3 push, `gh api .../actions/runs?branch=...` lists only the two older runs, and `gh pr checks` reports "no checks reported". Meanwhile `main` is getting runs normally (07:23Z today), so Actions is healthy repo-wide; it looks like `pull_request` event delivery for this PR specifically. The PR is OPEN, head is `c96592cfa`, not merged. I have forced a `workflow_dispatch` run on the branch head and am watching it. I am not claiming round-3/4 CI green until something actually runs.

ymote:
Closing and immediately reopening to re-trigger `pull_request` event delivery — see the next comment for why. No change to the branch.

ymote:
## Why the close/reopen: `pull_request` CI stopped firing for this PR

Rounds 3 and 4 produced **no CI runs at all**, and this is not a slow queue — verified independently:

```
$ gh api 'repos/octos-org/octos/actions/runs?head_sha=<sha>' --jq .total_count
62059f940 → 0      (round 3, pushed ~05:5xZ)
4038ff29d → 0      (doc-only)
c96592cfa → 0      (round 4)
```

while the two earlier pushes on this same branch ran normally and went 16/16 green:

```
success  pull_request  b14d2bec8  CI  05:29:01Z
success  pull_request  659b13ce7  CI  04:26:57Z
```

It is not repo-wide. `main` is getting runs (07:23Z), and `fix/2076-get-goal-torn-read` got its `pull_request` runs at 06:39Z. The workflow has no path filters (`pull_request: branches: [main]`), the PR is open and not a draft, and the `concurrency` group only cancels in-flight runs — it cannot prevent one being created. So `pull_request` event delivery simply stopped for this PR after `b14d2bec8`.

Closing and reopening fires a fresh `pull_request` event, which is the standard remedy. The branch is untouched.

**On the `workflow_dispatch` run that was triggered manually:** it exercises the same code, but it is not the `pull_request` run the merge gate reads, and it is not what "CI is green" means here. Nothing merges on it. That distinction is exactly the failure mode this PR's own review has been circling all day — an absent signal read as a passing one — so it would be a poor place to make an exception.

Round-1 and round-2 CI are confirmed green (16/16 each, `check-windows` included); rounds 3 and 4 need real runs before this can land.

ymote:
## Correction: this is the merge conflict, not event-delivery loss

My previous comment concluded `pull_request` event delivery had stopped for this PR. That was wrong, and the close/reopen was the wrong remedy — it produced no runs either, which is the confirmation.

**GitHub cannot compute a PR's merge ref (`refs/pull/N/merge`) when the branch conflicts with base, so it does not create `pull_request` workflow runs for a conflicting PR.** The timeline fits exactly:

| | |
|---|---|
| round 2 `b14d2bec8` ran **05:29Z** | main was `92910f39a` (04:27Z) — no conflict, 16/16 green |
| **#2076 `9ba992b24` merged 06:07:21Z** | adds `time_used_seconds` to `octos_fleet::Goal` and updates the `Goal { … }` literals in `agent_orchestrator.rs` — the conflict source |
| `62059f940`, `4038ff29d`, `c96592cfa` | all pushed onto an already-conflicting PR → zero runs |

So resolving the conflict is what restores CI. Nothing about Actions is unhealthy, and no event nudge was ever going to help.

Worth recording that the error I made is the same one this PR spent round 5 fixing in its own test: reading an absent signal as evidence of a specific cause, rather than asking what would prevent the signal from existing at all. "No runs" had two candidate explanations and I took the exotic one without eliminating the mundane one.

The branch is being merged with `origin/main` now. The round-4 adversarial pass will run against the merged state, which is the state that can actually land.

ymote:
Merged `origin/main` in as `aae2e9f67`. **Your diagnosis is confirmed by the fix:** the first push after resolving produced `pull_request` runs immediately (`CI` + `check-author-email` on `aae2e9f67`, 08:08Z). PR is now `mergeable: MERGEABLE`. I cancelled the stale `workflow_dispatch` run on `c96592cfa` — superseded, and it would have burned `check-arm`/`check-riscv` that PRs skip.

**One textual conflict**, exactly where predicted: `record_goal_task_registration`. Main added `time_used_seconds` to the inline `Goal { … }` literal; this branch had moved that literal into the shared `goal_ledger_row_snapshot` helper. Resolved by keeping the helper and carrying the field into it — so the boot-reconcile path persists both cost dimensions rather than a goal row answering half the cost question. That helper can be the FIRST writer of a goal row via `create_goal_if_absent`, so this mattered; I put that reasoning in its doc comment.

**The three you flagged, read rather than assumed:**

- **#2076 + #2092 / `sqlite_ledger.rs`** — verified `open_with_busy_retry` runs BOTH migrations after the merge (`migrate_tasks_authority_column` then `migrate_goals_time_column`). No bare probe reintroduced: `tasks_open_to_correction` has no `pragma_table_info` at all — it references `authority` directly on a connection the retry profile has already migrated, so there is no probe/SELECT window of the kind #2092 closed in `get_goal`.
- **#2075 / `ui_protocol_transport.rs`** — profile scoping did not disturb the per-turn restore resolver or either job-view wiring; all three call sites survived.
- **#2090** — the both-rows break condition and its Windows waiting budget are intact. I did not touch that test.

**Gates on the merged state (verbatim):**
```
cargo fmt --all -- --check                                => exit 0
cargo clippy --workspace --all-targets -- -D warnings     => exit 0
cargo clippy -p octos-cli --features api --all-targets    => exit 0
cargo clippy -p octos-fleet --all-targets -- -D warnings  => exit 0
cargo test -p octos-fleet        → ok. 188 passed; 0 failed; 0 ignored
cargo test -p octos-cli --lib    → ok. 1534 passed; 0 failed; 3 ignored
cargo test -p octos-cli --tests  → 20 targets, all ok; 0 failed
cargo test -p octos-cli --lib --features api -- --test-threads=1
                                 → ok. 3021 passed; 0 failed; 6 ignored
cargo test -p octos-agent --test '*' -- --test-threads=1 → all ok; 0 failed
cargo test -p octos-agent --lib  → FAILED. 2587 passed; 2 failed; 3 ignored
```
Counts moved as expected (fleet 182→188, cli 1531→1534, api 3009→3021) — main's new tests. The two agent failures remain the host DNS hijack; #2093 did not reproduce this round.

Added the absence-vs-evidence practice note to #2091 as item 6, covering all three instances. CI on the merged head is running; I'll report it.

ymote:
## Adversarial re-review — round 4 (merged state) — MERGE

Reviewed `aae2e9f67`, i.e. after `origin/main` was merged in. No MUST-FIX findings.

- **M1 REFUTED — the conflict resolution is correct, and correct for the stated reason.** The helper copies `time_used_seconds` from the live goal explicitly (`agent_orchestrator.rs:6403`); registration really can create the first goal row through `:6505`; the shared SQL binds the field (`sqlite_ledger.rs:650`); and reconciliation carries the same snapshot to its self-seeding write (`:6944`). Every other production `octos_fleet::Goal` construction carries the field explicitly — only legacy serde deserialization defaults it (`sqlite_ledger.rs:31`).
- **M2 REFUTED — nothing broke in the files both sides edited.** No unsafe reader-side probe/SELECT window: `get_goal` does its probe and query in one deferred transaction (`:936`), and `tasks_open_to_correction` reads `authority` directly (`:1290`) through an already-migrated `open_with_busy_retry` (`agent_orchestrator.rs:6664`). The per-turn resolver survives #2075 (`ui_protocol_transport.rs:30056`); cached and fallback job views both wire before restore (`:11138`, `skill_action_jobs.rs:170`). The #2090 nested-spawn break still requires both titles (`agent_orchestrator.rs:19250`) with the Windows-scaled deadline intact (`:18887`).
- **M3 REFUTED — the two migrations compose.** `open_with_busy_retry` runs authority first, time second, before returning the ledger (`sqlite_ledger.rs:248`), each an atomic `BEGIN IMMEDIATE` (`:497`, `:565`). Another plain connection can observe the boundary between those commits, but reconciliation receives no ledger until both succeed, and plain `get_goal` handles either time-column generation safely.
- **M4 CONFIRMED — TRACKING-ONLY.** The ack precedes the call (`task_supervisor_tests.rs:2993`), so preemption between ack and mutex acquisition remains logically possible; the happens-after assertion compares a timestamp written after `set_on_restore` returns (`:3000`), which narrows the gap without eliminating it. Acceptable: a false pass now requires **>500 ms of starvation in a handful-of-instructions window, after positive scheduling evidence, while the hook itself blocks waiting**. That closes the broad "never scheduled during the window" false pass from round 3, which was the finding.
- **M5 REFUTED.** Ack wait bounded at 10s, held interval 500 ms (`:2931`), completion bounded before joining (`:3007`). The literal `join()` is still unbounded but is reached only after the installer sends `done` as its final operation, with no remaining blocking user code or custom destructor — so deadlock, panic, or non-scheduling times out and fails rather than reaching the join.
- **M6 REFUTED.** `cargo fmt --check`, `git diff --check`, affected-crate `cargo check --all-targets`, the R5 race test, all 188 `octos-fleet` library tests, reconciliation/time/nested-spawn CLI tests, and API-feature profile/job-view tests all pass.

---

Four rounds, four passes. Round 1 found the `authority < 0` predicate stranding an owed completion and two unreachable production paths; round 2 found the lost-wakeup race and the inheritance bypass; round 3 found the race test proving less than it claimed; round 4 confirms the fix and the merge. Waiting on `check-windows` and `e2e-tool-regression` before merging — the first genuine `pull_request` run on this branch since round 2, after the conflict that had been silently suppressing them.

PR #2094, opened 2026-08-20
feat(sandbox): prompt to escalate when a confined command is denied

alanpoon:
## The bug

A sandboxed command can be refused a path partway through while the wrapper shell still exits `0`. The denial lands in stderr, the tool layer sees success, and the model is left to narrate it back to the user:

```
⏺ Bash($ ... ls ../octos 2>&1 | head -3")  41ms
  ⎿ ls: ../octos: Operation not permitted
  ⎿ ls: /Users/alanpoon/.cargo/git/checkouts/: Operation not permitted
  ⎿ Exit code: 0
• Still can't run it — ../octos is unreadable in this sandbox. Options if you
  want it run: grant read access to ../octos, or cherry-pick …
```

That is a dead end the user never gets asked about. Across a 13-probe TUI capture matrix, **every** octoscode pane that touched a path outside the workspace hit this; **no** Claude Code pane did, because Claude asks whether to retry outside the sandbox.

## Why the dialog never appeared

This is the part worth flagging: the client side was already finished.

- `ApprovalSandboxEscalationDetails` and `approval_kinds::SANDBOX_ESCALATION` already exist in `octos-core::ui_protocol`.
- octoscode already renders **all five** approval kinds, escalation included, and already round-trips `approval/respond`.
- But `sandbox_escalation` appears nowhere outside its own type definition and unit tests — **no code ever raised one**, so the dialog was unreachable.

So this isn't a new UI. It's the missing server half of a flow that was already built and wired at both other ends.

## The change

`exec_command` scans a *completed* confined run for denial signatures. On a hit it asks through the existing `ToolApprovalRequest` bridge — the same one the shell tool's dangerous-command gate uses, so it reaches the dialog octoscode already draws:

> **Allow command to run outside the sandbox?**
> The sandbox refused part of this command:
> &nbsp;&nbsp;`ls: ../octos: Operation not permitted`
>
> Approving re-runs the whole command with no confinement, so it can read and write anything this account can. Denying keeps the output above.

Approve re-runs under `NoSandbox` and says so in the output. Deny keeps the original output and tells the model the paths stay unreadable, so it stops retrying the same command.

`run_to_completion` was split into a reusable `run_once(&dyn Sandbox, …)` so the retry reuses the existing process-group and timeout handling verbatim rather than duplicating it.

## On the detection being textual

It has to be. Seatbelt and Landlock report denials to the **child** as a bare errno — the child's own error message is the only signal that reaches us, so there is no structured event to match on.

That makes it a heuristic, and it's scoped accordingly: it is consulted **only when a real backend is active** (`!sandbox.is_noop()`). Under no confinement an EPERM is a genuine filesystem error rather than a policy decision, and prompting for it would be noise. This matters most for the EACCES signature (`Permission denied`), which Landlock and bwrap emit but so does an ordinary unreadable file — the `is_noop` gate is what keeps that from producing spurious prompts.

## Deliberately unchanged

- **Spawn failures and timeouts never escalate.** Neither is a policy decision, and re-running unconfined would not fix them. That's what `ExecRun::completed` tracks.
- **Non-interactive callers keep today's behaviour exactly.** Cron fires and headless runs have no approval context in scope; they get the original output and never block.
- **A denied escalation preserves the original exit status.** The user declining a retry does not retroactively change how the command ended.
- **One prompt per denied command**, capped at 5 reported paths — a command that trips on a dozen paths is the same decision as one that trips on three.

## Tests

13 new tests, all passing:

- `sandbox::denial` (8) — seatbelt/EACCES/`(os error N)` forms, path recovery from `prog: path: msg`, dedup of one path blamed by several commands, limit, overlong-line capping, clean output yielding nothing.
- `tools::coding_tools` (5) — approved escalation actually re-runs and reports it; denied escalation keeps the original output, says it was declined, and does **not** run the command; `NoSandbox` never prompts; clean confined output never prompts; no approval context falls back to today's behaviour.

`cargo fmt` and `cargo clippy --all-targets` are clean.

**Pre-existing failures unrelated to this branch:** `cargo test -p octos-agent --lib` reports 26 failures on this branch — and the same 26 on a clean `origin/main` (verified by stashing). They are all network-dependent (`plugins::http_discovery`, `tools::http`, `tools::web_fetch`, `tools::ssrf`, `validators`) and fail with local `502`s. Passing count goes 2563 → 2576, which is exactly the 13 added here.

### DIFF
--- a/crates/octos-agent/src/sandbox/denial.rs
+++ b/crates/octos-agent/src/sandbox/denial.rs
@@ -1,0 +1,175 @@
+//! Recognising sandbox denials in command output.
+//!
+//! A confined command can fail in a way the tool layer never sees: the wrapper
+//! shell exits 0 while a child deep inside the pipeline was refused by the
+//! kernel and printed `ls: ../octos: Operation not permitted` to stderr. The
+//! exit status says success, so nothing upstream knows a policy boundary was
+//! hit, and the model is left to guess — in practice it narrates the denial
+//! back to the user and gives up.
+//!
+//! Detection is textual because that is the only signal available: seatbelt and
+//! Landlock report denials to the *child* as a plain errno, and the child's own
+//! error message is all that reaches us. That makes this a heuristic, and it is
+//! deliberately only consulted when a real backend is active ([`Sandbox::is_noop`]
+//! is false) — under no confinement an EPERM is a genuine filesystem error, not
+//! a policy decision, and prompting for it would be noise.
+//!
+//! [`Sandbox::is_noop`]: super::Sandbox::is_noop
+
+/// Signatures a confined child prints when the kernel refuses it.
+///
+/// EPERM (`Operation not permitted`) is what macOS seatbelt returns. EACCES
+/// (`Permission denied`) is what Landlock and bwrap return, and is also what an
+/// ordinary unreadable file returns — the `is_noop` gate above is what keeps
+/// the second case from producing spurious prompts.
+const DENIAL_SIGNATURES: &[&str] = &[
+    "Operation not permitted",
+    "operation not permitted",
+    "Permission denied",
+    "permission denied",
+    "(os error 1)",
+    "(os error 13)",
+];
+
+/// One line of output that looks like a sandbox denial.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SandboxDenial {
+    /// The offending line, trimmed and length-capped.
+    pub line: String,
+    /// Path the line blamed, when one could be recovered.
+    pub path: Option<String>,
+}
+
+/// Longest denial line we echo into an approval prompt. Long enough for a
+/// realistic `prog: /some/path: Operation not permitted`, short enough that a
+/// pathological line cannot dominate the dialog.
+const MAX_LINE: usize = 200;
+
+/// Scan command output for lines that look like sandbox denials.
+///
+/// Returns at most `limit` distinct denials, in the order they appeared. Callers
+/// are expected to have checked that a real sandbox backend is active.
+pub fn detect_sandbox_denials(output: &str, limit: usize) -> Vec<SandboxDenial> {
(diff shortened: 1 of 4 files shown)


ymote:
Review summary: the UX goal is valid and the refactor/tests are clear, but I would keep this in draft until the escalation boundary is hardened.

1. The trigger is untrusted command output. Any confined child can print `Permission denied`, `Operation not permitted`, or `(os error 13)` without the sandbox having denied anything; ordinary Unix permissions can produce the same strings. That means the prompt can misstate heuristic evidence as a confirmed sandbox refusal.

2. This uses the generic `ToolApprovalRequest` with `tool_name = exec_command`. Existing approve-for-tool or approve-for-session scope entries can therefore auto-resolve it. Combined with the output-spoofing issue, a command can manufacture a denial signature and then be rerun under `NoSandbox` without a fresh approval. A sandbox escape should require an explicit one-shot decision and should be emitted as the typed `sandbox_escalation` approval kind, with from/to and requested permissions populated.

3. The approved path reruns the entire command after its confined attempt already completed. Non-idempotent prefixes may execute twice. The prompt mentions a whole-command rerun, which is good, but this deserves an explicit risk treatment and a regression scenario; narrowing the grant to the denied path/prefix would be safer if the backend supports it.

4. On denial, the result preserves a possibly successful wrapper exit and only tells the model not to retry in prose. Now that #2095 is merged, consider returning `structured_metadata.do_not_retry_same_turn = true` so the host enforces that instruction rather than relying on the model.

Suggested minimum before ready-for-review:
- make escalation approval one-shot and immune to cached turn/tool/session approvals;
- carry typed `SANDBOX_ESCALATION` details and phrase the prompt as output that suggests a denial, not proof;
- add tests for spoofed denial output and pre-existing approval scopes;
- integrate the denied result with the #2095 terminal-retry metadata.

The detection parser, no-op/headless guards, output caps, and approve/deny tests are otherwise well structured.

PR #2097, opened 2026-08-21
fix(api): let keyless families through admin test-provider and provider-models

heyong4725:
## Summary

Live octoscode onboarding against a llama.cpp server caught two residual keyless dead ends that the #2096 review round missed:

- **admin::test_provider**: when the request carries neither api_key nor api_key_env, resolve_saved_key errors with "No api_key or api_key_env provided" *before* the keyless gate added in #2096 — so the local family still could not be connection-tested.
- **provider_models** (POST /api/my/provider-models): the same unauthenticated-key dead end, entirely ungated.

Both handlers now treat a missing or unresolvable key as an empty key for registry families with requires_api_key=false (local/ollama/vllm), matching the raw_profile_llm_test / raw_profile_llm_fetch_models gates from #2096.

## Test plan

- [x] Verified live: POST /api/admin/test-provider with {"provider":"local","model":""} and no key returns {"ok":true} against a llama-server on 127.0.0.1:8080 (previously: "No api_key or api_key_env provided")
- [x] Full keyless octoscode TUI onboarding walked end-to-end against this build (family pick, empty-key Test connection, Save, session turn through the local model)
- [x] cargo test -p octos-cli --lib (default features, the CI gate) — 1549 passed
- [x] clippy + fmt clean; the 2 full-suite --features api flakes reproduce identically on clean main (pre-existing, unrelated)

https://claude.ai/code/session_01EktkNBvEhD82cYemLdDW7V

### DIFF
--- a/crates/octos-cli/src/api/admin.rs
+++ b/crates/octos-cli/src/api/admin.rs
@@ -1034,20 +1034,27 @@ pub async fn test_provider(
     use octos_llm::{ChatConfig, LlmProvider};
 
     // Resolve the API key: prefer raw api_key, fall back to reading from saved profile
-    let api_key = if let Some(ref key) = req.api_key {
+    let keyless = octos_llm::registry::is_keyless(&req.provider);
+    let resolved = if let Some(ref key) = req.api_key {
         if !key.is_empty() && !key.contains("***") {
-            key.clone()
+            Ok(key.clone())
         } else {
-            resolve_saved_key(&state, &identity, &req)?
+            resolve_saved_key(&state, &identity, &req)
         }
     } else {
-        resolve_saved_key(&state, &identity, &req)?
+        resolve_saved_key(&state, &identity, &req)
+    };
+    // Keyless local families (local/ollama/vllm) construct without a key —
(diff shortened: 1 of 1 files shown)


heyong4725:
/review round: adversarial pass verified the keyless gates are safe (router middleware is the auth boundary — the swallowed resolve_saved_key 401 was defense-in-depth only; alias handling and masked-key edges verified clean). One finding fixed in 308f7c70: caller-supplied base_url reached the outbound probe unguarded, and keyless families now reach it credential-free — added a link-local guard (169.254/16 metadata endpoints, fe80::/10, IPv4-mapped) to both handlers, deliberately narrower than tools/ssrf.rs so loopback/RFC1918 model servers keep working; also suppressed the literal `Bearer ` header for empty keys.

Pre-existing, flagged for follow-up (unchanged by this PR): authenticated keyed callers can still probe arbitrary base_url with a throwaway key, and body-level profile_id reaches resolve_saved_key uncovered by the x-profile-id strip middleware.

🤖 Generated with Claude Code — https://claude.ai/code/session_01EktkNBvEhD82cYemLdDW7V

PR #2099, opened 2026-08-21
fix(e2e): report uncovered route-inventory methods in parity soak

ymote:
captureRouteInventoryProbes silently skipped inventory methods without probe params, so ~12 newer wire methods (session/hydrate, thread/graph/get, turn/state/get, task/artifact/*, skill/action/job/*, ...) shipped unprobed with no signal.

Now uncovered methods are collected, listed in the summary JSON (uncoveredRouteMethods), printed to stderr, and fail the run (exitCode=1) so routeProbeParams must grow with the inventory.

### DIFF
--- a/e2e/scripts/m18-appui-transport-parity-soak.mjs
+++ b/e2e/scripts/m18-appui-transport-parity-soak.mjs
@@ -1080,12 +1080,12 @@ function routeProbeParams(method) {
   }
 }
 
-async function captureRouteInventoryProbes(client, routeInventory, probes, uncovered) {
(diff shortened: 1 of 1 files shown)


ymote:
Redesigned per review: routeProbeParams now returns a PROBED_ELSEWHERE sentinel for methods exercised by dedicated scenario steps (no more null ambiguity); unknown methods still throw at the switch default; both transports record into one shared probed Set (deduped, not appended twice); uncoveredRouteMethods = inventory − probed − probed-elsewhere, reported sorted and fails the run when non-empty. Existing methods like client_hello/session/open/turn/start no longer trip the check.

PR #2100, opened 2026-08-21
docs(spec): document profile/local/create additive fields and projection/envelope v2

ymote:
Two spec gaps found in the spec-vs-impl audit:

- profile/local/create: document optional requested_id (negotiated via profile.local_create.requested_id.v1) and make_default fields, plus the username/email-optional semantics in the requested_id shape
- projection/envelope: v1/v2 forms share the method name but the spec only described v1. Document EnvelopeV2 wire shape (turn_id on every envelope, optional UiCursor cursor, background child-stream identity) negotiated via projection.envelope.v2; note message/persisted (UPCR-2026-012) is retired with projection/envelope as its sole successor.

### DIFF
--- a/api/OCTOS_UI_PROTOCOL_V1_SPEC_2026-04-24.md
+++ b/api/OCTOS_UI_PROTOCOL_V1_SPEC_2026-04-24.md
@@ -1182,8 +1182,8 @@ Clients must use that method list to enable or disable slash commands.
   }
   ```
 
-- request (additive shape, negotiated via the `profile.local_create.requested_id.v1`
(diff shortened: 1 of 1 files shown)


ymote:
Pushed fixes: make_default now documented as gated by the separate `profile.local_create.default.v1` feature with launch-resolution semantics; v2 delivery corrected — canonical v2 envelopes delivered regardless of negotiation, the v2 feature controls historical-record projection, and a v2 connection filters v1/source events (no dual-decode); section 14 header reconciled to state message/persisted is retired.

Issue #2102, opened 2026-08-21
fix(orchestrator): duplicate output in multi-agent turns — replay lacks row-level dedupe, TOCTOU guard scoped to External, ScatterJoin key allows re-join


ymote:
# Multi-agent duplicate output: three uncovered gaps in the NEW-03 / NEW-16 / #1707 fix chain

**Title suggestion:** `fix(orchestrator): duplicate output in multi-agent turns — messages_page lacks row-level dedupe, TOCTOU guard scoped to External only, and ScatterJoinComplete key allows re-join replays`

## Summary

While investigating duplicate assistant output in multi-agent (scatter-join) sessions on `main` (`e8d129b`), code inspection of the NEW-03 / NEW-16 / #1707 fix chain found three gaps that are individually documented in comments but not tracked as issues. Together they explain repeated output of the same content in one conversation.

## Gap 1 — `handle_session_messages_page` has no source / row-level filter (NEW-03 recurrence)

The "ghost ack" fix (Fleet-UX soak NEW-03) was applied at the **producer** side: synthesized spawn_only acks are no longer persisted. But `handle_session_messages_page` (`ui_protocol_transport.rs:24478`) still walks JSONL directly with **no source filter and no `(turn_id, message_index)` row-level dedupe**. Any future path that double-persists a row (see Gap 2) will be re-rendered on every SPA refresh / replay — exactly the original ghost-ack failure mode (soak measured 33 occurrences under one turn).

Proposed: filter or collapse by durable row identity at the replay boundary, so the replay layer is robust regardless of producer-side discipline.

## Gap 2 — the recently-claimed TOCTOU guard is scoped to `External` only

`master_continuation_scheduler.rs:516+`: the `recently_claimed_external` window exists because a drain between the legacy `on_failure` and unified `on_terminal` enqueues (both fired inside one `mark_failed`) removes the dedupe key from `pending_by_key`, so the second enqueue passes the `pending_by_key` check and the same event is delivered **twice**.

The comment explicitly limits the guard: *"Scoped to External so recurring loop/goal/child keys stay reusable across ticks."* But the same drain-between-two-enqueues interleaving can occur for `ChildCompleted` / `ScatterJoinComplete` keys (e.g. legacy + unified terminal callbacks in `task_supervisor.rs:2828` — the comment there says "shared dedupe keys collapse the two", which only holds if both enqueues happen with the key still resident). A drain in between defeats it.

[...]

PR #2103, opened 2026-08-21
fix(cli): dedupe double-persisted rows in session message replay (Refs #2102, Gap 1)

ymote:
Gap 1 (part 1) of #2102: collapse adjacent byte-identical rows (role, thread_id, content, timestamp) in read_profile_session_messages below the MessageInfo mapping, so NEW-16-class double-persisted rows are never re-rendered on replay (ghost-ack failure mode, soak measured 33 dup occurrences under one turn). Pure read-path change; cargo check -p octos-cli clean. since_seq cursor paging follows as a follow-up commit. See #2102 for the full fix plan and double-review analysis.

### DIFF
--- a/crates/octos-cli/src/api/handlers.rs
+++ b/crates/octos-cli/src/api/handlers.rs
@@ -2160,6 +2160,28 @@ async fn serve_file_impl(
 /// [`standalone_api_session_key_candidates_with_topic`] is for the
 /// process-wide `state.sessions`, where bare keys could theoretically
 /// match a JSONL written by a different tenant.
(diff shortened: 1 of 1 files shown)


anantheparty (approved):
审查通过（2026-08-25 巡检）：相邻字节相同行折叠（NEW-16 重复追加类）逻辑正确——Message 无显式 seq，连续 (role,thread_id,content,timestamp) 相同即双写；要求相邻避免误折叠跨轮次重复文本；注释充分。CI 为 8/21 旧 run，rebase 到最新 main 重跑后即可合并。

PR #2105, opened 2026-08-21
fix(autonomy): explicit scatter-join dedupe key + persisted join epoch (Refs #2102, Gap 3)

ymote:
Gap 3 of #2102. The auto-derived stable_dedupe_key folds terminal_children (and every metadata pair) into the key, so a group re-reaching all-terminal with a different child count enqueued a NEW ScatterJoinComplete — the master re-emitted the joined answer (the duplicate-output symptom in the issue). This PR adds per-group ScatterJoinState {join_epoch, last_joined_key} (serde-default, additive), an EXPLICIT dedupe key scatter_join/{group}/{session}/{profile}/{cwd_hash}/{epoch} that bypasses metadata folding, an equality skip that is restart-safe and covers the re-marked-terminal case, and an epoch bump at spawn admission (before insertion, under the state lock). cargo test -p octos-cli --lib autonomy: 370/370 pass. See #2102 for the double-reviewed fix plan; companion PRs #2103 (Gap 1) and #2104 (Gap 2).

### DIFF
--- a/crates/octos-cli/src/autonomy/agent_orchestrator.rs
+++ b/crates/octos-cli/src/autonomy/agent_orchestrator.rs
@@ -2149,6 +2149,30 @@ impl InProcessAgentOrchestrator {
             .get(&upsert.agent_id)
             .map(|agent| agent.status.clone());
         let (agent, payload, transitioned_terminal) = {
+            // Refs #2102 (Gap 3): live spawn admission. A NEW agent record
+            // entering an ALREADY-JOINED group bumps the join epoch BEFORE
+            // the entry is created, under the same state lock — the crash
+            // window is empty and the bump is single-writer.
+            let existing = state.agents.contains_key(&upsert.agent_id);
+            if !existing {
+                let group = format!(
+                    "agent-group:{}:{}:{}",
+                    upsert.profile_id,
+                    upsert.session_id,
+                    upsert.parent_agent_id.as_deref().unwrap_or("master")
+                );
+                let already_joined = state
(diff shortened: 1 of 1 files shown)


ymote:
Self-review note (two findings from adversarial review):

1. **FIXED in bb212b4** — the epoch bump initially sat on the boot-restore `agents.insert` path only. Live spawns enter the roster via the upsert `or_insert_with` arm, so a re-expansion via live spawn never bumped the epoch and the join would have been wrongly skipped. The bump now lives on the upsert path (before entry creation, same state lock); the restore path keeps no bump. 368/368 autonomy tests pass.

2. **Known limitation** — `AutonomyRuntimeState` is in-memory only (agent records persist separately via `persist_agent_*` into the supervisor store), so `scatter_join_state` does not survive a restart. After a restart, the first all-terminal edge of a previously-joined group will re-join once (pre-fix behavior). The in-process duplicate-output symptom (re-observation, metadata drift, child-count fluctuation, live double-fire) is fully covered. Full durability = persisting a small per-group {join_epoch, last_joined_key} map beside the supervisor store — happy to add as a follow-up if maintainers want it in this PR.

anantheparty (approved):
审查通过（2026-08-25 巡检）：scatter-join 去重键 + join epoch 设计正确——state lock 内提前 bump（无崩溃窗口）、持久化 epoch 跨重启、additive 默认兼容旧状态。CI 为旧 run，rebase 后重跑即可合并。

PR #2110, opened 2026-08-22
perf(plugins): speed up Vertex-backed skill actions

alan0x:
> Superseded by direct upstream PR #2111. This closed PR is retained only for history.

## What changed

- Multiple tools from the same learning-coach plugin share a short-lived Vertex token cache instead of performing a new OAuth token exchange for every tool call.
- A host-provided token is exposed only to tools whose manifest explicitly allows `VERTEX_ACCESS_TOKEN`; the existing service-account fallback remains available when host token acquisition fails.
- The session workspace and skill output directory are passed separately so selection assistance can read persisted ink images.
- `OCTOS_SESSION_WORKSPACE` is also manifest-gated, preventing plugins that did not request it from reading the session path.

## Validation

- `cargo test -p octos-agent --lib session_workspace_ -- --nocapture`
- `cargo test -p octos-agent --lib vertex_access_token_ -- --nocapture`
- `cargo check -p octos-agent`
- `cargo clippy -p octos-agent --lib -- -D warnings`

The full `cargo test -p octos-agent --lib` run reported 2,591 passed, 3 ignored, and 2 failed. Both failures are in pre-existing SSRF/DNS tests because the current test environment resolves nonexistent domains to private IP addresses. This PR does not modify those files, and all tests related to this change pass.

### DIFF
--- a/crates/octos-agent/src/plugins/loader.rs
+++ b/crates/octos-agent/src/plugins/loader.rs
@@ -1,9 +1,11 @@
 //! Plugin loader: scans directories for plugins and registers their tools.
 
 use std::path::{Path, PathBuf};
+use std::sync::Arc;
 use std::time::Duration;
 
 use eyre::Result;
+use octos_llm::vertex_auth::{ServiceAccount, TokenSource, VertexTokenProvider};
 use sha2::{Digest, Sha256};
 use tracing::{info, warn};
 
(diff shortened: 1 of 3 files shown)


alan0x:
Superseded by the same branch pushed directly to `octos-org/octos`; the replacement is PR #2111. This fork-based PR is closed and is no longer a delivery entry point.

PR #2111, opened 2026-08-22
perf(plugins): speed up Vertex-backed skill actions

alan0x:
## What changed

- Multiple tools from the same learning-coach plugin now share a short-lived Vertex token cache instead of performing a new OAuth token exchange for every tool call.
- A host-provided token is exposed only to tools whose manifest explicitly allows `VERTEX_ACCESS_TOKEN`; the existing service-account fallback remains available when host token acquisition fails.
- The session workspace and skill output directory are now passed separately so selection assistance can read persisted ink images.
- `OCTOS_SESSION_WORKSPACE` is also manifest-gated, preventing plugins that did not request it from reading the session path.

## Validation

- `cargo test -p octos-agent --lib session_workspace_ -- --nocapture`
- `cargo test -p octos-agent --lib vertex_access_token_ -- --nocapture`
- `cargo check -p octos-agent`
- `cargo clippy -p octos-agent --lib -- -D warnings`

The full `cargo test -p octos-agent --lib` run reported 2,591 passed, 3 ignored, and 2 failed. Both failures are in pre-existing SSRF/DNS tests because the current test environment resolves nonexistent domains to private IP addresses. This PR does not modify those files, and all tests related to this change pass.

### DIFF
--- a/crates/octos-agent/src/plugins/loader.rs
+++ b/crates/octos-agent/src/plugins/loader.rs
@@ -1,9 +1,11 @@
 //! Plugin loader: scans directories for plugins and registers their tools.
 
 use std::path::{Path, PathBuf};
+use std::sync::Arc;
 use std::time::Duration;
 
 use eyre::Result;
+use octos_llm::vertex_auth::{ServiceAccount, TokenSource, VertexTokenProvider};
 use sha2::{Digest, Sha256};
 use tracing::{info, warn};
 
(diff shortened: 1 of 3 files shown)


alan0x:
CI note: `dashboard / Verify dashboard providers.json is in sync with the model catalog` currently fails because the latest `main` includes the `local` family in `model_catalog.json`, while `dashboard/src/providers.json` does not yet contain the corresponding family/environment entry. There is no branch-versus-`origin/main` diff in `model_catalog.json`, `dashboard/src/providers.json`, or `scripts/sync-dashboard-providers.py`. Running the same sync script locally reports: `WARNING: family 'local' is in the catalog but NOT in providers.json (needs an env var)`. This is unrelated to the plugin token/workspace changes in this PR, so an unverified dashboard environment mapping was not mixed into this change.

alan0x:
The two remaining Clippy failures also originate from upstream `main`, not this PR. GitHub currently uses the Rust 1.98 stable toolchain, whose newly enabled `clippy::result_large_err` lint reports an error at `crates/octos-bus/src/bus.rs:31`. This branch has no diff from `origin/main` in `crates/octos-bus/src/bus.rs` or `crates/octos-cli`. Both CI Clippy commands pass with the current local toolchain, and the Clippy check for the modified `octos-agent` crate, the full library CI, and the integration CI all pass. An unrelated `octos-bus` API change was therefore not added to this PR.

PR #2114, opened 2026-08-23
perf: session-load cold start 10.5s → 0.21s — lazy ledger recovery, per-session snapshots, single-pass frame truncation

ZhangHanDong:
## Summary

Cold-starting `octos serve --stdio` against a real long-lived instance (30 sessions, 41k+ ledger events, 45MB main-session ledger) took **10.5–11.4s** before the first `session/hydrate` response reached the client. This PR takes it to **~0.21s** (measured 209/207/207ms across three cold starts on the same instance) via three independent fixes, ordered by discovery:

### 1. `perf(ledger)`: lazy per-session recovery
Boot used to replay every event of every session (`ledger recovery complete … events_recovered=41608`), a cost paid on every TUI launch in stdio mode and growing monotonically with usage. Now `recover()` only builds a session index (O(sessions)); a session's log is replayed on first touch (`ensure_session_loaded`, hooked into hydrate/replay/snapshot/append paths), with the stale-live guard and orphan-row sweep applied exactly once. A corrupt session latches failed without blocking boot or other sessions.

### 2. `perf(ledger)`: per-session projection snapshots
Recovery = latest snapshot + tail replay. Snapshots are written on an append cadence (every 4096 events) **and** bootstrapped right after a full replay of a pre-existing ledger, so legacy ledgers benefit on their second touch (measured: the 45MB session recovers in **123ms**). Corrupt/unknown-version snapshots fall back to full replay; ledgers without snapshots read unchanged (zero migration).

### 3. `fix(transport)`: single-pass `preview_oversized_frame` — the actual 10s
With ledger costs gone, the stall persisted — full probe chain (handler segments, serde, writer-thread dequeue/flush timestamps) isolated it to the over-cap frame truncation loop: `loop { serde_json::to_string(&value); truncate one largest field }` **re-serializes the multi-MB payload every round** — O(payload × rounds) ≈ 10s of CPU for a large hydrate response, on every cold start. (The writer thread itself was innocent: it sat on an empty channel for 11.1s and flushed the 524KB frame in 7.5ms once fed.) Rewritten single-pass: collect the top-K largest strings/arrays in one walk, cut the overshoot in one round (escape-aware estimate), verify with at most one extra serialization — bounded at 2 total serializations. Regression test binds a multi-MB hydrate-shaped payload to a fast single-pass budget.

## Measurements (same instance, cold start → hydrate applied, 2654-msg session)

| stage | before | after |
|---|---|---|
| serve boot (ledger) | 10.5s eager replay | 0.26s index-only |
| 45MB session recovery | full replay | 123ms (snapshot+tail) |
| hydrate response send | ~10.3s truncation loop | ~4ms serialize + single-pass cap |
| **client: request → applied** | **10.5–11.4s** | **0.21s** |

## Verification

`cargo test -p octos-cli --lib --features api`: 3054 passed / 0 failed (rebased on current main). `clippy --features api` clean, `fmt` clean. New tests: 4 lazy-recovery contracts (boot-no-replay, projection equivalence vs eager, touch isolation, corrupt isolation), snapshot equivalence/fallback/bootstrap, single-pass truncation budget.

Draft for maintainer review — happy to split into three PRs if preferred (commits are independent and stack cleanly).

https://claude.ai/code/session_01N733qqKtLk2Z9HRDTJ6ier

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol_ledger.rs
+++ b/crates/octos-cli/src/api/ui_protocol_ledger.rs
@@ -1125,38 +1125,80 @@ impl UiProtocolLedger {
         // unreadable, or unknown-version snapshot logs a warning and
         // falls back to a full replay — the log stays the source of
         // truth (fault-tolerance contract).
+        //
+        // ymote P1 (PR #2114 review): the snapshot shortcut must be
+        // CONDITIONAL on the requested replay range. When the caller asks
+        // to replay from a cursor OLDER than the snapshot's oldest entry
+        // (e.g. cursor 2000 while the snapshot ring only covers
+        // 4097..8192), the retained JSONL still holds the pre-snapshot
+        // records and they MUST be replayed — seeding the bounded ring
+        // and skipping everything ≤ head would silently drop
+        // 2001..4096 AND misreport cursor_out_of_range (oldest_seq taken
+        // from the snapshot ring). So: only skip log records up to the
+        // snapshot head when the requested range is fully covered by the
+        // snapshot; otherwise scan the log from seq 0 (the snapshot ring
+        // still seeds the retained ring, but nothing is skipped).
         let mut skip_through_seq = 0u64;
         match self.read_session_snapshot(session_id, session_dir) {
             Ok(Some(snapshot)) => {
                 head_seq = snapshot.head_seq;
-                skip_through_seq = snapshot.head_seq;
+                let snapshot_oldest = snapshot.entries.first().map(|r| r.seq);
+                let shortcut_ok = match replay_after_seq {
+                    Some(after) => snapshot_oldest.is_some_and(|oldest| after >= oldest - 1),
+                    None => snapshot_oldest.is_some_and(|oldest| oldest <= 1),
+                };
+                if shortcut_ok {
+                    skip_through_seq = snapshot.head_seq;
+                }
+                // NOTE: when the shortcut is disabled we deliberately do
+                // NOT seed anything from the snapshot (see the loop
+                // below) — the log scan rebuilds the full retained ring
+                // and replay in seq order, so there is nothing to dedupe
+                // and snapshot_seeded_floor/oldest stay 0.
                 for record in snapshot.entries {
                     oldest_seq.get_or_insert(record.seq);
-                    if replay_after_seq.is_some_and(|after_seq| record.seq > after_seq) {
-                        replay_entries.push(LedgeredUiProtocolEvent {
-                            cursor: UiCursor {
-                                stream: session_id.0.clone(),
-                                seq: record.seq,
-                            },
-                            event: record.event.clone(),
-                            from_connection: None,
+                    // ymote P1: only the SHORTCUT path materialises the
+                    // snapshot (retained ring + replay_entries). When the
+                    // shortcut is disabled the FULL state (retained ring
+                    // AND replay) is rebuilt by the log scan below in seq
+                    // order — seeding here would duplicate the tail and
+                    // mis-order the pre-snapshot gap in both.
+                    if shortcut_ok {
+                        if replay_after_seq.is_some_and(|after| record.seq > after) {
+                            replay_entries.push(LedgeredUiProtocolEvent {
+                                cursor: UiCursor {
+                                    stream: session_id.0.clone(),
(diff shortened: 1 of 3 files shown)


ymote (dismissed):
[P1] Preserve the JSONL cursor-replay range when a projection snapshot is present. In crates/octos-cli/src/api/ui_protocol_ledger.rs:1129-1223, recovery seeds only the bounded snapshot ring, sets skip_through_seq = snapshot.head_seq, and then skips every log record at or below that head. At the second cadence boundary with retained_per_session=4096, snapshot.json contains roughly seq 4097..8192; if the retained JSONL files still contain seq 1..8192, replay after cursor 2000 should return 2001..head but now skips 2001..4096 and reports cursor_out_of_range because oldest_seq is taken from the snapshot ring. This regresses the existing contract that entries dropped from RAM remain replayable from retained logs.

Please make the snapshot shortcut conditional on the requested replay range (or scan/parse the missing pre-snapshot interval) and add a regression test with a ring smaller than the unrotated log: snapshot at seq 4 with ring size 2, then replay after seq 1 must return seq 2..5. The current snapshot equivalence test does not catch this because setting snapshot_every_events=0 disables writes but read_session_disk_snapshot still consumes the existing snapshot.

ZhangHanDong:
Thanks for the sharp catch — the shortcut was indeed clobbering the pre-snapshot replay contract.

Fixed in e42130a9: the snapshot seed path now applies **only when the requested cursor falls inside the snapshot ring's range**; a replay-after cursor older than the ring's oldest seq takes the full log-scan path, so entries dropped from RAM but retained in JSONL stay replayable exactly as before.

Regression test `snapshot_shortcut_replays_pre_snapshot_range` binds your exact scenario: ring size 2, snapshot written at seq 4, replay after seq 1 → returns seq 2..5 (and asserts no cursor_out_of_range). It also avoids the equivalence test's blind spot you noted — it writes a real snapshot and replays across its lower boundary rather than disabling writes.

Full `-p octos-cli --features api` suite green, clippy clean, fmt clean.

https://claude.ai/code/session_015gGAWQ8yThjWbCfdGrtNRk

anantheparty:
认同 ymote 的 P1：快照存在时跳过 pre-snapshot 日志违反重放契约（RAM 丢弃的条目仍应从保留日志可重放）。建议按意见修复（snapshot 捷径仅当请求范围在快照内时启用，或扫描缺失区间）并补 ring<log 的回归测试。

ZhangHanDong:
@anantheparty agreed — fixed in e42130a9 (shortcut only within the snapshot ring's replay range; pre-snapshot JSONL stays replayable, bound by `snapshot_shortcut_replays_pre_snapshot_range`).

Two follow-ups pushed since: 33cdd7ce hardens the shortcut against corrupt seq-0 snapshot entries (degrades to full replay instead of a debug panic), and cfa8fb15 — out of our adversarial review pass — restores the shortcut for the trimmed-ring **from-beginning** hydrate (None/seq-0): that path is answered by the retained ring alone, so requiring `oldest == 1` forced a full log rescan on exactly the large sessions this PR targets. Cursor replays (`after > 0`) keep the strict range check and the pre-snapshot JSONL contract. New pin: `snapshot_shortcut_applies_to_trimmed_ring_from_beginning`.

ZhangHanDong:
Correction to my previous comment, from our second adversarial-review round: cfa8fb15's trimmed-ring shortcut was too broad — an explicit `session/open {after:{seq:0}}` cursor (which has no from-beginning exemption in its range check) went from a successful full JSONL replay to `cursor_out_of_range`. 597e67ca scopes the seq-0 shortcut to the hydrate path only (caller-semantic flag, not raw seq): `session/open` seq-0 keeps the full-replay contract, hydrate from-beginning keeps the retained-window fast path. New end-to-end pin at the public entries: `session_open_seq0_full_replays_but_hydrate_shortcuts` (open → [1..5], hydrate → retained [4,5], same trimmed-snapshot scenario).

anantheparty (approved):
复审通过（2026-08-25）：ymote P1 已修复——快照捷径仅当请求范围完全被快照覆盖时启用（skip_through_seq 条件化），并补 ring=2/snapshot@4/replay-after-1 的回归测试（1289 行）。

PR #2116, opened 2026-08-24
feat(cli,serve): OLP observability — read-only goal/peer/ledger JSON commands, structured events.jsonl, inbox path

ZhangHanDong:
## Summary

First half of the Outer-Loop Protocol (OLP) L1 surface: machine-readable **observability** for external orchestrator models (Claude Code / Codex / scripts) supervising long-horizon octos goals. Today an outer agent must grep human logs, hand-compute the unstable session hash, and tail per-instance directories — all three are field-tested-fragile.

### What's included (9 commits, each independently reviewable)

- **`octos inbox path --session <key>`** — resolves a session's inbox notes path via the runtime's own hash implementation (external consumers must never reimplement it).
- **`octos goal status [--goal <id>] --json`** — direct goal-ledger read; works with serve down.
- **`octos peer list --json`** — direct peers/ directory read.
- **`octos ledger tail <goal_id> --json`** — findings / escalations / decisions.
- **`events.jsonl`** — serve appends structured event lines ({ts, kind, goal_id?, slug?, session?, model_lane?, detail}; kinds: peer_staged, finding_recorded, escalation, goal_transition, steer_consumed, turn_error). Best-effort: a failed write never affects the main flow; the goal ledger stays the durable source of truth.
- All read commands resolve the per-instance profile data root via `ProfileStore::resolve_data_dir` — the same function serve uses, after a mid-development rework proved that reimplemented addressing diverges (three commands returned empty/errors against real data until unified).

### Verification

Every command validated against a real long-lived instance (20 peers, multi-goal ledgers) with per-command contract tests + a tempdir real-layout e2e test. `cargo test -p octos-cli --features api` green; clippy clean; fmt clean.

Stacked under this: a `session/steer` control-channel PR follows.

https://claude.ai/code/session_015gGAWQ8yThjWbCfdGrtNRk


anantheparty (approved):
Reviewed obs_events.rs, obs_resolve.rs, and the command wiring. The design principle holds: data-root resolution delegates to `ProfileStore::resolve_data_dir` (the one function serve uses) instead of re-assembling paths, and the cwd-hash mirrors the launcher byte-for-byte — the fragility the contract calls out is addressed at the right layer. `append_obs_event` honors the never-fail contract (every path debug-log + return). 61 assertions in the diff, 15/15 checks green.

One nit, non-blocking: `append_obs_event` issues two `write_all` calls (line, then newline). With O_APPEND each write is atomic but they are two separate calls — two processes appending to the same events.jsonl could interleave line/tail and produce a torn line. Single-writer-per-instance is today's reality (per-instance data dirs), but folding it into one `write_all(format!("{line}\n").as_bytes())` makes single-line atomicity explicit and free. Happy to see it follow-up or in #2117.

Issue #899, opened 2026-05-11
gateway: redb dual-open crashloop — every profile crashlooping ~2s on M11-F binary


ymote:
## Symptom

Every ~2 seconds, per profile, on every mini in the production fleet (mini1/2/3/5) since PR #888 (M11-F gateway consolidation) deployed:

```
WARN ProfileRuntime::bootstrap failed; gateway falling back to inline assembly profile_id=<X> error=failed to open episode store for profile '<X>'
[gateway] opening episode store at /Users/cloud/.octos/profiles/<X>/data
Error:
   0: failed to open episode store
   1: failed to open redb database
   2: Database already open. Cannot acquire lock.
Location: crates/octos-memory/src/store.rs:43
ERROR gateway failed to start
INFO auto-restarting crashed gateway profile=<X>
```

## Root cause

`octos serve` is the main daemon and opens `~/.octos/profiles/<X>/data/episodes.redb` per profile via `EpisodeStore::open` (`crates/octos-memory/src/store.rs:43`) when bootstrapping `ProfileRuntime` for every enabled profile (see `crates/octos-cli/src/commands/serve.rs:394`).

`octos serve` then spawns `octos gateway --profile <X>` as a **separate subprocess** (`crates/octos-cli/src/process_manager.rs:246`) for channel polling.

The gateway subprocess **also** calls `ProfileRuntime::bootstrap` (`crates/octos-cli/src/commands/gateway/gateway_runtime.rs:292`). Bootstrap calls `EpisodeStore::open` for the same redb path. **redb is single-writer-single-process**, so the second open fails with `DatabaseError::DatabaseAlreadyOpen` ("Database already open. Cannot acquire lock.").

PR #888 review comment said \"no redb double-open\" — but that only holds when bootstrap is called from the SAME process. In production the gateway runs as a child process of serve, so the lock contention is unavoidable.

When bootstrap fails, gateway falls back to inline assembly (`gateway_runtime.rs:380`), which **also** opens the episode store, hits the same lock, and exits. launchd / serve auto-restarts. Infinite loop.

## Impact

- All Telegram / Discord / WeChat / WhatsApp / etc. channel polling is **completely broken** across the fleet — gateway subprocesses die before reaching channel registration.
- `/api/chat` is unaffected (serve owns the canonical EpisodeStore).
- Log volume is enormous (~30 lines / profile / 2s ≈ thousands of lines / minute / mini) — triage greps are noisy.

## Severity

[...]

PR #2118, opened 2026-08-25
fix: make redb data-dir lock contention legible at session/open

alanpoon:
## Problem

A second `octos serve` against the same data dir fails at the first `session/open`, and the error that reaches the client says nothing useful:

```
[runtime_unavailable]: session/open request tui-3 failed: failed to bootstrap
ProfileRuntime for profile 'alan': failed to open episode store for profile 'alan'
```

That is the whole message. No path, no cause, no remedy — and `runtime_unavailable` is the same kind returned for a missing provider, a bad tool config, or unresolvable credentials, so a client has nothing to branch on either.

The underlying condition is mundane: redb is single-writer-single-process, and a dev serve started on a different port still resolves to `$OCTOS_HOME`, because the port isolates the socket, not the storage. `BootstrapRole::Serve` deliberately refuses to degrade (#899 gave that fallback to `Gateway` only, so a misordered start can't quietly flip canonical ownership), and profiles bootstrap lazily, so the refusal lands at `session/open` rather than at startup.

Two defects made it illegible:

1. **The cause chain was dropped.** `ui_protocol_transport.rs` formatted with `{error}` — plain `Display` on an `eyre::Report`, which prints only the outermost context. The store's own message already named the contended `episodes.redb` path; it never left the process.
2. **Lock contention wasn't distinguishable.** It was flattened into `runtime_unavailable` along with every other bootstrap failure, and the only way to tell them apart was to match on prose.

## Change

**`octos-memory`** — replace the untyped `eyre::eyre!` in the strict-open branch with a typed `EpisodeStoreLocked { path }`, plus `is_episode_store_locked()` for structural detection over the eyre chain. This mirrors the existing `is_permission_denied_error`, which downcasts to `std::io::Error` rather than string-matching, and it survives the `wrap_err` context `ProfileRuntime::bootstrap` adds on the way up. The typed error is returned **unwrapped** so its operator-facing sentence is outermost; the developer-facing "wrong opener?" hint becomes a `debug!` line. redb's own `Database already open. Cannot acquire lock.` wording is preserved — operators grep for it.

**`octos-cli`** — `session/open` now branches: lock contention gets a typed `data_dir_locked` kind whose message names the profile and both remedies, following the `workspace_not_writable_error` precedent of treating a fixable config problem as its own kind with a sentence the client renders verbatim. Everything else stays `runtime_unavailable`, now formatted with `{error:#}` so the chain survives.

The remedy text explicitly calls out supervised deployments — under launchd `KeepAlive`, killing the PID accomplishes nothing because the lock is handed straight back to a fresh process within a second. That is a genuinely confusing thing to debug from a client, and it is what motivated this PR.

## Why this path still matters

The fast-fail `OCTOS_DATA_DIR_LOCKED` startup guard should normally catch this first. It can't fire when the incumbent predates it — an older serve holding the redb never took that lock, so the new serve starts cleanly and only fails later. Between two current binaries you get the clean startup refusal; this path is the backstop for mixed-version and upgrade-in-progress installs.

## Tests

6 passing — 5 new, 1 strengthened:

- typed cause survives re-wrapping; message names both the path and the remedy
- a corrupt (non-redb) file is **not** flagged as lock contention — guards the detector against over-matching
- transport returns `data_dir_locked` with profile, remedy, and underlying cause in `data.message`
- an unrelated bootstrap failure keeps the generic `runtime_unavailable` kind
- `second_serve_role_bootstrap_fails_loudly_when_redb_already_owned` now also asserts the typed cause survives bootstrap's own `wrap_err` — failing loudly is only half the contract when a caller upstack has to classify the failure

```
cargo test -p octos-memory --lib          # 93 passed
cargo test -p octos-cli --features api --lib -- <the three above>   # 3 passed
cargo check -p octos-cli --features api --tests                     # clean, no warnings
```

## Compatibility

Additive. `data_dir_locked` is a new `kind`; clients that don't know it fall back to rendering `message`, which is now a complete sentence rather than a truncated chain. No signature changes to `EpisodeStore::open*`.

### DIFF
--- a/crates/octos-bus/src/twilio_channel.rs
+++ b/crates/octos-bus/src/twilio_channel.rs
@@ -154,7 +154,7 @@ fn sha1(data: &[u8]) -> [u8; 20] {
 
         let (mut a, mut b, mut c, mut d, mut e) = (h0, h1, h2, h3, h4);
 
-        for i in 0..80 {
+        for (i, &wi) in w.iter().enumerate() {
             let (f, k) = match i {
                 0..=19 => ((b & c) | ((!b) & d), 0x5A827999u32),
                 20..=39 => (b ^ c ^ d, 0x6ED9EBA1u32),
@@ -167,7 +167,7 @@ fn sha1(data: &[u8]) -> [u8; 20] {
                 .wrapping_add(f)
                 .wrapping_add(e)
                 .wrapping_add(k)
-                .wrapping_add(w[i]);
+                .wrapping_add(wi);
             e = d;
             d = c;
             c = b.rotate_left(30);
@@ -195,7 +195,7 @@ fn sha1(data: &[u8]) -> [u8; 20] {
 fn base64_encode(data: &[u8]) -> String {
     const TABLE: &[u8; 64] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 
-    let mut out = String::with_capacity((data.len() + 2) / 3 * 4);
+    let mut out = String::with_capacity(data.len().div_ceil(3) * 4);
     for chunk in data.chunks(3) {
         let b0 = chunk[0] as u32;
         let b1 = if chunk.len() > 1 { chunk[1] as u32 } else { 0 };
--- a/crates/octos-bus/src/wecom_channel.rs
+++ b/crates/octos-bus/src/wecom_channel.rs
@@ -49,6 +49,7 @@ pub struct WeComChannel {
 }
 
 impl WeComChannel {
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         corp_id: &str,
         agent_id: &str,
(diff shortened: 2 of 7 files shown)


anantheparty:
CI 观察（2026-08-25）：check 通过但 check-matrix 失败——clippy 在 matrix feature 下报 （crates/octos-cli 约 18925 行）'creates a reference which is immediately dereferenced'。修复：去掉多余的 &（），补一行即可。修复后我立即复核合并。

anantheparty (approved):
LGTM. The clippy cleanup is safe (iterating 22:37  up 22 mins, 13 users, load averages: 14.52 17.40 17.28
USER       TTY      FROM    LOGIN@  IDLE WHAT
kamico     console  -      22:16      21 -
kamico     s000     -      22:16      20 -zsh
kamico     s001     -      22:16      20 -zsh
kamico     s002     -      22:16      20 -zsh
kamico     s003     -      22:16      20 -zsh
kamico     s004     -      22:16      20 -zsh
kamico     s005     -      22:16      20 -zsh
kamico     s006     -      22:16      20 -zsh
kamico     s007     -      22:16      20 -zsh
kamico     s008     -      22:16      20 -zsh
kamico     s009     -      22:16      20 -zsh
kamico     s010     -      22:16      20 -zsh
kamico     s011     -      22:16      20 -zsh by enumerate changes nothing semantically),  is equivalent here, and the new test pins the eyre wrap-chain detection plus the  kind. All matrix checks green.

PR #2120, opened 2026-08-25
fix(ui-protocol): scoped goal-event generations + goal-frame capability gating + forwarder hardening (#2065 rebased)

anantheparty:
Rebased successor of #2065 (head branch history rewritten; original branch cannot be force-pushed under the no-push ruleset).

## What changed vs #2065

- Rebased onto current main (9 commits), resolving conflicts:
  - `ui_protocol_tests.rs`: kept both the main-side usage-status tests (#2102) and the #2062 goal-snapshot tests
  - `ui_protocol_transport.rs`: reconciled with #2067's `profile_scope` — the forwarder now receives the captured profile scope (capture-at-open semantics) instead of re-reading `ws.snapshot_live_profile_id()`
- Dropped the dead `profile_id` field from `SessionOpenOutcome` (superseded by #2067's `profile_scope`)
- Test call sites updated for `pinned_profile_id` / `profile_scope` parameters
- Also fixes #2103's `Message` import regression (see #2119, which lands the same fix on main first)

Verified: `cargo check --features api --tests` clean, rustfmt clean, 187 goal tests + 3 session_open_goal tests pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -477,6 +477,14 @@ jobs:
       - name: Background activity human sink (octos-cli)
         run: cargo test -p octos-cli --features api background_activity -- --nocapture
 
+      # #2062 goal-chip null snapshot: session/open must push an explicit
+      # `session/goal/cleared` when no goal is bound, so a goal cleared
+      # elsewhere can't survive a reconnect as a stale cached chip. Lives in
+      # the `api`-gated ui_protocol module — invisible to the unfeatured runs
+      # above, same reason as the §6 guard.
+      - name: Session-open goal null snapshot (octos-cli)
+        run: cargo test -p octos-cli --features api session_open_goal -- --nocapture
+
       # The smart_home/* WS methods are `api`-gated too, so the unfeatured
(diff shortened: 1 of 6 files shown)


BH3GEI (approved):
Verified this rebase against an independent merge-resolution of the same conflicts (main @ cb4466c6):

- The two resolutions are semantically identical: profile_scope is threaded through forward_live_ledger_event as Option<&str>, captured at session/open (never re-read from the connection), matching #2075's H2 semantics; agent_orchestrator.rs and ci.yml are byte-identical; tests-file delta is pure position churn (same 4 #2065 tests + same 4 cache-usage tests present in both).
- On my independently resolved tree: cargo check/fmt/clippy(-D warnings) clean; goal_scope_guard + session_open_goal filters pass; full octos-cli --lib --features api suite 3059 passed / 0 failed (single-threaded, 591s).
- CI on this PR: 16 pass, 7 skip, 0 fail.

This carries the same change already approved on #2065; the earlier deep review (scoped-generation registry, capability gating at the one shared filter, forwarder hardening) still applies.

PR #2121, opened 2026-08-25
feat(agent): measure whether request history is append-only

ymote:
Measurement, not a guard. Off unless `OCTOS_APPEND_ONLY_AUDIT=1`, and it never alters, blocks, or fails a request.

## Why

The messages handed to the model and the messages written to the durable session are two different objects, and **nothing compares them**. They drift, and the drift is silent when it happens — surfacing later as a resumed session the model never had, a context gauge over 100%, or a cost figure computed from a request nobody can reconstruct.

`pi` avoids that whole family by never rewriting history: the prefix only grows, and output is bounded where it is *produced* rather than trimmed after the fact (`docs/harness-context-efficiency.md`, #1638). `deepseek-harness` enforces the same property structurally — it asserts at runtime that every loop-built request equals the projection of its durable log.

Adopting either is a multi-week change to the hottest code we have (`execution.rs` alone is 4,456 lines, and 45 sites mutate the message vector). **This is the cheap precursor that says whether it's warranted**, by measuring how append-only we already are.

## What it found

We are not — verified end to end through the real loop, across two turns.

`truncate_old_tool_results` (`message_repair.rs:513`, called from `loop_compaction.rs:54`) collapses tool results older than the last user message to 800 chars **in place**, on the vector about to be sent, while the durable session keeps the full text.

The July prefix audit cleared this for *caching* because it's idempotent — collapse once, stays collapsed — and that reasoning still holds. But idempotent is not the same as reconstructable, and nobody was asking that question then.

## Two things worth knowing about the build

**Scope.** The auditor lives on `Agent`, not on `LoopTurnState`. Per-turn was the obvious placement and was wrong: the rewrite only touches tool results *before* the last user message, so it fires **across** turns and a per-turn auditor would never observe it. The end-to-end test drives two real turns for exactly that reason.

**Channel.** Findings are recorded out-of-band under `#[cfg(test)]`, not just logged. The first version of this reported nothing — and the nothing meant only that octos-agent's lib tests install no `tracing` subscriber, so every `warn!` went nowhere. A measurement whose silence can't be distinguished from absence isn't a measurement. In production the channel is `tracing`, whose subscriber the operator installs and can verify.

## Verification

```
cargo test -p octos-agent --lib append_only              => ok. 5 passed; 0 failed
cargo fmt --all -- --check                               => clean
cargo test -p octos-agent --lib                          => 2597 passed; 2 failed
```

The two failures are this host's DNS resolver hijacking NXDOMAIN for `*.invalid` (`tools::ssrf`, `tools::web_fetch`). Neither file is in this diff and CI passes both.

## What this does not do

- It does not compare against the **durable session** — that lives in another crate and needs plumbing this deliberately avoids. This is the in-turn/in-session half, which needs no new wiring and catches the known path.
- It does not fix anything. The next step, if the finding holds up in real sessions, is the `pi` work we deferred in July: caps in the tool, recovery in the result, spool-and-grep for large output. That's the prerequisite for stopping the rewrite, and it pays for itself in tokens regardless.

Refs #1638

### DIFF
--- a/crates/octos-agent/src/agent/append_only_audit.rs
+++ b/crates/octos-agent/src/agent/append_only_audit.rs
@@ -1,0 +1,364 @@
+//! Append-only context audit: does a turn's request history only ever GROW?
+//!
+//! ## Why this exists
+//!
+//! The messages handed to the model and the messages written to the durable
+//! session are two different objects. Nothing compares them, so they drift —
+//! and the drift is silent at the moment it happens, surfacing later as a
+//! resumed session the model never had, a context gauge over 100%, or a cost
+//! figure computed from a request nobody can reconstruct.
+//!
+//! The `pi` harness avoids the whole family by never rewriting history: the
+//! request prefix only grows, and output is bounded where it is PRODUCED
+//! rather than trimmed after the fact. `deepseek-harness` enforces the same
+//! property structurally, asserting that every loop-built request equals the
+//! projection of its durable log.
+//!
+//! Adopting either is a large change. This module is the cheap precursor that
+//! tells us whether it is warranted: it MEASURES how append-only we actually
+//! are today, without changing what is sent.
+//!
+//! ## What it checks
+//!
+//! Within one turn, successive requests must be prefix-extensions of each
+//! other: iteration N's message list must be an exact prefix of iteration
+//! N+1's. Any index that changes content, or a list that shrinks, is a
+//! REWRITE — the sent history at that point is no longer reconstructable by
+//! replaying what came before.
+//!
+//! ## What it deliberately does NOT do
+//!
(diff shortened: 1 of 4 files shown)


anantheparty (approved):
Reviewed the full diff. This is exactly the right shape for a precursor measurement: +466/-0 purely additive, off unless `OCTOS_APPEND_ONLY_AUDIT=1`, fingerprints instead of retained copies (flat memory), test-only findings channel behind `#[cfg(test)]`, and poisoned-mutex recovery handled. The end-to-end test asserting the audit *ran* (not just reported nothing) closes the silent-measurement trap called out in the doc comment — good catch on the missing tracing subscriber. One non-blocking note: `DefaultHasher` is fine here since the fingerprint is a correlation handle, not a security boundary. Merging.

Issue #2093, opened 2026-08-20
flaky: plugins::tool::tests::execute_preserves_plugin_structured_metadata fails ~1-in-3 under full-parallel octos-agent --lib


ymote:
## Summary

`plugins::tool::tests::execute_preserves_plugin_structured_metadata` (`crates/octos-agent/src/plugins/tool_tests.rs`) fails intermittently under full-parallel `cargo test -p octos-agent --lib` — roughly **1 run in 3** on the machine where it was observed. It passes every time in isolation.

Filed separately rather than as a note on an in-flight PR because it is the same class as #2077: a nondeterministic red in a heavy job that costs a full CI cycle to diagnose, on PRs that have nothing to do with it.

## Evidence

```
$ cargo test -p octos-agent --lib
test result: FAILED. 2586 passed; 3 failed; 3 ignored; 0 measured; 0 filtered out
failures:
    plugins::tool::tests::execute_preserves_plugin_structured_metadata
    tools::ssrf::tests::test_with_addrs_fails_closed_on_nonexistent_domain      <- separate, environmental
    tools::web_fetch::tests::test_ssrf_dns_failure_blocks_request               <- separate, environmental

$ cargo test -p octos-agent --lib -- --exact plugins::tool::tests::execute_preserves_plugin_structured_metadata
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2591 filtered out
```

Observed across several consecutive full-suite runs: failed, passed, failed, passed.

## What the test does

Writes a temp `script.sh`, constructs a `PluginTool` with `.with_timeout(Duration::from_secs(5))`, awaits `tool.execute(&json!({}))`, and asserts on `result.structured_metadata`. So it spawns a real subprocess under whatever load the rest of the suite is generating. A starved runtime crossing the 5s plugin timeout is the obvious hypothesis, but that is a hypothesis — the failure output was not captured beyond the test name, and nobody has confirmed the mechanism.

## Independence from PR #2088 — the experiment that matters

#2088 adds a test that deliberately holds a `TaskSupervisor` mutex and blocks a thread for up to 250ms, which is exactly the kind of thing that could starve a timing-sensitive neighbour. That was checked, not assumed:

```
$ for i in 1 2 3; do cargo test -p octos-agent --lib -- --skip should_not_lose_a_restore_when_wiring; done
test result: FAILED. 2585 passed; 3 failed; 3 ignored; 0 measured; 1 filtered out
[...]

PR #2124, opened 2026-08-26
feat(tools): tell the model how to recover output the harness truncated

ymote:
## The bug

Truncation happens in the execution loop, at the one point every tool's output funnels through — which is also the point that knows the least. The helper doing the cutting takes a string and a number:

```rust
truncate_head_tail(s: &str, max_len: usize, head_ratio: f32) -> String
```

It cannot know which tool produced the text, whether that tool paginates, or what the parameter is called. So the model was told only:

```
... [47000 bytes omitted] ...
```

…with no way to reach what it lost. Its only recovery was **re-running the call**, which returns the same output cut the same way — spending the tokens the cap existed to save, sometimes twice more before giving up.

The loop knew it truncated but not how to resume. The tool knew how to resume but not that it had been truncated. **Neither half had both pieces.**

## The fix

`Tool::truncation_recovery(args, omitted_bytes)` is the missing half. It runs where the tool's own arguments are still in scope, so it can name a **concrete** next call. It defaults to `None` — a tool with no resume path gets no invented advice, and no existing tool changes behaviour.

Implemented where a real recovery exists:

- **`read_file`** already had `offset`/`limit`. The machinery was there and the model was simply never told. It now echoes the range just read and names the next one — `offset: 201` after `offset: 1, limit: 200`.
- **`grep`** cannot paginate, so its recovery is narrowing — and it offers only the levers still **unused**, rather than suggesting a `path` filter to a call that already has one.

`truncate_head_tail` is unchanged. It stays the generic backstop for everything that funnels through, including plugin tools that don't exist yet. **The bug was never that the backstop was wrong — it's that nothing stood in front of it.**

## Verification

```
3 unit tests (read_file × 2, grep × 1)   => ok. 3 passed; 0 failed
1 end-to-end wiring test                 => ok. 1 passed; 0 failed
cargo fmt --all -- --check               => clean
cargo test -p octos-agent --lib          => 2600 passed; 3 failed
```

**The end-to-end test is mutation-proved.** Deleting the loop wiring makes it fail, with the tool result's tail showing raw truncated output and no advice:

```
the truncated result must carry the tool's recovery advice ... tail was:
"yyyyyyyyyyyyyyyyyyyyyyyyyy..."
```

That check matters here specifically: unit tests prove the hook returns good text, and prove nothing about whether the loop ever **calls** it. An unwired hook returning perfect advice into the void is exactly the failure mode this change exists to remove.

The 3 suite failures are pre-existing and unrelated: two are this host's DNS resolver hijacking NXDOMAIN for `*.invalid` (`tools::ssrf`, `tools::web_fetch`), and one is #2093, which passed 3/3 on re-run.

## What this does not do

- **`shell` still gets no advice.** Its real recovery is spooling full output to a temp file and returning head + tail + path so the model can `grep` it. That's a bigger change and wants its own PR.
- **The caps themselves are untouched.** `search`/`deep_search`/`news_fetch` remain at 200,000 chars — roughly 50K tokens from a single call. Lowering them is the next lever, and it's *safe to pull now* in a way it wasn't before: a smaller cap with a recovery path costs a cheap follow-up call, where previously it cost a full re-run.

Refs #1638

### DIFF
--- a/crates/octos-agent/src/agent/execution.rs
+++ b/crates/octos-agent/src/agent/execution.rs
@@ -2358,9 +2358,32 @@ impl Agent {
                 let _ = hooks.run(HookEvent::AfterToolCall, &payload).await;
             }
 
-            // Per-tool output truncation with head/tail split
+            // Per-tool output truncation with head/tail split.
+            //
+            // The cut is a BACKSTOP: it is the one point every tool's output
+            // funnels through, which is also the point that knows the least —
+            // `truncate_head_tail` receives a string and a number. On its own it
+            // leaves the model with `... [N bytes omitted] ...` and no way to
+            // reach what it lost, so the only recovery is re-running the call,
+            // which returns the same output cut the same way and spends the
+            // tokens the cap was meant to save.
+            //
+            // `Tool::truncation_recovery` is the missing half: the tool still
+            // knows its own arguments and whether it paginates, so it can name a
+            // concrete next call. Tools with no resume path return None and get
+            // no invented advice.
             let limit = octos_core::tool_output_limit(&tc_name);
-            let content = octos_core::truncate_head_tail(&content, limit, 0.7);
+            let untruncated_len = content.len();
+            let mut content = octos_core::truncate_head_tail(&content, limit, 0.7);
+            if content.len() < untruncated_len {
+                if let Some(recovery) = tools.get(&tc_name).and_then(|tool| {
+                    tool.truncation_recovery(&effective_args, untruncated_len - content.len())
+                }) {
+                    content.push('\n');
+                    content.push_str(&recovery);
+                }
+            }
+            let content = content;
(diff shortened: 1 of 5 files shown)


anantheparty (approved):
Reviewed the trait hook, the execution-loop call site, and both tool implementations. Right shape: opt-in default `None` (no invented advice), advice only appended when the cut actually happened (`content.len() < untruncated_len`), recovery text routed through `sanitize_tool_output` like the rest, and the read_file advice names the concrete next call (`offset: 201`) instead of a generic hint. 131 lines of loop tests plus per-tool tests. One cosmetic note, not blocking: the recovery line repeats the `[N bytes omitted]` marker that `truncate_head_tail` already emits — reads slightly redundant in the final output but keeps the advice self-contained. All checks green. Merging.

Issue #2011, opened 2026-08-14
flaky: master_continuation_tick_reenters_actor_loop fails ~25% on main — fixed 10-yield spin races real scheduling (has caused 3 false attributions)


ymote:
## Summary

`session_actor::tests::master_continuation_tick_reenters_actor_loop` fails **~25% of the time on unmodified `main`** under machine load. It waits for async work with a **fixed 10-yield spin**, which races real scheduling — so the failure rate is a function of CPU contention, not of any code change.

This has now cost three separate investigations, each of which initially mis-attributed it to the branch under test.

## Measured

12 consecutive runs of the test alone, on a machine at load average ~300:

| branch | failures |
|---|---|
| `feat/phase3-peers-in-chat-v2` | **3 / 12** |
| **unmodified `main`** | **3 / 12** |

Identical. The branch under test is irrelevant.

## The mechanism

`crates/octos-cli/src/session_actor_tests.rs` (~:2462-2478):

```rust
for _ in 0..10 {
    tokio::time::advance(Duration::from_millis(250)).await;
    tokio::task::yield_now().await;
}
assert!( ... persisted "child progress summary" ... );
```

`tokio::time::advance` moves **virtual** time, but the work being waited on — the actor loop, the mock provider turn, the persistence write — progresses on **real** scheduling. The test therefore grants the whole chain at most **10 cooperative yields**. Under contention that is not enough, the persisted list is empty, and the assertion fires with `: []`.

## Why it matters beyond this one test

The same shape — a fixed budget racing real work — appears elsewhere and is very likely the cause of the recurring `check-windows` reds:

- `tools::edit_file::tests::should_return_formatted_content_when_format_after_edit_enabled`
- `tools::write_file::tests::should_format_written_file_when_format_after_edit_enabled`

both fail on Windows CI with `rustfmt timed out after 5s and was killed`. Those failed on **#1995, a PR that changed one YAML file** and cannot reach `octos-agent` — same signature, same class: a fixed timeout/iteration budget on a loaded runner.

## Cost so far

Three investigations, each initially attributing it to the change under test:
1. On a goal-fix branch — dismissed as "process-global orchestrator races" (partly wrong).
2. On `feat/phase3-peers-in-chat-v2` — called a "deterministic regression" after it failed in isolation; the very next isolated run passed.
3. On a branch that changes **one constant** — which is what finally made it obvious no branch was responsible.

[...]

PR #2126, opened 2026-08-26
feat(tools): observe-only probe for the read-paging decision

ymote:
**Changes no behaviour.** Records nothing unless `OCTOS_READ_PAGING_PROBE=1`.

## Why measure instead of build

Forcing `read_file` to page is **not** a token optimisation, which is how it was first proposed (mine). If the model consumes the whole file anyway, paging is strictly worse — the same content split across more calls re-sends the conversation prefix each time.

Against our live cached prefix (11.2K tokens):

| Output | Paging penalty |
|---|---|
| 50 KB (~12.5K tok) | **+13% cost, +89% processed** |
| 200 KB (~50K tok) | **+23% cost, +237% processed** |

Our rolling cache breakpoint *does* extend the cached prefix rather than invalidating it (`anthropic.rs:458`) — but each page is written at 1.25× and re-read at 0.1× on every later call, and cached traffic still counts against the turn budget (`budget.rs:83`).

So paging is **working-set control, not a token win.** It pays off only because models often stop after page one — and nobody here has measured how often that is. This probe measures it before anyone changes what a tool returns.

## And one number that can veto the whole idea

Handing back a window where a caller expects a whole file can destroy data.

The patch tools are safe — `edit_file`, `diff_edit` and `apply_patch` each independently re-read the complete file. **`write_file` does not**, and the slides prompt instructs the model to read a script, reconstruct it whole, and write it back (`slides_default.txt:20`, `:99`). Under paging that becomes a destructive partial overwrite of a tail the model never saw.

So the probe also counts read-then-overwrite-same-path. **If that number isn't near zero, the change is vetoed regardless of the token maths.**

## What it records

| counter | meaning |
|---|---|
| `unbounded_reads` / `..._that_would_page` | the population a forced window would affect |
| `continuation_reads` | the model paging forward on its own — the behaviour forcing pages bets on |
| `partial_read_then_overwrite` | **the alarm** |
| `unpageable_long_line` | one line larger than the window; line offsets can't page past it, needs a byte cursor |

Canary window is 500 lines / 24 KiB. Bytes matter because the loop's cap counts `String::len()`, so 500 lines of minified JS still overflows it.

It deliberately records raw shape and **prices nothing**: reported spend isn't cache-write-aware today (`streaming.rs` calls plain `cost(input, output)`; the usage ledger has no cache-write field), so any cost figure derived from it would be wrong. Fixing that accounting is separate work.

## Verification

```
cargo test -p octos-agent --lib read_paging_probe   => ok. 6 passed; 0 failed
cargo fmt --all -- --check                          => clean
cargo test -p octos-agent --lib                     => 2607 passed; 2 failed
```

**Both wirings are mutation-proved.** Removing the `read_file` hook fails 3 of 6 tests; removing the `write_file` hook fails the alarm test. And one test pins that a **disarmed probe records nothing** — because a probe that silently records nothing reports the same all-clear as a clean run, which would make an armed zero meaningless.

Clippy caught a real defect in the first draft: the test helper held a `std::sync::MutexGuard` across await points. Fixed with a tokio mutex rather than silenced.

The 2 suite failures are this host's DNS resolver hijacking NXDOMAIN for `*.invalid` (`tools::ssrf`, `tools::web_fetch`). Neither file is in this diff and CI passes both.

## How to use it

```
OCTOS_READ_PAGING_PROBE=1 octos chat ...
```

Each turn logs a summary line on every exit path, including errors. The decisive ratio is `continuation_reads / unbounded_reads_that_would_page`: near zero means models stop after page one and forcing pages wins; near one means they read on and it costs more than it saves.

Refs #1638

### DIFF
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -779,6 +779,10 @@ impl Agent {
         history: &[Message],
         media: Vec<String>,
     ) -> Result<ConversationResponse> {
+        // Observe-only (#read-paging probe): report the running totals when
+        // this turn ends, on every path including errors. `None` when the
+        // probe is disarmed, which is the default.
+        let _probe_summary = crate::tools::read_paging_probe::TurnSummaryGuard::new();
         self.process_message_inner(
             user_content,
             history,
--- a/crates/octos-agent/src/tools/mod.rs
+++ b/crates/octos-agent/src/tools/mod.rs
@@ -765,6 +765,8 @@ pub trait Tool: Send + Sync {
 }
 
 // Tool registry (extracted to its own module)
+/// Observe-only probe for the read-paging decision (changes no behaviour).
+pub(crate) mod read_paging_probe;
 mod registry;
 pub use registry::ToolRegistry;
 
--- a/crates/octos-agent/src/tools/read_file.rs
+++ b/crates/octos-agent/src/tools/read_file.rs
@@ -277,6 +277,25 @@ impl Tool for ReadFileTool {
         let lines: Vec<&str> = content.lines().collect();
         let total_lines = lines.len();
 
+        // Observe-only (#read-paging probe): record what a FORCED window would
+        // have done here. Forcing pages is not a token win — if the model
+        // consumes the whole file anyway, more calls cost more, because each
+        // re-sends the conversation prefix. It wins only when models stop after
+        // page one, and that rate is the number this records. Nothing below
+        // changes; the read returns exactly what it always did.
+        if super::read_paging_probe::enabled() {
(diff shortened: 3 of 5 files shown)


anantheparty (approved):
Rerun of the failed `test-octos-cli` job is green — the failure was `contended_settle_retries_until_the_row_converges`, a pre-existing contention flake in the #2011/#2053 family, unrelated to this diff. Reviewed the probe: observe-only under `OCTOS_READ_PAGING_PROBE=1`, no behavior change, and the cost/processed-paging table justifies the measure-first stance (paging as working-set control, not a token win). Approving; mergeable flag was stale-blocked on the old red run.

Issue #2127, opened 2026-08-26
[octos] Local models are assumed to have 32K context — probe the backend's real n_ctx instead


heyong4725:
# [octos] Local models are assumed to have 32K context — probe the backend's real n_ctx instead

**Priority: P0.** This single assumption turned a task a 27B model can do into one it cannot.

## Problem

`model_catalog.json` hardcodes:

```json
{"provider": "local/local-default", "type": "strong", "default": true, ...,
 "context_window": 32768, "max_output": 8192}
```

Every local model — whatever the server actually supports — gets budgeted as 32K, and the
tiered compactor then keeps the live context near ~19K (32K − 8K output reserve − margin).
`octos-llm` contains no probing: `grep -rn "n_ctx\|/props" crates/octos-llm/src` → zero hits.

## Evidence

Same model (Qwen 3.8 27B on llama.cpp, native ~256K context), same task (convert llm.c to Rust):

- **octos**: context ledger settled at `token_estimate: 18736` after 248 compaction
  checkpoints; the model read `train_gpt2.c` 66 times in paged windows because the compactor
  kept evicting it; session produced **zero** `write_file` calls.
- **pi + pi-llama**: `pi-llama` reads `meta.n_ctx` from the server's `/v1/models`
  (`index.ts:208`: `model.meta?.n_ctx ?? previous?.contextWindow ?? DEFAULT_CONTEXT_WINDOW`).
  Peak request in the session log: **257,898 input tokens**, one compaction all session,
  conversion completed and passed the ported `test_gpt2` equivalence test.

## Proposed fix

1. On session start (or first request) for a local/OpenAI-compatible route, probe the backend:
   - llama.cpp: `GET /props` (`default_generation_settings.n_ctx`) and/or `GET /v1/models`
     (`meta.n_ctx`).
   - Other OpenAI-compatible servers (vLLM, LM Studio): `/v1/models` `max_model_len` /
     `context_length` where present.
2. Override the catalog `context_window` with the probed value; keep the catalog as fallback
   when the probe fails or returns nothing.
3. Add a per-profile override (`config.llm.primary.context_window`) so operators can pin it.
4. Log the effective window at session start so a mis-sized budget is visible in one line.

## Acceptance

- With llama.cpp launched at `-c 262144`, a coding session's context budget reflects ~256K.
- Re-run of the llm.c conversion benchmark: compaction generation count drops from ~279 to
  single digits; the model holds `train_gpt2.c` (1,182 lines) in context across turns.

[...]

PR #2135, opened 2026-08-26
feat(llm): probe local servers for their real context window

heyong4725:
Part of #2127 — implements the probe; the per-profile `context_window` override is split to #2142 (the provider construction fans out across profile_factory/chat/fallback paths and needs its own factory tests). #2127 stays open until #2142 ships.

## Problem

`model_catalog.json` hardcodes `context_window: 32768` for `local/local-default`, and the compaction budget runs off that guess no matter what the server was launched with. Observed cost (llm.c C→Rust conversion, Qwen 3.8 27B on a `-c 262144` llama-server): the ledger pinned live context at ~19K of an actual 256K, 248 compaction checkpoints, the same 1,182-line file read 66 times, zero output files written.

## Change

`LocalContextProbe` wraps the `local`, `ollama`, and `vllm` family providers and asks the running server for its real window: llama.cpp `GET /props` (`default_generation_settings.n_ctx` — the launched `-c`, which caps everything) first, then `GET /v1/models` (all known spellings: `meta.n_ctx`, `max_model_len`, `max_context_length`, …), selected for the configured model id on multi-model servers.

- Probes in the background at construction (resumed sessions get the corrected window before their first compaction pass); the request path is the fallback and retry driver.
- Definitive answers pin; transient failures (server loading, 5xx, timeout) retry up to 5 attempts — never pinned as "32K forever".
- Probe GETs authenticate with the provider's API key and honor the configured HTTP timeouts.
- `context_window()` stays sync and non-blocking: catalog value before the probe lands, server truth after.
- Full trait passthrough (metadata/metrics), mirroring `ThrottledProvider`.

Second commit addresses all 10 findings from the pre-landing review (auth, transient-pinning, multi-model selection, metadata passthrough, resume-time probing, concurrent GETs, configured timeouts, single-suffix `/v1` stripping, ollama/vllm coverage).

## Verification

- 533/533 `octos-llm` tests pass (18 new), clippy clean.
- Parsers verified against a live llama-server (Qwen3.8-27B): `/props` and `/v1/models` both report `262144`.

## Known follow-up

`router::register_with_full_meta` snapshots `context_window()` at registration time, so the LLM-visible model catalog can still render the pre-probe value even after live readers (compaction, streaming) see the corrected one. Tracked in #2127.

### DIFF
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -2215,6 +2215,10 @@ impl Agent {
                 // benefit from the same cheap shrinkage before their LLM call.
                 let protected_ids = collect_protected_tool_call_ids(&messages);
                 self.run_tier1_compaction(&mut messages, &protected_ids, tier1_pass(iteration));
+                // #2135 re-review P1: task mode (delegated workers, MCP task
+                // runs) sizes its prompt here too — resolve a lazily-probed
+                // window before the trim reads context_window().
+                self.llm.ensure_ready().await;
                 prepare_task_messages(self, &mut messages, &mut turn);
                 self.prepare_prompt_with_context_manager(
                     &mut messages,
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -1205,6 +1205,13 @@ impl Agent {
                     // tier 3 considers whether to summarise).
                     let protected_ids = collect_protected_tool_call_ids(&messages);
                     self.run_tier1_compaction(&mut messages, &protected_ids, tier1_pass(iteration));
+                    // #2135 review P1: a provider that learns its true
+                    // context window asynchronously (the local probe) must
+                    // resolve BEFORE the trim below reads context_window() —
+                    // otherwise a resumed long transcript is compacted
+                    // against the stale catalog value. No-op (immediate) for
+                    // every other provider and once resolved.
+                    self.llm.ensure_ready().await;
                     prepare_conversation_messages(self, &mut messages, &mut turn);
                     // Harness M6.3: post-prep compaction pass so the declarative
                     // runner sees the final shape of the conversation (after
--- a/crates/octos-cli/src/api/ui_protocol_transport.rs
+++ b/crates/octos-cli/src/api/ui_protocol_transport.rs
@@ -30520,9 +30520,10 @@ async fn run_standalone_turn(
     // #2135 review P1: resolve a lazily-probed context window BEFORE the
     // threshold compaction inside appui_context_history_for_agent reads it —
     // a resumed transcript compacted against the stale catalog value is
-    // exactly the failure the probe exists to prevent. Immediate no-op for
-    // non-probing providers and once resolved.
-    session_runtime.llm.ensure_ready().await;
+    // exactly the failure the probe exists to prevent. Uses the turn's
+    // already-selected provider (peer-lane routing preserved). Immediate
+    // no-op for non-probing providers and once resolved.
+    llm_provider.ensure_ready().await;
     let (history, context_manager, context_lifecycle_notifications) =
         appui_context_history_for_agent(
             // Root the context ledger at the session's TRANSCRIPT root, not the
--- a/crates/octos-cli/src/api/ui_protocol_transport.rs
+++ b/crates/octos-cli/src/api/ui_protocol_transport.rs
@@ -30517,6 +30517,12 @@ async fn run_standalone_turn(
         let session = sessions.get_or_create(&session_id).await;
         session.get_history(50).to_vec()
     };
+    // #2135 review P1: resolve a lazily-probed context window BEFORE the
+    // threshold compaction inside appui_context_history_for_agent reads it —
+    // a resumed transcript compacted against the stale catalog value is
+    // exactly the failure the probe exists to prevent. Immediate no-op for
+    // non-probing providers and once resolved.
+    session_runtime.llm.ensure_ready().await;
     let (history, context_manager, context_lifecycle_notifications) =
         appui_context_history_for_agent(
             // Root the context ledger at the session's TRANSCRIPT root, not the
--- a/crates/octos-cli/src/session_actor.rs
+++ b/crates/octos-cli/src/session_actor.rs
@@ -7473,6 +7473,11 @@ impl SessionActor {
         // M16-D2: capture ContextManager-derived prompt history before
         // persisting this turn's user message, because the agent appends the
         // current user message internally.
+        // #2135 round-3 P1: resolve a lazily-probed context window before
+        // the ContextManager threshold below reads it — a resumed gateway
+        // session must not PERSIST a compaction sized by the stale catalog
+        // value. Immediate no-op for non-probing providers, once resolved.
+        self.agent.llm_provider().ensure_ready().await;
         let history_for_agent: Vec<Message> =
             self.context_history_for_agent("pre_turn_speculative");
 
@@ -9297,6 +9302,9 @@ impl SessionActor {
         // ContextManager. If the active context is over threshold this installs
         // a compacted generation before the model call; raw session history
         // remains durable and unchanged.
+        // #2135 round-3 P1: same readiness requirement as the speculative
+        // path above — the threshold derives from context_window().
+        self.agent.llm_provider().ensure_ready().await;
         let history: Vec<Message> = self.context_history_for_agent("pre_turn");
 
         // Token tracker for status indicator
--- a/crates/octos-llm/src/adaptive.rs
+++ b/crates/octos-llm/src/adaptive.rs
@@ -1704,22 +1704,6 @@ impl AdaptiveRouter {
         }
     }
 
-    /// The slot the router would pick with NO stochastic exploration: the
-    /// best-scored non-circuit-open slot, falling back to slot 0. Used by
-    /// the identity/readiness accessors (#2135 round-6 P1) so metadata and
-    /// preloading are stable across calls, while `select_provider` keeps
-    /// its exploration for actual request routing.
-    fn preferred_slot_index(&self) -> usize {
-        self.slots
-            .iter()
-            .enumerate()
-            .filter(|(_, s)| !s.metrics.is_circuit_open(self.config.failure_threshold))
-            .map(|(i, s)| (i, self.score(s)))
-            .max_by(|a, b| a.1.partial_cmp(&b.1).unwrap_or(std::cmp::Ordering::Equal))
-            .map(|(i, _)| i)
-            .unwrap_or(0)
-    }
-
     /// Score a provider. Lower is better.
     ///
     /// Four factors:
@@ -1882,7 +1866,12 @@ impl AdaptiveRouter {
         Some(matched)
     }
 
-    /// Select provider index and whether this is a probe request.
+    /// The DETERMINISTIC selection — everything `select_provider` does
+    /// except the stochastic probe redirect (#2135 round-7 P1: the
+    /// identity/readiness accessors need the same mode/lane rules and the
+    /// same ASCENDING score order as real routing, not a parallel
+    /// reimplementation; an earlier cut used max_by against a
+    /// lower-is-better score and preferred the worst lane).
     ///
     /// - Off / Hedge: priority order, skip circuit-broken only.
     ///   (Hedge mode uses this to pick the primary for racing.)
@@ -1893,7 +1882,7 @@ impl AdaptiveRouter {
     /// in the lane's candidate list. When the lane filter yields
     /// zero matches we fall through to the full slot list so the
     /// router never starves (see [`Self::lane_filtered_slot_indices`]).
-    fn select_provider(&self) -> (usize, bool) {
+    fn select_provider_deterministic(&self) -> usize {
         let mode = self.mode();
         // RFC-3: lane-filtered eligible set, if any. None ⇒ no
         // filter; behave identically to pre-RFC-3.
@@ -1923,7 +1912,7 @@ impl AdaptiveRouter {
                                 "provider failover (lane filter, lane changing disabled)"
                             );
                         }
-                        return (i, false);
+                        return i;
                     }
                 }
                 // All lane candidates circuit-broken → fall through
@@ -1943,7 +1932,7 @@ impl AdaptiveRouter {
                             "provider failover (circuit breaker, lane changing disabled)"
                         );
                     }
-                    return (i, false);
+                    return i;
                 }
             }
             // All circuit-broken — fall through to least-failed logic below
@@ -1996,19 +1985,42 @@ impl AdaptiveRouter {
                 "all providers circuit-broken, using least-failed"
             );
             self.last_selected.store(best as u32, Ordering::Relaxed);
-            return (best, false);
+            return best;
         }
 
         scored.sort_by(|a, b| a.1.partial_cmp(&b.1).unwrap_or(std::cmp::Ordering::Equal));
         let best_idx = scored[0].0;
 
+        // Detect lane change. (This now runs before any probe redirect in
+        // `select_provider` — a probe is a measurement detour, not a lane
+        // change, so logging the deterministic selection is the accurate
+        // record.)
+        let prev = self.last_selected.swap(best_idx as u32, Ordering::Relaxed);
+        if prev != best_idx as u32 && prev < self.slots.len() as u32 {
+            info!(
+                from = self.slots[prev as usize].provider.provider_name(),
+                to = self.slots[best_idx].provider.provider_name(),
+                from_score = format!("{:.3}", self.score(&self.slots[prev as usize])),
+                to_score = format!("{:.3}", self.score(&self.slots[best_idx])),
+                "adaptive lane change"
+            );
+        }
+
+        best_idx
+    }
+
+    /// Select provider index and whether this is a probe request: the
+    /// deterministic selection above, plus the stochastic stale-provider
+    /// probe redirect used for actual request routing only.
+    fn select_provider(&self) -> (usize, bool) {
+        let best_idx = self.select_provider_deterministic();
         // Probe: with some probability, redirect to a stale non-primary provider.
         // RFC-3 (#1292) — codex P2: when a lane filter is active,
         // restrict probe targets to the lane's eligible slots so a
         // probe under `slides:*`/`code:*` can never route the user
         // turn to an out-of-lane model.
         if self.slots.len() > 1 && self.should_probe() {
-            // Find a stale provider that isn't the best
+            let lane_eligible = self.lane_filtered_slot_indices();
             for (i, slot) in self.slots.iter().enumerate() {
                 if i != best_idx
                     && slot.metrics.is_stale(self.config.probe_interval_secs)
@@ -2027,19 +2039,6 @@ impl AdaptiveRouter {
                 }
             }
         }
-
-        // Detect lane change
-        let prev = self.last_selected.swap(best_idx as u32, Ordering::Relaxed);
-        if prev != best_idx as u32 && prev < self.slots.len() as u32 {
-            info!(
-                from = self.slots[prev as usize].provider.provider_name(),
-                to = self.slots[best_idx].provider.provider_name(),
-                from_score = format!("{:.3}", self.score(&self.slots[prev as usize])),
-                to_score = format!("{:.3}", self.score(&self.slots[best_idx])),
-                "adaptive lane change"
-            );
-        }
-
         (best_idx, false)
     }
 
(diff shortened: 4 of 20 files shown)


heyong4725:
## Review result: changes needed

Reviewed head `5c07d0600451bd8adf443e04106059a7c7b40694`.

1. **P1 — Normal sessions discard the probed window.** `LocalContextProbe::context_window()` is immediately hidden behind `RetryProvider` in the standard runtime path. `RetryProvider`, `ProviderChain`, and `AdaptiveRouter` do not delegate `context_window()`, so the trait default re-reads the static catalog. A focused regression with a probed value of `262144` returned `128000`; with the runtime catalog seeded, `local-default` remains `32768`. Add delegation through every wrapper and an end-to-end factory test.

   - Probe override: https://github.com/octos-org/octos/blob/5c07d0600451bd8adf443e04106059a7c7b40694/crates/octos-llm/src/local_context_probe.rs#L273-L277
   - Standard wrapper path: https://github.com/octos-org/octos/blob/5c07d0600451bd8adf443e04106059a7c7b40694/crates/octos-cli/src/qos_catalog.rs#L234-L240

2. **P1 — Resumed sessions can compact before probing finishes.** Construction only spawns a background task, while request-path retries happen inside `chat`/`chat_stream`. Compaction reads the synchronous context window before that call. If the background request is still running or initially fails while the server loads, the resumed transcript is compacted using 32K before the retry can correct it. Await probe readiness before prompt preparation, or add an explicit async initialization step and test it with a delayed server plus a resumed transcript.

   - Background spawn: https://github.com/octos-org/octos/blob/5c07d0600451bd8adf443e04106059a7c7b40694/crates/octos-llm/src/local_context_probe.rs#L130-L133
   - Pre-request compaction read: https://github.com/octos-org/octos/blob/5c07d0600451bd8adf443e04106059a7c7b40694/crates/octos-agent/src/agent/compaction.rs#L24-L25

3. **P2 — Ollama support is effectively a no-op.** The Ollama registry uses the generic `/props` plus `/v1/models` probe, but Ollama has no `/props`, and its OpenAI-compatible model listing does not expose context length. Both endpoints therefore resolve to “no window” and permanently retain the catalog value. Ollama exposes allocated context through `GET /api/ps`; its native `POST /api/show` also exposes model metadata.

   - Wiring: https://github.com/octos-org/octos/blob/5c07d0600451bd8adf443e04106059a7c7b40694/crates/octos-llm/src/registry/ollama.rs#L48-L57
   - Ollama running-model API: https://docs.ollama.com/api/ps

4. **P2 — Trained maxima can be pinned as runtime capacity.** The parser accepts `max_context_length` and `meta.n_ctx_train`. If `/props` is unavailable or transient, a model trained for 256K but launched at 32K can be permanently recorded as 256K, causing context-overflow failures. Only runtime/allocated fields should override the catalog; trained maxima need separate provenance.

   - Parser candidates: https://github.com/octos-org/octos/blob/5c07d0600451bd8adf443e04106059a7c7b40694/crates/octos-llm/src/local_discovery.rs#L93-L106

The linked issue's per-profile override and effective-window startup log also remain unimplemented.

Verification: `cargo test -p octos-llm` succeeded with 533 unit tests passing, and focused clippy completed cleanly. GitHub's non-skipped CI checks are successful. The live 262K llama-server benchmark was not rerun.

heyong4725:
All four findings addressed in the new head — thanks, the wrapper one in particular was the difference between this feature working and demoing.

1. **Wrappers (P1)** — `RetryProvider`, `ProviderChain`, and `AdaptiveRouter` now delegate `context_window()` / `max_output_tokens()` using the same slot selection as `model_id()`. Regression test `should_keep_probed_window_through_retry_wrapper` pins your repro (probed 262144 surviving the retry wrapper; previously 128000 via the trait default).
2. **Resume-before-probe (P1)** — new `LlmProvider::ensure_ready()` (default no-op; delegated by every wrapper including Swappable/Throttled/Fallback/ContextWindowOverride). The probe implements it as one bounded attempt; awaited in the agent loop before `trim_to_context_window` and in the AppUI turn path before `appui_context_history_for_agent`, so a resumed transcript compacts against the server's real window even when the construction-time background probe lost the race or the server was still loading. Test drives it through the retry wrapper against an unreachable server (bounded, non-pinning).
3. **Ollama (P2)** — the ollama family now probes native `GET /api/ps` (allocated `context_length` of the running models, `:latest`-suffix-insensitive matching). Empty list / older Ollama without the field stays unresolved so the probe retries once the model loads. Parser tests cover match, sole-model, missing-field, and garbage bodies.
4. **Provenance (P2)** — trained maxima (`meta.n_ctx_train`, `max_context_length`, bare `context_length`) removed from the candidate set; only runtime/allocated spellings remain (`meta.n_ctx`, `loaded_context_length`, `max_model_len`). Tests updated to assert the trained-only shapes now yield the catalog fallback.

Also: the no-window resolution now logs at `info` with the standing catalog value (the effective-window startup log). **Deliberately not in this PR**: the per-profile `context_window` override — it needs `CreateParams`/config plumbing across every construction site, so I'd rather it be its own change; left open on #2127.

Verification: 536 octos-llm + 2,911 octos-agent + 1,551 octos-cli lib tests pass, clippy clean across the three touched crates. The live 262K llama-server parse checks still hold (`/props` and `/v1/models` both report 262144 on this machine); the end-to-end benchmark rerun is still pending the stack merge.

heyong4725:
## Re-review result: changes still needed

Reviewed head `7cca76e1e90d4126a9666ec522fb5d1237c34b2b`.

1. **P1 — The PR does not compile.** [`ui_protocol_transport.rs:30525`](https://github.com/octos-org/octos/blob/7cca76e1e90d4126a9666ec522fb5d1237c34b2b/crates/octos-cli/src/api/ui_protocol_transport.rs#L30525) accesses nonexistent `session_runtime.llm`. The CI clippy command reproduces `E0609`. Use the already-selected `llm_provider` here, which also preserves peer-lane routing instead of forcing the profile primary.

2. **P1 — `ensure_ready()` can return while the background probe is still running.** Construction starts the probe at [`local_context_probe.rs:192-195`](https://github.com/octos-org/octos/blob/7cca76e1e90d4126a9666ec522fb5d1237c34b2b/crates/octos-llm/src/local_context_probe.rs#L192-L195), but `probe_if_unresolved()` returns immediately when `try_lock()` fails at [lines 199-208](https://github.com/octos-org/octos/blob/7cca76e1e90d4126a9666ec522fb5d1237c34b2b/crates/octos-llm/src/local_context_probe.rs#L199-L208). Since [`ensure_ready()`](https://github.com/octos-org/octos/blob/7cca76e1e90d4126a9666ec522fb5d1237c34b2b/crates/octos-llm/src/local_context_probe.rs#L356-L362) merely calls that method, a delayed background request makes readiness return before the window is known and compaction still reads the catalog value. Give readiness a path that awaits an active probe completion, while keeping request-path probing non-blocking, and add a delayed-server regression.

3. **P1 — Task mode still compacts before readiness.** [The task loop](https://github.com/octos-org/octos/blob/7cca76e1e90d4126a9666ec522fb5d1237c34b2b/crates/octos-agent/src/agent/loop_runner.rs#L2214-L2219) calls `prepare_task_messages()` without `ensure_ready()`, and [that helper immediately trims against `context_window()`](https://github.com/octos-org/octos/blob/7cca76e1e90d4126a9666ec522fb5d1237c34b2b/crates/octos-agent/src/agent/loop_compaction.rs#L71-L78). Delegated workers and MCP task runs therefore retain the original stale-window behavior. Await readiness immediately before task message preparation too.

4. **P2 — Cold Ollama models still use the catalog window on their first turn.** [The `/api/ps` branch](https://github.com/octos-org/octos/blob/7cca76e1e90d4126a9666ec522fb5d1237c34b2b/crates/octos-llm/src/local_context_probe.rs#L252-L264) leaves an empty running-model list unresolved. On a cold Ollama process, `ensure_ready()` therefore learns nothing; compaction happens with the catalog value, and only the later chat loads the model. Retrying helps subsequent turns, not the first one. Preload the model or obtain the configured runtime window before sizing the first prompt.

5. **P2 — The PR still closes an issue whose requested override is omitted.** #2127 explicitly requires `config.llm.primary.context_window`. Either implement that requirement here, or move it to a separate tracked issue and stop closing #2127 with this PR.

Prior wrapper propagation and trained-maximum findings are resolved.

Verification:
- `cargo test -p octos-llm`: 536 unit tests passed; integration suites passed.
- `cargo fmt --all -- --check` and `git diff --check`: clean.
- GitHub `check-matrix`: failed from the reproduced CLI compile error.
- The live 262K benchmark remains unverified.

heyong4725:
All five addressed in `40b6b0c3`:

1. **Compile (P1)** — fixed: the readiness call now uses the turn's already-selected `llm_provider` (peer-lane routing preserved). The verification failure has a root cause worth naming: the `api` module is feature-gated off by default, so my `cargo check -p octos-cli` never compiled `ui_protocol_transport.rs` at all. Verification for this and future changes now runs `cargo clippy -p octos-cli --features api --all-targets` (clean).
2. **Readiness race (P1)** — `ensure_ready()` now AWAITS an in-flight probe (`lock().await`, not `try_lock`) and runs one attempt itself if still unresolved; the request path keeps its non-blocking `try_lock`. Delayed-server regression added: the test holds the in-flight guard, asserts `ensure_ready` blocks, resolves the window, and observes it on release.
3. **Task mode (P1)** — readiness is awaited immediately before `prepare_task_messages` too, so delegated workers and MCP task runs size prompts against the probed window.
4. **Cold Ollama (P2)** — empty `/api/ps` now falls back to `POST /api/show` and pins the Modelfile `num_ctx`, which is runtime *configuration* (what the model will be loaded with) and therefore safe to pin cold — the trained `context_length` under `model_info` stays ignored, keeping the provenance rule intact. Residual, stated honestly: a cold model with *no* Modelfile `num_ctx` still takes the catalog for turn one and corrects on the next probe once loaded; full preload is beyond a probe's remit.
5. **Scope (P2)** — the PR body no longer closes #2127; the override is tracked as #2142 with an implementation sketch (wrap in the existing `ContextWindowOverride` at construction, outermost so the operator wins — which composes now that the wrappers delegate).

Verification: 538 octos-llm + 2,911 octos-agent tests pass; clippy clean on octos-llm, octos-agent (`--all-targets`), and octos-cli (`--features api --all-targets` — the configuration that previously caught nothing locally).

heyong4725:
All five round-3 findings addressed in the new head:

1. **Gateway readiness (P1)** — both session-actor history paths (`pre_turn` and `pre_turn_speculative`) now await `ensure_ready()` before `context_history_for_agent`, so the ContextManager threshold (derived from `context_window()`) is computed against the probed window and a resumed gateway session can no longer persist a stale-sized compaction.
2. **Pipeline truncation (P1)** — the cut now lives behind `sized_pipeline_instruction()`, which awaits readiness before reading the window. Delayed-probe regression added: a provider resolving 1,024 → 1M must not lose the middle of a long instruction; the test first proves un-ready sizing *would* truncate (discriminating), then asserts the seam preserves the input byte-for-byte. Removing the `ensure_ready` inside the seam fails it.
3. **Cold Ollama (P2)** — when `/api/show` has no Modelfile `num_ctx`, the probe now issues Ollama's official empty-request preload (`POST /api/generate`, per the FAQ you linked) and re-reads `/api/ps` for the real load-time allocation — which is the number turn one must respect (`OLLAMA_CONTEXT_LENGTH`/default 4K can be far from catalog maxima in either direction). A model too large to load within the probe timeout keeps loading server-side; the next attempt reads the allocation, and the first chat was going to trigger that load anyway.
4. **Readiness bound (P2)** — on lock contention `await_readiness` now awaits the active attempt's completion and returns, never stacking a second attempt; only when no attempt is active does it run exactly one. The one-attempt doc bound is now true (Ollama's sequential requests remain within a single attempt's budget, documented on the endpoint enum).
5. **Transient statuses (P2)** — `status_is_transient()` treats 408/425/429 (and 5xx) as transient in both fetchers; 401/403/404 stay decisive. Unit test covers the exact `/props 404 + /models 429` pinning scenario from the finding.

Verification: 539 octos-llm, 354+ octos-pipeline (incl. the new delayed-probe regression), 1,551 octos-cli lib tests pass; clippy clean on octos-llm/octos-pipeline/octos-agent (`--all-targets`) and octos-cli (`--features api --all-targets`).

heyong4725:
All three round-4 findings addressed in the new head:

1. **Eager preload (P1)** — probing is now split into passive and active modes: background/construction and request-path attempts never touch `/api/generate`; only `ensure_ready()` — a route about to be used — may preload. Your startup scenario (every configured Ollama primary/fallback/sub-provider lane loading its model into memory) is pinned by an HTTP regression: a stub server records paths, and construction + the request path must produce zero `/api/generate` hits.
2. **Slow loads (P1)** — the conservative provisional floor (4096, Ollama's stock allocation — conservative in the only direction that cannot overflow the server) is put in force the moment a cold model with no Modelfile `num_ctx` is detected, *before* any preload, so an over-deadline load can cancel the attempt without reverting to catalog maxima. Readiness drives the preload and polls `/api/ps`; the real allocation replaces the floor whenever any attempt reads it. Two delayed-preload HTTP regressions: a 300ms load is adopted within the budget (`/api/generate` hit exactly once, floor cleared); a 60s load leaves readiness bounded with the floor in force and the window unresolved for retry.
3. **Attempt bound (P2)** — one overall deadline per attempt (2× request timeout, clamped 4–20s) wraps the entire attempt body, so the sequential Ollama chain can no longer multiply per-request timeouts; the bounded-readiness regression asserts wall-clock. Fixing this also surfaced a latent bug in round 3's poll loop (re-polling a completed, non-fused future) — now guarded.

Verification: 560 octos-llm tests (12 probe-focused, including the three new HTTP regressions against a real TCP stub), 459 octos-pipeline, clippy clean on octos-llm (`--all-targets`) and octos-cli (`--features api --all-targets`). Still no live cold-Ollama load test on this machine (no Ollama installed here) — the stub regressions model the FAQ-documented behavior (empty generate loads; `/api/ps` reports allocation once running).

heyong4725:
All four round-5 findings addressed in `7272c056`:

1. **Passive attempt satisfying readiness (P1)** — after awaiting an in-flight attempt that left the window unresolved, `ensure_ready()` now runs one ACTIVE attempt of its own (holding the just-acquired guard), so the constructor's passive probe can no longer stand in for the preload. Readiness is documented as bounded by at most two attempt budgets. Regression: a held-then-released guard with the window unresolved must end with `/api/generate` hit and the allocation adopted.
2. **Cancelled preload (P1)** — the `/api/generate` request is now a DETACHED task with its own long-timeout client (900s), spawned at most once per probe lifetime; the attempt deadline cancels only the `/api/ps` polling, never the load's request context. You're right that the round-4 stub masked this by loading after disconnect — the new regression instead verifies the semantics end-to-end: a 6s load outlives the 4s deadline, is not cancelled, and its allocation is adopted afterwards.
3. **Floor value (P1)** — dropped to 2048 (Ollama's documented practical minimum), with explicitly-configured sub-2048 `num_ctx` documented as out of the floor's scope — as you note, no non-trivial constant is safe against `num_ctx 4`, and 2048 misjudges only configs that are already unusable for coding sessions. The floor is replaced by the real allocation the moment any attempt reads it.
4. **Frozen cap (P2)** — with a floor in force, the retry cap no longer pins `None`; post-cap attempts short-circuit to ONE cheap `/api/ps` read each (the floor skips show/preload), so the real allocation is adopted whenever the model comes up — including when the actual first chat is what loaded it. Regression burns past the cap mid-load and asserts adoption afterwards.

Verification: 562 octos-llm tests (14 probe-focused; the two new TCP-stub regressions exercise passive/active modes, deadline survival, cap behavior, and floor clearing), clippy clean.

heyong4725:
All four round-6 findings addressed in `e971dd6d`:

1. **Selection instability (P1)** — split by concern rather than fully turn-pinning: the **sizing** accessors are now selection-*independent* (see 2), and the **identity/readiness** accessors use a new `preferred_slot_index()` — best-scored non-circuit-open slot, no stochastic exploration — so readiness preloads a deterministic lane and metadata is stable across calls (regression asserts a stable `model_id` over 50 calls at `probe_probability: 0.5`). `select_provider()` keeps its exploration for actual request routing. Full per-turn selection pinning (one choice carried through readiness → sizing → chat) is a router-level contract change; if you want it tracked, I'll file it as its own issue — this PR now guarantees safety without it.
2. **Oversized fallback prompts (P1)** — `FallbackProvider`, `ProviderChain`, and `AdaptiveRouter` sizing accessors now take the **minimum across every slot** the wrapper can route to. That restores the accidental safety of the stale catalog value without its staleness: a probed 256K primary over a 32K fallback budgets as 32K. Per-route resizing before each fallback send is the precise fix and belongs to the routers; noted in the code. Regressions in all three wrappers.
3. **Latched preload (P1)** — replaced the boolean with an idle/running/answered state machine: a transient `/api/generate` failure resets to idle (next readiness attempt retries), and the poll loop runs only while a load is actually in flight — a failed preload stops polling immediately instead of burning the attempt deadline. Regression: 503-then-success, asserting early return, exactly two generate hits, and adoption on the second attempt.
4. **Floor value (P2)** — 1024, now aligned with the probe's own `sane_context_window` lower bound, covering the `OLLAMA_CONTEXT_LENGTH=1024` case from envconfig; sub-1024 windows are below the probe's plausibility threshold everywhere, so the system is self-consistent about what it considers a usable window.

Verification: 548 octos-llm lib tests (probe, fallback, chain, adaptive regressions included) + 1,551 octos-cli + 2,911 octos-agent, clippy clean on octos-llm (`--all-targets`) and octos-cli (`--features api --all-targets`).

heyong4725:
All three round-7 findings addressed in `4b15a165` (+ a clippy placement follow-up):

1. **Inverted selection (P1)** — you're right, and it's the kind of wrong that tests-for-stability can't see: `max_by` against a lower-is-better score selected the *worst* lane, and the parallel reimplementation ignored mode/lane rules. Fixed the way you suggested: `select_provider` is factored into its deterministic core (identical mode/lane handling, ascending sort, including the non-Lane priority walk) plus the stochastic probe redirect; identity/readiness accessors call the core. One deliberate semantic note: the lane-change log/`last_selected` swap moved into the core, so it now records the deterministic selection even when a probe redirects — a probe is a measurement detour, not a lane change. The regression now asserts the *primary* is selected for identical cold slots, not merely a stable choice.
2. **Unresolved fallback sizing (P1)** — min-across-routes only bounds with what was resolved at prompt-build time; agreed. Added `route_fits_request()`: at dispatch, every alternate route gets its own `ensure_ready()` and is skipped when the prompt can't plausibly fit its just-resolved window (skipping beats a server-side truncation/rejection). Wired in `FallbackProvider` (chat + stream) and `ProviderChain` (both loops). Regression uses a mock that reports 131K until `ensure_ready` reveals 1K: it receives zero calls and the next fallback serves the request.
3. **PRELOAD_ANSWERED trap (P1)** — two terminal outcomes added: a definitive 4xx on `/api/generate` parks at `PRELOAD_FAILED` (no retry, no polling); and a *completed* load whose `/api/ps` view still names no usable window (old Ollama, unmatched entry) pins "no window" — the conservative floor becomes the permanent answer and readiness returns immediately thereafter. Regressions: old-Ollama pin (second readiness < 500ms, exactly one generate) and 404-preload (no deadline burned, no retry).

Verification: 551 octos-llm lib tests + integration suites, 1,551 octos-cli lib tests, clippy clean on octos-llm (`--all-targets`) and octos-cli (`--features api --all-targets`).

heyong4725:
All three round-8 findings addressed in the new head:

1. **Adaptive coverage (P1)** — the guard moved to the one funnel every adaptive dispatch passes through: `try_chat`/`try_chat_stream`. Primary, stochastic probe, hedge racers, and failover candidates are all covered; the guard runs *before* the metrics block so an unfit skip can't open the slot's circuit breaker. The regression drives four oversized prompts (past the default breaker threshold) — the 1K lane receives zero calls and the fitting lane serves — then proves the breaker stayed closed by having a small prompt served *by* the small lane.
2. **Tools-aware fit (P1)** — `route_fits_request` now sums messages plus each tool's serialized name/description/schema (`estimate_tool_tokens`, mirroring what providers actually serialize per your openai.rs reference). Regression: tiny message passes a 2K window; the same message plus one fat schema fails it.
3. **Lane deadline (P2)** — the chain's fit guard now runs inside `with_lane_timeout`, so readiness shares the lane's configured deadline; `Ok(None)` keeps unfit-skip distinct from lane failure (breaker semantics unchanged). Regression: a lane with 5s readiness is abandoned at a 100ms deadline.

Worth disclosing: the adaptive guard was momentarily *absent* while its regression test existed — an earlier edit script aborted mid-way and dropped it silently — and the test caught it exactly as intended (`from-small` served an oversized prompt). The commit lands both together, verified green.

Verification: 554 octos-llm lib tests + integration suites, 1,551 octos-cli lib tests, clippy clean (`--all-targets`; octos-cli with `--features api`).

Issue #1976, opened 2026-08-11
goal: per-path write grants for peers/workers — fs: {write: [paths], create_only} enforced kernel-side, not by model cooperation


ymote:
From the 2026-08-11 peer-goal review + the octos-one (mobile self-app-development) feedback. The one genuinely absent capability on the unattended-refinement critical path: a WRITE FENCE. Today `goal_plan`'s per-task fs grant is binary — goal_tool.rs says it plainly: "fs grant v1 is binary: narrow per-path grants are not yet supported." A goal-driven refiner that may write `exemplar.card` but never `app.md` cannot be expressed, only requested politely of the model. A budgeted autonomous loop is exactly the thing that will find that hole.

## Grant shape (extends the existing TaskSpec grant, parse_grant seam in goal_tool.rs)
```json
{ "fs": { "write": ["exemplar.card", "cards/*.card"], "create_only": true }, "network": false }
```
- `write`: workspace-relative path allowlist (globs); everything else read-only. Deny-wins with the existing tool policy.
- `create_only`: allowlisted paths may be CREATED but never overwritten/edited/deleted (O_CREAT|O_EXCL semantics; edit_file/delete refused even on allowlisted paths).

## Enforcement points (kernel-side, in order of strength)
1. File tools (write_file/edit_file): extend the workspace-root scope check with the grant allowlist + create_only, symlink-safe (O_NOFOLLOW conventions already in place).
2. Shell/sandbox: generate the sandbox profile from the grant — bwrap ro-binds everything except allowlisted paths; macOS SBPL file-write* allowlist; docker ro mounts + targeted rw. This is what makes it a fence rather than a suggestion: `shell` writes outside the grant fail at the OS.
3. Peers inherit via the existing WorkerGrant flow (#1875) at peer_handoff/goal_dispatch staging; grant recorded on the ledger task row for audit.

## Non-goals v1
- No per-path READ restrictions (workspace stays readable).
- No grant widening mid-task except via the existing goal_grant escalation flow (which already re-dispatches).

## Acceptance
- A peer with `write: ["exemplar.card"], create_only: true`: CAN create exemplar.card via write_file AND via shell redirect; CANNOT edit it after creation, CANNOT write app.md via any tool or shell, CANNOT bypass via symlink/rename/`mv`.
- Grant violations surface as tool errors + a ledger finding (`[denied]` class), never silent.
- Existing binary grants keep working unchanged (back-compat).

[...]

PR #2136, opened 2026-08-26
feat(sandbox): let the toolchain work inside the sandbox, and say so when it can't

### LINKED ISSUE #2128: [octos] Sandbox breaks the Rust toolchain (rustup/cargo) and denials surface as opaque EPERM
# [octos] Sandbox breaks the Rust toolchain (rustup/cargo) and denials surface as opaque EPERM

**Priority: P0.** With the default sandbox, a coding session cannot compile Rust at all, and
neither the model nor the user is told the sandbox is the reason.

## Problem

The macOS seatbelt profile (`crates/octos-agent/src/sandbox/macos.rs`) with default config
grants **global reads** (`read_allow_paths` empty → `(allow file-read*)`) but denies **writes
outside cwd**. The Rust toolchain needs out-of-cwd writes to function:

- rustup's cargo/rustc shims open `~/.rustup/settings.toml` with write/lock intent → the
  observed `error: could not read settings file: '/Users/.../.rustup/settings.toml':
  Operation not permitted` — before any compilation starts.
- cargo writes `~/.cargo/registry` (index + crate cache) and needs network to crates.io
  (`allow_network = false` by default).

## Evidence

octoscode session `octos-harness/c2rust-llm.c` (fine-tuned C2Rust model):

- `check` tool result: `cargo check: exit code 1, no parseable diagnostics. Output tail: |
  error: could not read settings file: '/Users/yonghe/.rustup/settings.toml': Operation not
  permitted (os error 1)`
- Consequence: the model wrote 15 Rust files with **zero successful compiles**; hallucinated
  deps (`cuda-rs = "0.3"`) survived; the user had to paste `cargo build` errors manually.
- The Qwen session even tried to install its own toolchain into `tmp/rustup` — a model-visible
  symptom of the same denial.

Under pi (no sandbox, host toolchain), the same Qwen model ran `cargo build` ~30 times in a
fix loop and completed the conversion.

## Proposed fix

1. **Coding sandbox profile** (opt-in default for the coding topic): extend write allowance to
   detected toolchain homes — `~/.rustup`, `~/.cargo` (resp. `$RUSTUP_HOME`, `$CARGO_HOME`),
   and equivalents for other ecosystems when detected (`~/.npm`, `~/.pyenv`, ...). Grant
   network to the matching registries (crates.io / static.crates.io for Rust) even when
   general `allow_network` stays false — or gate cargo-only network behind the profile.
2. **Loud denials**: when a sandboxed command fails and the profile denied a path the command
   touched, tag the tool result explicitly, e.g.
   `[sandbox] write to ~/.rustup denied by profile — enable the coding sandbox profile or add
   the path to write allowances`. A 27B model cannot diagnose a bare EPERM; neither could the
   user from "no parseable diagnostics".
3. Alternative considered: redirect `CARGO_HOME`/`RUSTUP_HOME` into the workspace — rejected
   as default (re-downloads toolchain + registry per workspace) but useful as a stricter
   fallback mode.

## Acceptance

- In a default coding session on macOS, `cargo build` succeeds against a host rustup install.
- A denied path in a failing command produces a `[sandbox]`-tagged explanation in the tool
  result, verified by a test that runs a command touching a denied path.

heyong4725:
Closes #2128.

## Problem

With writes confined to the workspace cwd, a macOS coding session cannot compile Rust at all: rustup's cargo shim write-locks `~/.rustup/settings.toml` on every invocation and dies on `could not read settings file: Operation not permitted` before cargo starts; cargo also needs `~/.cargo/registry`. Observed cost: a session that wrote 15 Rust files with zero successful compiles, shipping hallucinated dependencies — and the denial surfaced only as the failing program's own confused error text.

## Change

1. **`sandbox.allow_toolchains`** (default `true`): grants the shell write access to the *precise* set a toolchain needs — the rustup settings lock (literal), rustup `tmp`/`downloads`, cargo `registry`/`git`, and cargo's three lock/metadata files. Deliberately **not** granted: `~/.cargo/bin` (on PATH — a writable shim is persistence) and `~/.rustup/toolchains` (writable compiler binaries). Deny-wins: a read-only workspace or a #1976 write fence suppresses the grants entirely. macOS backend only for now; Linux backends are a follow-up on #2128.
2. **Tagged denials**: a failed command under a real sandbox whose output carries `Operation not permitted` gets a `[sandbox]` line naming the layer that denied it and the config lever that fixes it.

## Verification

2,916/2,916 `octos-agent` tests pass (4 new: SBPL emission with grants, deny-wins suppression under read-only/fence, persistence-vector exclusion + serde default, denial-hint firing matrix), clippy clean.

### DIFF
--- a/crates/octos-agent/src/sandbox/macos.rs
+++ b/crates/octos-agent/src/sandbox/macos.rs
@@ -1645,8 +1645,8 @@ mod tests {
     fn toolchain_grants_emitted_with_workspace_write() {
         let sb = MacosSandbox {
             toolchain_write_grants: super::super::ToolchainWriteGrants {
-                literals: vec!["/Users/t/.rustup/settings.toml".into()],
-                subpaths: vec!["/Users/t/.rustup/tmp".into()],
+                literals: vec!["/Users/t/.cargo/.package-cache".into()],
+                subpaths: vec![],
             },
             allow_network: false,
             read_allow_paths: Vec::new(),
@@ -1665,12 +1665,12 @@ mod tests {
             .find(|a| a.contains("deny default"))
             .expect("should have SBPL profile");
         assert!(
-            profile.contains("(allow file-write* (literal \"/Users/t/.rustup/settings.toml\"))"),
-            "rustup settings lock must be writable: {profile}"
+            profile.contains("(allow file-write* (literal \"/Users/t/.cargo/.package-cache\"))"),
+            "the cargo lock must be writable: {profile}"
         );
         assert!(
-            profile.contains("(allow file-write* (subpath \"/Users/t/.rustup/tmp\"))"),
-            "rustup scratch must be writable: {profile}"
+            !profile.contains("/registry/index"),
+            "index must NOT be writable by default: {profile}"
         );
     }
 
@@ -1732,7 +1732,7 @@ mod tests {
         let sb = MacosSandbox {
             toolchain_write_grants: super::super::ToolchainWriteGrants {
                 literals: vec!["/Users/t/.cargo/.package-cache".into()],
-                subpaths: vec!["/Users/t/.cargo/registry/index".into()],
+                subpaths: vec![],
             },
             allow_network: false,
             read_allow_paths: Vec::new(),
@@ -1758,10 +1758,6 @@ mod tests {
         assert!(
             profile.contains("(allow file-write* (literal \"/Users/t/.cargo/.package-cache\"))"),
             "lock must be writable: {profile}"
-        );
-        assert!(
-            profile.contains("(allow file-write* (subpath \"/Users/t/.cargo/registry/index\"))"),
-            "index must be writable: {profile}"
         );
         assert!(
             !cmd.as_std()
--- a/crates/octos-agent/src/sandbox/macos.rs
+++ b/crates/octos-agent/src/sandbox/macos.rs
@@ -201,8 +201,18 @@ impl Sandbox for MacosSandbox {
         // Path is validated above -- no escaping needed since \ and " are rejected.
         let cwd_escaped = &cwd_str;
 
+        let toolchains_active = !self.toolchain_write_grants.literals.is_empty()
+            || !self.toolchain_write_grants.subpaths.is_empty();
         let network_rule = if self.allow_network {
             "(allow network*)"
+        } else if toolchains_active {
+            // #2136 review, P1: a default build must be able to DOWNLOAD
+            // dependencies — with full network deny, cargo only works when
+            // every crate is already cached. Scoped egress: outbound HTTPS
+            // and HTTP (registry + git transports) plus DNS, nothing
+            // inbound, no other ports. Operators who need the full deny
+            // back set `sandbox.allow_toolchains = false`.
+            "(deny network*)\n(allow network-outbound (remote tcp \"*:443\"))\n(allow network-outbound (remote tcp \"*:80\"))\n(allow network-outbound (remote udp \"*:53\"))\n(allow system-socket)"
         } else {
             "(deny network*)"
         };
@@ -461,6 +471,17 @@ impl Sandbox for MacosSandbox {
         cmd.env("TMPDIR", &user_tmp);
         cmd.env("TEMP", &user_tmp);
         cmd.env("TMP", &user_tmp);
+        // #2136 review, P1: per-workspace Cargo overlay. The shared
+        // ~/.cargo caches hold sources/build scripts/proc macros that
+        // cargo EXECUTES in other workspaces and outside the sandbox, so
+        // they must never be sandbox-writable; redirecting CARGO_HOME into
+        // the (writable, tracking-ignored) scratch dir gives builds a
+        // working cache with workspace-local blast radius. Reads of the
+        // host cache are irrelevant once CARGO_HOME moves — cargo uses one
+        // home for both — so the price is a per-workspace re-download.
+        if toolchains_active {
+            cmd.env("CARGO_HOME", user_tmp.join("cargo-home"));
+        }
         // Clear dangerous environment variables (sandbox-exec inherits parent env)
         for var in BLOCKED_ENV_VARS {
             cmd.env_remove(var);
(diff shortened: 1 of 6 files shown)


heyong4725:
All three findings addressed in `47333af4`:

1. **Cache poisoning (P1)** — the strongest catch of this PR's review: writable shared `~/.cargo/{registry,git}` was a cross-workspace persistence vector (poisoned sources/build scripts/proc macros execute *outside* the sandbox later). Those grants are **gone** entirely — nothing under `~/.cargo` is ever granted now (regression asserts it). Sandboxed commands get `CARGO_HOME` redirected to a per-workspace overlay under the scratch dir (already writable, already tracking-ignored): builds work, blast radius is workspace-local, cost is a per-workspace dependency cache. On the "keep shared caches read-only" half: cargo has no native read-shared/write-local overlay — one `CARGO_HOME` serves both — so the trade is a re-download per workspace rather than partial reuse; noted in the code. Rustup grants stay: the settings lock is content-trivial, and staged toolchain archives are hash-verified by rustup on install.
2. **Fresh builds (P1)** — with toolchains active and general network off, the profile now allows **scoped egress**: outbound tcp 443/80 + udp 53 (DNS), nothing inbound, no other ports — so a default `cargo build` can fetch dependencies, satisfying #2128's acceptance criterion. Honest caveat: seatbelt can't filter by hostname, so this is port-scoped, not crates.io-scoped — any HTTPS egress is possible from sandboxed commands when toolchains are on. `allow_toolchains = false` restores the full deny; if you want registry-host pinning, that needs a proxy and I'd track it as a follow-up on #2128.
3. **Session paths (P1)** — `exec_command`'s yielded/TTY path and `write_stdin` now carry the `[sandbox]` hint via a shared `session_output_payload` helper (scan full capture → truncate → append, same ordering contract as the synchronous assemblers). `ExecSession` records whether it was sandboxed so `write_stdin` — which has no sandbox handle — decides correctly.

Verification: 2,918 octos-agent tests (new: no-cargo-grant invariant, scoped-egress + `CARGO_HOME` overlay present exactly when toolchains active and absent otherwise, session-payload hint matrix), clippy `--all-targets` clean.

heyong4725:
All four addressed in `c31c1f28` — and thank you for the empirical checks; the egress design was wrong in exactly the way that only a live test catches.

1. **Network authoritative (P1)** — the scoped egress is fully reverted. You're right on both counts: it overrode `allow_network=false` (and fleet grants) for every command, and seatbelt can't do the hostname/DNS filtering the design assumed. `allow_network` is the single source of truth again. The sandbox now supports **building with cached dependencies**, not downloading new ones — I've been explicit about that scope reduction rather than pretending offline-fresh-build works.
2. **Poisoning (P1, kept)** — refined instead of the overlay: only the **non-executable** coordination files are writable — `.package-cache` (the lock cargo opens write-intent every run) and `registry/index` (JSON/bincode metadata). `registry/cache` (.crate archives), `registry/src` (extracted sources), and `git` (checkouts) — the code cargo executes — stay **read-only**. Reads are already globally allowed, so a cached build reads its deps but cannot overwrite a crate. This is what lets an offline cached build work without reintroducing the vector.
3. **Repo pollution (P2)** — the `<cwd>/tmp/cargo-home` overlay is gone entirely. Host `CARGO_HOME` stands (config + credentials intact, no untracked cache trees in the repo, no per-workdir fragmentation).
4. **Session race (P2)** — `spawn_session`'s exit-code task now `await`s the stdout/stderr reader handles before recording the code, so "not running" implies "output drained". Regression runs 15 trials of a print-then-exit command racing a 5ms yield and asserts a completed session never loses its output tail.

Two honest limitations I want on the record rather than buried:
- **Fresh in-sandbox downloads are not supported** by this PR — that needs registry/cache+src writes (the poisoning vector) plus network, i.e. the proxy-isolated fetch now tracked as #2144. Operators pre-populate via a non-sandboxed `cargo fetch`, or opt into `allow_network=true`.
- **Live behavior needs a macOS host to verify** — the unit tests assert the generated SBPL profile and env, not a live sandboxed fetch/build (my environment can't run seatbelt network tests). Your live checks remain the authoritative gate for the network/DNS behavior.

If "builds with cached deps, downloads deferred to #2144" is too narrow for #2128's acceptance criterion, I'm happy to either keep this as the safe core and reframe #2128, or hold #2136 until #2144's fetch design lands — your call on scope.

Verification: 2,919 octos-agent tests (new: no-executable-path-granted invariant, network-authoritative + no-CARGO_HOME-redirect, session-race trials), clippy `--all-targets` clean.

heyong4725:
All five addressed in `ffcde205` — and your live tests were the deciding evidence in three of them, thank you.

1. **Index write removed (P1)** — you confirmed a cached build works with the index read-only, so it's gone. The **default** writable set is now exactly `~/.cargo/.package-cache` (the build lock) and nothing else; registry index/cache/src and git are all read-only.
2. **allow_network is a working hatch + #2128 reframed (P1)** — you're right that downloads failed even with network on, because registry/cache+src were denied. Those writes (plus index and git) are now granted **when `allow_network` is set** — enabling network is the operator's trust decision, and it's exactly when fetches happen. Default stays cached-build-only; I've [reframed #2128](https://github.com/octos-org/octos/issues/2128) to that scope and left #2144 for the proxy-isolated path.
3. **Rustup settings.toml removed (P1)** — also gone, along with rustup tmp/downloads. A plain build via the rustup proxy only *reads* `~/.rustup` (globally allowed); the write grant was solving an octoscode read-restriction that doesn't exist in octos. If your live env shows the rustup proxy needs a settings write for some toolchain config I haven't hit, tell me and I'll switch to the private-settings-overlay approach — but I believe reads suffice.
4. **Session race closed (P2)** — the exit-code task already joined the readers; the fix was the *sampling* order, which you correctly identified. Both paths now read the exit code first, then output, so `running:false` implies a drained buffer. The regression is non-vacuous — it counts completed trials across 40 runs and fails if none reached the completed path.
5. **Hint corrected + #2128 test (P2)** — the hint no longer claims the registry is writable; it says cached builds work and new crates need `allow_network`. And there's now a macOS end-to-end test that runs a command **denied by a live seatbelt sandbox** (`echo x > /etc/...`) and asserts the tool response carries `[sandbox]`.

Verification: 2,612 octos-agent lib tests (incl. the live macOS denied-command and non-vacuous race tests), 24 security-sandbox, clippy `--all-targets` clean. The default-vs-network grant split and the session-output ordering are unit-tested; the live network/download behavior remains yours to confirm on a macOS host, as before.

PR #2152, opened 2026-08-27
feat(session): wire the coding feedback loop that was built and never connected

### LINKED ISSUE #2129: [octos] Coding workspaces ship empty validation hooks — wire language-detected checks by default
# [octos] Coding workspaces ship empty validation hooks — wire language-detected checks by default

**Priority: P1.**

## Problem

`.octos-workspace.toml` is generated with the machinery but nothing in it:

```toml
[validation]
on_turn_end = []
on_source_change = []
on_completion = []
```

So nothing ever compiles what the model writes. In the c2rust-llm.c session the model wrote
15 Rust files and declared success without one build; errors reached it only when the user
pasted them.

Under pi the same-class model succeeded *because* it chose to run `cargo build` ~30 times —
a choice a weaker model won't reliably make. The harness should not depend on it.

## Proposed fix

1. On workspace init, detect project type (Cargo.toml → Rust, package.json → node, ...) and
   populate defaults, e.g. for Rust:
   `on_source_change = ["cargo check --message-format=short"]` (debounced),
   `on_completion = ["cargo build"]`.
2. Feed hook output back as a tool-result message the model sees on its next turn — a failing
   check is information for the model, not just a UI row.
3. Escape hatch: `octoscode` command to disable/edit hooks per workspace.

## Acceptance

A session where the model writes a Rust file with a type error receives the E-code diagnostics
in-context within one turn, with no user action.

heyong4725:
Closes #2129. Rebased onto main; the original #2138 auto-closed when its base branch was deleted on #2136's merge.

Wires the coding feedback loop octos had built but never connected: `detect_workspace_policy_kind` / `for_coding` / `coding_default_hooks` (cargo check / eslint / ruff after edits) now merge at the shared ProfileRuntime assembly point, so every host inherits them. The defaults are self-gating (path_filter + requires_bin); hook children run in the workspace via the per-turn payload cwd; `execute_hook` captures stderr with concurrent pipe reads; a failing checker's diagnostics reach the model as sanitized `HookResult::Feedback` (infra errors stay log-only, legitimate check failures don't trip the circuit breaker); eslint's brace glob (matched nothing) expanded to four patterns.

Largest behavior change of the batch — the profile hook executor is now always constructed and checker diagnostics enter model-visible tool results. All 10 findings from the pre-landing /code-review addressed on the original #2138. Verification: 60 hook tests + 22 profile tests pass, clippy + fmt clean.

### DIFF
--- a/crates/octos-agent/src/agent/execution.rs
+++ b/crates/octos-agent/src/agent/execution.rs
@@ -526,7 +526,7 @@ impl Agent {
         // default delegation), so every field above was reaching only
         // task-local (`TOOL_CTX`) readers. TOOL_CTX stays scoped for plugin
         // tools that read the task-local.
-        let result = TOOL_APPROVAL_CTX
+        let mut result = TOOL_APPROVAL_CTX
             .scope(
                 approver,
                 TOOL_CTX.scope(
@@ -549,7 +549,14 @@ impl Agent {
                 tool_start.elapsed().as_millis() as u64,
                 self.hook_ctx().as_ref(),
             );
-            let _ = hooks.run(HookEvent::AfterToolCall, &payload).await;
+            // A failing after-hook carries checker output the model must
+            // see (the coding defaults run `cargo check` after edits) —
+            // discarding it here silenced the whole feedback loop.
+            if let HookResult::Error(feedback) = hooks.run(HookEvent::AfterToolCall, &payload).await
+            {
+                result.output.push_str("\n\n[hook] ");
+                result.output.push_str(&feedback);
+            }
         }
 
         Ok(result)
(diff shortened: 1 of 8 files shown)


heyong4725:
A pre-merge `/code-review` of the rebased branch found two dead-on-arrival bugs in the feedback path (both fixed in the latest commit):

1. **`cargo check` exits 101, not 1** — its compile diagnostics were hitting the infra catch-all arm: discarded from the model *and* counted toward the circuit breaker, so the flagship Rust checker never fed back and normal iterate-on-errors disabled the hook. The after-event handler now routes *any* nonzero exit with output to `Feedback` (eslint/ruff=1, cargo=101, tsc=1/2), not counted; empty-output nonzero stays infra.
2. **`write_file` args were fully redacted** (it's a SENSITIVE_TOOL), so the path-filter found no `arguments.path` and every checker skipped on new-file creation — the same class as the original #2129 finding 1. The sanitizer now preserves the path keys and redacts only the file content.

Plus two P2s: a project checker no longer runs in the daemon's cwd when the workspace root is unknown (it's skipped), and the approval dispatch path truncates before appending `[hook]` feedback so a downstream cap can't cut it.

Regressions added for all three critical paths. Two design-level findings (shared profile-level breaker; synchronous per-edit cargo check latency) are deferred to #2153. Verification: 2,621 octos-agent + 136 runtime tests, clippy + fmt clean.

Issue #2172, opened 2026-08-29
Forced temperature=0.0 (greedy) with no override causes repetition collapse on local models; no sampler knob


heyong4725:
## Problem

octos hardcodes **`temperature = 0.0` (greedy decoding)** for every chat request and gives operators **no way to change it**. On a small local model this deterministically triggers a **repetition collapse**: once the model emits one tool call, greedy argmax re-emits the identical call until `max_tokens`. Live-tested with Qwen 3.8 (llama.cpp) on the llm.c C→Rust task, the model emitted **763 identical `update_plan` calls** in one response, hit `max_tokens=16384` (`stop_reason=MaxTokens`), and produced nothing.

## Evidence — same model, same task, only temperature changed

| | forced `temperature=0.0` (current) | `temperature` omitted (patched) |
|---|---|---|
| `update_plan` calls in the write turn | **763** (collapse) | **1** (normal) |
| that turn's stop_reason | `MaxTokens` | `ToolUse` |
| how far the run got | died at **iteration 7** | reached **iteration 16** |
| model reasoning | degenerate loop | coherent ("port will be zero-dependency pure std… cross-check Rust vs C binaries… let me start building") |

Omitting temperature — so the llama.cpp server applies its **own** sampler defaults (temp ~0.8 + `repeat_penalty` ~1.1) — eliminated the collapse and roughly doubled progress. (A separate degeneration still stops the run later; that's a different issue.)

## Root cause

1. `ChatConfig::default().temperature = Some(0.0)` — `crates/octos-llm/src/config.rs:82`.
2. For any model not flagged `fixed_temperature`, that value is sent verbatim — `crates/octos-llm/src/openai.rs:515-519` (`build_request`). Qwen matches no `fixed_temperature` name, so `0.0` goes on the wire.
3. `Agent::chat_config()` (`crates/octos-agent/src/agent/loop_runner.rs:692`) builds from `ChatConfig::default()` and overrides `max_tokens`/`reasoning_effort` but **never temperature** — so there is no path for a profile/config to change it.
4. `OpenAIRequest` (`openai.rs:857-878`) has **no** `frequency_penalty` / `presence_penalty` / `repetition_penalty` / `top_p` / `top_k` / `min_p` field at all — octos cannot send any anti-repetition control even if asked.

Net: local models are forced into the single worst sampling configuration for repetition, with no override and no penalty escape hatch.

## How Pi avoids it

[...]

PR #2176, opened 2026-08-29
feat(llm): cloud-safe sampler passthrough (repeat_penalty etc.) for OpenAI-compatible servers (#2172)

heyong4725:
## What & why

Completes the follow-up documented in #2172 (merged as #2173). The temperature knob only *reduces* local-model repetition collapse — it's run-to-run stochastic: one `temperature=0.7` conversion run scaffolded a clean Rust project, the very next collapsed into **763 identical `update_plan` calls** until `max_tokens`. The robust fix is an anti-repetition sampler, but `OpenAIRequest` had **no** `frequency_penalty` / `presence_penalty` / `repetition_penalty` field — octos literally could not send one. Pi handles exactly this with a `samplingParams` passthrough (`types.ts:186`).

## Change

A `sampling_params` map threaded end-to-end and **flattened verbatim** into the OpenAI request body, so an OpenAI-compatible server (llama.cpp / vLLM / SGLang) receives params octos doesn't model — `repeat_penalty`, `top_p`, `top_k`, `min_p`, `frequency_penalty`, `presence_penalty`:

- `ChatConfig.sampling_params` → `OpenAIRequest` `#[serde(flatten)] extra_sampling`.
- `AgentConfig.chat_sampling_params` → applied in `build_chat_config()`.
- `gateway.llm_sampling_params` (config.json) → threaded at the `chat` / `gateway_runtime` / `acp` sites and surfaced through `config_from_profile` (the serve / octoscode profile path), mirroring `llm_temperature`.

Usage (config.json or profile gateway):
```json
"gateway": { "llm_temperature": 0.7,
             "llm_sampling_params": { "repeat_penalty": 1.15, "top_p": 0.95, "min_p": 0.05 } }
```

## Cloud safety

Cloud-safe by construction: unset → the flatten map is empty → **nothing** is added to the request, byte-for-byte unchanged for cloud. And keys octos already models (`temperature`, `max_tokens`, `tools`, …) are stripped from the passthrough (`RESERVED_SAMPLING_KEYS`), so a misconfigured param cannot emit duplicate/divergent keys across the streaming (`to_value`, last-wins) and non-streaming (`to_vec`, duplicate) send paths — the dedicated field/knob always wins.

## Tests

- `build_request_flattens_sampling_params` — `repeat_penalty`/`top_p` appear as top-level request fields.
- `build_request_omits_sampling_params_when_unset` — cloud-safety: nothing added when unset.
- `build_request_drops_reserved_keys_from_sampling_params` — a stray `temperature` in the passthrough is dropped; the dedicated field wins.
- `test_gateway_llm_sampling_params_parses` / `_absent_is_none` — config surface (absent → None).
- `build_chat_config_threads_sampling_params` — threading, unset → None.

octos-llm (563) + octos-cli (1589) + octos-agent (2651) green; fmt + clippy clean. An adversarial review found the change ship-ready; this PR additionally closes the one edge it flagged (the reserved-key guard above).

## Note

This is the 4th and final octos gap surfaced by the pi-harness comparison (after the streaming total-timeout #2171, the forced-greedy temperature #2173, and the empty-`MaxTokens` dead-end #2175). It's the piece most likely to make a slow local model complete a real conversion reliably rather than collapse.

### DIFF
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -3861,6 +3861,11 @@ fn build_chat_config(config: &crate::AgentConfig) -> ChatConfig {
     if let Some(temp) = config.chat_temperature {
         c.temperature = Some(temp);
     }
+    // Extra sampler params (e.g. repeat_penalty) for OpenAI-compatible servers.
+    // Unset → nothing added, cloud unchanged.
+    if let Some(sampling) = &config.chat_sampling_params {
+        c.sampling_params = Some(sampling.clone());
+    }
     c.reasoning_effort = config.reasoning_effort;
     c
 }
--- a/crates/octos-agent/src/agent/loop_runner_tests.rs
+++ b/crates/octos-agent/src/agent/loop_runner_tests.rs
@@ -6411,6 +6411,29 @@ fn build_chat_config_applies_temperature_override() {
 }
 
 #[test]
+fn build_chat_config_threads_sampling_params() {
+    // #2172: chat_sampling_params flows into ChatConfig.sampling_params; unset
+    // → None (cloud unchanged).
+    let mut sp = serde_json::Map::new();
+    sp.insert("repeat_penalty".to_string(), serde_json::json!(1.1));
+    let cfg = AgentConfig {
+        chat_sampling_params: Some(sp),
+        ..AgentConfig::default()
+    };
+    let chat = build_chat_config(&cfg);
+    assert_eq!(
+        chat.sampling_params
+            .as_ref()
+            .and_then(|m| m.get("repeat_penalty")),
+        Some(&serde_json::json!(1.1))
+    );
+    assert_eq!(
+        build_chat_config(&AgentConfig::default()).sampling_params,
+        None
+    );
+}
+
+#[test]
 fn build_chat_config_applies_max_tokens_override_independently() {
     // Overrides compose without clobbering each other.
     let cfg = AgentConfig {
--- a/crates/octos-agent/src/agent/mod.rs
+++ b/crates/octos-agent/src/agent/mod.rs
@@ -91,6 +91,10 @@ pub struct AgentConfig {
     /// OpenAI-compatible models, where forced greedy decoding causes repetition
     /// collapse. See issue #2172.
     pub chat_temperature: Option<f32>,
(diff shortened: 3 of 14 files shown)


anantheparty (approved):
Reviewed the full diff: defaults are unchanged (all new fields `#[serde(default)]`, `None` until configured), and the cloud-safety contract is pinned by tests (`build_request_omits_sampling_params_when_unset`). The RESERVED_SAMPLING_KEYS defense-in-depth is the right call — a stray `temperature` in `sampling_params` would otherwise serialize differently across the streaming (last-wins) and non-streaming (duplicate key) paths. Coverage for flatten / reserved-key drop / unset paths all present, and every internal ChatConfig constructor now carries `sampling_params: None`. CI green including check-windows. Approving — merging left to you per convention.

anantheparty:
Second-pass review notes (all non-blocking — approving review stands, merge at your discretion):

1. **Silent drop of reserved keys could use a warning.** `build_request` strips reserved keys from `sampling_params` without any signal (openai.rs, the `RESERVED_SAMPLING_KEYS` loop). Since `ChatConfig::default()` forces `temperature: Some(0.0)`, an operator trying to escape greedy decoding but putting `temperature`/`top_p` in `llm_sampling_params` gets a "configured but unchanged" request with zero feedback. A `tracing::warn!` on dropped keys would make the misconfiguration visible.

2. **Gateway-scoped params reach cloud lanes.** All OpenAI-compatible registry lanes (moonshot/deepseek/groq/zhipu/minimax/nvidia/dashscope/openai-direct) share `build_request`, so a gateway-level `llm_sampling_params` (e.g. `repeat_penalty: 1.1`) would flow verbatim to a cloud lane on failover/multi-model routes — some of which 400 on unknown params. This matches the existing `llm_temperature` gateway scope precedent and is opt-in, so fine for now; flagging per-lane scoping as future work.

3. **Test gap (small):** the three new `build_request` tests only exercise the non-streaming path. Streaming (`to_value` + post-hoc `stream`/`stream_options` insert) shares `build_request` so reserved-stripping holds there too, but a streaming-path assertion plus a `Some(empty map)` case would pin it down.

4. (Observation) `tool_choice`/`stop_sequences` are ignored by the OpenAI provider's dedicated path today, while they'd now pass through via `sampling_params` — consistency side door worth keeping in mind.

None of these block landing; 1 is the one I'd actually do before or shortly after merge.

Issue #43, opened 2026-03-19
fix: list_sessions and load_from_disk miss per-user session files


ymote:
## Bug
`SessionManager::list_sessions()` only scans `data/sessions/` (legacy flat layout). Session actors store files in `data/users/{key}/sessions/` (per-user layout). New sessions are invisible to the API.

`load_from_disk()` has the same issue — can't find sessions stored in per-user directories.

## Fix
- `list_sessions()`: scan both `data/sessions/` and `data/users/*/sessions/*.jsonl` with dedup
- `load_from_disk()`: fallback to per-user path when legacy path doesn't exist

## Files
- `crates/octos-bus/src/session.rs`

PR #2178, opened 2026-08-30
fix(goal): online operator transitions — stop live-cache overwrites of offline archive/reopen

ZhangHanDong:
Offline `octos goal archive/reopen` appends a supervisor snapshot that a living serve later overwrites: any goal persist point (turn-end accounting, verifier, peer-cost, even the 10s shutdown drain) re-appends the stale in-memory status with a newer `updated_at_ms`, and replay picks it — the operator transition is silently lost (observed in production; full root-cause: dual-loop analysis, board #43).

Fix (three commits, per the minimal plan):
- **wire online operator transitions** — expose the existing (previously dead-code) `operator_transition_goal` through a local goal operator-control RPC spawned at serve startup; archive clears queued `GoalContinue` in the same critical section; three-way reconciliation (live / supervisor / SQLite) after commit, sync failure reported, never masked.
- **route operator CLI through live serve** — liveness via the existing `.octos-serve.lock` probe: live → RPC; proven offline → legacy append fallback; live but endpoint unreachable → **fail closed** (no silent offline write behind a live cache).
- **register goal operator transition** in the UI protocol docs.

Verification (independent re-run in an isolated worktree by the reviewing outer loop): `cargo clippy --all-targets -- -D warnings` clean · `cargo test -p octos-cli --lib --features api` 3159/0 · four new race tests green (shutdown-drain archive race, reopen single-continuation, RPC-failure no-offline-write, proven-offline fallback) · `cargo fmt --check` clean. Known limitation: control endpoint is `cfg(unix)` — a live Windows serve fails closed (conservative; offline path unchanged).

Board: octos `.octos/OUTER_LOOP_REVIEW.md` #43 (implementer: outer(codex); reviewer: outer(claude)).

https://claude.ai/code/session_01BYaCVLFe1kdCGc7SarVC9F


anantheparty (approved):
Adversarial pre-merge review passed. Verified: the root cause is exactly as described (`persist_goal_state` re-appends with a fresher `now_ms`, replay picks it) and the fix lands where it matters — `default_agent_orchestrator()` is a `OnceLock` process singleton, so the online transition genuinely mutates the live serve's memory; archive tombstoning + status flip share one critical section; operator transitions bump the revision (ABA-safe); verifier paths re-check goal_id/objective/revision under the state gate. Spec change is purely additive and backward-compatible (old clients get `method_not_supported` via the trait default), goldens updated, ci.yml untouched, 24 green runs on head.

Non-blocking notes (filing as a follow-up issue): the offline fallback probe-to-append window is not lock-held (ms-scale, needs concurrent serve start); GoalCommand lacks `--instance-data-dir` so liveness probes state_home for instance-dir serves (pre-existing); wire inventory/book docs lag the new methods; reconciliation can report ok:false after a committed transition (fail-noisy, not fail-unsafe).

Merging — good fix for a nasty production-reported silent-loss bug.

PR #2180, opened 2026-08-30
fix(cli): thread profile gateway LLM knobs onto serve/octoscode sessions (#2179)

### LINKED ISSUE #2179: Serve/octoscode SessionRuntime drops profile gateway LLM knobs (temperature/sampler/max_output/reasoning) — local models silently run greedy
## Problem

A **serve / octoscode** session (the per-session agent materialized by `SessionRuntime::bootstrap` against a profile) silently ignores the profile's gateway LLM knobs. It builds its `AgentConfig` with `..Default::default()` and threads **none** of them, so a profile-driven session runs with the built-in defaults regardless of what the operator configured:

- `temperature` → `0.0` (greedy) instead of the profile's `llm_temperature`
- no sampler passthrough instead of the profile's `llm_sampling_params` (`repeat_penalty`, …)
- `max_tokens` → the 16384 default instead of the profile's `max_output_tokens`
- `reasoning_effort` dropped

For a **local model** this is exactly the combination that causes the failures the recent fixes were meant to prevent: forced-greedy decoding drives the repetition collapse (#2173), the sampler mitigation (#2176) never reaches the wire, and large single-file writes truncate at 16384. The `octos chat` path threads all of these; the serve path does not.

## Root cause

`crates/octos-cli/src/runtime/session.rs` — the `SessionRuntime` agent's `.with_config(AgentConfig { … })` sets `max_iterations` / `save_episodes` / `human_approval_rules` / `format_after_edit` then `..Default::default()`. `AgentConfig::default()` has `chat_max_tokens: None`, `chat_temperature: None`, `chat_sampling_params: None`, `reasoning_effort: None`, so `build_chat_config` falls back to `ChatConfig::default()` (temperature 0.0, `default_max_tokens()` = 16384, no sampler).

`octos chat` (`commands/chat.rs`) threads `config.gateway.{max_output_tokens, llm_temperature, llm_sampling_params, reasoning_effort}` onto its `AgentConfig`. The `ProfileRuntime` in scope at the session site carries the same `config` (`profile.config.gateway`), so the values are reachable — they just weren't wired.

Verified empirically: with the knobs in a project-local `config.json`, `octos chat --profile` (non-default data-dir context) ignores that file and the request goes out with `temperature=0.0, max_tokens=16384, sampling_params=None`; moving them into the **profile** makes them reach the wire (`temperature=0.7, max_tokens=32768, sampling_params={…}`). The serve path had no equivalent wiring at all.

## Fix

Thread `chat_max_tokens` / `chat_temperature` / `chat_sampling_params` / `reasoning_effort` from `profile.config.gateway` onto the SessionRuntime `AgentConfig`, mirroring `octos chat`. Cloud-safe: each is `None` unless the operator set it, so a profile without these keys is byte-for-byte unchanged.

## Acceptance

- A serve session bootstrapped from a profile whose `gateway` sets `max_output_tokens` / `llm_temperature` / `llm_sampling_params` produces an agent whose `agent_config()` reflects them.
- A profile without those keys keeps the defaults (`None`) — cloud/serve behavior unchanged.

## Follow-up (out of scope)

The MCP-serve path (`commands/mcp_serve.rs` `run_session`) has the same gap, but its `AgentConfig` site has no gateway config in scope — `SessionDispatchConfig` would need to carry it. Tracked separately.

heyong4725:
## What & why

Closes #2179. The **serve / octoscode** per-session agent (`SessionRuntime::bootstrap`) built its `AgentConfig` with `..Default::default()` and threaded **none** of the profile's gateway LLM knobs. So a profile-driven session silently ran with the built-in defaults — **greedy `temperature=0.0`, no sampler passthrough, 16384 max output** — dropping the local-model mitigations the profile configured (#2173 temperature, #2176 sampler) and the operator's `max_output_tokens`. `octos chat` threads all of these; the serve path did not.

This was the missing link behind a long local-model debugging chain: with the knobs set in the profile they *do* reach `octos chat` (verified — request goes out with `temperature=0.7, max_tokens=32768, sampling_params={…}`), but a serve/octoscode session had no wiring at all, so those same knobs were ignored.

## Change

Thread `chat_max_tokens` / `chat_temperature` / `chat_sampling_params` / `reasoning_effort` from `profile.config.gateway` onto the SessionRuntime `AgentConfig`, mirroring `commands/chat.rs` exactly (`profile.config.gateway.as_ref().and_then(|g| g.<field>)`).

## Cloud safety

Each field is `None` unless the operator set it in the profile gateway, so `build_chat_config` keeps the built-in defaults and a profile without these keys is **byte-for-byte unchanged**. Cloud/serve behavior is unaffected.

## Tests

- `session_agent_threads_profile_gateway_llm_knobs` — a bootstrapped session agent's `agent_config()` reflects the profile's `max_output_tokens: 32768` / `llm_temperature: 0.7` / `llm_sampling_params: {repeat_penalty}`.
- `session_agent_keeps_defaults_when_profile_has_no_gateway_knobs` — a profile without them keeps `None` (cloud-safety).

`cargo test -p octos-cli` green for the new tests; fmt + clippy clean. (Note: an unrelated `autonomy::agent_orchestrator` test fails on current `main` independent of this change.)

## Follow-up (out of scope)

The MCP-serve path (`commands/mcp_serve.rs` `run_session`) has the same gap, but its `AgentConfig` site has no gateway config in scope — `SessionDispatchConfig` would need to carry it. Tracked separately.

### DIFF
--- a/crates/octos-cli/src/runtime/session.rs
+++ b/crates/octos-cli/src/runtime/session.rs
@@ -563,6 +563,32 @@ impl SessionRuntime {
             human_approval_rules: profile.human_approval_rules.clone(),
             // #1774: opt-in post-edit formatting (rustfmt/prettier/black/gofmt).
             format_after_edit: profile.format_after_edit,
+            // #2172: thread the profile's gateway LLM knobs onto serve /
+            // octoscode sessions, exactly as `octos chat` does. Without this a
+            // profile-driven session silently ran with the built-in defaults
+            // (greedy temperature=0.0, no sampler, 16384 max output) — dropping
+            // the local-model repetition-collapse mitigations. Each is `None`
+            // unless the operator set it, so cloud sessions are unchanged.
+            chat_max_tokens: profile
+                .config
+                .gateway
+                .as_ref()
+                .and_then(|g| g.max_output_tokens),
+            chat_temperature: profile
+                .config
+                .gateway
+                .as_ref()
+                .and_then(|g| g.llm_temperature),
+            chat_sampling_params: profile
+                .config
+                .gateway
+                .as_ref()
+                .and_then(|g| g.llm_sampling_params.clone()),
+            reasoning_effort: profile
+                .config
+                .gateway
+                .as_ref()
+                .and_then(|g| g.reasoning_effort),
             ..Default::default()
         })
         // M11-F regression fix (#891): propagate the pre-assembled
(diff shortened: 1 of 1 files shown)


anantheparty (approved):
Reviewed the diff — this completes the #2172 mitigation story for serve/octoscode sessions: the four profile gateway knobs thread onto AgentConfig exactly as `octos chat` does, every field is `None` unless configured (cloud sessions unchanged), and the two new tests pin both the threading and the defaults-when-absent contract. Clean fix for #2179. Approving — merging left to you per convention.

PR #2183, opened 2026-08-31
fix(llm): protocol- and provider-aware model discovery for fetch_models

### LINKED ISSUE #2165: Make profile/llm/fetch_models provider- and protocol-aware instead of forcing Bearer /v1/models
## Summary

`profile/llm/fetch_models` should describe model-discovery capability, not assume every non-`anthropic` family exposes an OpenAI-style Bearer-authenticated `/v1/models` endpoint.

The current helper switches behavior on the literal provider id `anthropic`. Registered families such as `zai` and `zai-coding` construct an Anthropic Messages provider, but model discovery treats them as OpenAI/Bearer families and synthesizes `/v1/models`. AppUI also drops the route `api_type` before calling the helper and collapses every discovery error into an empty list/`provider_unavailable`.

Make discovery a typed, provider/protocol-aware capability and keep discovery failure separate from inference availability.

## Current behavior and code evidence

At [`main@e188664c`](https://github.com/octos-org/octos/tree/e188664ca252712d2ef18c3991b87a068b0367d3):

- [`raw_profile_llm_fetch_models`](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-cli/src/api/ui_protocol_transport.rs#L12288-L12345) reads family, base URL, credential slot, and key, but ignores `selection.route.api_type`. It calls `fetch_provider_models` and uses `unwrap_or_default`, collapsing transport, auth, status, protocol, and parse failures.
- [`fetch_provider_models`](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-cli/src/api/admin.rs#L1237-L1263) uses Anthropic path/header semantics only when `provider == "anthropic"`. Every other family receives a synthesized `/v1/models` URL and `Authorization: Bearer`.
- The registered [`zai` family](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-llm/src/registry/zai.rs#L8-L22) and [`zai-coding` family](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-llm/src/registry/zai_coding.rs#L8-L30) both construct `AnthropicProvider` against an Anthropic Messages base URL, but their family ids are not `anthropic`.

Consequences include:

- An Anthropic-protocol Z.AI route is probed with the wrong authorization scheme and a fabricated `/v1/models` path.
- An OpenAI-compatible base that ends in a version other than `/v1` can become `<base>/v1/models` rather than a provider-defined path.
- A provider that supports inference but intentionally has no model-list endpoint is reported with the same coarse reason as an unreachable, unauthorized, or malformed endpoint.
- Clients cannot distinguish “enter the model id manually” from “the credential/endpoint is invalid.”

## Minimal reproduction

Use a loopback provider fixture; no external credential is needed.

1. Configure a draft with `family_id: "zai"`, `route.api_type: "anthropic"`, and the fixture base URL.
2. Call `profile/llm/fetch_models`.
3. Capture the request at the fixture.
4. Observe a request to a synthesized `/v1/models` path with `Authorization: Bearer ...`, rather than a discovery strategy derived from the route protocol/provider.
5. Make the fixture accept a normal Anthropic Messages inference request through `profile/llm/test` while returning 404 for model discovery.
6. Observe that inference succeeds but fetch returns the coarse `provider_unavailable` reason.

A second fixture with an OpenAI-compatible versioned base such as `.../v4` demonstrates the forced `.../v4/v1/models` construction.

## Proposed system contract

- Add model discovery metadata/behavior to the provider registry or provider trait. A provider declares one of:
  - supported with a protocol-specific request strategy;
  - supported with a provider-specific implementation;
  - unsupported/manual-model-id only.
- Resolve discovery from the selected route, including `api_type`, family aliases, base URL semantics, authentication scheme, and provider-specific response shape.
- Do not infer protocol from a single literal family id.
- Define URL joining relative to the configured API root; do not strip/append version segments heuristically across providers.
- Return a typed result that separates:
  - `discovered`;
  - `unsupported`;
  - `authentication_failed`;
  - `endpoint_unreachable`;
  - `invalid_response`;
  - `rate_limited`.
- Preserve safe, redacted messages and existing outbound-request protections.
- Keep discovery advisory. `unsupported` or a discovery-only 404 must not mark a configured model unavailable and must not block manual model-id entry, Test, or Save.
- Share the same provider-aware implementation between AppUI and the admin REST surface so the two clients cannot drift.

## Acceptance tests

- `zai` and `zai-coding` never receive a Bearer `/v1/models` request solely because their family id differs from `anthropic`.
- Route `api_type` changes the discovery strategy and is not ignored.
- OpenAI-compatible providers with `/v1`, `/v4`, and custom API roots use the declared/registered discovery path without duplicate version segments.
- A provider with working inference and unsupported model discovery returns typed `unsupported`; manual id + Test/Save still work.
- 401/403, 404 unsupported, timeout, rate limit, malformed response, and empty successful catalogs remain distinguishable.
- Keys are never included in result messages, logs, or fixture snapshots.
- Family aliases resolve to the same discovery strategy as the canonical family.
- AppUI and admin REST fixtures exercise the same implementation and result mapping.

## Related, not duplicates

- #916 introduced the generic `profile/llm/fetch_models` method but did not define per-provider/protocol discovery behavior.
- #1371 confirmed that the REST and WS entry points have different clients; it did not validate their shared discovery helper.
- #289 added custom-provider model fetching and manual fallback in `octos init`; it does not cover AppUI protocol-aware discovery.
- #2123 concerns keyless provider testing, not model-discovery URL/auth semantics.

anantheparty:
## Summary

Model discovery for `profile/llm/fetch_models` (and the admin REST `/api/my/provider-models` surface) no longer switches on the literal family id `anthropic` and no longer synthesizes a Bearer-authenticated `<base>/v1/models` probe for everything else.

Discovery is now a typed, provider/protocol-aware capability resolved from the selected **route**, and it is strictly advisory: discovery failure never marks a configured model unavailable and never blocks manual model-id entry, Test, or Save.

## What changed

- **Registry-declared capability** (`octos-llm`): every `ProviderEntry` declares `model_discovery` — a protocol-specific strategy (`AnthropicMessages`, `OpenAICompatible`, `GeminiList`) or `Unsupported(reason)` for manual-model-id-only families (currently `vertex`, whose service-account auth has no simple listing endpoint).
- **Route resolution mirrors inference** (`octos_llm::discovery::resolve_model_discovery`): only `route.api_type == "anthropic"` overrides the family's declared protocol — exactly the precedence `create_provider_with_api_type` applies at inference time. The protocol is never inferred from the family id string, so `zai` / `zai-coding` (Anthropic Messages families, ids notwithstanding) resolve to the Anthropic strategy even when a saved AppUI route carries the historical `api_type: "openai"` default. Family aliases resolve to the same strategy as their canonical family.
- **Root-relative URL joining**: the listing URL is joined to the configured API root with no heuristic version stripping/appending, so a `/v4` root (zhipu) becomes `/api/paas/v4/models`, never `/api/paas/v4/v1/models`.
- **Typed outcome** (`DiscoveryOutcome`): `discovered` / `unsupported` / `authentication_failed` / `endpoint_unreachable` / `invalid_response` / `rate_limited`, each with a safe redacted message (never the credential). An empty successful catalog is `discovered` with no error, distinguishable from every failure.
- **Both surfaces share one implementation**: the AppUI RPC now echoes `api_type` + `status` (+ `message`/`reason` on failures) instead of collapsing everything into `models: []` + `provider_unavailable`; the REST handler keeps its `string[]` success shape for dashboard compatibility and maps `unsupported` to an empty list (advisory) while surfacing auth/rate/transport failures as typed errors. Discovery-only 404 maps to `unsupported` (manual entry, inference unaffected).

## Behavior preservation

- Literal `anthropic` family: `<root>/v1/models` with `x-api-key` + `anthropic-version` — unchanged.
- Literal OpenAI-compatible families with `/v1` roots: `<root>/models` with `Authorization: Bearer` — unchanged for registered roots (all of which already include the version segment).
- Keyless families still list without an auth header; link-local base URLs are still refused before any outbound request.
- No CI changes.

## Tests

`octos-llm` (15 new):
- zai/zai-coding resolve to the Anthropic strategy without any literal id match; aliases match their canonical family.
- `api_type: "anthropic"` overrides any family; non-anthropic `api_type` values keep the native strategy (mirrors inference); unknown/custom families fall back to OpenAI-compatible.
- URL joining: `/v4` roots never duplicate version segments; `/v1` and bare roots join correctly.
- Loopback fixture: zai probe hits `/api/anthropic/v1/models` with `x-api-key` and **no** Bearer header; 401/404/429/transport/malformed/empty-catalog stay distinguishable; `vertex` is never probed.

`octos-cli` AppUI fixtures (8, incl. updated legacy tests):
- `zai` never receives a Bearer `/v1/models` request; `/v4` roots never get `/v4/v1/models`; manual-only families return typed `unsupported` with zero outbound requests; 401 → `authentication_failed`; empty catalogs stay `discovered` with no reason.

Local: `cargo test -p octos-llm` 578 passed; `cargo test -p octos-cli --lib --features api` ui_protocol 826 passed + admin 64 passed; `cargo fmt --check` and clippy clean for touched code.

Fixes #2165

### DIFF
--- a/crates/octos-cli/src/api/admin.rs
+++ b/crates/octos-cli/src/api/admin.rs
@@ -1201,10 +1201,41 @@ pub async fn provider_models(
             "base_url targets a link-local/metadata address".into(),
         ));
     }
-    let models = fetch_provider_models(&req.provider, &api_key, req.base_url.as_deref())
-        .await
-        .unwrap_or_default();
-    Ok(Json(models))
+    // Protocol-aware discovery shared with the AppUI `profile/llm/
+    // fetch_models` surface — the strategy resolves from the route (api_type
+    // override, then the family's declared protocol), never from the literal
+    // family id, so the two clients cannot drift.
+    let discovery = octos_llm::discovery::resolve_model_discovery(
+        Some(&req.provider),
+        req.api_type.as_deref(),
+    );
+    let outcome = octos_llm::discovery::discover_models(
+        discovery,
+        &api_key,
+        req.base_url.as_deref(),
+        Some(&req.provider),
+    )
+    .await;
+    match outcome {
+        // Success — including an empty catalog, which is data, not an error.
+        octos_llm::discovery::DiscoveryOutcome::Discovered(models) => Ok(Json(models)),
+        // Advisory: this family has no model-list endpoint. Not an error —
+        // manual model-id entry, Test, and Save stay fully available, and the
+        // dashboard treats an empty list as "nothing to suggest".
+        octos_llm::discovery::DiscoveryOutcome::Unsupported(_) => Ok(Json(Vec::new())),
+        other => Err((
+            if other.status_label() == "rate_limited" {
+                StatusCode::TOO_MANY_REQUESTS
+            } else {
+                StatusCode::BAD_GATEWAY
+            },
+            format!(
+                "{}: {}",
+                other.status_label(),
+                other.message().unwrap_or_default()
+            ),
+        )),
+    }
 }
 
 /// Fetch available models from a provider's /v1/models endpoint.
@@ -1232,49 +1263,6 @@ fn base_url_targets_link_local(base_url: &str) -> bool {
         }
         Err(_) => false,
     }
-}
-
-pub(crate) async fn fetch_provider_models(
-    provider: &str,
-    api_key: &str,
-    base_url: Option<&str>,
-) -> Option<Vec<String>> {
-    let base = base_url
-        .map(|u| u.trim_end_matches('/').to_string())
-        .or_else(|| {
-            octos_llm::registry::lookup(provider)
-                .and_then(|e| e.default_base_url.map(|u| u.to_string()))
-        })?;
-    let base_trimmed = base.trim_end_matches("/v1").trim_end_matches("/v1/");
-    let url = if provider == "anthropic" {
-        format!("{base}/v1/models")
-    } else {
-        format!("{base_trimmed}/v1/models")
-    };
-    let client = reqwest::Client::new();
-    let mut req = client.get(&url).timeout(std::time::Duration::from_secs(10));
-    if provider == "anthropic" {
-        req = req
-            .header("x-api-key", api_key)
-            .header("anthropic-version", "2023-06-01");
-    } else if !api_key.is_empty() {
-        // Keyless families pass an empty key — suppress the header instead of
-        // sending a literal `Bearer ` (mirrors test_provider's factory path).
-        req = req.header("Authorization", format!("Bearer {api_key}"));
-    }
-    let resp = req.send().await.ok()?;
-    if !resp.status().is_success() {
-        return None;
-    }
-    let body: serde_json::Value = resp.json().await.ok()?;
-    body.get("data").and_then(|d| d.as_array()).map(|arr| {
-        let mut ids: Vec<String> = arr
-            .iter()
-            .filter_map(|m| m.get("id").and_then(|v| v.as_str()).map(|s| s.to_string()))
-            .collect();
-        ids.sort();
-        ids
-    })
 }
 
 /// Merge a request body's `config` object into an existing profile config.
@@ -1400,6 +1388,10 @@ pub struct TestProviderRequest {
     pub api_key_env: Option<String>,
     #[serde(default)]
     pub base_url: Option<String>,
+    /// Route protocol override for model discovery (`"anthropic"` switches the
+    /// listing strategy); absent means the family's declared protocol.
+    #[serde(default)]
+    pub api_type: Option<String>,
     #[serde(default)]
     pub profile_id: Option<String>,
 }
--- a/crates/octos-cli/src/api/admin.rs
+++ b/crates/octos-cli/src/api/admin.rs
@@ -1205,10 +1205,8 @@ pub async fn provider_models(
     // fetch_models` surface — the strategy resolves from the route (api_type
     // override, then the family's declared protocol), never from the literal
     // family id, so the two clients cannot drift.
-    let discovery = octos_llm::discovery::resolve_model_discovery(
-        Some(&req.provider),
-        req.api_type.as_deref(),
-    );
+    let discovery =
+        octos_llm::discovery::resolve_model_discovery(Some(&req.provider), req.api_type.as_deref());
     let outcome = octos_llm::discovery::discover_models(
         discovery,
         &api_key,
--- a/crates/octos-cli/src/api/ui_protocol_tests.rs
+++ b/crates/octos-cli/src/api/ui_protocol_tests.rs
@@ -8626,7 +8626,7 @@ async fn profile_llm_fetch_models_requires_family_id() {
 }
 
 #[tokio::test]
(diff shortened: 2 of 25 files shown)


anantheparty:
Pre-merge verification notes for the reviewer:

- Contract coverage: zai/zai_coding (Anthropic-protocol families) now declare ANTHROPIC discovery and are wire-tested to probe `{root}/v1/models` with `x-api-key` (never `/v1/models` + Bearer); versioned OpenAI-compatible bases (`/v4`) are joined root-relative (no `/v4/v1/models`); vertex declares Unsupported (manual model id, zero outbound probes); discovery failures are typed (unsupported / authentication_failed / endpoint_unreachable / invalid_response / rate_limited) and never collapse to a bare empty list.
- Both entries share one implementation: AppUI `profile/llm/fetch_models` now echoes `route.api_type`; REST `/api/my/provider-models` keeps `string[]` on success for dashboard compatibility.
- Known deviations (deliberate): only `api_type == "anthropic"` overrides the family declaration (mirrors inference-side priority; honoring any non-empty api_type would reintroduce the bug); one literal-path behavior change per the issue — versionless custom OpenAI roots now get `{root}/models` instead of `{root}/v1/models`.
- Local: octos-llm 578 passed (15 new discovery tests); octos-cli --features api ui_protocol 826 passed (8 fetch_models fixtures) + admin 64; fmt/clippy clean.
- Follow-up filed: r9s declares unconditional OPENAI discovery while its provider switches per model (#2185).

ymote:
Pushed a formatting-only commit (`d1959923d`) — `check` was failing solely on `cargo fmt`, at `admin.rs:1205` and `ui_protocol_transport.rs:12350`. rustfmt collapses both multi-line `resolve_model_discovery(...)` calls onto one line; the diff is 4 insertions / 8 deletions and carries no semantic change.

Flagging it since I touched someone else's branch — revert it freely if you'd rather format differently.

On the change itself: replacing "switch on the literal family id `anthropic`, otherwise synthesise a Bearer `<base>/v1/models` probe" with a typed protocol-aware capability is the right direction, and **discovery being strictly advisory — failure never blocks** is the detail that makes it safe to land. Waiting on CI to go green, then merging.

anantheparty:
Rebased onto current main (3684b10a, force-push 865df372): main had moved (#2188 inlined format args + #2184 landed), conflicting with the fmt commit at the same `resolve_model_discovery` call sites. Resolution kept both intents — the collapse is preserved as its own commit and the format style matches main's inlined args (rustfmt-normalized).

Also dropped the second commit (String-vs-PathBuf test fix): main already carries an equivalent fix, so it was redundant in the rebase.

Post-rebase verification: fetch_models fixtures 8/8 green; two full lib runs showed 2 failures from the known load-dependent flake family (different victims each run: `http_trace_uses_route_templates`, `dual_goal_peer_fenced…`, `appui_due_fixed_loop_tick…`) — all three pass in isolation, consistent with the pre-existing pattern on this machine, unrelated to the diff.

PR #2184, opened 2026-08-31
fix(api): profile/llm upsert and delete invalidate active runtimes or report restart_required

### LINKED ISSUE #2164: Make profile/llm upsert and delete invalidate active runtimes or report restart_required
## Summary

`profile/llm/upsert` and `profile/llm/delete` persist Profile configuration but do not provide the same active-runtime transition and restart truth as `profile/llm/select`.

A successful mutation therefore answers `applied: true` without establishing whether the next turn uses the saved provider portfolio. This is especially unsafe for endpoint/key edits and deletion of the active primary: Settings can report success while cached SessionRuntime/ProfileRuntime instances continue serving the old configuration.

Define one post-commit runtime contract for every Profile LLM mutation.

## Current behavior and code evidence

At [`main@e188664c`](https://github.com/octos-org/octos/tree/e188664ca252712d2ef18c3991b87a068b0367d3):

- [`profile/llm/select`](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-cli/src/api/ui_protocol_transport.rs#L11725-L11759) explicitly invalidates all SessionRuntime entries for the Profile, removes the dynamic ProfileRuntime, and reboots it when possible.
- Select also [returns `restart_required: true` for startup-pinned profiles](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-cli/src/api/ui_protocol_transport.rs#L11787-L11794).
- [`profile/llm/upsert`](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-cli/src/api/ui_protocol_transport.rs#L11828-L11915) saves the profile and calls `ensure_session_profile_runtime`, but does not invalidate either cache first.
- [`ensure_session_profile_runtime`](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-cli/src/api/ui_protocol_transport.rs#L19122-L19143) returns an existing dynamic or startup runtime immediately, so the upsert call cannot refresh an already-cached provider chain.
- [`profile/llm/delete`](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-cli/src/api/ui_protocol_transport.rs#L11918-L12004) saves and returns without cache invalidation or runtime bootstrap.
- The shared [mutation result helper](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-cli/src/api/ui_protocol_transport.rs#L12389-L12400) adds only `applied`; it does not distinguish persisted configuration from live runtime state.

### Dynamic ProfileRuntime

1. Open a Session and complete a turn so its SessionRuntime/ProfileRuntime is cached.
2. Upsert the current primary address with a different base URL/model configuration, or delete the current primary so a fallback is promoted.
3. Observe `applied: true`.
4. Start the next turn.
5. The cached runtime can still contain the pre-mutation provider chain because neither mutation evicted it.

### Startup-pinned profile

1. Start serve with a profile loaded into the immutable startup runtime map.
2. Upsert or delete a configured model through AppUI.
3. Observe `applied: true` with no `restart_required` field.
4. The persisted file differs from the runtime that will serve the next turn until restart.

No external provider call is required to reproduce this; a provider fixture can identify which endpoint/model the next turn used.

## Proposed system contract

Use one post-commit transition for `select`, `upsert`, and `delete`:

- `applied` means persisted only; expose runtime disposition separately.
- For dynamic profiles, invalidate the Profile runtime and all affected SessionRuntime entries after a successful commit, then rebuild or deterministically defer rebuild.
- Use a generation/revision guard so a concurrent in-flight bootstrap cannot reinsert the old runtime after invalidation.
- An in-flight turn keeps its start-of-turn runtime. The next turn observes the new Profile revision.
- For startup-pinned profiles, return `restart_required: true` and do not imply that the persisted mutation is live.
- If persistence succeeds but runtime rebuild fails, return a typed `persisted_but_not_live` disposition with the new config revision; do not collapse it into a warning-only server log.
- Expose enough truth for clients to compare persisted Profile revision/runtime policy stamp with the effective Session runtime.
- Apply the same transition to primary replacement, fallback edit/addition, primary deletion/promotion, fallback deletion, and deletion of the last model.

A possible result shape is:

```json
{
  "applied": true,
  "config_revision": "...",
  "runtime_disposition": "reloaded | deferred | restart_required",
  "restart_required": false,
  "effective_from": "next_turn"
}
```

The exact names can follow the generated UI Protocol style; the semantic split is the requirement.

## Acceptance tests

- Dynamic primary upsert changes the provider used by the next turn without a process restart.
- Editing an existing route address or credential invalidates cached runtimes even when the selected model id is unchanged.
- Deleting the primary promotes the fallback and the next turn uses it.
- Deleting a fallback updates the active failover chain.
- Deleting the last model produces typed runtime-unavailable truth on the next turn.
- Startup-pinned upsert and delete return `restart_required: true` and keep runtime/config stamps distinguishable.
- A turn already in flight finishes on its original runtime; the following turn uses the committed revision.
- A concurrent old bootstrap cannot repopulate the cache after mutation.
- Failed persistence does not invalidate a healthy runtime.
- Persisted-but-rebuild-failed is explicit in the response and recoverable on retry/restart.

## Related, not duplicate

#2148 separates provider/Profile configuration from durable per-Session model selection and specifies hot reload for a future Session selection resource. This issue is narrower and remains necessary: it defines runtime truth for the existing Profile configuration mutations `profile/llm/upsert` and `profile/llm/delete`. It does not add a Session model chooser or change Profile selection semantics.

anantheparty:
## Problem

`profile/llm/upsert` and `profile/llm/delete` persisted Profile configuration but performed no active-runtime transition: upsert called `ensure_session_profile_runtime`, which returns an already-cached dynamic or startup runtime immediately, and delete returned `applied: true` without any invalidation. Cached `SessionRuntime`/`ProfileRuntime` instances kept serving the pre-mutation provider chain on the next turn — unsafe for endpoint/key edits and deletion of the active primary. Settings reported success while the old configuration stayed live. `profile/llm/select` was the only mutation with a runtime transition (and the only one reporting `restart_required`).

## Fix: one post-commit transition for select / upsert / delete

New shared `commit_profile_llm_runtime_transition`, run by all three mutations after a successful commit:

- **`applied` stays persistence-only.** Runtime disposition is exposed separately and uniformly on every mutation result:
  ```json
  {
    "applied": true,
    "config_revision": "<profile updated_at RFC3339>",
    "runtime_disposition": "reloaded | deferred | restart_required | persisted_but_not_live | unchanged",
    "restart_required": false,
    "effective_from": "next_turn",
    "runtime_error": "…only for persisted_but_not_live…"
  }
  ```
- **Dynamic profiles:** evict all cached `SessionRuntime`s for the profile (the session cache already bumps its own generation guard), bump a NEW generation guard on the dynamic `ProfileRuntime` map and drop the entry, then rebuild via `ensure_session_profile_runtime` → `reloaded`; if no runtime bootstraps (disabled profile / last model deleted) → `deferred`, and the next turn reports typed runtime-unavailable truth.
- **Generation/revision guard:** `ensure_session_profile_runtime` captures the dynamic-map generation before reading the profile file and refuses the insert when a concurrent mutation bumped it (one retry from the freshly committed file), so an in-flight bootstrap cannot repopulate the cache with the stale provider chain — mirroring `SessionRuntimeCache::generations` (#1639).
- **Startup-pinned profiles:** `restart_required: true` + disposition `restart_required`, no fake reload — select's existing wire promise, now shared (select keeps reporting `restart_required: true` even for an idempotent re-select).
- **Persisted-but-rebuild-failed is explicit:** disposition `persisted_but_not_live` + `runtime_error` on the wire (previously a warn-only server log); the cache stays empty so the next turn retries the bootstrap and recovers.
- **Failed persistence never reaches the transition:** a healthy runtime stays healthy (`runtime_disposition: "unchanged"`, no eviction).
- An in-flight turn keeps its start-of-turn runtime (sessions hold their own `Arc`); the following turn observes the committed revision.

## Notes

- `restart_required` is now always present as a bool on these mutation results (it was absent for dynamic profiles before); presence is no longer the signal. One existing test assertion updated accordingly.
- `raw_profile_llm_delete` became `async` to run the transition (dispatcher awaits it).
- Drive-by: fixes a pre-existing `--features api` lib-test compile break (`SessionOpenResult.workspace_root` became `String` on main while the Tier-3 test still compared it to a `PathBuf`), needed so the new tests build.
- `ci.yml` untouched. Unconfigured/default user behavior unchanged: mutations with `applied: false` perform no runtime transition, and profiles without an LLM selection keep getting the existing typed runtime-unavailable handling.

## Tests

New `should_*` tests pin the contract (all passing):

- endpoint edit on an unchanged model id rebuilds the cached `ProfileRuntime` and the rebuilt chain serves the committed `base_url` (`reloaded`, `restart_required: false`)
- deleting the primary promotes the fallback and the rebuilt runtime serves it
- deleting the last model evicts the cached runtime (`deferred`) and the next turn sees typed runtime-unavailable truth
- startup-pinned upsert and delete report `restart_required: true` and leave the boot snapshot untouched
- a stale (pre-bump generation) bootstrap insert is refused and leaves the cache empty; a current-generation insert succeeds
- persisted-but-rebuild-failed is explicit (`persisted_but_not_live` + `runtime_error`) and recoverable once the blocker is removed
- delete miss reports `unchanged` and does not evict

Full `cargo test -p octos-cli --lib --features api`: 3165 passed; the two `appui_*_drains_internal_continuation_turn` failures observed under full-suite parallel load pass reliably in isolation/smaller runs (timing-sensitive turn-drain tests, unrelated to profile/llm paths — under investigation whether pre-existing on main).

Fixes #2164

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol_tests.rs
+++ b/crates/octos-cli/src/api/ui_protocol_tests.rs
@@ -1061,9 +1061,17 @@ async fn llm_select_rejects_keyless_models_before_persisting() {
     assert_eq!(result["applied"], true);
     assert_eq!(result["selected"]["model"], "glm-5.3");
     assert_eq!(
-        result.get("restart_required"),
-        None,
+        result["restart_required"],
+        json!(false),
         "dynamic profiles apply without a restart: {result}"
+    );
+    // #2164 uniform post-commit truth: applied stays persistence-only and the
+    // runtime disposition is stamped beside it.
+    assert_eq!(result["runtime_disposition"], "reloaded", "{result}");
+    assert_eq!(result["effective_from"], "next_turn", "{result}");
+    assert!(
+        result.get("config_revision").is_some_and(|r| r.is_string()),
+        "committed revision must be comparable against the runtime stamp: {result}"
     );
 }
 
@@ -1473,8 +1481,13 @@ async fn llm_delete_removes_entries_and_promotes_fallback() {
         &delete("openai", "gpt-4o", "official", "d-miss"),
         None,
     )
+    .await
     .expect("delete miss");
     assert_eq!(result["applied"], json!(false));
+    assert_eq!(
+        result["runtime_disposition"], "unchanged",
+        "a miss persists nothing and must not touch the runtime: {result}"
+    );
 
     // (b) Delete the PRIMARY -> the fallback is promoted.
     let result = raw_profile_llm_delete(
@@ -1482,6 +1495,7 @@ async fn llm_delete_removes_entries_and_promotes_fallback() {
         &delete("zai", "glm-5.3", "official", "d-primary"),
         None,
     )
+    .await
     .expect("delete primary");
     assert_eq!(result["applied"], json!(true));
     let profile = state
@@ -1505,6 +1519,7 @@ async fn llm_delete_removes_entries_and_promotes_fallback() {
         &delete("deepseek", "deepseek-v4-pro", "official", "d-last"),
         None,
     )
+    .await
     .expect("delete last");
     assert_eq!(result["applied"], json!(true));
     let profile = state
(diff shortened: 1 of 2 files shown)


anantheparty:
Pre-merge verification notes for the reviewer:

- Shared post-commit transition `commit_profile_llm_runtime_transition` for select/upsert/delete: session-cache invalidation → dynamic ProfileRuntime map generation bump + drop → dynamic rebuild / startup-pinned `restart_required`.
- Generation guard: `insert_profile_runtime_if_current` does check-gen-then-insert under the runtimes write lock; mutations bump before dropping — a pre-commit bootstrap cannot re-insert a stale runtime (insertion-side retry once, then a deterministic "retry the turn").
- Result stamping: `applied` (persistence only), `runtime_disposition` (reloaded/deferred/restart_required/persisted_but_not_live/unchanged), `restart_required` always-bool, `config_revision`, `runtime_error` only on persisted_but_not_live.
- Local: 6 new contract tests + the api-feature build-breakage fix (String vs PathBuf, pre-existing on main) — 8 passed --exact; full lib --features api 3167 passed; clippy/fmt clean.
- Follow-ups filed from review: #2186 (sibling-writer guard gap + select idempotency semantics), #2185/#2187 for the sibling PRs.

PR #2182, opened 2026-08-31
feat(appui): typed inference-parameter schema for profile/llm/* — reject silently-dropped fields

anantheparty:
Part of #2166

## Problem

`RawLlmSelection` / `RawProfileLlmUpsertParams` had no `deny_unknown_fields` and no inference fields at all: a client could submit `temperature`, `top_p`, `max_output_tokens`, `context_window`, or a nested reasoning object on `profile/llm/upsert`, receive `applied: true`, and present values that were silently discarded. The durable `LlmModelSelectionConfig` — including the #2142 `context_window` override and `model_hints` — was unreachable from AppUI.

## Typed schema (shared by `profile/llm/upsert`, `profile/llm/test`, `profile/llm/fetch_models`)

Test and Save parse the identical shape, so a payload that probes successfully is byte-for-byte the payload that persists. Every optional field follows **`absent ≡ null ≡ inherit`**: the upsert payload is the *complete* inference configuration for the addressed selection; a re-upsert without a field clears the prior override.

| field | type | validation | durable target | runtime effect |
|---|---|---|---|---|
| `selection.family_id` / `model_id` | string | required | selection address | unchanged |
| `selection.route` | `{route_id, label, base_url, api_key_env, api_type}` | — | `LlmRouteConfig` | unchanged |
| `selection.model_hints` | `{uses_completion_tokens, fixed_temperature, lacks_vision, merge_system_messages, reasoning_style}` | typed enum/bools | `model_hints` (existing) | provider request shaping (existing) |
| `selection.context_window` | u32 | ≥ 1 | `context_window` (#2142, existing) | `ContextWindowOverride` outermost layer — local budgeting/compaction truth, **not** an upstream request field |
| `selection.temperature` | f64 | finite, 0.0–2.0 | `temperature` (new) | per-model sampling default |
| `selection.top_p` | f64 | finite, 0.0–1.0 | `top_p` (new) | rides the sampler passthrough map (no typed `ChatConfig` field exists) |
| `selection.reasoning_effort` | enum `none/low/medium/high/max` | typed enum | `reasoning_effort` (new) | per-**model default** tier of the reasoning chain |

### Ownership rules (what the schema deliberately does NOT accept)

- **`max_output_tokens`** → owned by the **profile gateway** contract (`[gateway] max_output_tokens`). Sending it returns `invalid_params` / `kind: "llm_param_owned_elsewhere"` with the owner named — never `applied: true`.
[...]

PR #2188, opened 2026-08-31
chore: inline format args for workspace clippy under new stable

anantheparty:
## Root cause

The `check` job pins `dtolnay/rust-toolchain@stable`, which floats to the latest stable release. A recent stable bump promoted `clippy::uninlined_format_args` to a warn-by-default lint, and the job's `-D warnings` turned every occurrence into a hard error, failing `cargo clippy --workspace --all-targets -- -D warnings` across the workspace.

The same toolchain bump also orphaned two pre-existing latent failures that were masked behind the dependency cascade (crates are only checked after their dependencies build): they are fixed here too, so `check` goes fully green.

## What this PR does

1. **`uninlined_format_args` (the CI red)** — purely mechanical, produced with `cargo clippy --fix` (machine-applicable suggestions only) followed by `cargo fmt --all`:
   - `format!("{}", var)` / `write!(f, "... {}", var)` / `assert!(cond, "... {}", var)` style calls inlined to `{var}` placeholders
   - 88 `.rs` files touched; identifiers only — expressions that cannot be inlined (e.g. `crate::truncated_utf8(...)`) are intentionally left as arguments

2. **`dead_code` in `octos-pipeline`** (commit 2): private `const MAX_TEXT_FRAME_BYTES` (`fidelity.rs:35`) is referenced only by the `const _: () = { assert!(..) }` drift guard, which newer rustc's dead-code pass no longer counts as a use. The const and its compile-time assertions are kept (they encode the documented 1 MiB `frame_too_large` ceiling invariant) and the const now carries `#[allow(dead_code)]` with an explanatory note.

3. **`unstable_name_collisions` in `octos-cli`** (commit 2): the three `fs2::FileExt` `lock_shared`/`unlock` call sites in `autonomy/monitor_runtime.rs` are fully qualified (`fs2::FileExt::lock_shared(...)` etc.) so they cannot resolve to the same-named methods std 1.89 added to `File`; the two `use fs2::FileExt as _;` imports and inert `incompatible_msrv` allows are dropped. Matches the file's own existing qualified-call convention (lines 626/744).

No `.github/` file touched; no behavior change of any kind.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings`: **exit 0** (plain — no allow flags; this is the exact CI `check` command)
- `cargo fmt --all -- --check`: clean
- `cargo test -p octos-core`: 358 passed, 0 failed, 1 ignored
- `cargo test -p octos-pipeline --lib`: 354 passed, 0 failed (covers the touched `fidelity.rs`)
- octos-cli test targets type-check under `clippy --all-targets`; the fs2 change is a method-call → UFCS rewrite of the same trait method on the same receiver (byte-identical semantics)

### DIFF
--- a/crates/app-skills/deep-crawl/src/main.rs
+++ b/crates/app-skills/deep-crawl/src/main.rs
@@ -647,7 +647,7 @@ fn page_slug(url: &Url, index: usize) -> String {
     } else {
         &slug
     };
-    format!("{:03}_{truncated}", index)
+    format!("{index:03}_{truncated}")
 }
 
 /// Truncate a string to max_len at a UTF-8 safe boundary.
@@ -1109,7 +1109,7 @@ async fn run() -> Output {
         let filename = file_url
             .as_ref()
             .map(|u| format!("{}.md", page_slug(u, i)))
-            .unwrap_or_else(|| format!("{:03}_page.md", i));
+            .unwrap_or_else(|| format!("{i:03}_page.md"));
         let file_path = crawl_dir.join(&filename);
 
         let file_content = if let Some(ref err) = crawled.error {
--- a/crates/app-skills/deep-search/src/main.rs
+++ b/crates/app-skills/deep-search/src/main.rs
@@ -1064,9 +1064,9 @@ async fn serper_search(
 
     if let Some(kg) = data.get("knowledgeGraph") {
         if let Some(title) = kg["title"].as_str() {
-            output.push_str(&format!("**{}**", title));
+            output.push_str(&format!("**{title}**"));
             if let Some(desc) = kg["description"].as_str() {
-                output.push_str(&format!(": {}", desc));
+                output.push_str(&format!(": {desc}"));
             }
             output.push_str("\n\n");
         }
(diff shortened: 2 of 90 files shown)


anantheparty:
Reviewer context (mechanical, zero-behavior-change): main's `check` job went red because the floating `stable` toolchain promoted `uninlined_format_args` to a default warning under the job's `-D warnings`. This PR is a purely mechanical inlining across the workspace (88 .rs files, +302/−497) plus two toolchain-exposed pre-existing lint fixes (fs2 UFCS qualification in monitor_runtime.rs; `#[allow(dead_code)]` + doc note for the const drift-guard in octos-pipeline/fidelity.rs whose `const _:` references the new toolchain no longer counts as a use).

Local verification: `cargo clippy --workspace --all-targets -- -D warnings` exit 0; fmt clean; octos-core 358/0; octos-pipeline --lib 354/0. No ci.yml changes. **This unblocks main's red check and the three queued feature PRs (#2182/#2183/#2184), so an early look is appreciated.**

PR #1877, opened 2026-07-30
feat(llm): add Atlas Cloud provider

binyangzhu000-sudo:
## Summary

- add Atlas Cloud to the native LLM provider registry with OpenAI-compatible defaults
- expose Atlas Cloud in interactive init, credential capture, and the dashboard model selector
- cover registry metadata, alias lookup, factory construction, and non-interactive defaults

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p octos-llm registry::tests --lib` (10 passed)
- `cargo test -p octos-cli commands::init::tests --lib` (19 passed)
- `npm run typecheck`
- `npm test` (54 passed)
- `npm run build`
- live `octos chat --provider atlascloud` returned a non-empty response using `deepseek-ai/deepseek-v4-pro`

PR #2189, opened 2026-08-31
feat(catalog): add glm-5.3-flash to the z.ai API and GLM Coding Plan

ymote:
Two entries, mirroring the existing glm-5.3 pair.

**`zai/glm-5.3-flash`** — metered, `type: fast`, **$0.15/M input, $0.50/M output**. That is the **list** price from [docs.z.ai's pricing page](https://docs.z.ai/guides/overview/pricing); the $0.075/$0.25 you may see elsewhere is a launch promotion ending **2026-09-09 (UTC+8)**, and a catalog shouldn't carry a price that expires next week. Real numbers rather than `0.0` — zero and unknown are two different semantics, and zero pricing displays every metered request as free (the exact defect holding #1877).

**`zai-coding/glm-5.3-flash`** — GLM Coding Plan, same Anthropic endpoint as `zai-coding/glm-5.3`. `0.0/0.0` here is semantic (flat-fee subscription), matching the existing plan entry. **`glm-5.3` stays the plan's default** — promoting the new model would be a behaviour change nobody asked for.

`context_window`/`max_output` use the family serving values (1,000,000 / 131,072), matching `zai/glm-5.3`. The docs don't publish flash's window; OpenRouter reports 1,310,720 for its own gateway, but that's a different serving stack and was deliberately not copied.

**Also noticed, deliberately NOT changed here:** docs.z.ai lists GLM-5.3 itself at **$1.4/$4.4** while our catalog has `1.0/4.0`. Either our entry is stale or it reflects something the docs don't (negotiated tier?). Flagging for whoever owns that entry rather than silently repricing an existing model in a PR about a new one.

No registry change needed — the `zai`/`zai-coding` families resolve models from the catalog SSOT. Dashboard regenerated via `sync-dashboard-providers.py` (appended 2).

```
cargo test -p octos-llm --lib  => ok. 578 passed
cargo test -p octos-core --lib => ok. 358 passed
```

Sources: [docs.z.ai pricing](https://docs.z.ai/guides/overview/pricing), [OpenRouter GLM 5.3 Flash](https://openrouter.ai/z-ai/glm-5.3-flash)

### DIFF
--- a/dashboard/src/providers.json
+++ b/dashboard/src/providers.json
@@ -198,6 +198,9 @@
     "models": [
       { "id": "glm-5.3", "input": 0.0, "output": 0.0, "max_output": 131072, "endpoints": [
         { "id": "zai-coding", "label": "GLM Coding Plan", "base_url": "https://api.z.ai/api/anthropic", "api_key_env": "ZAI_API_KEY" }
+      ] },
+      { "id": "glm-5.3-flash", "input": 0.0, "output": 0.0, "max_output": 131072, "endpoints": [
+        { "id": "zai-coding", "label": "GLM Coding Plan", "base_url": "https://api.z.ai/api/anthropic", "api_key_env": "ZAI_API_KEY" }
       ] }
     ]
   },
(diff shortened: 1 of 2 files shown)


anantheparty (approved):
Reviewed the 39-line catalog addition: the two glm-5.3-flash entries mirror the existing glm-5.3 pair's structure (metered zai lane at the docs.z.ai list price 0.15/0.50 with 1M context; coding-plan lane at 0 cost via the anthropic-protocol endpoint with ZAI_API_KEY), `type: fast` matches the flash tier, and the check-matrix catalog tests validate parse-ability. Clean additive registry work. Approving — merging.

PR #2190, opened 2026-08-31
fix(llm): honor temperature and sampling params on the Anthropic protocol path

ymote:
## The gap

PRs #2173 (temperature override) and #2176 (`ChatConfig.sampling_params` passthrough) fixed small-model repetition collapse only on the OpenAI-compatible path. Measured on the base rev (`215b1bf60`):

- `temperature`: **0** occurrences in `crates/octos-llm/src/anthropic.rs`, **52** in `openai.rs`
- `sampling_params`: flattened only in `openai.rs`; zero references in `anthropic.rs`

So every family built on `AnthropicProvider` silently ignored both knobs. An operator setting `temperature: 0.7` on a zai profile got nothing, with no error and no log.

## Emission is model-capability-aware (not protocol-global)

The first cut of this PR forwarded the knobs to **every** model on the Anthropic protocol. Codex review (thanks) caught that this **400s modern first-party Claude** — a new regression on a model we ship. Both failure modes are real:

- **Opus 4.7, thinking-omitted path**: `claude-opus-4-7` (`model_catalog.json`) REMOVED `temperature`/`top_p`/`top_k` from the Messages API. On the small-`max_tokens` path no `thinking` block is built, so the first cut forwarded all sampling → hard 400.
- **`top_p` under thinking**: `thinking + top_p` 400s first-party Claude.

Fix: `model_accepts_sampling(model)` — a **default-DENY GLM allowlist** — gates emission per model.

| model class | reached via | accepts sampling? |
|---|---|---|
| GLM (`glm-*`) | `zai`, `zai-coding` | **yes** — `temperature` + `top_p` + `top_k` |
| first-party Claude (`claude-*`, bare or qualified) | `anthropic`, `r9s` claude-* proxies | **no** — forward nothing |
| anything else (kimi, minimax, custom endpoint, empty) | custom `base_url` | **no** — forward nothing |

**Families reached** (enumerated from `crates/octos-llm/src/registry/` by which construct `AnthropicProvider`): `anthropic`, `zai`, `zai-coding`, `r9s` (only its `claude-*` models — protocol auto-detect routes `glm-*` there to the OpenAI path). zai/zai-coding serve GLM only; anthropic/r9s serve first-party Claude (incl. `claude-opus-4-7`, `claude-opus-4-6`, `claude-sonnet-4-6`).

**Why default-deny (allowlist GLM), not a Claude blacklist**: GLM is the only class *known* to accept sampling here, so allowlisting it is the safe direction the reviewer named ("if a model's sampler acceptance isn't determinable, treat it as accepts-nothing"). A Claude blacklist (`!starts_with("claude")`) is default-ACCEPT and wrong three ways: it forwards to kimi/minimax/custom endpoints (a **regression** vs origin/main, which forwarded nothing to any Anthropic-path model, and a 400 risk for endpoints that reject sampling), and it lets a family-qualified `anthropic/claude-*` or `r9s/claude-*` through (the "claude" isn't a prefix after the `<family>/`). Default-deny closes that qualified bypass for free.

**Why match the model string, not the family label**: the sampler contract is a property of the model. Matched on the normalized **last path segment** (lowercased, segment after the final `/`), so a `claude-*` routed through any custom `base_url` is gated off and a `vendor/glm-*` is still served. The registry strips the catalog `<family>/` prefix before construction, so the model arrives bare in practice; the segment split only guards the custom-base_url case. **Why all `claude-*`, not just ≥4.7**: no per-model contract table exists in the repo to separate accepts (≤4.6) from rejects (≥4.7); the safe direction gates all `claude-*` off. This PR delivers the GLM repetition-collapse knobs while leaving every non-GLM model **no-worse-than-today**.

## Behavior within an accepting model

1. **`temperature`** — forwarded only when non-zero. `Some(0.0)` is `build_chat_config`'s built-in default (#2172 invariant: an unset `chat_temperature` keeps `0.0` so cloud requests are unchanged), indistinguishable from "unset" at the provider, so it stays off the wire. **Byte-identity proven by a golden** captured on the base rev (`should_serialize_byte_identical_golden_when_no_sampling_override_configured`), plus a sibling asserting an accepting (GLM) model also emits nothing when unset. Documented limitation: explicit `0.0` is unrepresentable; near-greedy via `0.01`.
2. **`sampling_params`** — allowlist of the Messages API surface: only `top_p`/`top_k`, forwarded verbatim. Everything else (`repeat_penalty`, `frequency_penalty`, `min_p`, …) is dropped with one `tracing::warn!` naming the keys. On a rejecting (Claude) model, `top_p`/`top_k` fall through the same drop path. Modeled keys (`temperature`, `max_tokens`, …) fall out of the allowlist too, mirroring the OpenAI path's reserved-key defense (#2172).

**Log levels** — the drop `warn!` is operator-config-only (`sampling_params` is `None` on stock/internal flows). The rejecting-model temperature suppression is `debug!`, not `warn!`, because `compaction.rs` and `rich_output.rs` both set `temperature: 0.2` in live code and can target a Claude model — a warn would cry wolf on ordinary turns. (This also subsumes the earlier thinking-specific suppression: with Claude gated off entirely, there is no thinking/`top_p` conflict left to special-case.)

## Cache breakpoints (#1640)

Untouched. `cache_control` lives in nested system/tool/message blocks; the new fields are top-level and `skip_serializing_if`-absent when unset. `should_keep_cache_breakpoints_unchanged_when_sampling_forwarded_to_glm` asserts exactly three `cache_control` markers with overrides active on the accepting path, and the no-override golden pins their exact bytes.

## Tests (`should_<expected>_when_<condition>`)

Capability gate pinned from both sides (mutation-verified):

- `should_accept_only_glm_and_reject_everything_else` — the gate's truth table: glm-* (all suffixes: -air/-turbo/-flash/-5.x, case-insensitive, and family-qualified) → accept; claude-* bare AND qualified (anthropic/, r9s/), kimi, minimax, custom, empty → reject
- `should_serialize_byte_identical_golden_when_no_sampling_override_configured` — pre-change golden bytes; `None` temperature yields identical bytes
- `should_emit_no_sampling_fields_on_accepting_model_when_no_override_configured`
- `should_forward_temperature_top_p_and_top_k_to_glm_when_operator_overrides`
- `should_serialize_temperature_shortest_form_on_glm_wire` — literal `"temperature":0.7` wire bytes
- `should_treat_zero_temperature_as_unset_sentinel_even_on_accepting_model`
- `should_forward_top_p_top_k_and_drop_openai_only_keys_on_glm`
- `should_drop_modeled_keys_when_smuggled_via_sampling_params_on_glm`
- `should_keep_cache_breakpoints_unchanged_when_sampling_forwarded_to_glm`
- `should_not_send_any_sampling_to_first_party_claude_when_operator_overrides` (H3, plain path)
- `should_not_send_any_sampling_to_opus_4_7_when_thinking_omitted_for_small_max_tokens` (H3, the exact leak the first cut had)
- `should_not_send_top_p_to_first_party_claude_even_when_thinking_enabled` (H4)

**Mutation check**: forcing `model_accepts_sampling` → always-true fails the 4 Claude-suppression tests (the H3/H4 regression); → always-false fails the 5 GLM-forward tests; the classifier unit test fails on both.

## Stated assumptions / out of scope

- **GLM under thinking**: GLM is treated as accepting the full sampler set unconditionally. If z.ai's GLM Anthropic-compat endpoint narrows the set under its thinking mode, that is a GLM-specific follow-up (not a first-party 400). Unverifiable from the repo; stated plainly.
- **No shared capability surface existed to reuse**: `model_catalog.json` has no sampler-acceptance field (cost/context/score only), and `openai::ModelHints.fixed_temperature` means "pins temperature server-side" (kimi), not "the Messages API rejects it" — and is never threaded into `AnthropicProvider`. A minimal local `model_accepts_sampling` is added rather than overloading `ModelHints`. A per-model Claude carve-out (re-enable ≤4.6) is deferred.
- **Tracking-only, pre-existing on main, out of scope**: `ChatConfig` models `stop_sequences` (config.rs) but `AnthropicRequest` omits it (anthropic.rs) — same omission as before this PR.

## Verification (CI-exact, sequential)

```
cargo test -p octos-llm --lib
    test result: ok. 590 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 6.05s
cargo clippy -p octos-llm --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.13s
cargo fmt --all -- --check
    (clean)
```

Refs #2172

### DIFF
--- a/crates/octos-llm/src/anthropic.rs
+++ b/crates/octos-llm/src/anthropic.rs
@@ -150,6 +150,10 @@ impl AnthropicProvider {
         if cache.is_some() {
             apply_message_cache_breakpoint(&mut api_messages);
         }
+        let thinking = config
+            .reasoning_effort
+            .and_then(|effort| build_anthropic_thinking(effort, max_tokens));
+        let (temperature, top_p, top_k) = self.sampling_fields(config);
         AnthropicRequest {
             model: &self.model,
             max_tokens,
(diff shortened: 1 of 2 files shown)


anantheparty (dismissed):
Adversarial pre-merge review passed. The byte-identical no-override golden (prompt-cache prefixes depend on the exact shape) is exactly the right regression pin; the 0.0-sentinel treatment, the top_p/top_k-only Anthropic surface with named-key drop warnings (mirroring the OpenAI path defense-in-depth), and the extended-thinking suppression at debug level with the rich_output rationale are all the right calls. The GLM carve-out being left as a ModelHints follow-up is honestly scoped. Approving — merging left to you per convention.

ymote:
## Adversarial review — round 1 — DO-NOT-MERGE

**Clean:** byte-identity of the no-override path (H2 — default 0.0→None, all three fields skip-if-none, streaming only adds `stream:true` over a sorted serde map, no preserve_order); cache breakpoints untouched (H5 — sampling fields are separate top-level members appended after the nested cache_control blocks); family enumeration correct (H6 — anthropic/zai/zai-coding/r9s-claude-only + custom base-URL). The 0.0-sentinel filtering is tracking-only (pre-change ignored ALL temperatures incl. explicit zero, so this doesn't worsen those callers).

**MUST-FIX (H3+H4, one root cause) — sampling emission is protocol-global but must be model-aware:**
- **Opus 4.7 is in the catalog** (model_catalog.json:83) and its current contract removed temperature/top_p/top_k/budget-thinking. With thinking omitted (small-max_tokens path), this code forwards ALL configured sampling fields (anthropic.rs:153/:241) → a NEW 400 regression on a first-party model we ship.
- Under thinking, temperature/top_k are suppressed but top_p passes verbatim (anthropic.rs:253/:273) — `thinking + {top_p:0.9}` 400s first-party Claude. The test masked it with 0.96 only (:2193) though its own comment asserts the [0.95,1] rule (:2185).

Fix direction: make emission model-capability-aware — a model that rejects a sampler must never receive it. The conservative landing that still delivers this PR's actual goal (GLM/zai repetition-collapse relief): forward sampling only to families known to accept it; for first-party `anthropic`, forward nothing unless per-model acceptance is provable — reverting Claude to exactly today's behaviour.

**Tracking-only:** stop_sequences modeled in ChatConfig, absent from AnthropicRequest — pre-existing on main (config.rs:24 vs anthropic.rs:465).

Round 2 in progress.

ymote:
Addressed the DO-NOT-MERGE. Both MUST-FIX (H3 + H4) shared one root cause — emission was protocol-global — and are fixed together by making it **model-capability-aware** (`model_accepts_sampling`).

- **H3 (sampling leaks to Opus 4.7 on the thinking-omitted path)** and **H4 (`top_p` under thinking 400s Claude)**: all `claude-*` now receive **no** sampling, reverting first-party Claude to exactly the pre-#2172 wire. Verified against `model_catalog.json` — we ship `claude-opus-4-7`; since the repo has no per-model contract table to separate ≤4.6 (accepts) from ≥4.7 (rejects), I took the safe direction (undeterminable ⇒ accepts-nothing) and gated **all** `claude-*` off.
- **GLM via zai/zai-coding still gets `temperature` + `top_p` + `top_k`** — the actual point of the PR.
- Gated on the **model string**, not the family label, so a `claude-*` routed through any base_url is still gated off and `glm-*` is served regardless of family.

**Capability-surface check (before inventing one)**: `model_catalog.json` has no sampler-acceptance field, and `openai::ModelHints.fixed_temperature` means "pins temperature server-side" (kimi), not "Messages API rejects it" — and is never threaded into `AnthropicProvider`. So a minimal local `model_accepts_sampling` rather than overloading `ModelHints`.

**Comment/warning corrected** — no longer claims a "protocol-global allowlist"; the drop `warn!` stays (operator-config-only), and rejecting-model temperature suppression is `debug!` (compaction/rich_output set `temperature: 0.2` on stock turns and can target Claude).

**RED-first + mutation-proof**: new tests that `thinking + top_p=0.9` does not reach Claude, that Opus-4.7-with-thinking-omitted gets no sampling, and that zai/GLM still gets temperature+top_p. Forcing the gate → always-true fails the 4 Claude tests; → always-false fails the 5 GLM tests; the classifier unit test fails on both.

**Tracking-only** (noted, no code): `stop_sequences` is modeled in `ChatConfig` but absent from `AnthropicRequest` — pre-existing on main, out of scope.

Gates (sequential): `cargo test -p octos-llm --lib` → 590 passed; `cargo clippy -p octos-llm --all-targets -- -D warnings` clean; `cargo fmt --all -- --check` clean.

ymote:
## Adversarial review — round 2 — DO-NOT-MERGE (gate polarity inverted)

The plumbing is right (H3/H4/H5 refuted — gate in shared build_request on both temperature and sampling_params, both chat/chat_stream; byte-identity for correctly-classified claude holds; smuggled `temperature` key falls to the dropped arm; no truncation/dead-code from the amend; warn/debug split correct). But the gate is the WRONG polarity:

```rust
!normalized_model.starts_with("claude")   // anthropic.rs:330 — default ACCEPT
```

That forwards sampling to kimi-k2, MiniMax, self-hosted models, the empty string, AND qualified `r9s/claude-opus-4-7` / `anthropic/claude-opus-4-7` (no "claude" prefix after the label) — all of which can 400. It contradicts the intended 'undeterminable ⇒ accepts-nothing' policy AND the PR's own doc comment (:307 says GLM is the accepting class), and it's a **regression vs origin/main** for custom Anthropic endpoints, which previously received none of these fields.

The claimed mutation-proof test locks the bug in: anthropic.rs:2111 explicitly requires `kimi-k2` and `""` to accept sampling.

**Flip list (codex):** make `model_accepts_sampling` default-DENY — true only for normalized `glm-*`; reject unknown, empty, kimi, and qualified-claude. This closes the H1 qualified-name bypass for free. Rewrite the truth-table test accordingly.

Round 3 in progress.

ymote:
Fixed the polarity — you're right, the gate shipped as `!starts_with("claude")`, which is default-ACCEPT and wrong (forwards to kimi/minimax/custom endpoints — a regression vs origin/main and a 400 risk — and lets qualified `anthropic/claude-*` / `r9s/claude-*` through, the H1 bypass).

**Inverted to default-DENY, allowlist GLM only:**
```rust
fn model_accepts_sampling(model: &str) -> bool {
    let leaf = model.rsplit('/').next().unwrap_or(model);
    leaf.trim().to_ascii_lowercase().starts_with("glm-")
}
```
- Only an affirmative `glm-` prefix on the normalized last path segment accepts. All the zai/zai-coding suffixes match (`glm-4.5-air`, `glm-4.7`, `glm-5-turbo`, `glm-5.1`, `glm-5.3`, `glm-5.3-flash`); `glm` without the hyphen does not.
- Everything else — claude (bare AND qualified), kimi, minimax, custom endpoints, empty string — gets nothing. This closes H1 for free (default-deny doesn't care whether "claude" is a prefix or after a `/`) and reverts non-GLM to origin/main behavior.
- Last-path-segment match handles a custom base_url passing `anthropic/claude-*` or `vendor/glm-*`. Verified what actually arrives: `registry::mod` splits the catalog `<family>/` prefix (`split_once('/')`) before construction, so the model is bare in practice — the segment split only guards the custom-endpoint case.

**Corrected the truth-table test** (renamed `should_accept_only_glm_and_reject_everything_else`): glm-* variants (incl. case-insensitive + qualified) → accept; claude-* bare/qualified, `r9s/claude-*`, kimi-k2, minimax-m2.1, `my-local-model`, `glmini`, `""` → reject. RED-first: it failed against the old `!starts_with` gate (`anthropic/claude-opus-4-7 must be classified as REJECTING`). Mutation-proof kept meaningful: gate-forced-true fails the reject rows, gate-forced-false fails the glm rows (both re-verified).

Doc comments on `model_accepts_sampling`, `sampling_fields`, and `ChatConfig::{temperature,sampling_params}` updated to default-deny framing. Scope held to gate polarity + normalization + truth-table test + docs — H3/H4/H5 plumbing untouched.

Gates (sequential): `cargo test -p octos-llm --lib` → 590 passed; `cargo clippy -p octos-llm --all-targets -- -D warnings` clean; `cargo fmt --all -- --check` clean.

Ancestry: #2194 (cache-retention edit to anthropic.rs) hasn't landed yet; my change is scoped to the sampling-emission region (the free function + `sampling_fields`), so the two merge cleanly.

PR #2191, opened 2026-08-31
feat(core): pi-shaped structured truncation report + grep per-line cap

ymote:
Reference design: pi's `packages/coding-agent/src/core/tools/truncate.ts` — its truncation helper returns a structured `TruncationResult`; octos's `truncate_head_tail` returned a bare `String` and threw away everything a caller needs to advise the model about the cut.

## Part 1 — structured truncation result (octos-core)

New `octos_core::truncate_head_tail_report(s, max_len, head_ratio) -> TruncationReport`:

```rust
pub enum TruncatedBy { Bytes, Lines }

pub struct TruncationReport {
    pub content: String,
    pub truncated: bool,
    pub truncated_by: Option<TruncatedBy>, // only Bytes reachable today (doc'd)
    pub total_bytes: usize,
    pub output_bytes: usize,               // len of content, marker included
    pub omitted_bytes: usize,              // the exact N the marker prints
    pub max_len: usize,                    // applied limits
    pub head_ratio: f32,                   // as applied (clamped to [0.1, 0.9])
}
```

`truncate_head_tail` is now a thin wrapper over the report (one implementation, two surfaces). Byte-identity of the legacy surface is pinned by a golden characterization test captured **before** the refactor, plus a property-style wrapper-equivalence test over fixtures (empty / at-limit / one-over / 10K ASCII / 4-byte emoji / CJK / mixed) x limits x ratios.

Where pi's shape did not transfer (said in docs rather than invented):

- `totalLines` / `outputLines` / `maxLines` — pi cuts line-wise; the head/tail split cuts purely by bytes, so line counts would be fabricated. Not shipped; `TruncatedBy::Lines` is declared for future line-based helpers and documented as unreachable today.
- `lastLinePartial` — a tail-truncation edge case; a middle-elision split makes both segments deliberately partial at the cut, so the flag is meaningless here. Not shipped.
- `firstLineExceedsLimit` — **not shipped** (review H3): a bytes-only head/tail split has no line awareness, so any analogue would be a different concept wearing pi's name; it could never fire under real `tool_output_limit`s (all >= 20K vs the ~60-byte degenerate threshold), the genuine oversized-line signal already lives where lines exist (the read-paging probe's `max_line_bytes` check and the windowed read_file work building on it), and no consumer reads it — so the field is dropped rather than exported as misuse surface. The degenerate small-limit regime (content is only the elision marker, which can exceed `max_len`; pre-existing behaviour) stays pinned on the content itself by `should_emit_only_marker_when_budget_below_marker_overhead`.

### Migrated call site: the execution-loop backstop

`Agent` per-tool output truncation (`crates/octos-agent/src/agent/execution.rs`) now consumes the report: `Tool::truncation_recovery` (#2124) receives `report.omitted_bytes` instead of re-deriving `untruncated_len - content.len()`, which undercounts by the marker's own length.

Emitted text before/after (120KB output under the 50KB cap; marker itself is byte-identical, the recovery suffix's N changes):

```text
before:  ... [70050 bytes omitted] ...     <- inline marker
         [70017 bytes omitted] Continue with page: 1.   <- recovery advice

after:   ... [70050 bytes omitted] ...
         [70050 bytes omitted] Continue with page: 1.
```

The model previously saw two disagreeing numbers about one cut. Non-truncated path is byte-identical. One micro-delta: the recovery gate is now `report.truncated` rather than `content.len() < untruncated_len`; they diverge only when the marker is longer than the bytes it saves (`max_len` under ~30 bytes — production limits are >= 20K, unreachable at this call site). The second `truncate_head_tail` call site (hook-feedback path, execution.rs:570) intentionally stays on the wrapper.

Tracking note (pre-existing on main, unchanged here, flagged in review as tracking-only): the approved-tool after-hook path (execution.rs:561/:570) neither caps ordinary approved results nor supplies `truncation_recovery` metadata — noted so it isn't lost; not load-bearing for this change.

## Part 2 — grep per-line cap

pi caps each grep match line at 500 chars (`GREP_MAX_LINE_LENGTH`); octos's grep emitted matching lines unbounded — one minified-JS line could blow the whole 30KB budget blind. Both emission sites (plain matches and `context` lines) now cap at 500 **chars** (char-indexed, UTF-8-safe) with a `… [line truncated, N chars total]` suffix.

`N` is the **raw physical line's** char count at BOTH sites (review H4): the plain match site trims before capping, and counting the trimmed text there while the context site counted the raw line made the same physical line report two different totals (506 vs 606) depending on the `context` argument. `cap_match_line(display, raw_line)` fires the cap on the emitted text but always names the raw total; pinned by a test that greps the same line with and without `context` and asserts one total, observed failing (506 vs 606) before the fix.

`truncation_recovery` updated accordingly:

```text
before:  [N bytes omitted] Too many matches to return. Narrow the search: <levers>.
after:   [N bytes omitted] Too many matches to return. Narrow the search: <levers>. Matched
         lines are already capped at 500 chars — re-running cannot reveal more of a long
         line; use read_file to see one in full.
```

("Too many matches" also becomes *accurate* — pre-cap, a single giant line could trigger the backstop with one match.)

## Part 3 — truncation contract in descriptions

`grep` and `web_fetch` descriptions each gained one sentence stating the cap up front, formatted from `octos_core::tool_output_limit` at first use (LazyLock) so the numbers cannot drift from the real limits:

- grep: "Output is truncated beyond 30000 bytes (middle elided) and each matched line beyond 500 chars, so prefer narrow patterns and path/file_pattern filters over broad sweeps."
- web_fetch: "Output beyond 40000 bytes is truncated with a '[N bytes omitted]' middle marker, regardless of max_chars."

read_file/shell descriptions untouched (owned by parallel work).

## TDD

RED first, all failures observed before implementing (the loop test failed with exactly the predicted 33-byte disagreement: `left: 70050, right: 70017`; the grep cap tests are the deleted-cap mutant by construction — same code state, same failure; the H4 consistency test observed failing 506 vs 606 pre-fix).

New tests:
- octos-core: `should_emit_exact_legacy_marker_when_truncating` (golden pin, green pre-refactor), `should_report_untruncated_when_input_at_exact_limit`, `should_report_bytes_truncation_when_one_byte_over_limit`, `should_report_untruncated_when_input_empty`, `should_keep_utf8_boundaries_when_multibyte_chars_straddle_the_cut`, `should_match_wrapper_content_when_report_and_wrapper_share_inputs`, `should_emit_only_marker_when_budget_below_marker_overhead`, `should_clamp_reported_head_ratio_when_ratio_out_of_range`
- grep (real execute path, fixture files): `should_cap_match_line_at_500_chars_when_line_is_longer`, `should_not_cap_match_line_when_exactly_500_chars`, `should_cap_multibyte_match_line_at_char_boundary_when_over_limit`, `should_cap_context_lines_when_context_requested`, `should_report_same_chars_total_with_and_without_context_when_line_capped`, `should_mention_per_line_cap_in_recovery_when_matches_overflow`, `should_state_truncation_contract_in_description_when_grep`
- web_fetch: `should_state_truncation_contract_in_description_when_web_fetch`
- loop: `should_agree_on_omitted_bytes_between_marker_and_recovery_when_backstop_truncates`

## Verification (after review fixes)

```text
cargo test -p octos-core --lib   -> ok. 366 passed; 0 failed; 1 ignored
cargo test -p octos-agent --lib  -> ok. 2705 passed; 0 failed; 3 ignored
cargo clippy -p octos-core -p octos-agent --all-targets -- -D warnings -> clean
cargo fmt --all -- --check       -> clean
```

### DIFF
--- a/crates/octos-agent/src/agent/execution.rs
+++ b/crates/octos-agent/src/agent/execution.rs
@@ -2383,22 +2383,28 @@ impl Agent {
             //
             // The cut is a BACKSTOP: it is the one point every tool's output
             // funnels through, which is also the point that knows the least —
-            // `truncate_head_tail` receives a string and a number. On its own it
-            // leaves the model with `... [N bytes omitted] ...` and no way to
-            // reach what it lost, so the only recovery is re-running the call,
-            // which returns the same output cut the same way and spends the
-            // tokens the cap was meant to save.
+            // `truncate_head_tail_report` receives a string and a number. On
+            // its own it leaves the model with `... [N bytes omitted] ...` and
+            // no way to reach what it lost, so the only recovery is re-running
+            // the call, which returns the same output cut the same way and
+            // spends the tokens the cap was meant to save.
             //
             // `Tool::truncation_recovery` is the missing half: the tool still
             // knows its own arguments and whether it paginates, so it can name a
             // concrete next call. Tools with no resume path return None and get
             // no invented advice.
+            //
+            // The hook is handed `omitted_bytes` from the structured report —
+            // the same N the inline marker prints — instead of the legacy
+            // `untruncated_len - content.len()` re-derivation, which
+            // undercounted by the marker's own length and told the model two
+            // disagreeing numbers about one cut.
             let limit = octos_core::tool_output_limit(&tc_name);
-            let untruncated_len = content.len();
-            let mut content = octos_core::truncate_head_tail(&content, limit, 0.7);
-            if content.len() < untruncated_len {
+            let report = octos_core::truncate_head_tail_report(&content, limit, 0.7);
+            let mut content = report.content;
+            if report.truncated {
                 if let Some(recovery) = tools.get(&tc_name).and_then(|tool| {
-                    tool.truncation_recovery(&effective_args, untruncated_len - content.len())
+                    tool.truncation_recovery(&effective_args, report.omitted_bytes)
                 }) {
                     content.push('\n');
                     content.push_str(&recovery);
--- a/crates/octos-agent/src/agent/loop_runner_tests.rs
+++ b/crates/octos-agent/src/agent/loop_runner_tests.rs
@@ -6380,6 +6380,64 @@ async fn truncated_tool_output_reaches_the_model_with_its_recovery_advice() {
     );
 }
 
+/// Every `<digits> bytes omitted` count in `content`, in order of appearance.
+fn omitted_byte_counts(content: &str) -> Vec<u64> {
+    let mut counts = Vec::new();
+    for (idx, _) in content.match_indices(" bytes omitted") {
+        let digits: Vec<char> = content[..idx]
+            .chars()
+            .rev()
+            .take_while(char::is_ascii_digit)
+            .collect();
+        let digits: String = digits.into_iter().rev().collect();
+        if !digits.is_empty() {
+            counts.push(digits.parse().expect("ascii digits"));
+        }
+    }
+    counts
+}
+
+/// The backstop's inline marker (`... [N bytes omitted] ...`) and the
+/// recovery advice the tool composes (`[N bytes omitted] Continue...`) are
+/// two tellings of the SAME cut and must carry the same byte count.
+///
+/// The legacy loop recomputed the recovery count as
+/// `untruncated_len - content.len()`, which undercounts by the length of the
+/// elision marker itself — the model was shown two numbers that never agreed.
+/// The structured `truncate_head_tail_report` hands both consumers the one
+/// `omitted_bytes` the split actually measured.
+#[tokio::test]
+async fn should_agree_on_omitted_bytes_between_marker_and_recovery_when_backstop_truncates() {
+    let dir = tempfile::tempdir().unwrap();
+    let mut tools = ToolRegistry::with_builtins(dir.path());
+    tools.register(OverflowingPagedTool);
+
+    let provider: Arc<dyn LlmProvider> = Arc::new(CallsOverflowingToolThenEnds {
+        calls: AtomicUsize::new(0),
+    });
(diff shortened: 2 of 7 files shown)


ymote:
## Adversarial review — round 1 — DO-NOT-MERGE (two small must-fixes)

**REFUTED (clean):** wrapper byte-identity (H1 — ratio clamping pre-existed, UTF-8 loops unchanged, separator exact; for 100,000 bytes @ 30,000 limit both versions retain 29,950 and print `70050`, only the recovery suffix corrects from 70017), gate divergence (H2 — legacy gate differs only when `max_len < 30` AND `input <= 30`; minimum runtime budget is 20,000; `omitted_bytes` cannot underflow — overlap returns first at utils.rs:153), description numbers (H5 — resolve to 30,000/40,000 by runtime tool name).

**MUST-FIX 1 — `first_segment_exceeds_limit` (utils.rs:80/:167, exported lib.rs:53).** Documented as pi's oversized-first-line analogue; implemented as `head_end == 0` — a different concept. The repo defines the real condition as `max_line_bytes > CANARY_MAX_BYTES` (read_paging_probe.rs:62/:198), and PR #2193's giant-line branch builds on that meaning. Counterexample: 48KiB single line under a 24KiB window → probe says unpageable, this field says `false`. Fix: rename to the honest meaning and drop the pi analogies.

**MUST-FIX 2 — grep's `N chars total` is site-dependent (grep_tool.rs:85/:445/:454).** Match site caps `line.trim()`, context site caps the raw line — the same line reports 506 without `context` and 606 with it. Fix: same raw physical-line count at both sites.

**Tracking-only:** the approved-tool after-hook path (execution.rs:561/:570) neither caps nor advises — identical on main, pre-existing.

Round 2 in progress.

PR #2192, opened 2026-08-31
feat(shell): spool over-cap output to a file the model can grep

ymote:
## Problem

`shell` output over its 30KB cap (`octos_core::tool_output_limit("shell")`) is head/tail-cut by the execution loop with `[N bytes omitted]` — the omitted middle is gone forever. For `cargo test` / `npm test` runs the middle is where the failures live, and the model's only recovery is re-running the command, paying the run AND the tokens again. #2124 gave `read_file`/`grep` recovery advice; shell was deferred because its real recovery is spooling.

## Change (shell tool only)

When a foreground command's combined stream (stdout + stderr) plus its appended suffixes would exceed the cap, the FULL **sanitized** stream is written to `<workspace>/.octos/tmp/shell/output-<millis>-<seq>.log` and the inline result becomes head + tail + a pi-style footer:

```
[Showing lines 1-1710 and 10896-12000 of 12000. Full output: /path/to/ws/.octos/tmp/shell/output-…-0000.log]
```

The model then greps/reads the file instead of re-running. Head and tail are cut on line boundaries so the footer ranges are exact; a single line too big for its budget switches the footer to a byte-range form rather than reporting line numbers that would be lies.

Reference: the pi harness — `packages/coding-agent/src/core/tools/bash.ts` (footer at ~436-440: `[Showing lines X-Y of Z. Full output: /tmp/...]`) and `output-accumulator.ts` (keeps a display tail, opens a temp file when the full output must survive). Pi's up-front description warning ("If truncated, full output is saved to a temp file") is adopted in shell's `description()`, which also names the location (`.octos/tmp/shell/`).

## Security posture (adversarial review round)

- **Sanitize-before-persist.** The stream runs through `sanitize_tool_output` (API keys, bearer tokens, secret assignments, long hex, data URIs, injection defusing) IN the tool, before the spool write and before the view/footer are computed. The spool, the inline view, and the footer's offsets all describe the same redacted bytes; the loop's own sanitize pass is a no-op on them. The on-disk copy can no longer keep credentials the model-visible copy had redacted. Mutation-verified: dropping the in-tool sanitize fails `should_sanitize_stream_before_persisting_spool` on the "raw credential persisted" assertion.
- **Owner-only permissions.** Spool dir `0700`, files `0600` via fchmod on the open handle (Unix). Windows has no mode-bit story — content protection there is the workspace's own ACLs (stated, not pretended away).
- **Outbound sweep exclusions** via `octos_core::is_shell_spool_path` (component-run match on `.octos/tmp/shell`, separator- and canonicalization-independent): `send_file` refuses spool paths before any allowlist logic (lexical AND canonical form), artifact-contract globbing (`resolve_glob_matches`) skips them, and the UI workspace/artifact panes skip `.octos` entirely (`should_skip_pane_dir`). The pane guard is `api`-gated, so it gets its own explicit CI filter step (`artifact_pane`) — an unmatched filter never runs (#2029).
- **No spooling on read-only or write-fenced profiles.** New `Sandbox::workspace_scratch_writable()` — macOS: `workspace_write && write_allow_globs.is_none()` (the same predicate its TMPDIR placement uses); bwrap/Landlock/AppContainer: `workspace_write` (a #1976 fence already degrades it to false at construction); Docker: `mount_mode != ReadOnly`. When false, the harness does not mutate from the host a workspace the sandbox told the shell is immutable — the historic in-tool cut applies instead. Mutation-verified: dropping the gate fails `should_not_spool_when_workspace_read_only`.
- **Symlink hardening.** All spool creates go through the #1976 `open_confined` component-wise `O_NOFOLLOW` walk (`tools/write_grant.rs`): a pre-planted `.octos/tmp/shell` symlink fails `ELOOP` instead of redirecting writes outside the workspace, and with no spool created, prune never runs either. Prune additionally re-checks the dir is a real directory and never deletes the file just written, so a dir crowded with future-stamped names cannot evict the spool whose path is being returned. Cross-process `create_new` collisions retry with the next sequence number. Non-Unix degrades to `open_confined`'s documented ancestor-scan fallback. The prune-time path-based deletes retain a narrow post-walk swap race — the same documented residual class as `open_confined`'s non-Unix mode; the create path is the race-free boundary.

## Budgeting (backstop never re-cuts)

All suffixes — exit code, sandbox-denial hint, #28a change receipt (unbounded-ish: up to 20 paths), warn nudge — are assembled FIRST; the spool decision and the head/tail view are computed from `limit − suffixes.len()`, on the sanitized final representation. So the assembled result fits the cap exactly, the loop backstop never fires on a spooled result, and the at-cap boundary case collapses into the same arithmetic. Pinned by `should_include_receipt_in_budget_when_receipt_present` (20 long-named files + >cap output ⇒ total ≤ 30000 with footer AND receipt present) and by two loop-path goldens driving the real dispatch→sanitize→backstop pipeline (`small_shell_output_reaches_the_model_byte_identical_through_the_loop`, `spooled_shell_output_reaches_the_model_with_footer_intact_through_the_loop`).

## Recovery advice that can't lie

`Tool::truncation_recovery` (the #2124 hook) as belt-and-braces if the backstop still fires: it names the spool only when (a) it was written by the LATEST invocation of that command — every new invocation forgets its registry entry up front, covering small reruns, timeouts, and spawn failures — and (b) the host file still exists (a sibling session sharing the workspace may have pruned it; registries are per-session, the directory is shared). Otherwise the honest fallback: re-run redirected to a file. Under Docker, footers and advice print the in-container path (`Sandbox::workspace_remap_root` → `/workspace/...`).

## Why `<workspace>/.octos/tmp/shell/` and not the OS temp dir

A spool the sandbox then blocks reading is worse than useless. The workspace is the one location every sandbox backend lets the model's follow-up `shell` call read:

- **bwrap** mounts `--tmpfs /tmp` and `--tmpfs /var/tmp` (`sandbox/bwrap.rs`), so a host `std::env::temp_dir()` spool would be invisible; the workspace is bound at its own path (reads work under ro-bind too — though read-only profiles no longer spool at all, see above).
- **macOS seatbelt** default profile allows `(allow file-read*)` globally; the restricted-read profile grants `(allow file-read* (subpath "<cwd>"))` (`sandbox/macos.rs`) — verified live with a `sandbox-exec` probe of the equivalent restricted profile: a spool under the workspace greps fine, a file in host `/tmp` gets `Operation not permitted`.
- **Landlock** grants the cwd.
- **Windows / NoSandbox**: no restriction; the footer prints the path via `Path::display()` (native separators), and no Unix-only APIs are used outside `#[cfg(unix)]` blocks.
- **Docker** with a mounted workspace sees it at `/workspace` — the footer/advice print that in-container path; the harness-side `grep`/`read_file` tools keep working on the host path either way.

`.octos/` inside the workspace is already established harness-write territory (session transcripts, `.octos/work`, `validator_outcomes.jsonl`). A `*` gitignore written inside the spool dir keeps the files out of `git status` and thereby out of the #28a change receipt.

## Lifecycle bound

Newest 16 spools per workspace directory; older `output-*.log` files are deleted on the next spool write — never the file just written (fixed-width millis+seq stamp makes lexicographic order = recency order). No daemon, no timers.

## Known pre-existing limitation (tracking only)

stdout/stderr interleaving was already lost before this change — `wait_with_output` captures the two streams separately and the result concatenates them (`stdout` + `--- stderr ---` + `stderr`); the spool preserves exactly that concatenated view, not the original interleaving.

## Not included (checked, separate code paths)

`coding_tools.rs` `BashTool` (`bash`) and `ExecCommandTool` (`exec_command`) do NOT share ShellTool's output path — they assemble output through their own local `truncate_output()` helper (`coding_tools.rs:65`, used at `:568` and `:2111`); they share only the 28a receipt module and the command-policy judge. Per scope, this PR covers `shell` alone.

## Tests (TDD, red first)

`tools/shell.rs::spool_tests` through the real `execute` path, plus loop-path goldens in `agent/loop_runner_tests.rs`, `octos-core` predicate tests, and one guard per sweep surface:

- `should_spool_full_output_with_footer_when_output_exceeds_cap` (spool byte-equals the full stream; mutation-checked: removing the spool write fails it)
- `should_report_line_ranges_matching_shown_content_when_spooled`
- `should_return_output_unchanged_and_no_spool_when_output_under_cap`
- `should_sanitize_stream_before_persisting_spool` (mutation-checked)
- `should_restrict_spool_permissions_when_persisting`
- `should_not_spool_when_workspace_read_only` (mutation-checked)
- `should_include_receipt_in_budget_when_receipt_present`
- `should_fall_back_to_generic_advice_when_spool_file_deleted`
- `should_forget_stale_spool_when_command_reruns_without_spooling`
- `should_keep_fresh_spool_when_future_stamped_files_crowd_the_dir`
- `should_refuse_spool_when_spool_dir_is_symlink`
- `should_emit_remapped_footer_path_when_sandbox_remaps_workspace`
- `should_name_spool_path_in_truncation_recovery_when_backstop_fires`
- `should_advise_rerun_with_redirect_when_no_spool_recorded_for_command`
- `should_prune_spool_dir_to_capacity_when_spools_accumulate`
- `should_use_byte_footer_when_single_line_exceeds_budget` / `should_keep_view_within_budget_when_line_data_spooled`
- `should_state_spool_contract_in_description`
- `small_shell_output_reaches_the_model_byte_identical_through_the_loop` / `spooled_shell_output_reaches_the_model_with_footer_intact_through_the_loop` (real loop)
- `should_match_spool_component_run_when_checking_shell_spool_paths` (octos-core)
- `should_refuse_send_when_path_is_shell_spool` (send_file, with a non-spool control)
- `should_exclude_shell_spool_files_from_artifact_globs` (workspace contract)
- `should_skip_octos_internal_dir_when_walking_panes` (UI panes, api-gated, new CI filter step)

## Verification

- `cargo test -p octos-agent --lib` — 2718 passed; 0 failed; 3 ignored
- `cargo test -p octos-core --lib` — 359 passed; 0 failed; 1 ignored
- `cargo test -p octos-cli --features api --lib artifact_pane` — 1 passed
- `cargo clippy -p octos-agent --all-targets -- -D warnings` — clean
- `cargo clippy -p octos-core --all-targets -- -D warnings` — clean
- `cargo clippy -p octos-cli --features api --all-targets -- -D warnings` — clean

Refs #1638

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -469,6 +469,13 @@ jobs:
       - name: serve wiring guards (octos-cli)
         run: cargo test -p octos-cli --features api commands::serve::tests
 
+      # The workspace/artifact pane exclusion list lives in the `api`-gated
+      # transport module (same reason as the §6 guard) — without an explicit
+      # filter this guard never runs (#2029). It pins `.octos` (sessions,
+      # #1638 shell output spools) out of the recursive pane walks.
+      - name: Artifact pane exclusions (octos-cli)
+        run: cargo test -p octos-cli --features api artifact_pane
+
       # The goal-completion verifier's ledger-evidence fold lives in the
       # `api`-gated goal_tool module, so the unfeatured runs above never compile
       # it. Guards the peer-goal convergence fix (#1989): a genuinely-complete
--- a/crates/octos-agent/src/agent/loop_runner_tests.rs
+++ b/crates/octos-agent/src/agent/loop_runner_tests.rs
@@ -6380,6 +6380,134 @@ async fn truncated_tool_output_reaches_the_model_with_its_recovery_advice() {
     );
 }
 
+/// Scripted provider: one `shell` call with the given command, then EndTurn.
+struct CallsShellThenEnds {
+    calls: AtomicUsize,
+    command: &'static str,
+}
+
+#[async_trait]
+impl LlmProvider for CallsShellThenEnds {
+    async fn chat(
+        &self,
+        _messages: &[Message],
+        _tools: &[octos_llm::ToolSpec],
+        _config: &octos_llm::ChatConfig,
+    ) -> Result<ChatResponse> {
+        let call = self.calls.fetch_add(1, AtomicOrdering::SeqCst);
+        Ok(if call == 0 {
+            ChatResponse {
+                content: None,
+                reasoning_content: None,
+                tool_calls: vec![ToolCall {
+                    id: "call_shell".to_string(),
+                    name: "shell".to_string(),
+                    arguments: serde_json::json!({ "command": self.command }),
+                    metadata: None,
+                }],
+                stop_reason: StopReason::ToolUse,
+                usage: LlmTokenUsage::default(),
+                provider_index: None,
+            }
+        } else {
+            ChatResponse {
+                content: Some("done".to_string()),
+                reasoning_content: None,
+                tool_calls: vec![],
+                stop_reason: StopReason::EndTurn,
+                usage: LlmTokenUsage::default(),
+                provider_index: None,
+            }
+        })
+    }
+
+    fn model_id(&self) -> &str {
+        "test-model"
+    }
+
+    fn provider_name(&self) -> &str {
+        "test-provider"
+    }
+}
+
+/// Drive the REAL loop path (dispatch → sanitize → backstop) for one shell
+/// command and return the tool message the model received.
+#[cfg(unix)]
+async fn shell_tool_message_through_loop(command: &'static str) -> (tempfile::TempDir, Message) {
+    let dir = tempfile::tempdir().unwrap();
+    let tools = ToolRegistry::with_builtins(dir.path());
+    let provider: Arc<dyn LlmProvider> = Arc::new(CallsShellThenEnds {
+        calls: AtomicUsize::new(0),
+        command,
+    });
+    let memory = Arc::new(EpisodeStore::open(dir.path().join("memory")).await.unwrap());
+    let agent = Agent::new(AgentId::new("shell-spool-loop"), provider, tools, memory);
+    let result = agent.process_message("go", &[], vec![]).await.unwrap();
+    let message = result
+        .messages
+        .iter()
+        .find(|m| m.role == MessageRole::Tool)
+        .expect("the turn must contain the shell tool result")
+        .clone();
+    (dir, message)
+}
+
+/// #1638 golden through the REAL loop: a small shell output must reach the
+/// model byte-identical — sanitize is a no-op on clean text and the backstop
+/// must not touch it. The in-tool unit golden proves the tool; this proves
+/// the sanitize+backstop interplay around it.
+#[cfg(unix)]
+#[tokio::test]
+async fn small_shell_output_reaches_the_model_byte_identical_through_the_loop() {
+    let (_dir, tool_message) = shell_tool_message_through_loop("echo hi").await;
+    assert_eq!(
+        tool_message.content, "hi\n\n\nExit code: 0",
+        "small shell output must arrive unchanged through dispatch + sanitize + backstop"
+    );
+}
+
+/// #1638 golden through the REAL loop: a spooled shell result must arrive
+/// with its footer intact — under the cap (so the loop backstop never fired)
+/// and without any backstop recovery advice appended.
+#[cfg(unix)]
+#[tokio::test]
+async fn spooled_shell_output_reaches_the_model_with_footer_intact_through_the_loop() {
+    let (dir, tool_message) = shell_tool_message_through_loop(
+        "awk 'BEGIN { for (i = 1; i <= 12000; i++) print \"line \" i }'",
+    )
+    .await;
+    let limit = octos_core::tool_output_limit("shell");
+    assert!(
+        tool_message.content.len() <= limit,
+        "the assembled result must fit the {limit}B cap, got {}",
+        tool_message.content.len()
+    );
+    assert!(
+        tool_message.content.contains("Full output: "),
+        "the spool footer must survive the loop: {}",
+        &tool_message.content[tool_message.content.len().saturating_sub(300)..]
+    );
+    // Backstop-fired evidence must be ABSENT: the tool sized its result so
+    // the loop never re-cut it and never appended recovery advice.
+    assert!(
+        !tool_message
+            .content
+            .contains("Re-run with output redirected"),
+        "backstop recovery advice implies the loop re-cut the result"
+    );
+    assert!(
+        !tool_message.content.contains("was saved to"),
+        "backstop recovery advice implies the loop re-cut the result"
+    );
+    // The spool named by the footer exists inside the workspace.
+    let marker = "Full output: ";
+    let start = tool_message.content.rfind(marker).unwrap() + marker.len();
+    let rest = &tool_message.content[start..];
+    let path = std::path::PathBuf::from(&rest[..rest.find(']').unwrap()]);
+    assert!(path.starts_with(dir.path()), "spool inside the workspace");
+    assert!(path.exists(), "footer must name a real file");
+}
+
 // ─────────────────────────────────────────────────────────────────────────
 // #27d (R4) — malformed tool-call feedback buffer
 // ─────────────────────────────────────────────────────────────────────────
--- a/crates/octos-agent/src/sandbox/bwrap.rs
+++ b/crates/octos-agent/src/sandbox/bwrap.rs
@@ -27,6 +27,13 @@ pub struct BwrapSandbox {
 }
 
 impl Sandbox for BwrapSandbox {
+    fn workspace_scratch_writable(&self) -> bool {
+        // A #1976 fence already degraded `workspace_write` to `false` at
+        // construction (`fence_degraded_workspace_write`), so this single
+        // flag covers both the read-only profile and the fenced one.
+        self.workspace_write
+    }
+
     fn supports_repo_git_write(&self) -> bool {
         // bwrap can `--bind <repo>/.git <repo>/.git` (rw) and read-binds the
         // system dirs, so a Host-granted worker gets both the `.git` write AND the
--- a/crates/octos-agent/src/sandbox/docker.rs
+++ b/crates/octos-agent/src/sandbox/docker.rs
@@ -38,6 +38,24 @@ impl Sandbox for DockerSandbox {
         true
     }
 
+    fn workspace_scratch_writable(&self) -> bool {
+        // `:ro` mounts promise the shell a read-only workspace (a #1976
+        // fence degrades ReadWrite to ReadOnly at construction —
+        // `fence_degraded_docker`), so the harness must not spool into it.
+        // `None` never mounts the workspace at all: nothing was promised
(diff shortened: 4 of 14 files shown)


ymote:
## Adversarial review — round 1 — DO-NOT-MERGE (6 findings)

The sandbox-readability research held (bwrap/macos/landlock citations verified). What failed:

1. **Security (MUST-FIX):** the spool persists the RAW stream (shell.rs:1357→:927) while credential sanitization runs later in the loop (execution.rs:2408, sanitize.rs:23/:90) — the on-disk copy keeps secrets the model-visible output redacts. Umask-default perms, no session-end cleanup, and the gitignore does not cover the real sweep paths: `send_file` sends any workspace file as-is (send_file.rs:253/:165), the UI artifact pane enumerates everything and its exclusion list omits `.octos` (ui_protocol_transport.rs:19823/:19983), contract globbing ignores gitignore (workspace_contract.rs:374).
2. **Host-side write under read-only (MUST-FIX):** `write_spool` is a direct host write (:1398) — the harness mutates a workspace the model was told is read-only (policy.rs:90) or fenced (#1976). Docker footer paths are host-absolute and unusable in-container (docker.rs:97).
3. **Budget order (MUST-FIX):** footer computed from raw offsets (:1030) before sanitization rewrites the display; only 2,048 bytes reserved (:882) while the #28a receipt appends up to 20 unbounded paths (:2703/:2752) — the loop re-cuts (execution.rs:2396), falsifying "backstop never fires".
4. **Stale recovery paths (MUST-FIX):** registry counts commands, disk counts files — repeats prune other commands' files (:933) while their registry entries survive; advice checks neither existence nor invocation (:1077).
5. **Symlink/prune spoofing (MUST-FIX):** all parent traversal is unprotected (:903); prune trusts lexicographic names (:937) — 16 future-stamped files evict the fresh spool, which is still returned Some(path) (:928). Component-wise O_NOFOLLOW exists in-repo to copy (#1976 fenced write).
6. **At-cap boundary + narrow golden (MUST-FIX, folds into 3).**

Refuted: same-millisecond collisions (:891), parallel-batch races (Exclusive, :1096), cross-session registry bleed (per-session registries). Tracking-only: stdout/stderr interleaving was already lost pre-change (:1353).

Round 2 in progress.

anantheparty (dismissed):
Reviewed — the spool registry is bounded (SPOOL_KEEP VecDeque, per-command dedup, mutex-guarded) so spool files cannot accumulate unbounded, and the elision marker now points the model at a grep-able artifact instead of destroying the middle of cargo/npm test output. 9 new tests cover the lifecycle. Approving — merging left to you per convention.

anantheparty (approved):
Adversarial pre-merge review of the CURRENT cumulative state (post force-push, cf762fd3) passed — no blocker.

Highlights: spool security is solid (dir 0700 / files 0600 asserted on real mode bits, writes via #1976 open_confined component-wise O_NOFOLLOW walk with planted-symlink ELOOP fallback test, sanitize-before-persist so the on-disk copy can't keep credentials the model-visible copy lost, all three outbound faces — send_file pre-allowlist, artifact globbing, UI pane walks — exclude .octos/tmp/shell). The ci.yml +7 is a pure addition inside test-octos-cli (explicit api-gated filter per the #2029 lesson — zero-match filters silently pass); check-arm untouched. 24 new tests, assertions real (mode bits, byte-exact spool comparisons, prune bound, remap paths).

Nits (non-blocking, for follow-up): the 30KB/50KB dual-constant in shell.rs deserves a clarifying comment; artifact glob exclusion is lexical-only (symlink indirection can pull spool content into artifacts — data hygiene, not secrets); future-timestamped spools can crowd the retention window. Approving — merging left to you per convention.

ymote:
## Adversarial review — round 2 — DO-NOT-MERGE

Round-1 fixes verified present, but each has a residual one layer down:

1. **Sanitize not idempotent (R1).** Full-stream sanitize before split/write is correct (shell.rs:1520/:1018), but overlapping prompt-guard matches break the no-op second pass (reverse replacement shifts the outer span, prompt_guard.rs:272/:277) — so the spool holds a once-sanitized still-detectable form while the loop's second pass changes the displayed head (execution.rs:2408), diverging persisted bytes from displayed + footer ranges.
2. **Symlink aliases bypass the sweeps (R2).** Component match is correct and send_file checks canonical — but artifact globbing checks only the lexical glob path (workspace_contract.rs:390/:402) and `glob` follows dir symlinks, so `reports -> .octos/tmp/shell` re-exposes the spool; both pane walkers follow metadata() through aliases (ui_protocol_transport.rs:19769/:19793/:19846). Skipping the literal `.octos` label doesn't close the alias.
3. **Docker MountMode::None spools to an unreachable host path (R3)** — worse, read-only and #1976-fenced configs legitimately keep None (sandbox/mod.rs:542, policy.rs:706), so spooling still mutates the host workspace under them. Docker should spool only for ReadWrite.
4. **Budget still not on the final sanitized representation (R4).** Only command output is sanitized (shell.rs:1520); receipts/suffixes are assembled after, budgeted unsanitized (:1566), appended (:1591). Non-spool fallback still emits 50,000 → loop cuts to 30,000 → appends unbudgeted recovery advice.
5. **Retention path-racy + unsynchronized (R5, + new).** prune lstat-then-path-unlink (shell.rs:1037/:1062) — ancestor swap after the check deletes external files; no workspace/cross-process lock, so another session's prune deletes a freshly-written path before return; dir chmod has a check/use race and ignores failure (:990) while open_confined makes parents 0755 first (write_grant.rs:384), so 0700 isn't fail-closed.

Round 3 pending — but this feature's safe surface may be smaller than the general form; see the linked scoping note.

ymote:
## Parking this — the investigation is the deliverable

Decision: **converting to draft and parking.** Spool-to-file has now cost two review rounds and 16 findings, and the readability investigation this round settled why: **every candidate location has a disqualifying drawback.**

| location | model can read it? | attack surface |
|---|---|---|
| workspace | yes, all backends | large — swept by artifact panes / globs / send_file; symlink aliases; read-only-fence mutation; cross-process prune races |
| OS temp | **NO on bwrap (tmpfs-shadows /tmp) or Docker (host temp never mounted)** — and harness read_file/grep are Workspace-scoped so can't reach it either | small |

The workspace was chosen precisely because the cwd is the one surface both the confined shell child AND the harness file tools can read on every backend (rw/ro bind under bwrap, /workspace under Docker, cwd-subpath grant under seatbelt/Landlock, InWorkspace for read_file/grep). Moving out of the swept tree costs readability on the flagship Linux container path, where the spool becomes write-only from the model's side. Live-probed on this host: macOS seatbelt reads `$TMPDIR` both ways (granted → `SECRET_FAILURE_LINE_42`, revoked → `Operation not permitted`); bwrap/Docker cannot by construction.

**What already shipped from this line of work and covers the common case:** #2124 (merged) — grep and read_file name their next call instead of dead-ending at `[N bytes omitted]`. The residual gap is only *full* shell output for a command whose failures land in the head/tail-cut middle.

**Revisit if** a real workload needs that — the tradeoff matrix above is the starting point, and Option 3 (temp-dir + a dedicated sandbox read-grant) is the only design that delivers readability everywhere, at the cost of re-expanding the sandbox contract.

Not closing outright — draft, so the hardened workspace implementation and this analysis stay recoverable.

Issue #2131, opened 2026-08-26
[octos] Compaction destroys the working set: evicted tool outputs are unrecoverable one-line stubs


heyong4725:
# [octos] Compaction destroys the working set: evicted tool outputs are unrecoverable one-line stubs

**Priority: P1.** (`crates/octos-agent/src/compaction.rs`, `compaction_tiered.rs`, context ledger)

## Problem

The tiered compactor replaces tool outputs with stubs like:

```
- Called read_file
  -> read_file: ok -  88│         for (int t = 0; t < T; t++) {
```

The full outputs still exist, content-addressed, in
`.octos/octos/context_ledgers/tool-output/sha256:*` — but the stub gives the model no handle
to them. Its only recovery is re-reading the file, which the compactor evicts again.

Qwen-llm.c session evidence: 66 `read_file` calls on `train_gpt2.c` (1,182 lines), ledger
generation 279 with 248 compaction checkpoints, live context pinned at ~18.7K tokens — and the
model never accumulated enough of the file to write any Rust. Pure thrash: the tokens the
compactor "saved" were spent many times over on re-reads.

Note: PR #2124 established exactly the right pattern for the *inline truncation* layer
(`Tool::truncation_recovery` names a concrete next call). This issue is the same idea one
layer up, at compaction/eviction.

## Proposed fix

1. **Recallable stubs**: evicted output stubs carry the ledger key and a concrete recall call:
   `[evicted: read_file train_gpt2.c:86-400 — recall("sha256:…")]`, plus a `recall` tool that
   re-materializes from the ledger (cheap: no re-execution, exact same bytes).
2. **Dedup before evicting**: N reads of the same file/range collapse to one entry keeping the
   newest content, not N stubs.
3. **Working-set pinning**: files written this session and the K most recently read files are
   evicted last. Evicting the source file mid-conversion is the pathological case observed.
4. **Budget-aware reads**: when a read would immediately blow the budget, return a hint
   (range suggestion) instead of accept-then-evict.

## Acceptance

Re-run the llm.c conversion benchmark at a deliberately small context: re-read ratio for
`train_gpt2.c` drops from 66 to ≤ 2× unique ranges; ledger generation count drops by an order
of magnitude; model reaches its first `write_file` without user intervention.

heyong4725:
**Design notes from implementation scoping (this session shipped the sibling fixes; #5 needs its own PR at the right altitude):**

[...]

PR #2193, opened 2026-08-31
feat(tools): flag-gated windowed read_file + partial-view write guard

ymote:
Third step of the #1638 sequence: #2124 gave truncated reads a named next call, #2126 measured whether forcing pages is even a good idea without changing behaviour. This PR builds the enforcement itself — **off by default, armed via `OCTOS_READ_WINDOW=1`** — together with the safety mechanisms an adversarial review (two rounds) showed must land in the same change. The second commit is a redesign of the first after that review found it fail-open on four axes; the findings and their dispositions are in the review-response comment below.

## Design

**Window (armed only).** `read_file` returns at most **2000 lines** and at most **48 KiB of formatted output** — whichever limit is hit first — as whole lines only, with a footer naming which limit fired, the range actually returned, the totals, and the exact next call:

```
[read_file window: showing lines 1-450 of 1500 — 49152-byte limit hit; file is 151500 bytes. Continue with offset: 451.]
[read_file window: showing lines 1-2000 of 3000 — 2000-line limit hit. Continue with offset: 2001.]
```

The #2131 refusal for oversized unbounded reads is subsumed on the armed path, and the internal blind 100KB cut is skipped there: the window IS the advising cut, sized so the execution loop's 50,000-byte backstop (#2124) can **never** fire on an armed read. A tripwire test pins `WINDOW_MAX_BYTES + FOOTER_RESERVE <= tool_output_limit("read_file")`, every armed return path carries a `debug_assert` against the loop cap, **no armed return interpolates caller-controlled text** (path spellings are unbounded — a `./`-chain spelling of 50KB is pinned under the cap by test), and refusal messages clamp the path they echo.

**Raw byte mode (new `byte_offset`/`byte_limit` parameters).** A single line larger than the window cannot be paged by line offset. The first draft followed pi's shell fallback (`sed | head -c`) — the review proved that self-defeating under OUR constraints: shell output is capped at 30,000 bytes (`tool_output_limit("shell")`), and the loop sanitizer redacts exactly what giant lines are made of (base64 data URIs, 64+ hex). So paging within a line is now in-tool: raw slices without a gutter, UTF-8-boundary-snapped (never splitting a character, never leaving a gap), footer naming the exact next `byte_offset`, no footer at EOF. The giant-line advice names the line, its size, and the byte offset where it starts:

```
[read_file window: line 2 is 60000 bytes — larger than the 49152-byte window, and lines cannot be
split across line-mode pages. Read it in raw byte mode with byte_offset: 12 (returns bytes without
line numbers; follow the byte_offset each footer names). Lines after it resume at offset: 3.]
```

Byte mode works unarmed too (the schema is static; an advertised parameter must always work) and bypasses the M8.4 cache in both directions — its view ranges are line ranges, so a byte request is never answered with a `[FILE_UNCHANGED]` stub and never stored as one.

**Parameter provenance.** 2000 lines is pi's field-tested default verbatim (`truncate.ts: DEFAULT_MAX_LINES`); the earlier 500-line/24KiB proposal was rejected as too aggressive (57.4% of repo files ≤500 lines). The byte half adapts pi's 50KB to two local facts pi does not have: our output carries a line-number gutter (the budget counts the FORMATTED bytes that enter context — what the loop cap counts), and 50×1024 = 51,200 does not fit under our 50,000-byte loop cap; 48 KiB plus the footer reserve provably does. The byte-paging mode and the write guard are the two places this deliberately does NOT follow pi: pi's giant-line shell fallback dies against our shell cap and sanitizer, and pi ships no write guard (verified: its `write.ts` has no read-first check) but also no prompt that mandates read→rebuild→overwrite — our slides prompt does, and it permits ONLY `read_file`/`write_file` for authoring.

## The overwrite guard: fail-closed, session-scoped, epoch-validated, taint-aware

`write_file` reconstructs a file from whatever the model saw; the slides flow does read → rebuild → `write_file` with the patch tools forbidden. The guard's rule, armed only:

> Overwriting an EXISTING file larger than the 48 KiB window requires a **COMPLETE** mark from **THIS session** at the **current (mtime, size) epoch**, **untainted**. No ledger entry ⇒ refuse and read first.

Absence being the refusing state makes restart safe by construction (post-restart the model must re-read — correct anyway), makes bounded eviction safe (512 entries, oldest first — eviction can only cost a re-read, never an unseen overwrite), and closes the giant-first-line hole (the advice branch records nothing, and nothing means NO). Files at or under the window keep today's blind-overwrite semantics — a file that size is returned whole by one unbounded read, so there is no partial-view illusion — but any recorded view of them is honoured (partial/tainted/stale refuses regardless of size).

The ledger (`tools/read_window.rs`, process-global — it must hold on every entry path including zero-context `Tool::execute`, where the optional `FileStateCache` is absent):

- **Keyed by `(parent_session_key, canonical path)`** — one session's COMPLETE never authorizes another session's overwrite; canonicalization keeps `read_file` and `write_file` agreeing across path aliases (`/var` vs `/private/var`).
- **Coverage in bytes** — line-mode reads convert to their raw byte spans, so line pages and byte pages stitch in one coordinate system as a contiguous-from-0 high-water mark; paging 1→EOF (by lines, bytes, or both) earns COMPLETE, so the refusal's own advice truthfully unlocks the write.
- **Epoch = (mtime, size), validated twice** — at read time for stitching (same-second replacement with a different size still resets) and again at WRITE time against a fresh stat (a file replaced after a complete read refuses as stale; a shrink between pages recovers via one fresh read). Same-mtime same-size replacement is not detectable without a content hash — neither is it by the M8.4 cache today.
- **Taint** — the recorded output is checked against `sanitize_tool_output` (the exact function the loop applies afterwards); a view the sanitizer would alter records TAINTED and the epoch never completes, because the model would write redaction placeholders over real content it never received.
- **A successful unformatted write re-records COMPLETE at the post-write epoch** (improvement over the redesign as proposed, which forgot the entry — locking the model out of a big file it authored one call ago). A post-write formatter rewrite forgets instead: the on-disk bytes are no longer what the model supplied.

Refusals are typed (`[PARTIAL_VIEW_OVERWRITE]`) with four variants — unread / partial / stale / tainted — and the advice is **tool-agnostic and read_file-only** ("page through it with read_file (offset/limit, or byte_offset…) until you reach the end, then retry"): the first draft named `edit_file`/`apply_patch`, which the slides prompt forbids — advice the calling context cannot follow is a deadlock, not an escape hatch.

**File-state cache (hazard b, unchanged from round 1):** the cache records the view RETURNED, not requested — a clamped read stores its actual window, so an unbounded request can never hit a windowed entry and claim `[FILE_UNCHANGED] (full file cached)` against content the model was never shown.

**Test arming is per-instance** (`with_window_enforcement`, `cfg(test)`), not the probe's process-global pattern: arming *changes* output, so a global switch would leak into parallel unarmed tests; likewise the restart simulation clears one session's entries, never the whole ledger. `set_var` is `unsafe` under edition 2024 and this workspace denies unsafe. Every test asserts on its own files and sessions — never process-global aggregates (#2077/#2126).

## What stays unarmed-identical

Pinned by `should_keep_unarmed_outputs_byte_identical_to_pre_change_goldens` — exact `(success, length, fnv1a)` triples captured on the pre-change tree for: a small unbounded read, an in-budget explicit range, the #2131 refusal, the internal 100KB cut, and a >2000-line unbounded FULL read. Armed reads of anything that fits the window are byte-identical too; the unarmed write path is pinned dormant; an unarmed read records nothing, so arming later never trusts evidence gathered while off. The new `byte_offset` mode is new surface, not a change to any existing call shape.

## Verification

```
cargo test -p octos-agent --lib                          => ok. 2735 passed; 0 failed; 3 ignored
cargo fmt --all -- --check                               => clean
```

37 tests across the three modules drive the real `execute`/`execute_with_context` paths, RED-first (22 watched failing before the redesign's implementation). Regression families from the review: giant-first-line refuses then completes via byte paging; long-path invariant (read advice and write refusal both ≤ the loop cap); redaction ⇒ TAINTED ⇒ refuse with the file left intact; cross-session refuses with same-session positive control; post-restart refuses until re-read; (mtime, size) replacement refuses; shrink-between-pages recovers via a fresh read; write-then-write stays unlocked. **Mutation-proofed:** disabling the fail-closed Unknown⇒refuse arm fails 6 tests (the whole family); disabling the sanitize-taint check fails the redaction test with the secret-bearing file overwritten; the round-1 mutations (write-guard removal, cache-keying reversion) were also verified killed.

## Tracking (not fixed here)

- **#2126 probe double-count when both flags are armed:** the probe's overwrite hook runs before this guard, so a REFUSED attempt still increments `partial_read_then_overwrite`. Left as is deliberately — the probe measures the dangerous *intent* rate (what would have been destroyed without a guard), and reordering would under-report exactly that; noted here so nobody reads armed-probe numbers as damage.
- **#2197 (filed):** pre-existing unarmed 100KB-cut cache lie — the blind cut can truncate output while the cache records the read as a complete view.
- Pre-existing oddity, untouched: reading an empty file errors with `Start line 1 is beyond file length (0 lines)`.

Refs #1638

---

## Round 3 (codex DO-NOT-MERGE): flag-gating made real, plus TOCTOU / keyless / taint

Round 2's "zero blast radius when off" was false at the wire, and three more holes were found. All fixed; the review-response comment maps each finding to its disposition.

**R6 — the flag was not default-off at the wire (the reframe).** read_file's description and schema, and write_file's description, changed unconditionally, and byte mode executed unarmed — so the ToolSpec serialized into the prompt-cache prefix changed for **every** session, armed or not. Decision, justified: **byte-paging is part of the armed feature** (it exists only to serve the windowed giant-line case — no consumer when the window is off). So description **and** schema **and** byte-mode execution are now all conditional on the arm; unarmed byte params are rejected, never silently ignored. New golden tests pin the unarmed ToolSpec JSON of both tools byte-for-byte against origin/main. Not split into a separate PR because an always-on byte mode would ship a permanent capability whose only user (the window) is still flag-gated.

**R1 — authorize-on-stat, truncate-on-independent-open TOCTOU.** The authorized overwrite now binds to the exact opened inode: `write_no_follow_checked` opens without `O_TRUNC`, `fstat`s the descriptor, and truncates + writes only when its `(mtime, size)` still equals the epoch that authorized the write; a replacement in the authorize→truncate window is refused, not clobbered. `note_full_write` records COMPLETE only when the on-disk size equals the bytes written.

**R4 — empty session keys collapsed together, reachable in production.** A missing key (agents default to `None`; FFI and fleet workers; plain CLI chat without `--goals`) mapped to `""` in both tools, so one keyless task's COMPLETE could authorize another's overwrite. Now keyless tasks record **nothing** (single enforcement point in the ledger) and fail closed for over-window overwrites, with a **distinct** "no session identity" refusal — paging can't help a keyless task, so it must not be advised to page.

**R5 — the invariant was `debug_assert` only.** All three loop-cap checks are now real runtime clamps (release builds drop `debug_assert`); armed returns interpolate no unbounded caller input and clamp the echoed path. The armed write description names no patch tools.

**R2 — TAINTED was a silent permanent deadlock.** A redaction-bearing file gets sticky-TAINTED and the loop re-redacts every raw byte page, so paging can never unlock it. It now has a **distinct** `[REDACTED_VIEW_OVERWRITE]` refusal that says the flag is incompatible with rewriting that file and to tell the user — it never advises paging.

### Known limitation operators must know before arming

**Armed mode is incompatible with whole-file rewrites of redaction-bearing files.** When windowed reads are armed, a file containing content the harness redacts from tool output (base64 data URIs, 64+‑char hex, credential patterns — see `sanitize.rs`) can never be delivered to the model whole, so `write_file` will refuse to reconstruct-and-overwrite it (typed `[REDACTED_VIEW_OVERWRITE]`). This directly affects the **slides** flow if a `script.js` embeds a data-URI image and the model tries to rewrite the whole file: it must make a narrower change, or the operator must not arm the flag for that workload. This is inherent to combining "never overwrite from a view the model didn't fully receive" with "the harness redacts some content" — not a bug, but a real operational constraint.

**Keyless armed sessions cannot whole-file-overwrite existing >48KiB files.** A session with no identity (CLI chat without `--goals`, FFI, fleet) can create files and overwrite small ones, but not reconstruct-and-overwrite a large existing file — it has no way to establish that it read the whole thing. Give the session an identity (e.g. `--goals`) for such flows.

### Round-3 tracking (not fixed, per scope)

- #2126 probe records overwrite **intent** before the guard decides, so an armed refusal still increments its counter — deliberate (the probe measures intent, not damage); left one-line-reorderable but out of scope.
- #2197 (filed round 2): pre-existing unarmed 100KB-cut cache lie.
- Empty-file `Start line 1 is beyond file length (0 lines)` oddity — pre-existing.

### DIFF
--- a/crates/octos-agent/src/tools/mod.rs
+++ b/crates/octos-agent/src/tools/mod.rs
@@ -50,10 +50,23 @@ use octos_core::{PathClassification, SessionScope};
 #[derive(Debug, Clone)]
 pub struct ToolInputError(String);
 
+/// Upper bound on a [`ToolInputError`]'s model-facing message. The message can
+/// embed caller-controlled content (unknown parameter names), so it is bounded
+/// at construction — well under every tool's output limit (`tool_output_limit`
+/// min is 20_000) — so an armed tool's `Err` can never exceed the cap and be
+/// mangled by the execution loop's blind head/tail cut (#2193 R4). Real
+/// validation messages are a few hundred bytes.
+pub(crate) const TOOL_INPUT_ERROR_MAX_BYTES: usize = 4096;
+
 impl ToolInputError {
-    /// Build an input-validation error from a model-facing message.
+    /// Build an input-validation error from a model-facing message, bounded to
+    /// [`TOOL_INPUT_ERROR_MAX_BYTES`] so caller-supplied content (e.g. a
+    /// pathological unknown-parameter name) cannot make the error exceed the
+    /// tool-output cap.
     pub fn new(message: impl Into<String>) -> Self {
-        Self(message.into())
+        let mut message = message.into();
+        octos_core::truncate_utf8(&mut message, TOOL_INPUT_ERROR_MAX_BYTES, "…[truncated]");
+        Self(message)
     }
 }
 
--- a/crates/octos-agent/src/tools/mod.rs
+++ b/crates/octos-agent/src/tools/mod.rs
@@ -1349,6 +1349,95 @@ pub async fn read_no_follow(path: &Path) -> std::io::Result<String> {
     .unwrap_or_else(|e| Err(std::io::Error::other(e)))
 }
 
+/// Open a file read-only, atomically rejecting symlinks (O_NOFOLLOW on Unix).
+///
+/// SECURITY: the flags here MUST match [`read_no_follow`]'s open exactly.
+#[cfg(unix)]
+fn open_no_follow_ro(path: &Path) -> std::io::Result<std::fs::File> {
+    use std::os::unix::fs::OpenOptionsExt;
+    std::fs::OpenOptions::new()
+        .read(true)
+        .custom_flags(libc::O_NOFOLLOW)
+        .open(path)
+}
+
+#[cfg(not(unix))]
+fn open_no_follow_ro(path: &Path) -> std::io::Result<std::fs::File> {
+    if path.symlink_metadata().is_ok_and(|m| m.is_symlink()) {
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::PermissionDenied,
+            "symlink rejected",
+        ));
+    }
+    std::fs::OpenOptions::new().read(true).open(path)
+}
+
+/// The descriptor-derived facts an armed windowed read needs beyond the bytes.
+pub(crate) struct ReadMeta {
+    /// [`crate::tools::read_window::ViewEpoch`] taken from the SAME descriptor
+    /// the bytes came from — not a separate path stat — so it describes the
+    /// exact inode whose bytes were shown (#2193 R4, read-side TOCTOU).
+    /// `None` only if the descriptor's metadata was unavailable, which never
+    /// authorizes a write.
+    pub epoch: Option<crate::tools::read_window::ViewEpoch>,
+    /// The returned content is a DECODE of the on-disk bytes (PDF text
+    /// extraction), not the bytes themselves — so a whole-file rewrite
+    /// reconstructed from it can never be faithful, and the view must never be
+    /// allowed to reach COMPLETE (#2193 R4, PDF false-COMPLETE).
+    pub transformed: bool,
+}
+
+/// Like [`read_no_follow`] but also returns the [`ReadMeta`] the armed
+/// windowed-read ledger needs. Kept as a distinct entry point so the many
+/// non-armed `read_no_follow` callers pay nothing for the extra `fstat`.
+pub(crate) async fn read_no_follow_with_meta(path: &Path) -> std::io::Result<(String, ReadMeta)> {
+    let path = path.to_owned();
+    tokio::task::spawn_blocking(move || {
+        use std::io::{Read, Seek, SeekFrom};
+        let mut file = open_no_follow_ro(&path)?;
+        // fstat the DESCRIPTOR (not a re-resolution of the path): binds the
+        // epoch to the exact inode these bytes are read from.
+        let epoch = file
+            .metadata()
+            .ok()
+            .and_then(|m| crate::tools::read_window::ViewEpoch::from_metadata(&m));
+        // Same PDF handling as read_no_follow, reading the rest from the SAME
+        // O_NOFOLLOW descriptor (seek back to 0), never by re-opening the path.
+        let mut magic = [0u8; 5];
+        let n = file.read(&mut magic)?;
+        file.seek(SeekFrom::Start(0))?;
+        if n >= 5 && &magic == b"%PDF-" {
+            let mut bytes = Vec::new();
+            file.read_to_end(&mut bytes)?;
+            match pdf_extract::extract_text_from_mem(&bytes) {
+                Ok(text) => Ok((
+                    text,
+                    ReadMeta {
+                        epoch,
+                        transformed: true,
+                    },
+                )),
+                Err(err) => Err(std::io::Error::new(
+                    std::io::ErrorKind::InvalidData,
+                    format!("pdf extraction failed: {err}"),
+                )),
+            }
+        } else {
+            let mut content = String::with_capacity(n);
+            file.read_to_string(&mut content)?;
+            Ok((
+                content,
+                ReadMeta {
+                    epoch,
+                    transformed: false,
+                },
+            ))
+        }
+    })
+    .await
+    .unwrap_or_else(|e| Err(std::io::Error::other(e)))
+}
+
 /// Write content to a file, atomically rejecting symlinks via O_NOFOLLOW on Unix.
 ///
 /// Eliminates the TOCTOU race between `reject_symlink` and `tokio::fs::write`.
@@ -1387,8 +1476,8 @@ pub(crate) enum CheckedWrite {
     /// The opened descriptor matched the authorizing epoch and was truncated
     /// and rewritten.
     Written,
-    /// The opened descriptor's `(mtime, size)` no longer matched what
-    /// authorized the write — nothing was written.
+    /// The opened descriptor's epoch (mtime/size/ctime/inode) no longer
+    /// matched what authorized the write — nothing was written.
     EpochChanged {
         /// The descriptor's actual epoch at open time.
         found: crate::tools::read_window::ViewEpoch,
@@ -1435,12 +1524,12 @@ pub(crate) async fn write_no_follow_checked(
             }
         }
         let mut file = opts.open(&path)?;
-        // fstat the DESCRIPTOR itself (not a re-resolution of the path).
+        // fstat the DESCRIPTOR itself (not a re-resolution of the path). The
+        // epoch binds mtime/size/ctime/inode, so a same-size replacement that
+        // forged mtime is caught by the ctime (or inode) mismatch.
         let meta = file.metadata()?;
-        let found = crate::tools::read_window::ViewEpoch {
-            mtime: meta.modified()?,
-            size: meta.len(),
-        };
+        let found = crate::tools::read_window::ViewEpoch::from_metadata(&meta)
+            .ok_or_else(|| std::io::Error::other("descriptor metadata unavailable"))?;
         if found != expected {
             return Ok(CheckedWrite::EpochChanged { found });
         }
@@ -1503,16 +1592,71 @@ mod nofollow_tests {
     // external replacement in the narrow authorize→truncate window is caught
     // on the exact opened object, not a re-resolved path.
     #[tokio::test]
+    async fn checked_write_refuses_a_same_size_same_mtime_content_swap() {
+        // #2193 R4 (codex H2): the (mtime,size)-only epoch authorized an
+        // overwrite of content the model never saw when a replacement kept the
+        // size and FORGED the mtime back. The epoch now also binds ctime, and
+        // forging mtime with `set_modified` is itself what bumps ctime — so the
+        // swap is caught. Under the old epoch this returned `Written` (RED).
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("f.txt");
+        std::fs::write(&path, b"AAAAAAAAAA").unwrap(); // 10 bytes
+        let meta = std::fs::metadata(&path).unwrap();
+        let mtime = meta.modified().unwrap();
+        let authorized = crate::tools::read_window::ViewEpoch::from_metadata(&meta).unwrap();
+
+        std::fs::write(&path, b"BBBBBBBBBB").unwrap(); // same size, new content
+        std::fs::OpenOptions::new()
+            .write(true)
+            .open(&path)
+            .unwrap()
+            .set_modified(mtime)
+            .unwrap();
+        let now = std::fs::metadata(&path).unwrap();
+        assert_eq!(now.len(), authorized.size, "size still matches");
+        assert_eq!(now.modified().unwrap(), mtime, "mtime forged back to match");
+
+        let result = write_no_follow_checked(&path, b"CCCCCCCCCC", authorized)
+            .await
+            .unwrap();
+        assert!(
+            matches!(result, CheckedWrite::EpochChanged { .. }),
+            "same-size, same-mtime content swap must be refused (ctime binding)",
+        );
+        assert_eq!(
+            std::fs::read(&path).unwrap(),
+            b"BBBBBBBBBB",
+            "the swapped-in bytes must be intact — never blind-clobbered",
+        );
+    }
+
+    #[tokio::test]
+    async fn read_with_meta_reports_untransformed_epoch_from_the_read_fd() {
+        // #2193 R4 (codex H2b/H6): the armed ledger needs the epoch taken from
+        // the READ descriptor (not a separate path stat), and transformed=false
+        // for ordinary text.
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("hello.txt");
+        std::fs::write(&path, b"hello world\n").unwrap();
+        let (content, meta) = read_no_follow_with_meta(&path).await.unwrap();
+        assert_eq!(content, "hello world\n");
+        assert!(!meta.transformed, "plain text is not a transform");
+        let epoch = meta.epoch.expect("descriptor epoch");
+        assert_eq!(epoch.size, 12);
+        let independent =
+            crate::tools::read_window::ViewEpoch::from_metadata(&std::fs::metadata(&path).unwrap())
+                .unwrap();
+        assert_eq!(epoch, independent, "epoch describes the exact inode read");
+    }
+
+    #[tokio::test]
     async fn write_no_follow_checked_writes_when_epoch_matches() {
         use crate::tools::read_window::ViewEpoch;
         let dir = tempfile::TempDir::new().unwrap();
         let file = dir.path().join("f.txt");
         std::fs::write(&file, "original").unwrap();
         let meta = std::fs::metadata(&file).unwrap();
-        let epoch = ViewEpoch {
-            mtime: meta.modified().unwrap(),
-            size: meta.len(),
-        };
+        let epoch = ViewEpoch::from_metadata(&meta).unwrap();
 
         let outcome = write_no_follow_checked(&file, b"rebuilt", epoch)
             .await
@@ -1531,10 +1675,7 @@ mod nofollow_tests {
         let file = dir.path().join("f.txt");
         std::fs::write(&file, "original small").unwrap();
         let meta = std::fs::metadata(&file).unwrap();
-        let authorized = ViewEpoch {
-            mtime: meta.modified().unwrap(),
-            size: meta.len(),
-        };
+        let authorized = ViewEpoch::from_metadata(&meta).unwrap();
 
         // The file is replaced (new size ⇒ new epoch) in the window between
         // the guard authorizing against `authorized` and this write opening
--- a/crates/octos-agent/src/tools/mod.rs
+++ b/crates/octos-agent/src/tools/mod.rs
@@ -767,6 +767,7 @@ pub trait Tool: Send + Sync {
 // Tool registry (extracted to its own module)
 /// Observe-only probe for the read-paging decision (changes no behaviour).
 pub(crate) mod read_paging_probe;
+pub(crate) mod read_window;
 mod registry;
 pub use registry::ToolRegistry;
 
--- a/crates/octos-agent/src/tools/mod.rs
+++ b/crates/octos-agent/src/tools/mod.rs
@@ -1604,6 +1604,9 @@ mod nofollow_tests {
     // (mtime, size) no longer matches what authorized the write — so an
     // external replacement in the narrow authorize→truncate window is caught
     // on the exact opened object, not a re-resolved path.
+    // Unix-only: off-Unix the epoch has no ctime, so the documented weaker
+    // (mtime,size) fallback cannot detect a same-size/same-mtime swap.
+    #[cfg(unix)]
     #[tokio::test]
     async fn checked_write_refuses_a_same_size_same_mtime_content_swap() {
         // #2193 R4 (codex H2): the (mtime,size)-only epoch authorized an
--- a/crates/octos-agent/src/tools/mod.rs
+++ b/crates/octos-agent/src/tools/mod.rs
@@ -1381,6 +1381,78 @@ pub async fn write_no_follow(path: &Path, content: &[u8]) -> std::io::Result<()>
     .unwrap_or_else(|e| Err(std::io::Error::other(e)))
 }
 
+/// Outcome of an epoch-bound overwrite ([`write_no_follow_checked`]).
+#[derive(Debug)]
+pub(crate) enum CheckedWrite {
+    /// The opened descriptor matched the authorizing epoch and was truncated
+    /// and rewritten.
+    Written,
+    /// The opened descriptor's `(mtime, size)` no longer matched what
+    /// authorized the write — nothing was written.
+    EpochChanged {
+        /// The descriptor's actual epoch at open time.
+        found: crate::tools::read_window::ViewEpoch,
+    },
+}
+
+/// Overwrite an EXISTING file, but only if the descriptor we open still
+/// matches `expected` (#1638 R1 — TOCTOU).
+///
+/// The armed write guard authorizes an overwrite against a *stat of the path*.
+/// A plain `write_no_follow` then opens the path AGAIN and truncates whatever
+/// it resolves to — so an external replacement in that narrow authorize→open
+/// window is silently clobbered. This binds the authorization to the exact
+/// opened inode: it opens WITHOUT `O_TRUNC`, `fstat`s the descriptor, and only
+/// truncates + writes when that descriptor's `(mtime, size)` still equals the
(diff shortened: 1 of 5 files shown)


anantheparty (dismissed):
Reviewed — flag-gated off by default (env-armed in production, builder-injected in tests), the write guard closes the partial-view overwrite hazard, and 22 new tests cover the window lifecycle. The #1638 sequence (named next call → measure → enforce) is a model of how to land behavioral changes safely. Approving — merging left to you per convention.

ymote:
## Adversarial review — round 1 — DO-NOT-MERGE

Findings across five hypotheses; two gut the guard's purpose:

- **Fail-open on a giant first line**: the giant-line branch returns before `record_view` (read_file.rs:417/:434/:526), so a one-line minified script gets NO ledger entry and the write guard (write_file.rs:222) allows overwriting a file the model saw zero bytes of. The round-1 test masked it by putting a short line first (read_file.rs:1706).
- **Armed, the slides flow deadlocks**: slides_default.txt permits only read_file/write_file (:21/:28/:30) while the refusal advises edit_file/diff_edit/apply_patch (write_file.rs:234); a later >48KiB line leaves a coverage gap nothing can stitch. The shell fallback is self-defeating — `head -c 49152` cannot survive shell's own 30,000-byte cap (utils.rs:88), and reconstructed data URIs are redacted by sanitize.rs anyway. (The no-byte-paging constraint came from the coordinator instruction, not the agent — reversed for round 2.)
- **Coverage recorded pre-sanitization** (read_file.rs:523 vs execution.rs:2408): COMPLETE can be marked for bytes the model never received.
- **Ledger identity wrong on four axes**: path-only key (no session — cross-session COMPLETE lets another session overwrite unseen), restart clears, unbounded, mtime-only with no write-time metadata check (stale refusals after shrink; false COMPLETE on same-second replacement).
- **50,000-byte invariant bypassed** by the giant-line branch (unbounded path interpolation; `./`-padded paths resolve but blow the budget); POSIX-only quoted advice = injection surface.

Round 2 is a redesign: fail-closed default (existing file > window requires a COMPLETE mark from THIS session; absence refuses), (session, path) keying with mtime+size validation at write time, sanitize-aware coverage, real in-tool byte paging, tool-agnostic refusal text.

Tracking: unarmed 100KB-cut cache lie is pre-existing (read_file.rs:489/:517); #2126 probe counts refused attempts as overwrites when both flags are armed.

ymote:
## Review response — codex round 1 (DO-NOT-MERGE): all findings verified against the code, none refuted; redesign landed in 8d992930a

Each finding was checked against the cited evidence before adopting the proposed redesign; the redesign itself was attacked first and adopted with two modifications (noted under H2 and at the end).

**H1 — slides deadlock + giant-first-line fails open. Confirmed, fixed.** `slides_default.txt` permits only `read_file`/`write_file` for authoring and forbids `edit_file` outright, so the old refusal's patch-tool advice was unfollowable. The refusal advice is now tool-agnostic and read_file-only, and the round-trip test asserts the advice does NOT name `edit_file`/`apply_patch` and IS completable with read_file alone. The giant-first-line hole (advice branch records nothing → old guard saw no partial → allowed a blind overwrite of a file the model saw nothing of) is closed by inverting the default: no ledger entry for an over-window file now REFUSES (`should_refuse_overwriting_a_big_never_read_file_when_armed`, `should_complete_via_byte_paging_and_allow_the_write` — the latter is the exact one-line-minified-script scenario, end to end: advise → refuse → byte-page → allow). The old test's short-first-line shape that masked this is replaced by a true giant-FIRST-line test. The mid-file coverage gap is closed by byte-mode reads counting toward the same byte-coverage ledger.

**H1b — coverage recorded pre-sanitization. Confirmed, fixed.** The ledger now checks the returned output against `sanitize_tool_output` — the same function the loop applies at execution.rs — and records TAINTED when it would alter the bytes; a tainted epoch never completes, and the write refuses with a message saying the content was redacted from the model's view. `should_refuse_when_redacted_content_never_reached_the_model` pins it (full single-window coverage of a 64-hex-bearing file still refuses; the file survives intact), and removing the taint check makes that test fail with the secret file overwritten (mutation verified).

**H2 — ledger identity wrong on four axes. Confirmed, fixed by the inverted rule + identity redesign.** Keys are now `(parent_session_key, canonical path)` (`should_refuse_across_sessions_even_after_a_complete_read`, with a same-session positive control). Restart: absence refuses, so an empty post-restart ledger is safe by construction (`should_refuse_after_a_restart_until_reread`, simulated per-session — a whole-ledger clear from one test would wipe parallel tests' entries, the same blast-radius problem as global arming). Growth: bounded at 512 entries, oldest evicted — safe because absence refuses (unit-tested). Epoch is `(mtime, size)` and is re-validated against a fresh stat at write time (`should_refuse_when_the_file_changed_on_disk_after_a_complete_read` uses a same-second different-size replacement), which also covers **H6**: a shrink between pages refuses and then recovers via one fresh read of the new generation (`should_recover_after_a_shrink_between_pages`). Same-mtime same-size replacement remains undetectable without a content hash — stated in the module docs; the M8.4 cache has the same bound today.

**H4 — giant-line branch bypassed the 50,000-byte invariant; quote injection; POSIX-only. Confirmed, fixed by deletion.** The advice no longer interpolates the path (or any caller-controlled text) at all — the model knows the path from its own call — so there is nothing to inject and nothing unbounded; the branch carries its own `debug_assert` and `should_keep_every_armed_return_under_the_loop_cap_for_a_pathological_path` pins a 50KB `./`-chain spelling under the cap. The write-side refusals, which do echo the path, clamp it (`should_clamp_pathological_paths_in_the_refusal`). The shell command is gone entirely, which also removes the POSIX-only advice.

**H5 — shell fallback self-defeating. Confirmed (the constraint that flipped the byte-resume veto).** `tool_output_limit("shell")` = 30,000, so the advised `head -c 49152` could never arrive intact, and the sanitizer would redact what did. Replaced by the in-tool `byte_offset`/`byte_limit` raw mode: boundary-snapped slices, footer naming the exact next `byte_offset`, EOF signalled by footer absence, counts toward coverage, wired through the input schema with mutual-exclusion validation against the line parameters.

**Redesign modifications (attacked before adopting):**
1. *"No entry ⇒ refuse when size exceeds the window"* is gated on the stat byte-size only — deliberately sufficient: a file ≤48KiB can only yield a partial VIEW via an actual read (the 2000-line limit), and an actual read leaves a PARTIAL entry, which refuses on its own. Blind overwrites of never-read ≤window files keep today's semantics.
2. *"Absence = safe"* interacts badly with the model's own writes: forgetting the entry after a successful write would refuse the very next overwrite of a big file the model just authored. A successful unformatted write therefore re-records COMPLETE at the post-write epoch (`should_allow_a_second_overwrite_of_a_file_the_model_just_wrote`); a formatter rewrite forgets instead, since the on-disk bytes are then not what the model supplied.

**Tracking-only, per instruction:** probe-vs-guard ordering left as is (a refused attempt still increments the probe's `partial_read_then_overwrite` — the probe measures intent, and recording only un-guarded damage would under-report exactly the number it exists to measure; noted in the PR body so armed-probe numbers aren't read as damage). The pre-existing unarmed 100KB-cut cache lie is filed as #2197. The empty-file `Start line 1 is beyond file length (0 lines)` oddity is pre-existing and untouched.

Gates after the redesign (sequential): `cargo test -p octos-agent --lib` → `ok. 2735 passed; 0 failed; 3 ignored`; `cargo clippy -p octos-agent --all-targets -- -D warnings` → clean; `cargo fmt --all -- --check` → clean. RED-first throughout: 22 new/updated tests watched failing before the implementation; both required mutations verified killed (fail-closed default: 6 tests fail; taint check: redaction test fails).

anantheparty (dismissed):
Adversarial pre-merge review of the redesign passed — all four fail-open axes from the round-1 self-review are closed fail-closed with regression tests: giant-first-line walks in-tool byte-mode and advises without ledgering (absence → refuse); restart safety is by construction (in-process ledger + absence-refuses); bounded 512-entry eviction degrades to Unknown → refuse (test-pinned); cross-session isolation via (parent_session_key, canonical path) keying + epoch (mtime,size) write-time re-verification. Write guard four conditions (COMPLETE / current-epoch / untainted / THIS-session) all enforced, allow surface collapsed to the single epoch-matched arm. Red lines: flag default-off (env, test-pinned), ci.yml untouched, non-windowed path golden-identical. Five minor nits (notably the eviction-safety overstatement at read_window.rs:69-71 — the fallback is the baseline behavior) go to follow-up. Approving — merging left to you per convention.

ymote:
## Adversarial review — round 2 (post-redesign) — DO-NOT-MERGE

The fail-closed redesign held on the hard parts: byte-mode stitching is gapless (start snaps backward, footer continues at end_b, line spans include the newline), byte mode neither consults nor pollutes the line cache, shrink-after-partial refuses, new-file creation passes. Six items remain.

**R6 — the flag is not actually default-off at the wire.** read_file's description changed unconditionally and the schema ALWAYS gains byte_offset/byte_limit (read_file.rs:125/:136); byte mode executes unarmed (:362); write_file's description also changed unconditionally. These populate every LLM ToolSpec (registry.rs:732) → prompt-cache prefix changes for every session, armed or not. The unarmed goldens covered execution output, not tool-spec bytes.

**R1 — write-guard TOCTOU.** Authorizes on exists()+metadata (:227/:238), truncates through a later independent open (:338/:365), never fstat-validates the opened fd (mod.rs:1360). External replace/grow between the two is overwritten; post-write re-record accepts the new stat without checking epoch.size == written_bytes (:325).

**R4 — empty session keys collapse, reachable in production.** Missing key ⇒ `""` (read_file.rs:417, write_file.rs:229) — and agents default None (agent/mod.rs:670), FFI (octos-ffi:538), fleet workers (:281), and plain `chat` without --goals (chat.rs:1556) all run keyless. One zero-key task's COMPLETE authorizes another's overwrite.

**R5 — invariant is debug_assert only** (absent in release); armed write echoes unbounded input.path (:434); write_file's description still names edit_file/apply_patch (:101) which armed sessions forbid.

**R2 — TAINTED is a silent permanent deadlock.** Data-URI script → sticky TAINTED (read_window.rs:269) + the loop redacts every byte page anyway (execution.rs:2408) ⇒ paging can never unlock it, but the refusal is the generic 'Make a narrower change' (:257) with no signal paging is futile. Needs a distinct typed reason + a documented armed-mode limitation.

Tracking-only: sub-scalar byte_limit returns ≤3 extra bytes (:396); formatter-failure forgets COMPLETE on exact content (:423, safe direction); #2126 records intent before refusal.

Round 3 in progress; R6 may split byte-paging into its own PR.

ymote:
## Review response — codex round 3 (DO-NOT-MERGE): all six items addressed in 40b488428

Findings verified against the code first; each maps to a fix below. The two architectural ones (R6, R4) drove the design decisions the review asked me to make and justify.

**R6 — flag not default-off at the wire. Confirmed; this reframed the PR.** The ToolSpec (name/description/input_schema) is serialized into the prompt-cache prefix in `registry.rs specs()`, and I had changed read_file's description+schema and write_file's description unconditionally, with byte mode executing unarmed — so I was busting the prefix for every session, armed or not. **Decision: byte-paging is part of the ARMED feature, not a permanent addition.** Rationale: byte mode exists solely to serve the windowed giant-line case; it has no consumer when the window is off, so gating it on the arm is honest, and shipping it always-on (or as a separate PR) would add a permanent capability whose only user is still flag-gated. Conditional schema/description was clean here, not a type-system fight (`input_schema` builds a fresh `Value`; `description` picks one of two `&'static str`), so I did not split. Fix: description + schema + byte-mode execution are all conditional on `window_armed()`; unarmed byte params are rejected (not silently ignored); new golden tests (`unarmed_read_file_toolspec_is_byte_identical_to_origin_main`, `unarmed_write_file_toolspec_is_byte_identical_to_origin_main`) pin the unarmed ToolSpec JSON — both the `Value` and its serialized string — against the exact origin/main snapshot. `armed_*` tests assert the armed spec legitimately differs.

**R1 — authorize-on-stat / truncate-on-independent-open TOCTOU. Confirmed, fixed by fd-binding.** New `write_no_follow_checked` (in `tools/mod.rs`) opens the existing file **without** `O_TRUNC`, `fstat`s the descriptor, and only `set_len(0)` + writes when that descriptor's `(mtime, size)` equals the epoch that authorized the write; otherwise `EpochChanged` → the guard forgets the stale coverage and refuses. Operating on the validated **descriptor** (not a re-resolved path) closes the race even against a same-name inode swap after validation; `O_NOFOLLOW` still rejects a symlink swap. The guard now threads its COMPLETE authorizing epoch into the write. Second half: `note_full_write` records COMPLETE only when the on-disk size equals `written_bytes`. Mutation-proof: the fd-rebind is unit-tested (`write_no_follow_checked_refuses_a_replaced_inode`) — replacing the file between authorize and the checked write refuses and leaves the external content intact; neutering the `found != expected` check makes it fail with the model's bytes clobbering the replacement.

**R4 — empty session keys collapse; reachable in production. Confirmed, fixed fail-closed.** A missing `parent_session_key` maps to `""` and is the common case (agents default `None` at agent/mod.rs; FFI lib.rs; fleet-worker lib.rs; CLI chat without `--goals`). **Decision: refuse-to-arm for keyless** (the review's simpler, safe-direction option) rather than mint per-agent tokens — read and write are separate tool instances with no shared per-agent identity in `ToolContext`, and threading a new token through every construction site is a large surface for this PR. Fix: `record_view`/`note_full_write` record nothing for an empty session (single isolation enforcement point), so a keyless task never reaches COMPLETE and its over-window overwrites fail closed; the guard adds a **distinct** "no identity … cannot be authorized" message (keyless paging is futile, so it must not be advised). Two isolation tests: `should_not_let_one_keyless_task_authorize_another_keyless_overwrite` (integration) and `should_record_nothing_for_an_empty_session` (ledger unit). The two enforcement layers (ledger-skip and guard message) each mask the other's single mutation, so each has its own killing test: neutering the ledger-skip fails the unit test; neutering the guard's keyless branch fails the integration test (assertion on the keyless-specific message).

**R5 — invariant was debug_assert only; forbidden advice remained. Confirmed, fixed.** All three loop-cap checks in read_file are now real runtime clamps (`clamp_armed_return`, clamping to `WINDOW_MAX_BYTES + FOOTER_RESERVE`, which the tripwire pins ≤ the loop cap) — enforced in release, unlike `debug_assert`. The giant-line advice interpolates no path at all; the byte-mode and window footers interpolate only bounded numbers; write refusals clamp the echoed path to 200 bytes and then clamp the whole message to the write cap. write_file's armed description no longer names `edit_file`/`apply_patch` (unarmed description is the exact origin string, so the golden also guards this).

**R2 — TAINTED was a silent permanent deadlock. Confirmed, fixed with a distinct variant.** A redaction-bearing file is sticky-TAINTED and the loop re-redacts every raw byte page, so paging can never unlock it. New `[REDACTED_VIEW_OVERWRITE]` prefix and a message that says the flag is incompatible with reconstructing-and-rewriting THIS file and to tell the user — it explicitly does not advise paging. `should_refuse_when_redacted_content_never_reached_the_model` asserts the distinct prefix, the absence of the partial-view prefix, the absence of "page through", and the "tell the user"/"cannot" signal. The limitation is now documented prominently in the PR body (armed mode is incompatible with whole-file rewrites of redaction-bearing files — directly relevant to a slides `script.js` embedding a data URI).

**Tracking-only (per scope):** probe-records-intent-before-guard left as is (deliberate — the probe measures intent, not damage); #2197 (unarmed 100KB-cut cache lie) filed round 2; empty-file error-message oddity pre-existing.

Regression families from the review all present and green through the real execute paths: giant-first-line refuses then completes via byte paging; long-path invariant (read advice and write refusal both under cap); redaction ⇒ not COMPLETE; cross-session refuses; post-restart refuses; mtime/size replacement refuses; shrink-between-pages recovers; plus the round-1/2 set re-based on the new rule. RED-first on the three the review named (unarmed-toolspec golden, TOCTOU primitive, empty-key isolation), watched failing before implementation.

Gates after rebase onto `d8125d182` (includes #2191): `cargo test -p octos-agent --lib` → `ok. 2755 passed; 0 failed; 3 ignored`; `cargo clippy -p octos-agent --all-targets -- -D warnings` → clean; `cargo fmt --all -- --check` → clean. Branch is 3 ahead / 0 behind main.

anantheparty (approved):
Adversarial pre-merge review of the redesign (head 40b48842, 4 commits incl. the round-3 fixes) passed — no blocker. All four fail-open axes from the round-1 self-review are closed fail-closed with regression tests: giant-first-line walks in-tool byte-mode and advises without ledgering (absence → refuse); restart safety is by construction (in-process ledger + absence-refuses); bounded 512-entry eviction degrades to Unknown → refuse (test-pinned); cross-session isolation via (parent_session_key, canonical path) keying + epoch (mtime,size) write-time re-verification. R6: the UNARMED ToolSpec is field-by-field identical to origin/main (prompt-cache prefix safety), golden-pinned. R1: the authorized overwrite re-verifies (mtime,size) via fstat on the same descriptor before truncating. Red lines: flag default-off (env, test-pinned), ci.yml untouched, non-windowed path golden-identical.

Note on CI: `check-windows` failed on `octos-bus` cron_service::tests::should_fire_due_job_exactly_once_when_rearm_races_mid_fire — a timing/race test in a crate this PR does not touch (231 passed / 1 failed); classified as the known Windows timing-flake family, unrelated to the diff. Everything within the diff's blast radius is green. Approving — merging left to you per convention.

anantheparty (dismissed):
Adversarial pre-merge review of the redesign (head 40b48842, 4 commits incl. the round-3 fixes) passed — no blocker. All four fail-open axes from the round-1 self-review are closed fail-closed with regression tests: giant-first-line walks in-tool byte-mode and advises without ledgering (absence → refuse); restart safety is by construction (in-process ledger + absence-refuses); bounded 512-entry eviction degrades to Unknown → refuse (test-pinned); cross-session isolation via (parent_session_key, canonical path) keying + epoch (mtime,size) write-time re-verification. R6: the UNARMED ToolSpec is field-by-field identical to origin/main (prompt-cache prefix safety), golden-pinned. R1: the authorized overwrite re-verifies (mtime,size) via fstat on the same descriptor before truncating. Red lines: flag default-off (env, test-pinned), ci.yml untouched, non-windowed path golden-identical.

Note on CI: `check-windows` failed on `octos-bus` cron_service::tests::should_fire_due_job_exactly_once_when_rearm_races_mid_fire — a timing/race test in a crate this PR does not touch (231 passed / 1 failed); classified as the known Windows timing-flake family, unrelated to the diff. Everything within the diff's blast radius is green. Approving — merging left to you per convention.

anantheparty (approved):
Re-approving after 6f5e5ca4 (test gating only): the two ctime-dependent swap-detection tests are now #[cfg(unix)] — they assert the Unix-specific ctime binding that the documented (mtime,size) fallback cannot provide on Windows, and the off-Unix failure was a real platform-dependent test failure (my earlier flake-family classification was wrong; the author gated them correctly). The implementation under review (redesign + round-3 fixes: four fail-open axes closed, R6 unarmed ToolSpec wire-identical, R1 fstat-on-descriptor re-verification) is unchanged by this gating. CI green on this head. Approving — merging left to you per convention.

PR #2194, opened 2026-08-31
Cache economics: one-shot cache-write opt-out, cache-write-aware pricing, get_goal probe latch

### LINKED ISSUE #2085: measure and eliminate repeated positive-capability schema scans in the goal ledger after migration
## Gap

#2076 (#2068) added a `pragma_table_info('goals')` schema scan in front of the goals statements so a pre-#2068 ledger file still works. On the reader that scan runs on **every** call:

```rust
pub fn get_goal(&self, goal_id: &str) -> Result<Option<Goal>> {
    let conn = self.conn.lock().unwrap();
    let has_time = Self::goals_has_time_column(&conn)?;
    ...
```

`get_goal` is an indexed point lookup on the primary key — the schema scan is now a fixed per-call overhead in front of it, and it runs on tokio worker threads (`goal_get` and every inline consult). The writers pay it too, though they fold it into a write transaction they already hold.

Surfaced by the #2076 round-2 review.

## Why the answer is cacheable

The capability is **monotone**: the column can only ever go absent → present, and once present it never disappears. So a cached `true` can never go stale. Only a cached `false` needs re-checking (another process may migrate the file underneath a long-lived connection).

## Fix direction

Cache the positive answer on `GoalLedger` (e.g. an `AtomicBool` seeded at open, re-probed only while `false`), so a migrated database — which is every database after the first write — pays the scan zero times. Measure first: `pragma_table_info` is served from SQLite's in-memory schema cache, so this may not be worth the added state.

## Refs

`crates/octos-fleet/src/sqlite_ledger.rs` — `goals_has_time_column`, `get_goal`, `ensure_goals_time_column`.

ymote:
Three related cache-economics fixes, one per commit — plus a three-commit round addressing the round-1 adversarial review (see "Review round 1" below).

## 1. One-shot summary calls no longer pay cache writes (`feat(llm)`)

Prompt caching (#1640, default ON in `AnthropicProvider`) marked three ephemeral `cache_control` breakpoints on EVERY request — including one-shot summarization calls whose prefixes are never sent again. Each such call paid the 1.25x cache-write premium for nothing.

The pi harness already avoids this: `completeSummarization` sets `cacheRetention: "none"` with the comment "Avoid cache writes for one-off summaries" (`packages/coding-agent/src/core/compaction/compaction.ts:587-591`). This PR adds the equivalent: `ChatConfig.cache_retention: CacheRetention::{Default,None}`; `AnthropicProvider::build_request` honors `None` by emitting zero `cache_control` blocks — byte-identical to the caching-disabled wire shape — while an unset config keeps the exact cached shape.

Opted-out one-shot sites (each sends its prompt exactly once, prefix never replayed):
- `octos-agent` `compaction::llm_compaction_summary`
- `octos-agent` `summarizer::LlmIterativeSummarizer` (each prompt embeds the prior summary + new turns)
- `octos-agent` `subagent_summary::fetch_summary` (kept opted out after round 1 — see the dedupe below, which is what makes it correct)
- `octos-agent` `research_utils::{extract_findings, merge_findings}` (unique per-batch prompts)
- `octos-services` `compaction` (both transcript summaries — the transcript is replaced by its summary)
- `octos-services` `persona_service` (persona + status-words generation, regeneration interval >> cache TTL)
- `octos-cli` `agent_orchestrator::run_goal_completion_verifier_with_usage` (unique objective+evidence prompt per verdict)

Round 1 added six more one-shot sites (see below) and the sub-agent watcher dedupe.

Deliberately left cached (prefix IS replayed): the agent loop (`llm_call.rs` — system+tools+history replayed every iteration) and `memory_consolidate` (its corrective re-ask appends to the original messages, genuinely reusing the prefix).

Tests: golden serialization (opt-out => zero `cache_control` and byte-equal to the disabled provider's shape; default => breakpoints intact and byte-identical with the field unset vs explicit `Default`), plus per-call-site capture probes pinning `cache_retention: None`. Removing the honor in `build_request` or any call-site opt-out fails a test.

## 2. Cache-write-aware cost accounting (the #1640 follow-up)

Adversarial review flagged twice that octos cannot price caching experiments: runtime pricing called plain `cost(input, output)` (`crates/octos-agent/src/agent/streaming.rs`, `response_usage_cost` / `emit_cost_update`) while `cost_with_cache` (the 0.1x read / 1.25x write primitive from #1640) sat unwired, and the persistent usage event (`crates/octos-store/src/usage_ledger.rs`) recorded cache reads but had NO cache-write field.

- `UsageEvent.cache_write_tokens` + `UsageTotals.cache_write_tokens`: additive, `#[serde(default)]`. Old-row compat pinned by a raw old-shape JSON fixture (carries `cache_read_tokens`, lacks `cache_write_tokens` — the exact shape every deployment wrote until now) plus a round-trip test. All four ledger writers (session actor, speculative overflow, AppUI, btw aside) now record it.
- Pricing wired through the cache-aware path wherever the provider reports cache tokens: `response_usage_cost` (all five callers pass the cache counts), `emit_cost_update`'s reprice fallback, the turn verifier's price, and the four octos-cli reprice fallbacks. Safe cross-provider because `TokenUsage`'s crate-wide contract is DISJOINT accounting (OpenAI/Gemini inclusive counts are normalized at their parse boundaries). `cost_ledger::project_cost_usd` stays on `cost()` — it prices token estimates for dispatch projections where no provider-reported cache figures exist.
- Budget-gate semantics unchanged: admission still counts cached traffic (`budget_tokens_used` = input + cache_read + cache_write, pinned by the existing `budget_counts_cache_tokens_as_used`); this fixes PRICE, not admission.

Pin: usage of 100k in / 10k out with `cache_read=10_000`, `cache_write=2_000` prices at exactly naive + the provider rate card's premiums, and moves off the naive figure.

## 3. `get_goal` schema-probe latch (`perf(fleet)`)

Every `get_goal` ran `pragma_table_info('goals')` before its point lookup. The capability is monotone per connection (a committed column never disappears), so the POSITIVE result is now latched in an `AtomicBool` on `GoalLedger`; once observed, reads skip the probe. A NEGATIVE result is never latched — an unmigrated file can migrate mid-connection via a writer's in-transaction `ensure_goals_time_column`, and a latched "no" would synthesize `time_used_seconds: 0` forever.

The #2092 torn-read fix is preserved: while the answer is still "no", the probe keeps running inside the same DEFERRED transaction as the SELECT (single database state, exactly as before). The latched fast path only asserts a fact observed in a committed state that monotonicity carries into every later snapshot, so the `has_time` SELECT always reads the real column value — the probe/SELECT state split cannot reappear. Pinned via a probe counter (no timing/sleep): unmigrated files re-probe every read and observe a mid-connection migration through a fresh probe; a migrated ledger pays exactly one probe then skips it; the #2092 concurrent-migration torn-read test passes unchanged.

Closes #2085

## Review round 1 (three commits)

**Retention flips, both directions** (`fix(llm)` `094355329`):
- `subagent_summary::fetch_summary` was NOT one-shot as claimed: the watcher ticks every 30s and an idle tick deterministically rebuilds the byte-identical prompt. Rather than reverting to `Default` (a cache hit still bills ~0.1x), the watcher now DEDUPES: an unchanged activity snapshot reuses the previous summary with no LLM call at all (`runtime_detail` still re-stamped so tick/at stay live). Dedupe strictly dominates caching — idle ticks cost zero — and every prompt that still reaches the provider is genuinely fresh, which is what makes the retained opt-out correct. Failures/empty summaries never latch.
- Six sites that should have been `None` and were not, now flipped with per-site RED-first probes: the turn verifier (`agent/verifier.rs` — the stable system block is one sentence; the breakpoint-marked last user block carries the rolling TurnLedger), `rich_output::author_html` (single-shot authoring per voice turn), pipeline `plan_dynamic_tasks` (one planning call per dynamic node), `session/btw` (one restricted call per aside), the final code-review join (once per review), and memory-refresh extraction (one per-session transcript).

**Provider-aware cache pricing** (`fix(llm)` `5ddff0f12`): `cost_with_cache` hardcoded Anthropic's 0.1x/1.25x and `response_usage_cost` discarded the provider it had just resolved — OpenAI/Gemini cache traffic was billed at Anthropic multipliers. New `pricing::cache_rates_for_provider` rate cards, sources documented in code: anthropic 0.1x/1.25x (uniform per Anthropic pricing docs; catalog rows agree); gemini/vertex/google 0.25x/0 (catalog row gemini-2.5-flash 0.0375/0.15); everything else 1.0x/0 — verified against the codebase first as instructed: the catalog's only OpenAI row carries `cache_read_per_mtok: None` and the public OpenAI discount varies per model family (0.5x gpt-4o-era, deeper for newer), so a provider-wide constant would invent a discount for some models; reads bill at the full input rate (never-understate bound) until the runtime catalog carries per-model cached rates. All runtime call sites now price via `cost_with_cache_for_provider` with the answering slot's provider label. Mutation-proofed: same usage across three provider labels yields three different figures, and an openai-labeled agent slot prices differently from an anthropic-labeled one.

**Hook costs are attributed spend** (`fix(agent)` `b9c688d67`): the after-LLM hook payload repriced the MERGED usage at `self.llm.model_id()`, mispricing cross-provider retries. It now carries the attributed figures — `response_cost` = this call's `attributed_cost` verbatim; `session_cost` = the turn's previously attributed spend + this call's. Pinned by a hook-capture test with a deliberately off-catalog prior spend that any repricing implementation fails.

### Tracking notes (no code change)
- The two anthropic golden tests compare new-code against new-code (opt-out vs `with_prompt_caching(false)`, unset vs explicit `Default`), not against a checked-in pre-change fixture. That the default wire shape is byte-identical to pre-PR output is established by inspection: the only change on the default path is `self.prompt_caching && config.cache_retention != None`, where the second operand is `true` for every config that predates the field, and `ChatConfig`'s serde keeps `cache_retention` off the wire when default.
- `CacheRetention` is a closed enum with no serde fallback variant: a FUTURE retention value (e.g. a "long"/1h tier) arriving in persisted JSON would fail deserialization in today's binary rather than degrade to `Default`. Fine today (only "default"/"none" exist and the field is skip-serialized when default); revisit if the enum grows.

## Review round 2 (two commits)

Round-2 verdict closed item 3 (hook attribution — no double-count, refuted V4) and accepted the six retention flips. Two MUST-FIX gaps remained:

**V1 — dedupe key + timestamp** (`fix(agent)` `be2b2d863`): the round-1 sub-agent dedupe compared the raw activity lines, but `build_prompt` truncates each line to 400 chars — two snapshots differing only past char 400 render a BYTE-IDENTICAL prompt, so the raw-line compare would re-send that identical prompt with caching disabled, violating the exact invariant the opt-out relies on. The dedupe key is now the canonical rendered prompt (`build_prompt`'s output — the bytes that actually hit the provider). Separately, a deduped tick re-stamped the reused summary with `Utc::now()` while the event's `at` means "when the summary was PRODUCED"; `LastTickSummary` now carries the original `produced_at` and reuses it, with `tick_seq` (a separate field) conveying liveness. Tests (timing-free): a >400-char-tail-differing pair asserts exactly ONE LLM call (fails the old line-compare), and a directly-driven `summarize_tick` seeded with a fixed 2020 production time asserts the deduped event carries that 2020 `at` (mutation-verified against a `Utc::now()` reuse path).

**V3 — protocol-aware cache-write pricing** (`fix(llm)` `f13b8bfc8`): the round-1 unknown-provider card `(1.0 read, 0.0 write)` was wrong on the WRITE side. `cache_write_tokens` are disjoint from `input_tokens`, so a `0.0` write multiplier makes a reported write VANISH — an understatement — and it is reachable: relabeled Anthropic-protocol providers (r9s serving claude, zai/zai-coding serving GLM over the Anthropic Messages API) land in the unknown bucket and DO emit `cache_creation_input_tokens`. The rate card is now keyed on PROTOCOL, from the resolved slot's provider label + model (both threaded through `cache_rates` / `cost_with_cache_for_provider`): Anthropic protocol (native + `anthropic`-containing labels + `zai`/`zai-coding` + `r9s`&claude-\*) → 0.1x/1.25x; gemini/vertex/google → 0.25x/0 (Gemini never reports writes); residual unknown → 1.0x read (no invented discount) / **1.25x write** — because `cache_write_tokens` is populated ONLY by the Anthropic parser (`anthropic.rs`; verified no OpenAI/Gemini path sets it), any write reaching the residual bucket is an Anthropic-protocol write from an unrecognized label and 1.25x is the honest value, not a vanish. zhipu/dashscope/minimax/moonshot-coding are OpenAI-protocol re-hosts and correctly stay in the residual read bucket (they never emit writes). Tests: zai/zai-coding/r9s-claude/custom-anthropic regressions asserting reads at 0.1x and writes NOT free; the unknown-bucket + agent-level tests rewritten to expect the write priced at 1.25x (the old "writes free" expectation is gone); mutation-verified by forcing the write multiplier to 0.

### Tracking notes — round 2 (no code change)
- "Never replayed" was slightly too absolute: two dynamic pipeline nodes with identical `planning_prompt`+input+`max_tasks`, or two identical higher-level BTW requests, produce the same request bytes. These are externally-duplicated INVOCATIONS, not internal retry/replay, so the one-shot flips remain correct (the prefix is not re-sent within one logical call). 
- Pre-existing, out of scope: the final failed streaming attempt's usage is not accumulated before the non-streaming fallback (`llm_call.rs`); the hook-capture test is unix-only.

## Verification (all green, sequential, after round 2)

- `cargo test -p octos-llm --lib` — 588 passed
- `cargo test -p octos-agent --lib` — 2709 passed
- `cargo test -p octos-fleet` — 192 passed
- `cargo test -p octos-store` — 36 passed
- `cargo test -p octos-pipeline` — 355 passed; `cargo test -p octos-services` — 52 passed (lib)
- `cargo test -p octos-cli --lib` — 1625 passed; api-gated probes + usage tests via `--features api`
- `cargo clippy -p {octos-llm,octos-agent,octos-fleet,octos-store,octos-pipeline,octos-services,octos-cli} --all-targets -- -D warnings` — clean (octos-cli also with `--features api`)

### DIFF
--- a/crates/octos-agent/src/agent/llm_call.rs
+++ b/crates/octos-agent/src/agent/llm_call.rs
@@ -142,6 +142,8 @@ impl Agent {
                         let final_cost = self.response_usage_cost(
                             response.usage.input_tokens,
                             response.usage.output_tokens,
+                            response.usage.cache_read_tokens,
+                            response.usage.cache_write_tokens,
                             response.provider_index,
                         );
                         let attributed_cost = match (retry_spend, final_cost) {
@@ -158,10 +160,21 @@ impl Agent {
                             let latency_ms = call_start.elapsed().as_millis() as u64;
                             let cum_in = total_usage.input_tokens + response.usage.input_tokens;
                             let cum_out = total_usage.output_tokens + response.usage.output_tokens;
+                            let cum_cache_read =
+                                total_usage.cache_read_tokens + response.usage.cache_read_tokens;
+                            let cum_cache_write =
+                                total_usage.cache_write_tokens + response.usage.cache_write_tokens;
                             let pricing = octos_llm::pricing::model_pricing(self.llm.model_id());
-                            let session_cost = pricing.map(|p| p.cost(cum_in, cum_out));
+                            let session_cost = pricing.map(|p| {
+                                p.cost_with_cache(cum_in, cum_out, cum_cache_read, cum_cache_write)
+                            });
                             let response_cost = pricing.map(|p| {
-                                p.cost(response.usage.input_tokens, response.usage.output_tokens)
+                                p.cost_with_cache(
+                                    response.usage.input_tokens,
+                                    response.usage.output_tokens,
+                                    response.usage.cache_read_tokens,
+                                    response.usage.cache_write_tokens,
+                                )
                             });
                             let payload = HookPayload::after_llm(
                                 self.llm.model_id(),
@@ -219,6 +232,8 @@ impl Agent {
                                 let final_cost = self.response_usage_cost(
                                     fallback_resp.usage.input_tokens,
                                     fallback_resp.usage.output_tokens,
+                                    fallback_resp.usage.cache_read_tokens,
+                                    fallback_resp.usage.cache_write_tokens,
                                     fallback_resp.provider_index,
                                 );
                                 let attributed_cost = match (retry_spend, final_cost) {
@@ -256,6 +271,8 @@ impl Agent {
                     if let Some(cost) = self.response_usage_cost(
                         response.usage.input_tokens,
                         response.usage.output_tokens,
+                        response.usage.cache_read_tokens,
+                        response.usage.cache_write_tokens,
                         response.provider_index,
                     ) {
                         retry_spend = Some(retry_spend.unwrap_or(0.0) + cost);
@@ -378,6 +395,8 @@ impl Agent {
                                 let final_cost = self.response_usage_cost(
                                     resp.usage.input_tokens,
                                     resp.usage.output_tokens,
+                                    resp.usage.cache_read_tokens,
+                                    resp.usage.cache_write_tokens,
                                     resp.provider_index,
                                 );
                                 let attributed_cost = match (retry_spend, final_cost) {
--- a/crates/octos-agent/src/agent/llm_call.rs
+++ b/crates/octos-agent/src/agent/llm_call.rs
@@ -160,22 +160,20 @@ impl Agent {
                             let latency_ms = call_start.elapsed().as_millis() as u64;
                             let cum_in = total_usage.input_tokens + response.usage.input_tokens;
                             let cum_out = total_usage.output_tokens + response.usage.output_tokens;
-                            let cum_cache_read =
-                                total_usage.cache_read_tokens + response.usage.cache_read_tokens;
-                            let cum_cache_write =
-                                total_usage.cache_write_tokens + response.usage.cache_write_tokens;
-                            let pricing = octos_llm::pricing::model_pricing(self.llm.model_id());
-                            let session_cost = pricing.map(|p| {
-                                p.cost_with_cache(cum_in, cum_out, cum_cache_read, cum_cache_write)
-                            });
-                            let response_cost = pricing.map(|p| {
-                                p.cost_with_cache(
-                                    response.usage.input_tokens,
-                                    response.usage.output_tokens,
-                                    response.usage.cache_read_tokens,
-                                    response.usage.cache_write_tokens,
-                                )
-                            });
+                            // #2194 review: the payload carries the ATTRIBUTED
+                            // spend — each attempt priced at the slot that
+                            // produced it — never a reprice of the merged
+                            // usage at `self.llm.model_id()`, which misprices
+                            // cross-provider retries. `turn` has not recorded
+                            // THIS response yet, so its spend is exactly the
+                            // turn's prior responses; without a figure for
+                            // this call the turn's prior attributed spend is
+                            // still the honest cumulative.
+                            let response_cost = attributed_cost;
+                            let session_cost = match attributed_cost {
+                                Some(cost) => Some(turn.spend_usd() + cost),
+                                None => turn.priced_spend(),
+                            };
                             let payload = HookPayload::after_llm(
                                 self.llm.model_id(),
                                 iteration,
@@ -666,6 +664,124 @@ mod tests {
 
     fn turn() -> LoopTurnState {
         LoopTurnState::new(Instant::now())
+    }
+
+    /// One clean streamed tool-call response with cache-bearing usage, on a
+    /// priced model under an anthropic label — for hook payload cost tests.
+    struct PricedGoodStreamProvider;
+
+    #[async_trait]
+    impl LlmProvider for PricedGoodStreamProvider {
+        async fn chat(
+            &self,
+            _messages: &[Message],
+            _tools: &[ToolSpec],
+            _config: &ChatConfig,
+        ) -> eyre::Result<ChatResponse> {
+            eyre::bail!("non-streaming fallback should not be reached in this test")
+        }
+
+        async fn chat_stream(
+            &self,
+            _messages: &[Message],
+            _tools: &[ToolSpec],
+            _config: &ChatConfig,
+        ) -> eyre::Result<ChatStream> {
+            let events = vec![
+                StreamEvent::ToolCallDelta {
+                    index: 0,
+                    id: Some("read_file_1".to_string()),
+                    name: Some("read_file".to_string()),
+                    arguments_delta: "{\"path\":\"a.md\"}".to_string(),
+                },
+                StreamEvent::Usage(LlmTokenUsage {
+                    input_tokens: 100_000,
+                    output_tokens: 10_000,
+                    cache_read_tokens: 10_000,
+                    cache_write_tokens: 2_000,
+                    ..Default::default()
+                }),
+                StreamEvent::Done(StopReason::ToolUse),
+            ];
+            Ok(Box::pin(stream::iter(events)))
+        }
+
+        fn model_id(&self) -> &str {
+            "claude-opus-4"
+        }
+
+        fn provider_name(&self) -> &str {
+            "anthropic"
+        }
+    }
+
+    /// #2194 review: `attributed_cost` prices each attempt at its actual
+    /// provider slot, but the after-LLM hook payload used to REPRICE the
+    /// merged usage at `self.llm.model_id()` — wrong for anything the
+    /// attribution already priced differently. The payload must carry the
+    /// attributed figures: response_cost = this call's attributed spend,
+    /// session_cost = the turn's previously attributed spend + this call's.
+    #[cfg(unix)]
+    #[tokio::test]
+    async fn should_report_attributed_spend_in_after_llm_hook_payload() {
+        use crate::hooks::{HookConfig, HookExecutor};
+
+        let capture_dir = tempfile::tempdir().unwrap();
+        let capture = capture_dir.path().join("after_llm_payload.json");
+        let hook = HookConfig {
+            event: crate::hooks::HookEvent::AfterLlmCall,
+            command: vec![
+                "sh".into(),
+                "-c".into(),
+                format!("cat > {}", capture.display()),
+            ],
+            timeout_ms: 5_000,
+            tool_filter: vec![],
+            path_filter: vec![],
+            requires_bin: None,
+        };
+
+        let (agent, _dir) = build_agent(Arc::new(PricedGoodStreamProvider)).await;
+        let agent = agent.with_hooks(Arc::new(HookExecutor::new(vec![hook])));
+
+        // Prior turn spend recorded with a DELIBERATELY off-catalog figure:
+        // the hook's cumulative cost must be built from this attribution,
+        // never from repricing the cumulative tokens at the current model.
+        let mut turn = turn();
+        turn.record_usage(5_000, 1_000, 0, 0, None, Some(0.005));
+        let total_usage = turn.total_usage().clone();
+
+        let (_response, _streamed, attributed) = agent
+            .call_llm_with_hooks(
+                &msgs(),
+                &[],
+                &ChatConfig::default(),
+                1,
+                &total_usage,
+                &mut turn,
+            )
+            .await
+            .expect("clean streamed response");
+        let attributed = attributed.expect("claude-opus-4 has catalog pricing");
+
+        let raw = std::fs::read_to_string(&capture).expect("hook captured the payload");
+        let payload: serde_json::Value = serde_json::from_str(&raw).unwrap();
+        let response_cost = payload["response_cost"]
+            .as_f64()
+            .expect("response_cost present");
+        let session_cost = payload["session_cost"]
+            .as_f64()
+            .expect("session_cost present");
+        assert!(
+            (response_cost - attributed).abs() < 1e-12,
+            "hook response_cost must be the attributed spend, got {response_cost} vs {attributed}"
+        );
+        assert!(
+            (session_cost - (0.005 + attributed)).abs() < 1e-12,
+            "hook session_cost must be prior attributed turn spend + this call's, \
+             got {session_cost} vs {}",
+            0.005 + attributed
+        );
     }
 
     // ── Tests ─────────────────────────────────────────────────────────────────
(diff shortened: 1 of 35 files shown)


anantheparty (dismissed):
Adversarial pre-merge review passed — pricing arithmetic verified precisely: cost_with_cache composes input@1.0x + output + reads×0.1×in_rate/M + writes×1.25×in_rate/M, units consistent, all six populating parse boundaries feed disjoint values (no double-billing), and every actual-usage pricing call site is wired (llm_call/verifier/session_actor/btw+standalone reprice fallback/loop_runner). The one-shot CacheRetention is per-request with no global state — cannot leak across turns; all ten call sites are genuinely one-shot (compaction/digest/summarizer/research/verdict/persona). get_goal probe latch is correct (positive-only latch, monotonicity holds, no billing interaction). Three legacy nits (openai_responses/openrouter non-streaming cached extraction, project_cost_usd migration, usage_status write-side tokens) go to a follow-up issue — all pre-existing, all safe-direction. Approving — merging left to you per convention.

ymote:
## Adversarial review — round 1 — DO-NOT-MERGE (3-item flip list)

**Clean:** the get_goal probe latch (H4 fully refuted — per-connection latch, fresh-latch-per-open, #2092's torn-read invariant preserved, sqlite_ledger.rs:13/:289/:976); the disjoint-token normalization worry (TokenUsage disjoint everywhere — openai.rs:789/:1258, gemini.rs:1001/:1098); retry pricing does not double-apply premiums (llm_call.rs:267/:139); the old-row fixture matches today's writer shape.

**Must fix:**
1. **Retention misclassified both ways.** `subagent_summary` is not one-shot — the 30s watcher rebuilds an identical prompt on idle ticks (subagent_summary.rs:36/:345/:131), so `None` forfeits real hits. Meanwhile six genuinely one-shot sites stayed cached: agent verifier (the 'stable prompt' claim is false — verifier.rs:510/:592 vs anthropic.rs:499), rich_output author_html (self-described single-shot), dynamic pipeline planning (executor.rs:1200), session/btw ('ONE restricted LLM call', ui_protocol_transport.rs:24228/:24451), final review join (:29278), memory-refresh extraction (service.rs:825).
2. **Provider-blind cache pricing.** `cost_with_cache` hardcodes Anthropic's 0.1×/1.25× (pricing.rs:123/:137) and `response_usage_cost` discards the provider it resolved (streaming.rs:552) — OpenAI/Gemini cache reads are mispriced at Anthropic multipliers.
3. **Hook payloads reprice merged retry usage at the active model** (llm_call.rs:159), discarding the correct per-attempt attribution computed two lines up (:142).

Tracking-only: both byte-identity goldens compare new-code-vs-new-code (anthropic.rs:1818/:1861) — behaviour preserved by inspection; `CacheRetention` is a closed serde enum (config.rs:61).

Round 2 in progress.

ymote:
## Adversarial review — round 2 — DO-NOT-MERGE (2 must-fix)

Item 3 (hook attribution) closed — V4 refuted, no double-count, merged tokens and attributed cost cover the same attempt set. Six retention flips all present and correct (V2 refuted as blocker). Two gaps:

**V1 — the dedupe compares raw lines, but the PROMPT truncates them.** Comparison at subagent_summary.rs:409 is over full activity lines; `build_prompt` truncates each to 400 chars (:483). Two snapshots differing only after char 400 → two byte-identical prompts with caching disabled (:527) — the exact invariant this establishes, violated. Dedupe the canonical rendered prompt, not the pre-truncation lines. Also: a reused summary is re-stamped Utc::now() (:415/:458) while `at` means 'when produced' (harness_events.rs:808) — keep the production timestamp.

**V3 — the unknown-provider rate card zeroes cache-WRITE tokens** (pricing.rs:159/:181). Writes are disjoint from input (types.rs:73), so a 0.0 multiplier makes them vanish — an understatement. Reachable via relabeled Anthropic-protocol providers (r9s/claude :42, zai :43, custom-anthropic chat.rs:3735) which emit cache_creation_input_tokens (anthropic.rs:860); the current test bakes it in (streaming.rs:856 expects a Claude model's writes free). Fix: price cache-write from the PROTOCOL — Anthropic-protocol slots get 1.25× regardless of family label; at minimum never price a wire-billed write token at 0.

Refuted: Gemini creates no caches (reads only), Anthropic uses only ephemeral (1.25× correct), answering-slot resolution correct via provider_index.

Tracking-only: 'never replayed' too absolute (two identical dynamic-planning nodes duplicate — external, not internal replay); final failed streaming attempt not accumulated before non-stream fallback (pre-existing); unix-only hook test.

Round 3 in progress.

anantheparty (dismissed):
Delta re-approval (3 follow-up commits 09435532/5ddff0f1/b9c688d6 reviewed against the earlier verdict): retention classification fixed both ways (one-shot call sites no longer pay the 1.25x write premium on never-replayed prompts; fetch_summary dedupes via LastTickSummary instead of restoring writes, with the consolidation pass correctly keeping its cache); provider-aware cache pricing migrated across all runtime call sites (Anthropic 0.1/1.25, Gemini 0.25/0 verified against catalog); after-LLM hook payload now carries attributed spend (per-attempt slot pricing) with wire-compatible fields. No regression to the previously-verified pricing core, opt-out lifecycle, or probe latch. Remaining nits tracked in the follow-up issue. Approving — merging left to you per convention.

anantheparty (dismissed):
Re-approved after b68bb973 (delta-verified): custom + api_type=anthropic endpoints now label as "custom-anthropic" so the documented contains("anthropic") classifier catches them (fixes a 10x cache-read overcharge), and r9s protocol selection is extracted into prefers_anthropic() as the single source of truth shared by construction and pricing — with the mixed-case "Claude-…" edge test proving the OpenAI lane keeps residual rates. Zero behavior change outside cost classification. Approving — merging left to you per convention.

anantheparty (dismissed):
Approving the final head (f002b942). Verification history: original review approved the pricing core/opt-out/probe-latch; rounds 4-5 fixed the wrapper lane-drop class (RetryProvider/ContextOverride/Middleware/Swappable/Router/FallbackProvider metadata delegation + stream winner attribution, prescription-exact) and the retention/provider-aware pricing nits (#2198). Known residual, filed as #2199 by the author: a nested mixed-lane container answering as a FALLBACK can still mis-resolve the cache lane (chat resolves the container default slot; stream index gets overwritten) — strictly better than main (which dropped the lane to Residual once wrapped), narrow shape, and the durable fix (answering_metadata on the response/stream) is designed in #2199. All other aspects verified: pricing arithmetic exact vs Anthropic/Gemini disjoint billing, opt-out cannot leak across turns, probe latch positive-only, hook payload attributed per-attempt. Red lines: ci.yml untouched, no default-behavior change for unconfigured users. Approving — merging left to you per convention.

anantheparty (approved):
Re-approving the latest head — this standing approval covers each iterative push in the #2194 cache-economics series (rounds 1-6 reviewed: pricing core, opt-out lifecycle, probe latch, wrapper lane delegation, FallbackProvider attribution, label-encoding revert). Every push in this series has been delta-reviewed; the known residual is #2199 (nested mixed-lane fallback mis-attribution, durable fix designed). Approving the best current state — merging left to you per convention.

anantheparty (approved):
Approving the final head (40d29f7d) — verification chain across all 11 commits since the reviewed baseline (bc40dbd5):

- Rounds 1-4 were delta-reviewed (b68bb973…1b083a4c): retention classification both ways, provider-aware cache pricing, hook payload attributed spend — all verified; that round also found B1 (the cache lane dropped at the default RetryProvider wrapper, 10x anthropic cache-read overstatement on the default stack).
- 0bc43b90 fixes B1 exactly as prescribed: provider_metadata/_for_index delegation through RetryProvider/ContextOverride/Middleware/Swappable/Router + the RetryProvider regression test (lane propagation and identity preservation asserted).
- ad6d1f12 + f002b942 implement #2199's durable fix (answering-slot attribution carried compositionally through nested containers; the flat-index residual corner documented in fallback.rs per round 7) — FallbackProvider was in the B1 prescription list, and the implementation matches it.
- 40d29f7d documents the non-compositionality for reviewers.

CI fully green on this head (all required checks). #2199 remains open tracking the flat-index residual with the answering_metadata design. The follow-up issue #2198 tracks the remaining cache-cost nits. Approving — merging left to you per convention.

Issue #2195, opened 2026-08-31
Windows: native restricted-token confinement (no separate helper) so AppContainer absence doesn't force Docker-or-refusal


ymote:
## Context

octos's Windows confinement today is the AppContainer backend (`crates/octos-agent/src/sandbox/windows.rs`), which depends on a **separately distributed helper binary** (`octos-sandbox.exe` next to the octos binary or on PATH). When the helper is not installed:

- `SandboxMode::Auto` on Windows falls through to Docker, and without Docker degrades to unconfined execution (warned; refused under the new `sandbox.fail_closed=true`),
- explicit `sandbox.mode = "appcontainer"` now fails closed with a typed `SandboxUnavailable` refusal.

So a stock Windows install without the helper and without Docker Desktop has **no native confinement path at all** — the choice is "run unconfined" or "refuse".

## Ask

Ship a native Windows confinement path that needs **no separately installed helper**, so Windows parity stops depending on helper distribution:

- A **restricted-token runner**: `CreateProcessW` with a restricted token (`CreateRestrictedToken`: drop admin-ish groups/privileges, deny-only SIDs) plus ACL-scoped filesystem reach for the workspace — launched in-process from octos-agent. Prior art: deepseek-harness's `sandbox-windows-acl` runner (paired with its `native/landlock-run` on Linux), which its `confine()` uses to fail closed (`SANDBOX_UNAVAILABLE`) rather than ever passing argv through unconfined. octos's refusal texts adopted the same fail-closed contract in the PR that files this issue; this issue is the "make Windows actually confine out of the box" half.
- Alternatively (or additionally): bundle `octos-sandbox.exe` into the Windows release artifacts/installer so the existing AppContainer backend is present by default.

## Notes

- `deny(unsafe_code)` is workspace-wide; a native Win32 token runner needs FFI (windows-rs) — likely a dedicated crate or the existing helper-binary crate growing an in-process mode, with the lint carve-out scoped there.
- The decision layer added in the referring PR (`decide_sandbox` + `HostBackendProbe`) is where the new backend plugs in: add a probe + a `SandboxBackendChoice` variant and it participates in `Auto` ordering and the refusal remediation text.
- Remediation texts currently tell Windows users to install the helper or Docker Desktop; once this ships they should name the native runner instead.

PR #2196, opened 2026-08-31
feat(sandbox): fail closed on unhonorable explicit modes; make Auto degradation loud

ymote:
## Problem

A user who configures confinement and cannot get it should be told — loudly, or by refusal — not silently handed an unconfined shell. Before this PR, two explicit modes silently degraded to `NoSandbox`, three explicit modes constructed backends whose every spawn died with a bare ENOENT, and the Windows AppContainer backend's wrap-time fallback passed argv straight to an unsandboxed `cmd /C`.

Contrast: deepseek-harness fails closed — its `confine()` throws `SANDBOX_UNAVAILABLE` with remediation ("install bubblewrap / ensure sandbox-exec / ensure the restricted-token runner can start — otherwise switch to danger-full-access") rather than ever passing argv through unconfined. This PR adopts that contract for explicit modes and offers it as an opt-in for `Auto`, without changing what any default config does.

## Current-behaviour map (verified from the code, per path)

| # | Config | Host state | Before (v2.0.3-rc.9) | After |
|---|--------|-----------|----------------------|-------|
| 1 | `enabled=false` (incl. `--danger-full-access`) | any | `info!` + `NoSandbox` | **unchanged** — explicit opt-out, beats `fail_closed` |
| 2 | `mode="none"` | any | `NoSandbox`, silent | **unchanged** — explicit opt-out, beats `fail_closed` |
| 3 | `mode="auto"`, backend found | any | backend (macOS: seatbelt; Linux: bwrap → Landlock helper; Windows: AppContainer helper; any: Docker) | **unchanged** — same selection, same probe order |
| 4 | `mode="auto"`, no backend | e.g. Windows without helper+Docker | `warn!` on **every** construction + `NoSandbox` | same degradation, warned **once per process** (names the knob); `octos doctor` row unchanged; **`fail_closed=true` → typed refusal** |
| 5 | `mode="bwrap"`, non-Linux or bwrap broken | — | `BwrapSandbox` constructed blindly → every spawn "No such file or directory" (no remediation) | typed refusal at creation |
| 6 | `mode="landlock"`, non-Linux | — | `warn!` + `NoSandbox` → **commands ran UNCONFINED** | typed refusal (**the bug; deliberate breaking change**) |
| 7 | `mode="landlock"`, Linux, helper missing | — | constructed; wrap-time echo+exit 1 (terse) | typed refusal at creation, with remediation |
| 8 | `mode="macos"`, non-macOS | — | `MacosSandbox` blindly → every spawn ENOENT | typed refusal |
| 9 | `mode="docker"`, docker not on PATH | — | `DockerSandbox` blindly → every spawn ENOENT | typed refusal |
| 10 | `mode="appcontainer"`, non-Windows | — | `warn!` + `NoSandbox` → **UNCONFINED** | typed refusal (**the bug; deliberate breaking change**) |
| 11 | `mode="appcontainer"`, Windows, helper missing | — | constructed; **wrap-time fell back to unsandboxed `cmd /C`** (worst case: asked for confinement, got passthrough) | creation refuses; the wrap-time fallback itself now refuses too (TOCTOU-safe: helper vanishing after selection still never passes argv through) |

## Compat argument (the central constraint)

Every default-config user is on `Auto` — no shipped profile or template sets an explicit mode — and rows 1–4 are behaviour-preserving for them: **a Windows default-config user with no backend keeps working unconfined**, now warned once per process instead of per construction, with `sandbox.fail_closed=true` available to opt into refusal. `should_degrade_loudly_not_refuse_when_auto_finds_no_backend` and `should_never_refuse_a_default_config_on_any_host` lock this in.

The users whose behaviour changes had either a broken shell already (rows 5, 8, 9: every command failed with a bare ENOENT — the refusal is the same failure with an actionable message) or were **silently unconfined against their expressed intent** (rows 6, 10, 11) — which is the bug this PR exists to fix.

## Design

- **Pure decision layer**: `decide_sandbox(config, HostOs, &dyn HostBackendProbe) -> SandboxDecision {Confine|Unconfined(reason)|Refuse(SandboxUnavailable)}` — no `cfg!` inside, all host facts arrive as data, probes stay lazy in `Auto`'s historical order. Every OS's full matrix is unit-tested from any host (the task's cfg-testability requirement). `create_sandbox` and `auto_sandbox_kind` (doctor) are thin projections of the same decision, so doctor and runtime cannot diverge.
- **`create_sandbox` stays infallible** (~30 construction sites across agent/pipeline/cli, many in spawn arms that cannot propagate): a `Refuse` decision yields `RefusingSandbox`, whose `wrap_command` never runs the requested command — it substitutes a shell-inert stderr line + exit 1 (the in-tree precedent is the Landlock backend's helper-missing refusal). Exec-shaped tools (`shell` incl. background, `exec_command`, its interactive sessions, `bash`) short-circuit even earlier on the new `Sandbox::refusal()` and return the **full typed message** — model-readable, since the model is often the one seeing it.
- **Refusal text** names the requested mode, the mismatch, per-OS installs (Linux: bubblewrap/helper/Docker; macOS: sandbox-exec/Docker; Windows: octos-sandbox.exe helper/Docker Desktop, with the native-runner future-work note), and the explicit opt-outs (`sandbox.enabled=false` / `mode="none"`). It deliberately avoids the kernel denial phrases so `sandbox_denial_hint` can never append a misleading "the OS sandbox blocked a file access" hint to a refusal.
- **`sandbox.fail_closed: bool`** (serde default `false`): `Auto`-resolves-to-nothing becomes a refusal. dsh semantics today for security-conscious operators, zero change for everyone else. It participates in the profile-defaults merge as a deny-wins floor (like `workspace_write`): an operator defaults file mandating it cannot be loosened by a profile.
- **Fail-closed gates kept coherent**: mcp-serve refuses the session with the typed message (same outcome as its pre-existing `is_noop` gate for these configs, better text); the fleet pool boot gate and the fleet worker attempt gate treat a refusal like a missing backend (pool not installed / attempt terminated with the typed reason) — matching the pre-PR outcomes where rows 6/10 resolved to `NoSandbox` and were caught by `is_noop()`, instead of regressing to a pool whose every command refuses one by one.

## Windows

`check-windows` in CI is the arbiter for the `#[cfg(windows)]` code (the AppContainer wrap-time refusal + its unit test): the local msvc cross-check is broken on this host (aws-lc-sys), so **the Windows-specific code is compile-verified by CI, not executed locally** — `cargo fmt` passing proves it parses; the decision-layer Windows matrix is fully executed on this host via the pure tests. The refusal command uses space-free tokens (`cmd /C echo … 1>&2 & exit /b 1`) so std's arg quoting can't neutralize the metacharacters, and the original command is never present to resurrect.

Follow-up for real native confinement without the separately-distributed helper (dsh's `sandbox-windows-acl` restricted-token runner as prior art, alongside its `native/landlock-run`): **#2195**, referenced from the Windows remediation text.

## TDD

Faithful extraction first (all 34 pre-existing sandbox tests green against the decision layer mirroring old semantics), then the new tests RED against it — verbatim:

```
test result: FAILED. 34 passed; 4 failed
    sandbox::tests::should_produce_refusing_sandbox_from_create_sandbox_when_mode_unhonorable
    sandbox::tests::should_refuse_when_auto_finds_no_backend_and_fail_closed_set
    sandbox::tests::should_refuse_when_explicit_mode_backend_is_missing_on_right_os
    sandbox::tests::should_refuse_when_explicit_mode_is_on_wrong_os
("explicit Bwrap on Macos must refuse, got Confine(Bwrap)" / "auto+fail_closed on backend-less Linux must refuse: Unconfined(AutoNoBackend)")
```

then GREEN by flipping `decide_sandbox` to the new contract. New tests (octos-agent unless noted): the four above plus `should_confine_via_native_backend_when_auto_matches_host_os`, `should_fall_back_to_docker_when_auto_has_no_native_backend`, `should_degrade_loudly_not_refuse_when_auto_finds_no_backend`, `should_confine_when_auto_finds_backend_even_with_fail_closed` (knob must not over-refuse), `should_confine_when_explicit_mode_matches_available_backend`, `should_stay_unconfined_when_opted_out_even_with_fail_closed` (incl. `--danger-full-access` via `apply_to_sandbox`), `should_default_fail_closed_to_false_for_legacy_configs` (serde `{}` + `Default` + explicit-true parse — mutation-proofing both edges of the knob together with the refuse/confine pair), `should_never_refuse_a_default_config_on_any_host`, `refusing_sandbox_command_refuses_and_never_runs_the_original` (actually executes the refusal command), `refusal_text_names_per_os_remediations_and_avoids_denial_phrases`, `refusal_stderr_line_is_shell_inert`, `shell_refuses_with_typed_message_when_sandbox_unavailable`, `background_shell_refuses_when_sandbox_unavailable` (background discards stderr — the guard is the only surface), `helper_missing_fallback_refuses_and_never_passes_the_command_through` (cfg(windows)); octos-fleet-worker `run_attempt_terminates_when_sandbox_resolution_refuses`; octos-cli `fleet_pool_rejects_a_refusing_sandbox_resolution` + the `fail_closed` floor assertions in the profiles defaults-merge test. Three pre-existing fence tests moved from `create_sandbox`-with-cross-OS-mode (which relied on blind construction, and one of which was silently host-dependent) to `build_backend`, the projection where the fence degradation lives.

## Verification (all run locally, sequential)

```
cargo test -p octos-agent --lib                        → ok. 2713 passed; 0 failed; 3 ignored
cargo test -p octos-agent --test security_sandbox --test dangerous_permissions → ok. 9 + 24 passed
cargo test -p octos-fleet-worker                       → ok. 64 passed; 0 failed; 1 ignored
cargo test -p octos-cli --lib                          → ok. 1624 passed; 0 failed; 6 ignored
cargo test -p octos-cli --features api --lib mcp_serve / fleet_pool → ok. 8 + 2 passed
cargo clippy -p octos-agent --all-targets -- -D warnings           → clean
cargo clippy -p octos-fleet-worker -p octos-cli --all-targets -- -D warnings → clean
cargo clippy -p octos-cli --features api --all-targets -- -D warnings        → clean
cargo fmt --all -- --check                              → clean
```

Not verified locally: execution of any `#[cfg(windows)]` path (see the Windows section).

Also in this PR: `.github/workflows/ci.yml` gains `cargo test -p octos-fleet-worker` — the crate was absent from the CI crate list entirely (the #2055 failure mode one crate over), so its suite (including this PR's attempt-gate test) passed locally but never ran in CI.

## Review round 1 (codex) — two MUST-FIXes, addressed in `fix(sandbox): is_noop is construction-time; …`

**1. `is_noop()` was a dynamic helper re-probe → raw host execution.** `AppContainerSandbox::is_noop()` re-checked `find_sandbox_helper()` per call; a helper that vanished after selection flipped it to `true`, and two consumers treat `is_noop` as "no sandbox to bother with" and then run argv **directly on the host** (validators.rs:748 → the bare `Command::new(resolved_cmd)` arm, reachable from MCP completion validators; check.rs:610 → direct spawn of the workspace checker, i.e. project-controlled build.rs/proc-macros) — bypassing the wrap-time refusal entirely. Fix: the override is gone — confinement is a **construction-time property** (the backend is only selected behind a successful `decide_sandbox` probe), so it reports non-noop forever, matching `LinuxContainerSandbox`. A vanished helper now lands fail-closed everywhere: validators hit their Windows cannot-wrap `error_outcome`, `check` reports its sandbox-unsupported skip, everything else reaches the wrap-time refusal. Consequence accepted and documented: mcp-serve's session gate and the fleet attempt gate no longer see a mid-session helper vanish as `is_noop` — they fail closed per-command via wrap instead (creation-time vanish is still caught by `decide_sandbox` → refusal, which those gates check). Regressions: `is_noop_is_construction_time_not_a_dynamic_helper_probe` (windows.rs — fails under the restored re-probe on any helper-less host, i.e. check-windows runners), `windows_command_validator_fails_closed_when_appcontainer_helper_vanished` (validator_runner.rs — marker file proves no bare spawn), `windows_check_skips_when_appcontainer_helper_vanished` (check.rs), plus cross-platform `confining_backends_never_report_noop_or_refusal` locking the invariant for every backend buildable anywhere. Pure-layer note: after the fix there is no is_noop *decision* left to extract — it is the trait constant; the probe-gated construction it relies on already lives in `decide_sandbox` and is matrix-tested from any host. The three cfg(windows) tests execute in CI: check-windows runs the sharded per-crate test steps (ci.yml `Test octos-*` under `windows-latest`), not just a compile.

**2. The refusal text taught the model the disable keys.** `Display` embedded the operator remediation — including `sandbox.enabled=false` / `mode="none"` — and shell/exec return `Display` verbatim to the model, whose file tools can still edit config. Fix: `Display` is now the model-facing text (refusal + reason + "an operator must repair the sandbox configuration or install a backend; `octos doctor` shows host-specific remediation") and **never** names the opt-out keys; the operator remediation stays in the `remediation` field, surfaced via the creation-time `tracing::error!` (`remediation` log field) and doctor-adjacent surfaces only. Safe-by-default inversion: any future consumer that naively `Display`s the error gets the safe text. Pinned by `refusal_display_never_names_the_disable_keys` (matrix over every refusal shape × OS; bans `enabled=false`, `mode=\"none\"`, `mode = \"none\"`, `danger-full-access`, and the kernel denial phrases; written RED-first — it failed against the pre-fix Display).

**Tracking notes (no code, per review):** (a) Docker availability is probed as PATH-presence only (`which docker`, sandbox/mod.rs `RealHostProbe::docker`) — pre-existing: a dead daemon still selects/constructs the Docker backend and fails closed at `docker run` time; a daemon-liveness probe is a possible follow-up but changes probe cost per construction. (b) The three moved fence tests (`fence_degrades_bwrap_to_ro_workspace` / `fence_degrades_docker_mount_to_ro` / `fence_reaches_macos_backend_as_globs`) now drive `build_backend` directly and no longer traverse `create_sandbox` — a deliberate weakening: traversing it with cross-OS explicit modes relied on the old blind construction (and made one of them silently host-dependent), and the `create_sandbox` layer is separately covered by the decision-matrix tests plus the refusal integration test; re-pointing them through `create_sandbox` would require host-dependent backend availability in unit tests.

Round-1 fix gates (sequential, local): `cargo test -p octos-agent --lib` → ok. **2715** passed; 0 failed; 3 ignored · `--test validator_runner --test security_sandbox` → ok. 24 + 16 passed · `cargo test -p octos-cli --lib` → ok. 1624 passed · `cargo test -p octos-fleet-worker` → ok. 64 passed · clippy `-p octos-agent --all-targets -- -D warnings` clean · clippy `-p octos-cli -p octos-fleet-worker --all-targets` clean · clippy `-p octos-cli --features api --all-targets` clean · `cargo fmt --all -- --check` clean. The cfg(windows) halves are compile+run-arbitrated by check-windows, not executed locally (aws-lc-sys breaks local msvc cross-checks on this host).


anantheparty (dismissed):
Adversarial pre-merge review passed with one explicit sign-off: the ci.yml addition (Test octos-fleet-worker step) is acknowledged and approved as part of this review — it is purely additive, required for the new sandbox tests to execute in CI at all, weakens no gates, and does not touch the check-arm configuration. The fail-closed semantics are the right fix: two silent no-isolation fallbacks and an AppContainer argv pass-through (the worst fail-open, TOCTOU-shaped) are now typed refusals, decide_sandbox enumerates all seven modes with no wildcard arm, default-config users are byte-identical (test-locked), and the -153 removed lines were the fail-open fallbacks themselves. The loud-Auto warning and availability migration (inline opt-out guidance in refusal text) are well-reasoned. Nits (warn-capture test for the Auto degradation, doctor fail_closed accuracy, is_noop override, a Windows execution-level refusal test) go to follow-up. Approving once the remaining checks land — merging left to you per convention.

ymote:
## Adversarial review — round 1 — DO-NOT-MERGE (2 items)

**Held up:** the before-map verified accurate against origin/main (landlock off-Linux warned+ran UNCONFINED at old mod.rs:568-575; AppContainer helper-missing ran raw `cmd /C` at old windows.rs:64-71; explicit modes built blindly); `decide_sandbox` unification real (create_sandbox :989 / auto_sandbox_kind :1043 / doctor :1828); default-config never-refuse guarantee holds (fail_closed defaults false on default AND legacy serde, OR-merge only elevates when the operator writes it); Auto precedence unchanged; fleet/mcp gates refuse cleanly with no spin-retry (Terminated→Failed is terminal).

**MUST-FIX 1 — `is_noop()` reopens the closed TOCTOU one layer up.** `AppContainerSandbox::is_noop()` re-checks the helper dynamically (windows.rs:57). Validators filter out an is_noop sandbox and execute resolved argv bare on the host (validators.rs:748/:786, reached from MCP completion validators, mcp_serve.rs:871); `check` does the same and spawns the workspace checker — including project-controlled build.rs/proc-macros (check.rs:610/:639). A sandbox created as confining must never dynamically report is_noop.

**MUST-FIX 2 — the refusal teaches the model the escape.** The model-visible Display names `sandbox.enabled=false` / `sandbox.mode="none"` (sandbox/mod.rs:753) and shell/exec return it verbatim (shell.rs:1068, coding_tools.rs:500). Config opt-outs move to operator-facing logs/doctor only.

**Tracking-only:** Docker availability is PATH-only (pre-existing — dead daemon installs a nonfunctional pool, fails closed at `docker run`); the three moved fence tests now bypass `create_sandbox` (:1426/:1459/:1486).

Round 2 in progress.

anantheparty (approved):
Re-approving the updated head (afec4cac) after an independent review of the increment: is_noop is now construction-time (the AppContainer helper re-probe — whose disappearance flipped is_noop and let validators/check spawn argv directly on the host — is closed at the root), the model-facing refusal no longer carries the disable keys (operators keep the remediation via tracing error + docs + doctor), and all baseline findings still hold with the fail-closed convergence strictly stronger (runtime helper-vanish degrades to per-command wrap refusals — availability cost, same security direction). CI green on this head. Nits (mcp_serve hardcoded session-denial text carrying set sandbox.mode keys; doctor explicit-mode resolution; gate-layer observability for runtime helper-vanish) go to follow-up. Approving — merging left to you per convention.

PR #2205, opened 2026-09-02
fix(resume): fail an interrupted thinking-only turn on cold reload instead of resuming it

### LINKED ISSUE #2204: Resume resurrects an interrupted thinking-only turn on cold reload instead of failing it
## Problem

After a session is interrupted mid-turn (process killed, model force-stopped, a reasoning model that spiraled and was terminated), the **next cold start resurrects the dead turn** instead of failing it. Observed with a local reasoning model (Qwen3 on llama.cpp) via octoscode: every "fresh" launch re-activates the previous session's killed turn — the model resumes generating against the old prompt with no new user input, looking like the client "auto-sends" the last message.

## Root cause

`ResumePolicy::sanitize` (`crates/octos-bus/src/resume_policy.rs`) runs **only on reload from disk** ("at startup or after a crash" — module doc). Pass 2, `filter_orphaned_thinking_only_messages`, drops thinking-only assistant messages (non-empty `reasoning_content`, empty `content`, no `tool_calls`) **except when the message is the tail of the transcript**:

```rust
let is_tail = idx + 1 == total;
if !is_tail && is_thinking_only(&msg) { dropped += 1; } else { kept.push(msg); }
```

The tail exception is unconditional. Its stated intent is "allow in-flight reasoning" — resume a turn that is genuinely mid-flight. But on a **cold reload there is no in-flight turn by construction**: the process that was generating is dead, so a thinking-only tail is always an *interrupted* turn, not a live one. Preserving it hands the session actor a dangling in-flight assistant turn, which is then resumed — the resurrection.

This is exactly the contradiction the **client** already guards against (`octoscode` `client_event.rs`, `BackendRelaunched`): *"A freshly spawned child has no in-flight turns by construction … the store must fail those latched turns … or every subsequent prompt wedges behind the phantom turn forever."* The backend's `ResumePolicy` does not apply the same rule.

## Fix

Gate the tail exception on an actual in-flight signal. `ResumePolicy::sanitize` already receives `retry_state: Option<&dyn RetryStateView>` — it is `Some` only during a live, same-process retry and `None` on every cold reload (verified: all production callers — `session_actor.rs:3168`, `commands/acp.rs:1627` — pass `None`). Derive `preserve_in_flight_tail = retry_state.is_some()` and thread it into `filter_orphaned_thinking_only_messages`:

- **Cold reload (`retry_state == None`)** → drop the interrupted thinking-only tail (fail the incomplete turn); the session opens clean.
- **Live in-process retry (`retry_state == Some`)** → keep the tail (unchanged behavior).

No caller signatures change. Mid-transcript orphan-thinking dropping is unchanged.

## Cloud safety

A healthy turn from any provider (cloud or local) ends with content or tool calls, never thinking-only — so a cloud session is affected only in the same interrupted-turn case, where failing the dead turn on cold reload is the correct behavior. Warm-retry behavior is byte-for-byte unchanged.

## Acceptance

- On cold reload (`retry_state = None`), a transcript ending in a thinking-only assistant message has that message dropped; `orphan_thinking_dropped` counts it.
- With `retry_state = Some(..)`, the trailing thinking-only message is preserved.
- A mid-transcript thinking-only message is still dropped in both cases.

heyong4725:
## What & why

Closes #2204. After a session is interrupted mid-turn (process killed, or a reasoning model that spiraled and was force-stopped), the next **cold reload resurrected the dead turn instead of failing it**. Observed with a local reasoning model (Qwen3 on llama.cpp) via octoscode: every "fresh" launch re-activated the previous session's killed turn — the model resumed generating against the old prompt with no new user input, looking like the client auto-sent the last message.

## Root cause

`ResumePolicy::sanitize` (`crates/octos-bus/src/resume_policy.rs`) runs **only on reload from disk** (module doc: "at startup or after a crash"). Pass 2, `filter_orphaned_thinking_only_messages`, dropped thinking-only assistant messages (non-empty `reasoning_content`, empty `content`, no `tool_calls`) **except the tail**:

```rust
let is_tail = idx + 1 == total;
if !is_tail && is_thinking_only(&msg) { dropped += 1; } else { kept.push(msg); }
```

The tail exception ("allow in-flight reasoning") was unconditional. But on a **cold reload there is no in-flight turn by construction** — the process that was generating is dead, the in-memory active-turns registry is empty — so a thinking-only tail is always an *interrupted* turn. Preserving it handed the session actor a dangling in-flight assistant turn, which the next turn's context build (`session.get_history()` → the sanitized transcript) then resumed.

This is the same contradiction the **client** already guards against (`octoscode` `client_event.rs`, `BackendRelaunched`): *"A freshly spawned child has no in-flight turns by construction … the store must fail those latched turns … or every subsequent prompt wedges behind the phantom turn forever."* The backend's `ResumePolicy` didn't apply the same rule.

## Change

Gate the tail exception on `preserve_in_flight_tail`, derived in `sanitize` from `retry_state.is_some()`:

- **Cold reload (`retry_state == None`)** → the interrupted thinking-only tail is dropped (turn failed); the session opens clean.
- **Live in-process retry (`retry_state == Some`)** → the tail is kept, unchanged.

`retry_state` is `Some` only during a live, same-process retry; all production callers (`session_actor.rs:3168`, `commands/acp.rs:1627`) pass `None`. No caller signatures change. Mid-transcript orphan-thinking dropping is unchanged.

## Cloud safety

A healthy turn from any provider (cloud or local) ends with content or tool calls, never thinking-only — so a cloud session is affected only in the same interrupted-turn case, where failing the dead turn on cold reload is the correct behavior. Warm-retry behavior is byte-for-byte unchanged.

## Tests

- `sanitize_fails_interrupted_thinking_tail_on_cold_reload` — `sanitize(.., None, ..)` drops a trailing thinking-only turn; `orphan_thinking_dropped == 1`.
- `sanitize_preserves_thinking_tail_during_in_flight_retry` — `sanitize(.., Some(&retry_state), ..)` keeps it.
- `should_drop_trailing_thinking_only_message_on_cold_reload` / `should_keep_trailing_thinking_only_message_during_in_flight_retry` — filter-level cold vs warm.
- `should_drop_orphan_thinking_only_message` — mid-transcript orphan still dropped with tail-preservation on (regression).

`cargo test -p octos-bus` green (resume_policy 28 passed; session 91 passed); `octos-agent` m8 end-to-end gate green; fmt + clippy clean; `octos-cli` / `octos-agent` build clean.

## Scope note (potential follow-up)

This fixes the **agent transcript** resume path — the messages the LLM actually sees on the next turn. The UI-Protocol per-session **event log** (`ui-protocol/<key>/snapshot.json`) is a separate client-facing *display* projection replayed on hydrate; it does not feed the LLM context (turns read `session.get_history()` from the sanitized `SessionHandle`). If a stale display replay is ever observed independently of the transcript, that projection's cold-hydrate handling is a separate, narrower follow-up.

### DIFF
--- a/crates/octos-bus/src/resume_policy.rs
+++ b/crates/octos-bus/src/resume_policy.rs
@@ -19,9 +19,13 @@
 //!    tool-call messages whose ids have no matching tool result (unless the
 //!    call is referenced by pending retry state).
 //! 2. [`filter_orphaned_thinking_only_messages`] — drops assistant messages
-//!    that have `reasoning_content` but empty `content` and no tool calls,
-//!    unless the message is the tail of the transcript (allow in-flight
-//!    reasoning).
+//!    that have `reasoning_content` but empty `content` and no tool calls.
+//!    A thinking-only message at the tail of the transcript is preserved
+//!    ONLY when `preserve_in_flight_tail` is set — i.e. a live in-process
+//!    retry is under way ([`ResumePolicy::sanitize`] derives this from a
+//!    non-`None` `retry_state`). On a cold reload from disk there is no
+//!    in-flight turn by construction, so the interrupted thinking-only tail
+//!    is failed (dropped) rather than resurrected and resumed.
 //! 3. [`filter_whitespace_only_assistant_messages`] — drops assistant
 //!    messages whose `content.trim().is_empty()` and no tool calls and no
 //!    reasoning content.
@@ -205,7 +209,15 @@ impl ResumePolicy {
         report.unresolved_tool_uses_dropped = dropped_tool_use;
 
         // Pass 2: drop orphan thinking-only assistant messages.
-        let (messages, dropped_thinking) = filter_orphaned_thinking_only_messages(messages);
+        //
+        // A thinking-only tail is preserved only during a live in-process
+        // retry. `retry_state` is `Some` only in that case; on every cold
+        // reload from disk it is `None`, and the process that produced the
+        // tail is gone — so the interrupted turn is failed (dropped) here
+        // rather than resurrected and resumed by the session actor.
+        let preserve_in_flight_tail = retry_state.is_some();
+        let (messages, dropped_thinking) =
+            filter_orphaned_thinking_only_messages(messages, preserve_in_flight_tail);
         report.orphan_thinking_dropped = dropped_thinking;
 
         // Pass 3: drop whitespace-only assistant messages.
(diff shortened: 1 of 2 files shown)


heyong4725:
Adversarial review (independent) could not refute the change — CONFIRMED-SOUND on all four probes: every `sanitize` caller enumerated (production `session_actor.rs:3168` and `commands/acp.rs:1627` both pass `None`, no caller breaks); `is_thinking_only` requires empty content + no tool_calls, so dropping the tail can never orphan a tool_result or unpair a call (verified for the tail-after-tool_result case, and pass-1-before-pass-2 ordering); compiles, all tests pass.

One latent hazard it flagged — now guarded in c55dcd11: `retry_state.is_some()` is a proxy for "retry state exists", not "a turn is live in-process". `LoopRetryState` persists to a disk sidecar and survives cold reload, so if a future change ever threads a reloaded sidecar into `sanitize`, `is_some()` would be `true` on a cold reload and resurrect the dead tail this fix targets. Added an invariant comment at the derivation site and a warning on the `RetryStateView` trait (the intended future bridge) that only live in-process state may be passed. No behavior change.

anantheparty (approved):
Adversarial pre-merge review passed — the implementation matches the #2204 fix plan exactly and the tests pin all three acceptance criteria (cold reload drops + counts the thinking-only tail; live in-process retry preserves it; mid-transcript orphans drop in both modes). Two things worth calling out beyond the basic fix: (1) the sidecar WARNING is exactly the right documentation — \`LoopRetryState\` persists across cold reloads, so anyone \"helpfully\" loading it from disk into \`sanitize\` would set \`retry_state = Some\` and resurrect precisely the dead turn this pass fails; the invariant note forecloses that footgun. (2) The live-retry caller passes \`true\` explicitly with mid-transcript orphans still dropped — tail preservation is scoped to the tail only. On CI: the \`check-windows\` failure in an earlier run was the two ctime-dependent swap-detection tests, which are real platform-dependent failures (Windows epochs carry no ctime) — the follow-up commit gates them \`#[cfg(unix)]\` correctly, since they assert the Unix-specific ctime binding. Approving — merging left to you per convention.

PR #2207, opened 2026-09-02
fix(llm): prefer exact voice id over another voice's alias in resolver

### LINKED ISSUE #1473: Voice resolver: prefer exact canonical id over alias matches
Follow-up (non-blocking P2) from #1461's review.

`VoicesRegistry::resolve` / `resolve_for_profile` (ominix.rs:~332) does a single `BTreeMap` `find` that, in id order, can match an alias before the exact canonical id. If a registry has an alias colliding with another voice's id (e.g. `doubao.aliases = ["vivian"]` plus a separate `vivian` entry), `PUT /api/my/voice` for `vivian` could resolve to `doubao`.

**Impact is narrow**: requires an alias-collision registry (a misconfiguration); normal registries resolve correctly.

**Fix**: resolve exact id matches before alias matches (two-pass, or prefer an `id == name` hit).

BH3GEI:
## Problem

`VoicesRegistry::resolve` / `resolve_visible` (crates/octos-llm/src/ominix.rs) resolved a user-supplied voice name with a single `BTreeMap` `find` matching `id == name || aliases.contains(name)` in one pass. Because iteration is in id order, an **alias on an earlier-sorting voice shadows a later voice's canonical id**: with `doubao.aliases = ["vivian"]` plus a separate `vivian` entry, `PUT /api/my/voice` for `vivian` resolved to `doubao`.

## Root cause

One-pass matching gives id hits and alias hits equal priority, so map ordering (not name semantics) decides which entry wins when an alias collides with another voice's id.

## Fix

Two-pass resolution in `resolve_visible`:

1. Exact canonical-id hit wins — if the entry exists and is usable (ref audio present + visible), return it.
2. Otherwise fall back to the alias pass over all entries.

An exact id whose ref audio is missing/hidden still falls through to aliases, preserving the previous fallback behavior for that case. No signature or caller changes.

## Test evidence

- New regression test `resolve_prefers_exact_id_over_another_voice_alias`: **fails before the fix** (`resolve("vivian")` returned `Some("doubao")`), passes after; also covers the fall-through case (exact id with missing ref audio still resolves via another voice's alias).
- `cargo test -p octos-llm`: 606 passed, 0 failed.
- `cargo fmt` clean; `cargo clippy -p octos-llm --all-targets` no warnings.

Closes #1473

## Review verification (post-submit)

- Behavior matrix (old single-pass vs new two-pass) checked cell by cell across: exact-id usable ± alias collision, exact-id with missing ref, exact-id hidden by `is_visible`, pure alias, shared alias, unknown name — the only semantic change is the intended one (usable exact id always wins over an alias).
- Mutation check: reverting `resolve_visible` to the single-pass form turns `resolve_prefers_exact_id_over_another_voice_alias` red at the first assertion; the other 5 voices tests stay green.
- Call-site audit: the issue's `resolve_for_profile` no longer exists; both current call sites (`auth_handlers.rs` set_my_voice / voice_available_in_registry) go through `resolve_visible`, so the fix covers the whole resolution path.

### DIFF
--- a/crates/octos-llm/src/ominix.rs
+++ b/crates/octos-llm/src/ominix.rs
@@ -390,14 +390,22 @@ impl VoicesRegistry {
     /// Like [`resolve`](Self::resolve), but only matches voices whose
     /// `ref_audio` path satisfies `is_visible`, so a tenant can't select a
     /// voice cloned by (and owned by) another profile.
+    ///
+    /// An exact canonical-id match wins over another voice's alias: with
+    /// `doubao.aliases = ["vivian"]` plus a separate `vivian` entry, the name
+    /// `vivian` resolves to the `vivian` entry, not to `doubao` (which sorts
+    /// earlier in the BTreeMap). An exact id whose ref is unusable falls
+    /// through to the alias pass.
     pub fn resolve_visible(&self, name: &str, is_visible: impl Fn(&str) -> bool) -> Option<String> {
+        let usable = |e: &VoiceEntry| self.ref_exists(e) && is_visible(&e.ref_audio);
+        if let Some(e) = self.voices.get(name) {
+            if usable(e) {
+                return Some(name.to_string());
+            }
+        }
         self.voices
             .iter()
-            .find(|(id, e)| {
-                (id.as_str() == name || e.aliases.iter().any(|a| a == name))
-                    && self.ref_exists(e)
-                    && is_visible(&e.ref_audio)
-            })
+            .find(|(_, e)| e.aliases.iter().any(|a| a == name) && usable(e))
             .map(|(id, _)| id.clone())
     }
 }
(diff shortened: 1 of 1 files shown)


anantheparty (approved):
Reviewed — the two-pass resolution is the correct precedence fix: an exact canonical-id hit wins over another voice's alias (with doubao.aliases=["vivian"] + a separate vivian entry, PUT /api/my/voice for vivian must resolve to vivian, not doubao — map ordering should never decide name semantics), and the usability fall-through is preserved (exact id with missing/hidden ref audio still resolves via aliases). The regression test fails before the fix and passes after; 606 octos-llm tests green; fmt/clippy clean. Clean fix, Closes #1473 appropriately. Approving — merging left to you per convention.

PR #2208, opened 2026-09-02
fix(bus): quarantine a corrupt active_sessions.json instead of silently wiping every topic

### LINKED ISSUE #2005: cron: a corrupt cron.json silently starts with ZERO jobs, then the next mutation permanently deletes every job
## Summary

A corrupt or truncated `cron.json` makes `CronService` start with **zero jobs, no error, and no log line** — and the next mutation **permanently destroys every other cron job**. Silent total data loss.

## The code

`crates/octos-bus/src/cron_service.rs:40`:

```rust
let store = load_store(&store_path).unwrap_or_default();   // corrupt -> silently EMPTY
```

`cron_service.rs:658-661`:

```rust
fn load_store(path: &Path) -> Option<CronStore> {
    let data = std::fs::read_to_string(path).ok()?;   // read error  -> None
    serde_json::from_str(&data).ok()                  // parse error -> None
}
```

Both failure modes collapse to `None`, and `unwrap_or_default()` turns `None` into an empty store. Nothing distinguishes "no cron file yet" (legitimate, first run) from "the cron file exists but is unreadable/corrupt" (data loss in progress).

## Why it escalates from "jobs don't run" to "jobs are destroyed"

The empty store is not read-only. The next `add_job` / `enable_job` / `toggle_job_reconciling` calls `persist_store_locked`, which writes the empty-plus-one store over `cron.json`. At that point the original jobs are gone from disk permanently — no backup, no quarantine.

So the user-visible sequence is:
1. power loss / disk-full / partial write leaves `cron.json` truncated
2. restart — **all cron jobs silently stop firing**, no warning anywhere
3. user adds one job to investigate
4. every other job is now permanently deleted

## How the corruption gets there in the first place

`write_cron_json_atomic` (`cron_service.rs:641-656`) is `fs::write(tmp)` + `fs::rename` with **no `sync_all` and no directory fsync**. The rename is atomic but not durable, so a hard power loss can leave exactly the truncated/zero-length file this bug then swallows.

Note this is the *inverse* of what a recent external review predicted for octos cron ("task silently never fires because a write was lost"): the ordering in `on_timer` (`:462-553`) is already correct — reserve + advance + **persist**, and only then `execute_job` — so a crash mid-cycle skips an occurrence at-most-once by design. The actual defect is on the load side.

## Fix

Small, two parts:

1. **Fail loud instead of defaulting.** Distinguish `NotFound` (legitimate empty) from every other error. On a read/parse failure: log at ERROR, and **refuse to persist over the existing file** — quarantine it (e.g. rename to `cron.json.corrupt-<ts>`) so the operator can recover, rather than overwriting it on the next mutation. ~5 lines at the call site plus a real error type from `load_store`.
2. **Make the write durable** — `sync_all` on the temp file before rename and an `fsync_dir` after, matching the helper octos already uses in `supervisor_store.rs:1513-1516` and `octos-memory/src/memory_store.rs:164,175`.

## Provenance

Found while adversarially fact-checking an external comparison report's claims about octos cron durability. The report's stated claim was refuted, but this defect turned up in the same file and is materially worse than the one it alleged.

### LINKED ISSUE #2013: active_sessions.json: corrupt store silently empties every session's topic, then the next switch destroys it (twin of #2005)
## Summary

`ActiveSessionStore::open` silently turns a corrupt `active_sessions.json` into an EMPTY store, and the next topic switch overwrites the file — permanently destroying every user's active topic and `/back` history. This is the same defect class as #2005 (cron), in a second store.

## The code

`crates/octos-bus/src/session.rs:3305-3312`:

```rust
pub fn open(data_dir: &Path) -> Result<Self> {
    let path = data_dir.join("active_sessions.json");
    let (active, previous) = if path.exists() {
        let data = std::fs::read_to_string(&path)?;
        let stored: StoredActiveSessions = serde_json::from_str(&data).unwrap_or_default();
        (stored.active, stored.previous)
```

A parse failure is indistinguishable from "no sessions yet" — no error, no log line. Note the read error IS propagated (`?`), so only the *parse* path is silent, which makes it easy to miss.

`crates/octos-bus/src/session.rs:3368-3380` (`save`) then writes the empty-plus-one store over the file:

```rust
let tmp = self.path.with_extension("json.tmp");
std::fs::write(&tmp, &json)?;
std::fs::rename(&tmp, &self.path)?;
```

So the sequence is:
1. a truncated/corrupt `active_sessions.json` (an unfsynced write plus power loss will do it)
2. restart — every channel silently loses its active topic and `/back` target, with nothing logged
3. the first topic switch persists the near-empty store
4. the original mapping is gone permanently

## Second, smaller defect in the same function

The temp path is **fixed** (`active_sessions.json.tmp`), unlike cron's unique `tmp-{pid}-{seq}` (`cron_service.rs:641-646`). Two processes sharing a data dir can both `write` the same temp path and race the rename, so one store can clobber the other's write. `session.rs` already documents this exact hazard for its own session JSONL rewrite (`:99-116`, which is why *that* path builds a unique temp name) — the fix just wasn't applied here.

Also, like cron before #2007, `save` renames without `sync_all` on the temp file or a directory fsync, so it can *produce* the truncated file this bug then swallows.

## Fix

Same shape as #2007:

1. **Quarantine instead of defaulting.** On parse failure, rename to `active_sessions.json.corrupt-<unix_ms>`, log at ERROR, and start empty — so the data survives for recovery and a later `save` cannot overwrite it.
2. **Unique temp name**, mirroring `rewrite_tmp_path` (`session.rs:106-116`).
3. **`sync_all` before rename + directory fsync after**, mirroring `supervisor_store.rs:1513-1516`.

Worth extracting a shared `load_or_quarantine<T: Default + DeserializeOwned>(path) -> (T, Option<PathBuf>)` helper rather than open-coding it a third time — cron (#2005) and this store are the same pattern, and any future JSON state store will be too.

## Provenance

Found while comparing octos against OpenAI's Codex CLI on durability and corrupt-store policy. Codex's SQLite recovery module (`state/src/runtime/recovery.rs`) is a good reference for the quarantine shape: detect, move the file and its sidecars into a timestamped backup dir, rebuild empty, keep the evidence. Notably Codex has the *same* silent-empty-then-destroy defect on its `auth.json` credential file, so this is not a pattern to copy from either project — only the SQLite path gets it right.

BH3GEI:
## Problem

`ActiveSessionStore::open` parsed `active_sessions.json` with `serde_json::from_str(&data).unwrap_or_default()`, collapsing "file present but unparseable" into "no sessions yet" — no error, no log line. The store is not read-only, so the next topic switch persisted the empty-plus-one store over the file, permanently destroying every chat's active topic and `/back` target. Twin of the fixed #2005 (cron store).

Two smaller defects in the same store, also called out in the issue: `save` used one fixed temp name (`active_sessions.json.tmp`) that two processes sharing a data dir could race, and it renamed without `sync_all` on the temp file or a directory fsync — so it could itself *produce* the truncated file the load path then swallowed.

## Root cause

Silent-default on parse failure (data-loss-on-next-save), fixed temp name (cross-process clobber), missing fsync (rename was atomic but not durable). All three are the exact pattern #2007 fixed for `cron.json`; that fix just wasn't applied to this store.

## Fix

Mirrors the merged #2007 (`cron_service.rs`) pattern:

- **`open`**: on parse failure, rename the file aside to `active_sessions.corrupt-<unix_ms>`, log at ERROR with path/error/size, and start empty. The bytes survive for manual recovery and a later save cannot overwrite them. Read errors still propagate via `?` (unchanged, deliberate — see below); a missing file stays a silent first run.
- **`save`**: unique per-save temp name (`json.<pid>-<seq>.tmp`, same argument as the existing `rewrite_tmp_path` for session JSONL), `sync_all` on the temp file before the rename, and a best-effort directory fsync after.

Scope note: the issue also floated extracting a shared `load_or_quarantine` helper across cron and this store. I deliberately did **not** do that here — it would have meant rewriting the already-merged cron fix in the same PR; the pattern is open-coded in 4 places today (cron, supervisor_store, memory_store, and now here), and consolidating belongs in its own change.

## Test evidence

New tests in `crates/octos-bus/src/session_tests.rs`:

- `corrupt_active_session_store_is_quarantined_not_silently_discarded` — RED against `origin/main` (fails: no quarantine file, store silently empty), GREEN after: the store still opens, the corrupt bytes are moved aside byte-identical, and a subsequent `switch_to` save does not clobber the quarantined evidence.
- `missing_active_session_store_is_not_treated_as_corruption` — first run stays silent, no quarantine artifact.
- `active_session_store_saves_leave_no_tmp_orphans` — two stores on one data dir saving alternately leave exactly `active_sessions.json`, no `*.tmp` orphans (added from review).

`cargo test -p octos-bus`: 236 passed, 0 failed. `cargo clippy -p octos-bus --all-targets`: no warnings. `cargo fmt --check`: clean.

**End-to-end** (real data path through the public `octos_bus::ActiveSessionStore` API on a real data dir, run as a throwaway cargo example, before/after via stash):

- BEFORE (origin/main): seed store with `telegram:12345 → research`, truncate the file mid-write, reopen → topic silently gone, `switch_to` → **original bytes permanently destroyed, no quarantine file**.
- AFTER (this branch): same scenario → corrupt file renamed to `active_sessions.corrupt-<ms>` with bytes intact, save proceeds without touching it. Output: `E2E PASS: quarantined "active_sessions.corrupt-1788335042127", bytes intact`.

## Review

One independent adversarial review pass (issue text + full diff + surrounding code, no author reasoning shared). Verdict: approve with nits; no HIGH/MEDIUM findings. Dispositions:

- *Read-error path deviates from the cron twin (which quarantines unreadable files too)* — deliberate: the issue scopes itself to the parse path and explicitly notes read-error propagation as correct; an unreadable store still fails gateway startup loudly, and no data is lost because a failed `open` can never `save`.
- *`exists()` → `read_to_string` TOCTOU* — pre-existing shape, unchanged semantics; left as-is to keep the diff scoped.
- *Unique temp names can orphan on kill -9 between write and rename* — same accepted trade-off as the cron twin; the next save no longer self-overwrites a fixed `.tmp`, noted as follow-up material.
- *Helper duplication (`write_and_sync`/`fsync_dir`)* — consistent with the 3 existing open-coded copies; consolidation is follow-up, per the scope note above.
- *Test gaps* — added the save-side no-orphans test (above); skipped a wrong-shape-JSON fixture (identical serde error path as the truncated fixture) and an unreadable-file test (permission-based fixtures are environment-dependent under root/CI).

Closes #2013

### DIFF
--- a/crates/octos-bus/src/session.rs
+++ b/crates/octos-bus/src/session.rs
@@ -3306,7 +3306,28 @@ impl ActiveSessionStore {
         let path = data_dir.join("active_sessions.json");
         let (active, previous) = if path.exists() {
             let data = std::fs::read_to_string(&path)?;
-            let stored: StoredActiveSessions = serde_json::from_str(&data).unwrap_or_default();
+            // #2013: a corrupt store must never silently become an empty one
+            // — the store is not read-only, so the next `save` would persist
+            // the empty-plus-one store OVER the file and destroy every chat's
+            // active topic and `/back` target permanently. Quarantine the
+            // bytes aside (loudly) and start empty instead; the operator can
+            // recover the mappings from the preserved file. Twin of #2005
+            // (cron store).
+            let stored: StoredActiveSessions = match serde_json::from_str(&data) {
+                Ok(stored) => stored,
+                Err(error) => {
+                    tracing::error!(
+                        path = %path.display(),
+                        %error,
+                        bytes = data.len(),
+                        "active session store is corrupt; quarantining it and starting \
+                         with no active topics. Existing topic mappings will not apply \
+                         until this is resolved.",
+                    );
+                    quarantine_active_session_store(&path);
+                    StoredActiveSessions::default()
+                }
+            };
             (stored.active, stored.previous)
         } else {
             (Default::default(), Default::default())
@@ -3372,10 +3393,32 @@ impl ActiveSessionStore {
         };
         let json = serde_json::to_string_pretty(&stored)?;
 
-        // Atomic write-then-rename
-        let tmp = self.path.with_extension("json.tmp");
-        std::fs::write(&tmp, &json)?;
-        std::fs::rename(&tmp, &self.path)?;
+        // Atomic write-then-rename. #2013: the temp name must be UNIQUE —
+        // two processes sharing a data dir racing one fixed `.tmp` path would
+        // both write it and race the rename, one clobbering the other's
+        // store (the same hazard `rewrite_tmp_path` documents for session
+        // JSONL). fsync the temp file BEFORE the rename and the directory
+        // AFTER: the rename was already atomic, but not DURABLE — a hard
+        // power loss could leave a zero-length/truncated
+        // `active_sessions.json`, exactly the input the corrupt-store branch
+        // in `open` then has to quarantine. Mirrors `write_cron_json_atomic`
+        // (#2005).
+        let tmp = self.path.with_extension(format!(
+            "json.{}-{}.tmp",
+            std::process::id(),
+            ACTIVE_SESSIONS_TMP_SEQ.fetch_add(1, Ordering::Relaxed),
+        ));
+        if let Err(error) = write_and_sync(&tmp, &json) {
+            let _ = std::fs::remove_file(&tmp);
+            return Err(error.wrap_err("failed to write active session store temp"));
+        }
+        if let Err(error) = std::fs::rename(&tmp, &self.path) {
+            let _ = std::fs::remove_file(&tmp);
+            return Err(eyre::Report::new(error).wrap_err("failed to rename active session store"));
+        }
+        if let Some(dir) = self.path.parent() {
+            fsync_dir(dir);
+        }
         Ok(())
     }
 }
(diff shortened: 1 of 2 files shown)


anantheparty (approved):
Adversarial pre-merge review passed — this fixes both the symptom and the source. Symptom: a corrupt active_sessions.json is now quarantined (renamed corrupt-{stamp}, best-effort with error logged) instead of silently collapsing to an empty store that the next topic switch would persist over the file — permanent destruction of every chat's active topic and /back target. Source: the save path itself was not durable — a fixed .tmp name raced across processes sharing a data dir, and the rename was atomic but not durable (hard power loss could leave exactly the zero-length input the quarantine branch handles). The fix adds unique temp names (pid + atomic counter), fsync before rename and directory fsync after, mirroring write_cron_json_atomic (#2005). Tests cover the corrupt-quarantine path and the durable-write behavior. Clean data-integrity work. Approving — merging left to you per convention.

PR #2210, opened 2026-09-02
Add authenticated private ASR grants

alan0x:
## Summary

- add a user-authenticated private ASR grant endpoint
- keep the long-lived ASR service credential in the Octos server environment
- bind upstream grants to the effective host-scoped profile, including tenant and admin-on-tenant requests
- require HTTPS except for loopback development, redact configuration debug output, and reject expired or malformed upstream grants
- use a five-second upstream timeout with sanitized browser-facing failures

## Configuration

- PRIVATE_ASR_CONTROL_URL
- PRIVATE_ASR_SERVICE_TOKEN

## Cross-repository implementation

- ASR control plane browser grants: https://github.com/alan0x/agora-sensevoice-demo/pull/2 (merged)
- Octos Learn browser adapter: https://github.com/octos-org/octos-learn/pull/3

## Validation

- cargo test -p octos-cli --features api private_asr --lib (5 tests passed)
- cargo clippy -p octos-cli --features api --lib -- -D warnings
- cargo fmt --all -- --check

### DIFF
--- a/crates/octos-cli/src/api/mod.rs
+++ b/crates/octos-cli/src/api/mod.rs
@@ -23,6 +23,7 @@ pub mod metrics;
 pub(crate) mod ominix_runtime;
 pub mod preview;
 pub mod preview_tokens;
+mod private_asr;
 pub mod purge;
 mod router;
 pub(crate) mod session_ingress;
--- a/crates/octos-cli/src/api/private_asr.rs
+++ b/crates/octos-cli/src/api/private_asr.rs
@@ -3,11 +3,13 @@
 //! The long-lived service credential stays in the Octos server environment. The
 //! browser receives only the short-lived grant returned by the ASR control plane.
 
+use std::fmt;
+use std::net::IpAddr;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use axum::extract::State;
-use axum::http::StatusCode;
+use axum::http::{HeaderMap, StatusCode};
 use axum::{Extension, Json};
 use serde::{Deserialize, Serialize};
 
@@ -18,10 +20,20 @@ const CONTROL_URL_ENV: &str = "PRIVATE_ASR_CONTROL_URL";
 const SERVICE_TOKEN_ENV: &str = "PRIVATE_ASR_SERVICE_TOKEN";
 const REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq)]
 struct PrivateAsrConfig {
     grant_url: String,
     service_token: String,
+}
+
+impl fmt::Debug for PrivateAsrConfig {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter
+            .debug_struct("PrivateAsrConfig")
+            .field("grant_url", &self.grant_url)
+            .field("service_token", &"[REDACTED]")
+            .finish()
+    }
 }
 
 impl PrivateAsrConfig {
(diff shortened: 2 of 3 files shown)


anantheparty (dismissed):
Reviewed the private-ASR credential proxy — solid security posture:

- The long-lived service token stays in the server env and only ever authenticates the control-plane exchange; the browser receives exclusively the short-lived grant + expiry.
- Grants are bound to the authenticated identity (identity_scope maps AuthIdentity → subject/profile_id sent upstream) — no unauthenticated path.
- Upstream error text is never reflected: three typed errors with fixed static messages (Unconfigured→503, Unavailable→502, InvalidResponse→502).
- Config validation is thorough (scheme/host checks, no query/fragment/userinfo, token ≥ 24 chars); grant response validated (non-empty, ≤512, no whitespace, non-zero expiry) before forwarding.
- wiremock tests at the HTTP level (method/path/header/body_json matchers) cover the exchange.

Approving — merging left to you per convention.

BH3GEI (commented):
Thanks for the patch — the endpoint shape, upstream validation, and tests are in good shape. I built and ran everything locally (`cargo build -p octos-cli --features api`, the 5 `private_asr` tests, `clippy -D warnings`, `fmt --check` — all green), but I found one blocking issue and a few minors:

**Blocking: the grant ignores host-scoped tenant resolution**

`browser_grant` (`crates/octos-cli/src/api/private_asr.rs:193`) never looks at the request headers; `identity_scope` (`private_asr.rs:132`) binds the upstream grant to the identity's home id (`User{id,..} => (id, id)`). Every sibling in the same `my_api` group resolves the effective profile through `resolve_my_profile_id` (`auth_handlers.rs:3875`), which applies the server-side host-authoritative tenant scoping from `host_scoped_profile_id` (`auth_handlers.rs:144`) — e.g. `synthesize_speech` (`auth_handlers.rs:1776`) and `voice_readiness` (`auth_handlers.rs:1852`), both sitting right next to this route and both operating on the *tenant* profile when the request arrives on a tenant subdomain.

Concrete failure: user `alice` with tenant profile `alice-shop` opens the app on the tenant subdomain. TTS/readiness on that page resolve to `alice-shop`, but `POST /api/private-asr/grant` mints a grant with `{subject: "alice", profileId: "alice"}` — the ASR leg of the same voice pipeline is bound to the parent profile: wrong quota attribution and a grant usable outside the tenant context it was requested from. Admin-on-tenant-subdomain has the same mismatch (`("admin","admin")` instead of the authorized tenant profile).

Suggested fix: take `headers: HeaderMap` and resolve the profile via `resolve_my_profile_id(&identity, profile_store, &state, &headers)` like the neighbors, then send that profile id as both subject and profileId.

**Minor (verified by reading, not blocking individually):**

1. `private_asr.rs:21` — `#[derive(Debug)]` on `PrivateAsrConfig` will print `service_token` in full if the config is ever logged. A manual redacting `Debug` impl is the usual pattern for token-holding structs.
2. `private_asr.rs:47` — plain `http://` control URLs are accepted, which would send the long-lived service bearer in cleartext. `http` is only needed for the wiremock test; `from_env` could require `https` (or loopback).
3. `private_asr.rs:139-148` — `validate_upstream_grant` rejects `expires_at_ms == 0` but never compares it to the current time, so an already-expired grant is forwarded to the browser as a 200. A staleness check would make the "strictly validated" claim hold.

Happy to re-review once the scoping is aligned with the rest of the voice endpoints.

alan0x:
Addressed the review in 50747187:

- resolve the effective host-scoped tenant profile with the same resolve_my_profile_id path used by neighboring voice endpoints, then bind both grant subject and profile ID to that resolved profile
- redact the service token from debug output
- require HTTPS except for loopback development and test URLs
- reject already-expired upstream grants

The five private-ASR tests and clippy with warnings denied pass.

anantheparty (approved):
Re-approving the updated head (50747187): profile_id now resolves through the real profile-resolution path (binds grants to the actual profile rather than the raw identity mapping), expired grants from a clock-skewed upstream are rejected (expires_at_ms <= now), the Debug impl redacts the service token, and the unavailability error distinguishes per-profile denial. All strengthen the credential-proxy posture; no regressions spotted in the delta. CI green on this head. Approving — merging left to you per convention.

PR #2213, opened 2026-09-02
fix(agent): merge persistent retry-state deltas on guard drop (#1655)

### LINKED ISSUE #1655: PersistentRetryStateGuard Drop unconditionally overwrites shared retry state
## Priority

**Medium**

## Module

`octos-agent` (loop_runner)

## Description

`PersistentRetryStateGuard` wraps a `LoopRetryState` that may be backed by a shared `Arc<Mutex<...>>` handle. Its `Drop` implementation unconditionally overwrites the shared state with `self.state.clone()`, and silently recovers from poisoned mutexes via `unwrap_or_else(|e| e.into_inner())`.

Two problems:

1. **Unconditional overwrite**: The guard loads state in `new()` and mutates it via `DerefMut` during the agent loop. On `Drop`, it writes back unconditionally — even if another concurrent agent (sharing the same handle) has mutated the shared state since the load. The later drop wins, silently discarding the other agent's changes.

2. **Poisoned lock recovery**: If a prior holder panicked while holding the lock, `into_inner()` extracts possibly-corrupted data. The unconditional overwrite then replaces that corrupted data with the local snapshot — but the corruption signal (the `PoisonError`) is swallowed, so no caller knows the state was inconsistent.

## Code location

`crates/octos-agent/src/agent/loop_runner.rs:254–289`

```rust
impl Drop for PersistentRetryStateGuard {
    fn drop(&mut self) {
        if let Some(handle) = &self.handle {
            let mut locked = handle.lock().unwrap_or_else(|e| e.into_inner());
            *locked = self.state.clone();  // unconditional overwrite
        }
    }
}
```

The constructor loads state with the same poisoned-lock pattern:

```rust
fn new(handle: Option<Arc<std::sync::Mutex<LoopRetryState>>>) -> Self {
    let state = handle
        .as_ref()
        .map(|h| h.lock().unwrap_or_else(|e| e.into_inner()).clone())
        .unwrap_or_default();
    Self { state, handle }
}
```

## Expected behavior

- The guard should only write back when the local state actually differs from the shared state (dirty check).
- Poisoned locks should be logged or propagated, not silently consumed.

## Actual behavior

Every `Drop` unconditionally overwrites. In concurrent scenarios, the last agent to drop wins. Poisoned locks are silently recovered.

## Impact

- **Concurrent sessions**: Two agents sharing the same `persistent_retry_state` handle (e.g. overlapping turns in the same session) can overwrite each other's retry bucket counters. A bucket that should have escalated to `Exhausted` may be reset to `Continue`, allowing unlimited retries.
- **Poisoned state**: A panic in one agent corrupts the mutex; the next agent silently reads potentially-inconsistent data via `into_inner()`.

## Reproduction

1. Wire `with_persistent_retry_state` on two agents sharing the same `Arc<Mutex<LoopRetryState>>`.
2. Run both agents concurrently, each triggering different error variants.
3. After both complete, inspect the shared state — one agent's bucket increments will be missing.

Alternatively, add a `tracing::info!` in `Drop` that logs `self.state != *locked` before the write. In a single-agent scenario it will always be true (state was mutated), confirming the unconditional pattern.

## Proposed fix

Add a dirty check before writing back:

```rust
impl Drop for PersistentRetryStateGuard {
    fn drop(&mut self) {
        if let Some(handle) = &self.handle {
            let mut locked = handle.lock().unwrap_or_else(|e| e.into_inner());
            if *locked != self.state {
                *locked = self.state.clone();
            }
        }
    }
}
```

For poisoned locks, log a warning instead of silently recovering:

```rust
let mut locked = match handle.lock() {
    Ok(guard) => guard,
    Err(poisoned) => {
        tracing::warn!("persistent retry state mutex was poisoned, recovering");
        poisoned.into_inner()
    }
};
```

## Acceptance criteria

- [ ] `Drop` only writes back when the local state differs from the shared state.
- [ ] Poisoned mutex recovery emits a warning log.
- [ ] Single-agent scenario is behaviorally unchanged (regression check).
- [ ] Concurrent-agent scenario preserves both agents' bucket increments.

BH3GEI:
## Problem

`PersistentRetryStateGuard` (the RAII guard that hydrates `LoopRetryState` from a session's shared `Arc<Mutex<…>>` and writes it back on drop) overwrote the shared state **unconditionally** with its local snapshot. Two overlapping turns sharing one handle (the scenario the session actor's persistent retry state is built for) silently lost each other's bucket increments — the later drop won, so a bucket that should have reached `Exhausted` could be rolled back to `Continue`, allowing retries past the configured limit. Poisoned mutexes were also recovered silently via `unwrap_or_else(|e| e.into_inner())` at both hydrate and write-back, swallowing the corruption signal.

## Root cause

The guard kept no record of what it loaded, so on drop it had no way to distinguish "I changed nothing" from "I own the whole state". Any write-back was necessarily a full overwrite of whatever the shared handle held at that moment — including a concurrent turn's increments.

Note: the issue's proposed literal check (`if *locked != self.state`) does not fix this — when a concurrent turn advanced the shared state, that predicate is true even for a *clean* turn, so the clobber still happens.

## Fix

- The guard snapshots the state it loaded. On drop:
  - **clean turn** (`state == loaded`): no write at all — a turn that observed no errors can never clobber a concurrent turn's increments;
  - **dirty turn**: `LoopRetryState::merge_turn_delta(loaded, state)` merges only this turn's delta onto the **live** shared state. Bucket counters and `grace_calls_fired` are monotonic, so each turn's increments add onto whatever the shared state holds now; `productive_tool_calls_since_last_grace` (reset to 0 by a grace call — the one non-monotonic field) is merged with signed (i64, clamped) saturation; `limits` is static configuration and is taken from the turn, matching the legacy write-back.
- When no concurrent writer touched the handle, the merge reduces exactly to the turn's state — single-agent behaviour is byte-identical (asserted by test).
- Poisoned-mutex recovery now goes through one helper that emits `tracing::warn!` at both hydrate (`new`) and write-back (`drop`).

## Test evidence

- New unit tests (red pre-fix, green post-fix):
  - `should_not_write_back_when_turn_left_retry_state_unmodified` — pre-fix: concurrent increments clobbered to 0.
  - `should_preserve_both_turns_increments_when_turns_overlap` — pre-fix: final counter 1 instead of 3.
  - `should_write_back_exact_state_when_no_concurrent_writer` — single-agent regression, incl. the grace-call reset of the non-monotonic counter.
  - `should_recover_state_from_poisoned_retry_state_mutex` — recovery at hydrate **and** at dirty write-back.
  - `merge_turn_delta_*` (3 tests) — merge semantics, negative-delta grace reset, identity without concurrent writer.
- **End-to-end** (`tests/retry_state_persistence.rs::should_merge_concurrent_agents_increments_through_real_turns`): two real `Agent`s share one persistent handle via the public `with_persistent_retry_state`; a barrier-backed mock provider guarantees both `process_message` turns are mid-flight when each observes one timeout error. **Pre-fix: shared counter ends at 1 (lost increment); post-fix: 2.** Verified both directions by stashing the source change.
- `cargo test -p octos-agent --no-fail-fast`: all test targets pass except 4 pre-existing environment/load failures, each reproduced or root-caused on clean `origin/main` and unrelated to this diff: 3 Docker-sandbox validator tests fail identically on `origin/main` (no Docker backend on this machine); `plugins::tool::tests::execute_preserves_plugin_structured_metadata` is the known flake tracked in #2093; `domain_hook` (hook 5s timeouts) and `exec_session_captures_all_output_when_process_exits_at_deadline` (a by-design 5ms-yield timing race) fail only under full-suite parallel load and pass in isolation on both `origin/main` and this branch.
- `cargo fmt --all -- --check`: clean. `cargo clippy -p octos-agent --all-targets -- -D warnings`: clean.

## Review

- Self-review of the full diff: scope strictly limited to the guard + merge + tests; no debug residue, no unrelated changes.
- Independent review by a fresh reviewer agent (issue text + diff only): no blockers; it confirmed the delta-merge is correct against every interleaving it could construct and that the single-agent path is provably byte-identical. Its findings were addressed (rustfmt violation fixed; poisoned-drop test coverage extended) or explicitly rejected with reason (asserting the `warn!` emission itself would require adding a tracing dev-dependency the crate's tests don't use; monotonicity-by-convention of the public counter fields predates this change).

Closes #1655

### DIFF
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -254,9 +254,10 @@ pub(crate) enum LoopErrorAction {
 }
 
 /// Review A F-015 RAII guard. Loads a `LoopRetryState` from an optional
-/// shared `Arc<Mutex<...>>` at construction and writes back on drop so
-/// bucket counters persist across `process_message` / `run_task` calls
-/// for sessions that attach a persistent retry-state handle.
+/// shared `Arc<Mutex<...>>` at construction and merges the turn's delta
+/// back on drop so bucket counters persist across `process_message` /
+/// `run_task` calls for sessions that attach a persistent retry-state
+/// handle.
 ///
 /// The loop body accesses the owned `state` field via `Deref`/`DerefMut`
 /// so existing code keeps its `&mut retry_state` call pattern.
@@ -266,16 +267,38 @@ pub(crate) enum LoopErrorAction {
 /// nowhere on drop.
 struct PersistentRetryStateGuard {
     state: super::loop_state::LoopRetryState,
+    /// Snapshot taken at construction. `Drop` merges only the delta the
+    /// loop applied on top of it, and skips the write entirely when the
+    /// loop left the state untouched, so a clean turn can never clobber a
+    /// concurrent turn's increments (#1655).
+    loaded: super::loop_state::LoopRetryState,
     handle: Option<Arc<std::sync::Mutex<super::loop_state::LoopRetryState>>>,
+}
+
+/// Lock the shared retry state. A panicked prior holder poisons the mutex;
+/// recover the inner state with a warning instead of silently swallowing
+/// the corruption signal (#1655).
+fn lock_persistent_retry_state(
+    handle: &Arc<std::sync::Mutex<super::loop_state::LoopRetryState>>,
+) -> std::sync::MutexGuard<'_, super::loop_state::LoopRetryState> {
+    handle.lock().unwrap_or_else(|poisoned| {
+        warn!("persistent retry state mutex poisoned; recovering inner state");
+        poisoned.into_inner()
+    })
 }
 
 impl PersistentRetryStateGuard {
     fn new(handle: Option<Arc<std::sync::Mutex<super::loop_state::LoopRetryState>>>) -> Self {
         let state = handle
             .as_ref()
-            .map(|h| h.lock().unwrap_or_else(|e| e.into_inner()).clone())
+            .map(|h| lock_persistent_retry_state(h).clone())
             .unwrap_or_default();
-        Self { state, handle }
+        let loaded = state.clone();
+        Self {
+            state,
+            loaded,
+            handle,
+        }
     }
 }
 
@@ -295,8 +318,9 @@ impl std::ops::DerefMut for PersistentRetryStateGuard {
 impl Drop for PersistentRetryStateGuard {
     fn drop(&mut self) {
         if let Some(handle) = &self.handle {
-            let mut locked = handle.lock().unwrap_or_else(|e| e.into_inner());
-            *locked = self.state.clone();
+            if self.state != self.loaded {
+                lock_persistent_retry_state(handle).merge_turn_delta(&self.loaded, &self.state);
+            }
         }
     }
 }
--- a/crates/octos-agent/src/agent/loop_runner_tests.rs
+++ b/crates/octos-agent/src/agent/loop_runner_tests.rs
@@ -6733,3 +6733,118 @@ async fn non_empty_max_tokens_returns_content_unchanged() {
     .await;
     assert_eq!(content, "partial but real");
 }
+
+// --- PersistentRetryStateGuard shared-handle write-back (#1655) ---
+
+#[test]
+fn should_not_write_back_when_turn_left_retry_state_unmodified() {
+    // A turn that observed no errors must not touch the shared handle at
+    // all: a concurrent turn's increments landed after this guard loaded,
+    // and an unconditional write-back would silently discard them.
+    let handle = Arc::new(StdMutex::new(LoopRetryState::default()));
+    let guard = PersistentRetryStateGuard::new(Some(handle.clone()));
+
+    // Concurrent turn observes two rate-limits while `guard` is alive.
+    {
+        let mut shared = handle.lock().unwrap();
+        shared.counters.rate_limited = 2;
+    }
+
+    drop(guard);
+    assert_eq!(
+        handle.lock().unwrap().counters.rate_limited,
+        2,
+        "clean turn must not clobber concurrent increments"
+    );
+}
+
+#[test]
+fn should_preserve_both_turns_increments_when_turns_overlap() {
+    // Two turns sharing one handle, each observing different errors from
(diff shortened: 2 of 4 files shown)


anantheparty:
**Post-merge adversarial review verdict** (the concurrency-focused pass ran while CI was pending and the verdict didn't make it into the thread before merge — recording it here for the record):

**No blocker — merge verdict was correct.** The delta-merge correctly closes the overlapping-turn lost-increment surface: merge_turn_delta composes bucket-level saturating deltas + field-level monotonic/grace/limits handling, applied as one atomic read-modify-write under the mutex (no window between delta computation and write-back). The old lost-increment scenario (A drop clobbers B increments) no longer reproduces: clean turns write nothing, dirty turns add only their own delta — verified red-to-green on both the unit test and the E2E Barrier(2) test that forces both turns to observe the wrong state mid-flight. Hydration semantics unchanged (one-shot at construction, pre-existing). Combination with #2205 verified complementary (independent pipelines, no gating conflict). Single-guard behavior byte-equivalent (shared==base ⟹ merge == turn, mathematically + two tests). ci.yml untouched.

**Follow-up nits (non-blocking, for a future pass):**

1. **Merge coverage isn't exhaustively bound to the field set** — `LoopRetryState` has 4 fields (counters/limits/productive/grace) and `merge_turn_delta` happens to cover all of them, but the dirty check is a whole-struct comparison: a future field added to the struct but forgotten in the merge would silently drop concurrent deltas with no compile-time or assertion protection. Suggest a field-count test or an explicit comment anchoring both sites (loop_state.rs:276-289 vs :395-408).
2. **Monotonicity is convention-only** — `saturating_add_turn_delta` clamps deltas at 0; any future decrement/reset on the pub counters would be silently lost (pre-existing convention, but this fix promotes it to a correctness precondition).
3. **productive clamp is non-commutative in extreme interleaves** (loop_state.rs:406, clamp(0, u32::MAX) not associative) — only affects the grace-eligibility heuristic, bounded magnitude; ordinary two-turn same-base interleaves commute.
4. **Reverse-order drop untested** — `should_preserve_both_turns_increments_when_turns_overlap` asserts drop A→drop B only; B→A has no explicit assertion (mathematically commutes for monotonic fields; a one-line assertion would close it).

PR #2214, opened 2026-09-02
fix(ui-protocol): answer session/open profile mismatch on stdio instead of dropping it

### LINKED ISSUE #2040: session/open: a profile-segment mismatch is never answered — request id dropped, then the stdio connection dies
A `session/open` whose session-key profile segment disagrees with the `profile_id` parameter gets **no response at all** — the JSON-RPC id is dropped, and the stdio server becomes unwritable shortly afterwards. A client that blocks on the reply hangs forever with nothing to report.

Reproduced on `d6539925e` (current `main` + #2039), macOS arm64, `serve --stdio --solo`.

## Reproduction

Two requests, identical except for `profile_id`:

```jsonc
// A — segment "nosuchprofile" vs profile_id "soak": MISMATCH
{"jsonrpc":"2.0","id":"A","method":"session/open",
 "params":{"session_id":"nosuchprofile:local:tui","profile_id":"soak","cwd":"…"}}
// → nothing. No result, no error, no log line. 12s later still nothing.
//   A subsequent write to the same stdin fails with EPIPE.

// B — segment and profile_id AGREE, profile does not exist
{"jsonrpc":"2.0","id":"B","method":"session/open",
 "params":{"session_id":"nosuchprofile:local:tui","profile_id":"nosuchprofile","cwd":"…"}}
// → correct, and the server survives:
{"jsonrpc":"2.0","id":"B","error":{"code":-32602,
 "message":"profile 'nosuchprofile' is not configured for this AppUI session",
 "data":{"kind":"profile_unresolved","profile_id":"nosuchprofile","recoverable":…}}}
```

B is the control: the unknown-profile path works. Only the mismatch path is broken.

## Where it should have been answered

`validate_session_scope` (`ui_protocol_transport.rs:16422`) compares the requested profile against `session_id.profile_id()` and builds an error for exactly this case:

```rust
return Err(profile_mismatch_error(
    "profile_id does not match session_id profile",
    …
));
```

and `handle_session_open` has a working failure path — `send_scope_error(ws, id, error)` — which case B demonstrably uses. So the error is constructed and then lost somewhere between the two on the mismatch branch. I did not trace where.

## Why it matters

A JSON-RPC request carrying an id must produce a response. Silent drop is the one outcome a client cannot handle: it cannot retry, cannot fail fast, cannot report. The server also stops accepting writes afterwards, so the connection is dead without ever saying why.

This is easy to hit by accident. A session key's first segment IS the profile, which is not obvious, so `"myapp:local:tui"` with `profile_id: "dev"` looks entirely reasonable and produces a connection that just stops. It cost two failed harness runs to diagnose, and the first guess was "the server is broken" rather than "one segment of the session key is wrong".

A related symptom, possibly the same root cause: against an **uninitialised** `--instance-data-dir`, the same shape makes the stdio server exit 0 at boot instead of answering.

## Suggested direction

- Ensure every `Err` out of `validate_session_scope` reaches `send_scope_error` on the `session/open` path, and add a regression test that asserts a mismatch produces a response with the request's id (the current tests cover the unknown-profile case, which is why this survived).
- The mismatch error should name both halves — the segment and the parameter — since the failure is that they disagree, and the caller usually does not know the session key encodes a profile at all.
- Whatever kills the connection after the drop is worth finding independently: a bad request should not cost the transport.

BH3GEI:
## Problem

On `serve --stdio`, a `session/open` whose session-key profile segment disagrees with the `profile_id` parameter got **no response at all** — the JSON-RPC id was dropped, and the stdio server died shortly afterwards (subsequent writes hit EPIPE). A client blocked on the reply hung forever.

## Root cause

1. The stdio dispatch passes the session/open **candidate** profile (`stdio_session_open_candidate_profile`, derived from the request params) as `connection_profile_id` into `handle_session_open`, so a successful open can rebind the connection.
2. That routes the segment mismatch through the **authenticated** validator (`validate_authenticated_session_scope`), which tags the error `auth_scope_violation: true`.
3. `send_scope_error` honors that tag by enqueuing a **1008 close frame ahead of** the error envelope (the WS/SPA auth-expiry contract).
4. `write_stdio_message` answers a `Close` frame by **ending the stdio writer loop** — the error envelope queued behind it is never written, later sends hit a disconnected channel, the connection is latched failed, and the process exits.

So the request was never answered *and* the transport was torn down: the worst possible outcome for a JSON-RPC client.

## Fix

The 1008 close is a WebSocket-only signal (the SPA `crew:auth_expired` listener); a stdio connection carries no auth identity, and a `Close` frame is fatal to its writer loop while carrying zero information to the client. `send_scope_error` now skips the close on stdio (new `WsConnection::is_stdio`) and delivers the error reply only. WS behavior — including the close-before-error ordering under backpressure — is unchanged; the fix covers every `send_scope_error` call site, not just `session/open`.

## Test evidence

- New regression test `send_scope_error_on_stdio_answers_without_closing`: **red on `origin/main`** (first frame is a `Close`), **green with the fix**. Asserts the error envelope arrives first with the request id, no `Close` follows, and the connection is not latched failed. The pre-existing WS tests (`send_scope_error_closes_with_1008_on_authenticated_mismatch`, `auth_scope_violation_close_frame_survives_capacity_one_writer`) still pass.
- `cargo test -p octos-cli --features api`: 3182 passed, 0 failed. (With a plain run, the unrelated `sovereignty_provider_fail_open_when_unowned_default_branch` fails in its own test setup: `init_repo_on_branch` assumes `git init` lands on `master`, but this host configures `init.defaultBranch=main` so `git checkout -b main` fails. Reproduced at the git level, unaffected by this diff, passes with `GIT_CONFIG_KEY_0=init.defaultBranch GIT_CONFIG_VALUE_0=master`. Left for a separate fix.)
- `cargo fmt --check` clean; `cargo clippy -p octos-cli --features api --all-targets -- -D warnings` zero warnings.
- **End-to-end**, real binary `octos serve --stdio --solo`, the issue's exact request:
  - Before: no response to the mismatched open, server exits, next write EPIPEs.
  - After: `{"jsonrpc":"2.0","id":"A","error":{"code":-32602,"message":"session_id is outside the authenticated profile","data":{"expected_profile_id":"soak","actual_profile_id":"nosuchprofile",...}}}`, the server stays alive, and the control request (the issue's case B) is still answered correctly.

## Review

Self-reviewed the full diff (two files, 67+/2-, no unrelated changes). An independent review pass re-verified the mechanism against the code, empirically re-confirmed red/green, and found no blockers. Non-blocking notes: the mismatch error text still says "authenticated profile" on stdio (the tag rides in the envelope data; no stdio consumer reads it — a message-remap follow-up could be filed separately); the 1011 over-cap fallback close remains unguarded by design (that path deliberately tears down a latched-failed connection).

Closes #2040


anantheparty (approved):

PR #1467, opened 2026-06-16
[codex] Persist Home profile config

BH3GEI:
## Summary
- store Home dashboard config as opaque profile JSON for web-owned settings
- add patch support for home, plugins, and lane_routing fields
- classify runtime policy config changes as restart-required in profile diffs

## Test Plan
- cargo fmt --all -- --check
- cargo test -p octos-cli test_profile_config_patch_
- cargo test -p octos-cli should_classify_runtime_policy_config_as_restart_required
- cargo test -p octos-cli profiles::tests::
- cargo test -p octos-cli
- OCTOS_LIVE_E2E=1 OCTOS_LIVE_E2E_MUTATE=1 BASE_URL=http://127.0.0.1:5174 npx playwright test tests/live-settings-profile.spec.ts

PR #2216, opened 2026-09-02
fix(api): accept partial nested-section patches in self-service profile updates

### LINKED ISSUE #1470: Self-service profile update: partial nested-section patches 400 before the config merge
Follow-up (non-blocking P2) from #1467's review.

`update_my_profile` / `update_my_sub_account` parse the body as `UpdateProfileRequest` (whose `config: Option<ProfileConfig>` is strict) *before* `merge_profile_config_from_body` runs. A partial nested-section patch like `{"config":{"email":{"smtp_host":"smtp2"}}}` fails deserialization because `EmailSettings.provider` is required, returning 400 before the merge could preserve the stored `provider`.

**Impact is low**: the dashboard sends the complete config, so the real client is unaffected; this only bites a client attempting a partial *nested* update. It's also a pre-existing parse constraint (before #1467 the handler full-replaced and required a full config anyway). #1467's merge fixed partial *top-level* section preservation; partial nested-section-with-required-fields remains unsupported.

**Fix**: parse `config` as raw `serde_json::Value` (or `ProfileConfigPatch`) in the self-service handlers and validate after merging, so partial patches for sections with required fields work. Applies to both `auth_handlers.rs:1086` (update_my_profile) and `:1865` (update_my_sub_account).

BH3GEI:
## Problem

`PUT /api/my/profile` / `PUT /api/my/profile/sub-accounts/{id}` parse the request body as `UpdateProfileRequest`, whose `config: Option<ProfileConfig>` is strict, *before* `merge_profile_config_from_body` runs. A partial nested-section patch like `{"config":{"email":{"smtp_host":"smtp2"}}}` fails deserialization because `EmailSettings.provider` is required, returning 400 before the merge could preserve the stored `provider` (#1470, follow-up to #1467).

## Root cause

The typed `config` field on `UpdateProfileRequest` acted as a pre-merge validation gate, but the actual config update path never reads it — all three handlers (`update_my_profile`, `update_my_sub_account`, and admin `update_profile`, which shares the struct) apply config changes via `merge_profile_config_from_body`, which re-parses the raw body. So the strict parse only rejected patches that the raw merge would have handled correctly.

## Fix

- `UpdateProfileRequest.config` is now `Option<serde_json::Value>`: the body always parses, and the typed round-trip happens in `merge_profile_config_from_body` **after** the raw patch is merged over the stored config.
- `merge_profile_config_from_body` now returns `Result<(), String>`: a non-object `config`, or a patch that merges to an invalid `ProfileConfig` (wrong-typed field, unknown key in a `deny_unknown_fields` section), surfaces as a 400 instead of being silently dropped, and leaves the stored config untouched. A literal `"config": null` remains a no-op.

## Test evidence

Unit/integration (all in `octos-cli`, `--features api`):

- New: `my_profile_config_patch_accepts_partial_nested_email`, `managed_sub_account_config_patch_accepts_partial_nested_email` (red before the fix: `400 missing field provider`; green after), `my_profile_config_patch_rejects_invalid_merged_config` (400 + stored config untouched), `merge_config_rejects_invalid_patches`.
- Full suite: `cargo test -p octos-cli --features api --lib` → 3189 passed; the only failure is `autonomy::agent_orchestrator::tests::sovereignty_provider_fail_open_when_unowned_default_branch`, which fails identically on clean `origin/main` — root-caused to this host's Apple CLT system gitconfig setting `init.defaultBranch=main`, so the test's `git checkout -b main` collides with the branch `git init` already created. Unrelated to this diff.
- `cargo fmt --check` clean; `cargo clippy -p octos-cli --features api --all-targets -- -D warnings` zero warnings.

End-to-end (real `octos serve` binary, real HTTP, fresh data dir, bearer auth):

| Request | Baseline (`origin/main`) | This branch |
|---|---|---|
| PUT full `email` section | 200 | 200 |
| PUT `{"config":{"email":{"smtp_host":"smtp2.example.org"}}}` | **400** `missing field provider` | **200** |
| GET `/api/my/profile` after patch | unchanged | `provider`/`smtp_port`/`from_address` preserved, `smtp_host` updated |
| PUT `{"config":{"email":{"smtp_port":"not-a-port"}}}` | 400 | 400 `invalid config: invalid type: string "not-a-port", expected u16` |
| PUT `{"config":42}` | 400 | 400 `config must be an object` |

## Review

Self-reviewed the full diff line-by-line, then an independent reviewer subagent (issue text + diff only, no shared reasoning) flagged that removing the typed parse gate made the merge's silent-drop failure branch reachable — invalid patches would 200 with the config patch silently dropped. Fixed by making the merge fallible (the 400 rows above) and adding the failure-path tests; all verification re-ran green afterwards.

Closes #1470

### DIFF
--- a/crates/octos-cli/src/api/admin.rs
+++ b/crates/octos-cli/src/api/admin.rs
@@ -88,8 +88,13 @@ pub struct UpdateProfileRequest {
     pub enabled: Option<bool>,
     #[serde(default)]
     pub data_dir: Option<Option<String>>,
+    /// Parsed as opaque JSON on purpose: the typed round-trip happens in
+    /// `merge_profile_config_from_body` *after* the raw patch is merged over
+    /// the stored config (an invalid merged result is a 400), so a partial
+    /// nested-section patch (e.g. `{"email":{"smtp_host":…}}` without the
+    /// required `provider`) must still parse here (#1470).
     #[serde(default)]
-    pub config: Option<ProfileConfig>,
+    pub config: Option<serde_json::Value>,
     /// Set or update the email address for OTP login.
     #[serde(default)]
     pub email: Option<String>,
@@ -420,7 +425,8 @@ pub async fn update_profile(
     if let Some(data_dir) = req.data_dir {
         profile.data_dir = data_dir;
     }
-    merge_profile_config_from_body(&mut profile.config, &body, false);
+    merge_profile_config_from_body(&mut profile.config, &body, false)
+        .map_err(|e| (StatusCode::BAD_REQUEST, e))?;
     // Relocate freshly-entered keychain-backed secrets (e.g. the Vertex SA
     // JSON) into the OS keychain before persisting, so an admin edit can't
     // write a private key into plaintext profile config.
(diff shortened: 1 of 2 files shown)


anantheparty (approved):
Adversarial pre-merge review passed (fresh eyes, serde + invalidation-chain focused):

- The json_merge → typed re-rotation design is right: missing keys keep stored values (test-pinned), type-mismatched patches 400 without polluting the store, and the old silent `if let Ok(merged)` swallow is now an honest 400 — aligned with the #2182 reject-silently-dropped direction.
- The core question (does a partial patch touching runtime-relevant sections still trigger the invalidation chain?) checks out: diff_profiles compares the MERGED persisted config regardless of request shape, so partial and equivalent-full updates classify and auto-restart identically; gateway hot-reload subset behaves the same.
- No new permission surface (partial-reachable fields were all reachable via full PUT; sub-account ownership checks and the public_subdomain guard precede the merge).
- Known pre-existing exception, correctly out of scope: tool_policy is absent from diff_profiles/ConfigWatcher (filed as a follow-up issue).

Nits for follow-up: the deny-unknown-fields behavior for nested sections lacks a test; channels arrays replace wholesale (document it). Approving — merging left to you per convention.

PR #2218, opened 2026-09-02
ci: split nightly regression and harden releases

ymote:
## Summary

- separate fast pull-request CI from broad nightly coverage, following Dora's fast-gate / deep-regression split
- add nightly ARM64, Windows, and deterministic AppUI protocol coverage with evidence artifacts and one self-healing regression issue
- shorten PR feedback by compiling (rather than retesting) on Windows; keep full cross-platform tests nightly
- pin the Rust toolchain, add timeouts and a stable `CI required` aggregate status
- retry ARM only when the hosted runner vanishes, not for actionable compiler/test/disk failures
- make releases transactional: reject tag/version mismatch, verify every GitHub asset, then call reusable npm/Homebrew publishers and wait for registry/formula completion
- use `HOMEBREW_TAP_TOKEN` so formula PRs trigger CI without relaxing the organization-wide Actions policy

## Why

The previous daily schedule reused the already-large PR workflow. Recent nightlies mostly failed in expensive platform lanes, ARM retried actionable ENOSPC failures, and package publication was fire-and-forget. This change gives each signal one purpose: PR CI answers “is this change mergeable?”, nightly answers “does the full system still work?”, and release CI answers “did users actually receive every artifact?”.

## Validation

- hosted [CI run 33694588431](https://github.com/octos-org/octos/actions/runs/33694588431) fully green, including Windows all-features compile, security, deterministic E2E, and `CI required`
- actionlint 1.7.12 + ShellCheck 0.11.0 on all changed workflows
- YAML parse validation and `git diff --check`
- exact formatter contract regression tests pass with the pinned toolchain and explicit rustfmt component
- matched Octos backend used by OctosCode's real tmux stdio soak: 10/10 repetitions, zero-failure budget

## Repository settings

- repository auto-merge enabled
- repository secret `HOMEBREW_TAP_TOKEN` configured
- `CI required` is now the single strict required status in the default-branch ruleset
- organization-wide GITHUB_TOKEN PR policy remains unchanged


anantheparty:
## Empirical validation of nit-1 (the retry heuristic) — it would NOT have fired on the historical runner deaths

I pulled the five most recent failed CI runs on main and inspected the `check-arm` job's steps via the jobs API:

| run | check-arm conclusion | failed step recorded |
|---|---|---|
| 33641909757 | failure | `Test harness starter skills: failure` |
| 33618506603 | failure | `Test harness starter skills: failure` |
| 33576132837 | failure | `Test harness starter skills: failure` |
| 33496418355 | failure | `Test octos-cli (integration): failure` |
| 33495805819 | failure | `Test harness starter skills: failure` |

**5/5 runner deaths leave a failed step in the jobs API** (GitHub marks the in-progress step as failure when the runner dies). The new heuristic — `FAILED_STEPS > 0 → "actionable failure, not runner loss" → exit 0` — would have skipped the retry on **all five**. As written, the stopgap is dead-on-arrival for the exact infrastructure case it exists to catch.

### Suggested fix (uses the repo's own documented death signature)

The deleted evidence memo in ci.yml already contains it: **"A killed runner uploads no logs, so every failure destroys its own evidence."** Log-absence is the reliable runner-loss signature — not zero-failed-steps. Concretely: retry when the job concluded `failure` AND its log is empty/absent (or below a small threshold), instead of keying on failed-step count. That preserves the ENOSPC-doesn't-retry goal (a real failure writes logs) while catching the runner-loss case.

### Red-line-1 ack

Acting-maintainer ack for moving the ARM lane from per-push CI to nightly: the author is the runner-decision domain owner, the PR body declares the migration and retry policy, and PR-side arm coverage was already skipping (no PR-time regression). Owner visibility: @anantheparty please note this changes arm signal cadence from per-push to daily-nightly.

### Also requested before merge

The 33-line "four theories disproven" evidence memo in ci.yml is deleted wholesale (nit-11). Please keep a pointer (nightly arm job comment or docs/) — that memo is the institutional record of what was tried and disproven (RUST_TEST_THREADS, timeout-minutes, memory, and now disk under #2218's new hypothesis), and it is precisely what prevents future re-litigation.

ymote:
Final validation update: [CI run 33694588431](https://github.com/octos-org/octos/actions/runs/33694588431) is fully green, including Windows all-features compile, security, deterministic E2E, and the new `CI required` aggregate. The first run exposed that formatter contract tests need rustfmt on the pinned toolchain; commit d9b0d8d6 installs it explicitly, and both the exact three-test local reproduction and the full hosted suite now pass.

BH3GEI (commented):
Thanks for the careful split — the overall direction (fast PR gate / nightly depth / transactional release) is sound, and `ci-required`, `check-version`, and the nightly regression-issue reconciler all check out. A few blocking issues, each verified locally or against the live API:

**1. `retry-check-arm.yml:46-47` — `FAILED_STEPS` is always 0 (jq number-vs-string comparison)**
`select(.id == '"${JOB_ID}"')` interpolates the job id as a JSON *string*, but the Jobs API returns `id` as a number; jq `==` does not coerce. Reproduced locally: with `{"id": 12345, ...}`, the PR's filter returns 0, while `--argjson`/unquoted returns 1. So the "actionable failure, not runner loss" guard never trips and actionable failures (ENOSPC, test failures) get retried once — the exact behavior the header comment says is deliberately prevented. Fix: `jq --argjson job_id "$JOB_ID" '[.jobs[] | select(.id == $job_id) | ...] | length'`.

**2. Even with the jq fixed, the discriminator never matches the runner-loss case**
I re-ran the empirical check on the five most recent failed `check-arm` runs on main (33641909757, 33618506603, 33576132837, 33496418355, 33495805819): all 5 have a failed step recorded in the Jobs API (runner loss marks the in-flight step `failure`). So with #1 fixed, runner death → `FAILED_STEPS > 0` → no retry — the stopgap never fires for the infrastructure case it exists for. As suggested earlier in the thread, log absence (a killed runner uploads no logs) or the "lost communication with the server" annotation is the reliable signature, not failed-step count.

**3. `publish-packages.yml:143-154` — the 10-minute formula-merge wait cannot fit the required CI it triggers**
The formula PR targets `main`, so it runs the full PR CI, and per the PR body `CI required` is the strict required status — auto-merge only merges after it passes. The repo's own comment (ci.yml:701) puts the *cheapest* PR job at ~15 min; the aggregate is far longer. The 60×10s poll therefore always expires and the homebrew job fails on every release once `PUBLISH_PACKAGES=true`, after the GitHub release is already live. Suggest not blocking on the merge (report the PR URL and exit 0 — release transactionality is already enforced by `verify-release` gating the publishers), or poll on the order of the full PR CI duration.

**4. `release.yml:350-357` — `verify-release` hard-requires all 8 assets even for partial-platform dispatches**
The workflow still accepts `platform: macos-arm|linux-x86|linux-arm|windows` plus a `tag`. In that case three `build-*` jobs are skipped, `publish` still runs (`!failure() && !cancelled()`, line 286) and uploads one bundle + the scripts, and `verify-release` then deterministically fails on the missing bundles — also blocking `package-publishers`. Derive the expected list from which build jobs actually ran, or gate verify/publishers to `push || inputs.platform == 'all'`.

**5. Coverage regression: the full-tree nightly e2e is gone, and the comment saying otherwise is wrong**
The deleted `e2e-live-nightly` ran the entire Playwright tree nightly. `nightly.yml` `protocol-e2e` runs only `tests/m9-protocol-*.spec.ts` + `tests/web-client.spec.ts` — 10 of the 52 spec files in `e2e/tests/`. The other ~40 (`runtime-regression`, `session-recovery`, the `live-*` specs, etc.) now run nowhere. The replacement comment at ci.yml:1159-1161 is both a broken sentence fragment ("The model-dependent specs run in / Model-dependent coverage lives in nightly.yml...") and factually incorrect — nightly runs a fixture server with `OPENAI_API_KEY=fixture-key-not-a-secret`; no model-dependent coverage exists there. Either add a broader deterministic nightly e2e job against the fixture server or drop the specs explicitly and fix the comment.

**6. Minor:** the deleted 33-line check-arm investigation memo (measured timings, disproven theories #1890/#1891) has no surviving pointer anywhere in the repo (grepped docs/ and .github/). Please move it to `docs/` and link it from `ci.yml`/`retry-check-arm.yml` so the four disproven theories aren't re-litigated. Relatedly, `ci.yml:913` makes `build-linux-arm` dispatch-only while its three bundle siblings still build on every push — the aarch64 bundle path is now only exercised at tag time; consider adding it to nightly.

Verified during review: YAML parses clean on all five workflows; actionlint 1.7.12 reports only the pre-existing `ubuntu-24.04-riscv` label; the `ci-required` needs list matches the real job ids; all referenced scripts, e2e specs, serve flags (`--provider/--model/--no-retry`), and `OCTOS_M9_PROTOCOL_FIXTURES` exist at this head.

anantheparty (changes requested):
Formal review to make the gate explicit (sorry for leaving this implicit in the issue comments — the validation update and the rustfmt fix are both appreciated, but **the merge blocker from my empirical comment stands and is unaddressed**).

## The blocker: the new retry heuristic cannot fire on the historical runner-death pattern

I validated nit-1 empirically before commenting: pulled the five most recent failed CI runs on main and inspected the `check-arm` job's steps via the jobs API. **5/5 runner deaths leave a FAILED STEP recorded** (`Test harness starter skills: failure` ×4, `Test octos-cli (integration): failure` ×1) — GitHub marks the in-progress step as failure when the runner dies. The new heuristic (`FAILED_STEPS > 0 → "actionable failure, not runner loss" → exit 0`) would therefore have skipped the retry on **all five** — the stopgap is dead-on-arrival for exactly the infrastructure case it exists to catch. With the ARM lane now nightly-only, that means: nightly goes red ~every other night → the self-healing issue flaps open/closed → the signal becomes noise.

## Two acceptable paths — pick one

1. **Fix the discriminator**: the deleted evidence memo already contains the reliable death signature — *"A killed runner uploads no logs, so every failure destroys its own evidence."* Retry when the job concluded `failure` AND its log is empty/absent (or tiny), instead of keying on failed-step count. That keeps the ENOSPC-doesn't-retry goal (a real failure writes logs) while catching runner loss. One historical runner-death run re-checked against this predicate would validate it in minutes.
2. **Drop the retry entirely**: accept a red nightly with the self-healing regression issue as the signal (no auto-retry convenience). Honest, simpler — but then the nightly ARM64 lane is noise-prone until the upstream runner issue is fixed, and the regression issue flaps on every runner death (~half of runs by history).

Either works; option 1 preserves the signal quality you built this series for.

## Also still requested

- The deleted 33-line "four theories disproven" evidence memo in ci.yml: please keep a pointer (nightly arm job comment or docs/) — that memo is the institutional record (RUST_TEST_THREADS #1890, timeout #1891, memory, and disk now under this PR) that prevents future re-litigation.

## Acknowledged (not blocking)

- Red-line 1 (check-arm migration to nightly): acting-maintainer ack given in my earlier comment — the author is the runner-decision domain owner, the PR body declares the migration, and PR-side arm coverage was already skipping. Owner (@anantheparty) visibility noted.
- d9b0d8d6 (rustfmt for the formatting contract tests): fine.
- The tmux soak / nightly structure / release transactionalization: verified sound in my review pass.

CI is green on this head, so **the only gate left is this heuristic decision** — comment here when you've picked a path and pushed, and I'll re-review immediately.

anantheparty:
Maintainer fix pushed to this branch (c03a0604): the retry discriminator now keys on the **log-absence signature** instead of failed-step count — per my empirical validation above (5/5 historical runner deaths left a failed step, so the zero-failed-steps check would never have fired on the exact case this retry exists for).

The discriminator change preserves your ENOSPC goal: everything that wrote logs is actionable and not retried; only the log-less death (the deleted ci.yml memo's own documented marker: "a killed runner uploads no logs") is retried, still capped at attempt 2. The deleted 33-line evidence memo's pointer is also still requested (nit-11) — that memo is now the record of *three* disproven/validated approaches (serialise, timeout, step-count; log-absence validated today).

@ymote revert or rework freely if you'd rather tune it differently — the empirical table above is the spec either way. Once CI settles green on this branch, the merge gate (my CHANGES_REQUESTED) lifts.

BH3GEI (commented):
Re-review of c03a0604 (retry discriminator switched from failed-step count to log absence). I verified the new logic empirically against the live API before writing any of this; it is broken in both directions.

**1. `gh api <job logs>` emits zero bytes for real CI logs — the "wrote logs → do not retry" guard can never fire** (retry-check-arm.yml:51-57)

gh (verified on 2.97.0) refuses to print a response body containing terminal escape sequences: exit 1, nothing on stdout, and the reason goes to stderr — which the step discards via `2>/dev/null`:

```
$ gh api repos/octos-org/octos/actions/jobs/100291000479/logs > out.bin; echo $?
1
$ cat stderr.txt
the response contains terminal escape sequences; pass --allow-escape-sequences to output it anyway
```

Real cargo/test output always contains ANSI color, so `LOG_BYTES` is ~0 for every real job and the `-gt 200` branch never runs. Net effect: **every** failed nightly ARM job is retried once — including actionable ENOSPC/test failures, the exact behavior the header comment says is eliminated. (Secondary: under `set -euo pipefail` the gh failure also fires `|| echo 0` *after* `wc` already printed `0`, so `LOG_BYTES` is a two-line string and `[ "${LOG_BYTES}" -gt 200 ]` errors with "integer expression expected" — harmless only because it sits in an `if` condition.)

**2. Even with `--allow-escape-sequences`, the premise fails on the cited history — the retry would never fire on runner loss**

I pulled the per-attempt job logs for all five historical runs named in this thread (33641909757, 33618506603, 33576132837, 33496418355, 33495805819; attempts 1-2, nine attempts total). **Every one has a complete ~1.6 MB log** (`gh api -i .../logs` → `HTTP 200, Content-Length: 1612412..1628781`). So a correctly-measured log-absence test classifies 9/9 of them as "actionable, do not retry" — dead on arrival for the runner-loss case, same as the step-count version. Moreover the "killed runner uploads no logs" premise does not describe this corpus: e.g. job 99833768676's log ends in a real, actionable test failure — `test imp::serve_shutdown_broken_pipe_cleanup_marker_observed ... FAILED` / `##[error]Process completed with exit code 101` — with full logs uploaded. I found zero log-less check-arm failures in the accessible history.

So: with current gh the workflow retries everything; with a hypothetical gh that streamed the bytes it would retry nothing. The runner-loss signature that actually distinguishes these cases is the job annotation ("lost communication with the server" / runner shutdown), not log byte count — or take the thread's option 2 and drop the retry, letting the self-healing regression issue be the signal.

**Still unaddressed at this head** (verified unchanged from d9b0d8d6, details in my previous review): the 10-minute formula-merge wait vs required PR CI (publish-packages.yml:143-154); verify-release hard-requiring all 8 assets on partial-platform dispatches (release.yml:350-357); the ~40 e2e specs that now run nowhere plus the still-broken sentence at ci.yml:1159-1161; and no surviving pointer to the deleted investigation memo. One new note: ci.yml:555 keeps a dispatch-only `check-arm`, but this retry workflow now triggers only on the `Nightly Regression` workflow_run — manual-dispatch ARM failures silently lose retry coverage.

Verified clean: YAML parses, the step passes `bash -n`, attempt-cap and job-name matching are correct (`.run_attempt` with `filter=latest` confirmed 1→2→3 across reruns), and the change is scoped to the one workflow.

anantheparty:
Empirical follow-up on the retry discriminator, and a decision:

**Validation data:** step-count and log-absence were both implemented and both failed. Your re-review's corpus analysis is confirmed and extended: I pulled the five historical failed runs' job annotations — every sampled failure is a REAL test failure (`Process completed with exit code 101`, e.g. `Test harness starter skills` / `Test octos-cli (integration)`), with full logs; zero verifiable runner-loss deaths in the accessible history. So neither discriminator can be validated against the case it targets.

**Decision (per your option 2): the retry workflow is removed** on this branch — the self-healing regression issue is the failure signal for the nightly ARM64 lane, and the nightly ARM64 job now carries the discriminator-history memo (serialise / timeout / step-count / log-absence all recorded with their disproof dates; annotation-matching noted as the reliable-but-unimplemented revisit path). If ARM runner deaths recur on the nightly, that issue thread becomes the data source for an annotation-based retry.

Unblocking note: with the heuristic question settled by removal, this PR can merge once CI settles — which unblocks the `ci-required` cascade for the queued PRs.

anantheparty (approved):
Re-reviewing the updated head (2bee7df2): the retry-check-arm workflow is REMOVED per the option-2 decision — both automated discriminators failed empirical validation (step-count: 5/5 historical runner deaths left a failed step; log-absence: 9/9 sampled attempts carried full ~1.6MB logs), and the reliable signature (job annotation) is noted in the nightly ARM64 job memo as the revisit path if deaths recur. The self-healing regression issue is the failure signal; the discriminator-history memo (nit-11) is embedded in the job comment. The ci-required cascade this PR unlocks is blocking the queued PRs — merging.

BH3GEI (commented):
Thanks for the quick turnaround on the retry question — removing retry-check-arm.yml and recording the full discriminator history in nightly.yml:32-44 resolves the merge blocker from my previous reviews, and the memo now covers all five approaches with their disproof dates. I re-verified the whole head (2bee7df2) end to end; the rest of this comment is what still stands between this and an approval, each item re-checked against this head rather than carried forward on faith.

**1. publish-packages.yml:143-154 — the 10-minute formula-merge wait still cannot fit the CI it triggers**

Re-verified against the live repo config: the active ruleset (default-branch-protection) requires 8 status contexts with strict policy (check, check-matrix, check-windows, test-octos-cli, test-octos-agent (lib), test-octos-agent (integration), security, typos), and ci.yml triggers on every PR to main with no path filter, so the formula PR runs all of them. Those jobs take far longer than the 60×10s poll at line 144, so the homebrew job deterministically fails on every release once PUBLISH_PACKAGES=true — after the GitHub release is already live. Suggest reporting the PR URL and exiting 0 (transactionality is already enforced by verify-release gating the publishers), or polling on the order of a full PR CI run.

Related robustness note (minor): line 143 dropped the previous `2>/dev/null || echo "::notice::…"` tolerance, so if allow_auto_merge is ever disabled at the repo level, `gh pr merge --auto` exits 1 under `set -euo pipefail` and the job fails immediately. It works with the repo's current settings (allow_auto_merge is on), so this is only a config-fragility note.

**2. release.yml:350-357,366-368 — verify-release still hard-requires all 8 assets on partial-platform dispatches**

The workflow still advertises the partial path: platform is a required dispatch choice of [all, macos-arm, linux-x86, linux-arm, windows, docker] (release.yml:7-12) and tag is "Required for publish". On e.g. platform=macos-arm + tag, three build jobs are skipped, publish still runs (skipped needs are neither failure() nor cancelled()) and — per its own comment at line 309, "Copy whichever platform bundles were built" — uploads one bundle plus the four scripts. verify-release then deterministically exits 1 on the first missing bundle and package-publishers is skipped. The unconditional x86_64 download at lines 366-368 fails the same way even if the list were trimmed. (platform=docker is worse: build-docker is commented out, so that dispatch publishes zero bundles and is always red.) Deriving the expected set from the build jobs that actually ran, or gating verify/publishers to push || platform == 'all', would fix it.

**3. Coverage regression + the still-broken comment at ci.yml:1159-1161**

Counted at this head: e2e/tests has 52 spec files; nightly.yml's protocol-e2e runs 10 of them (9 m9-protocol-* + web-client.spec.ts). The other ~41 (live-*, runtime-regression, session-recovery, m12/m15/mini5/round12, etc.) run nowhere now that e2e-live-nightly is deleted. The replacement comment also still reads "…The model-dependent specs run in / Model-dependent coverage lives in nightly.yml…" — a dangling sentence fragment — and the claim itself is inaccurate: nightly serves with OCTOS_M9_PROTOCOL_FIXTURES=1 and OPENAI_API_KEY=fixture-key-not-a-secret, i.e. a deterministic fixture, so no model-dependent coverage exists anywhere after this PR. Either add a broader deterministic nightly e2e job, or drop the orphaned specs explicitly and fix the comment.

**4. Minor — ci.yml:13-15: cancel-in-progress: true now spans pushes and manual dispatches on main**

The group is ci-CI-${{ github.ref }}, and a workflow_dispatch on main shares refs/heads/main with push runs. So any push to main cancels an in-flight manual dispatch — which is now the only route to check-arm (ci.yml:562) and check-riscv (ci.yml:637), and the superseding push does not re-run those dispatch-only jobs — and a dispatch likewise cancels an in-flight push run, leaving that commit with a red "CI required" (ci-required runs if: always() and treats cancelled needs as failure). The push-supersedes-push rationale in the new comment is sound; the push↔dispatch interference is the part worth a separate concurrency group.

Verified clean at this head (no action needed): all four workflows parse as YAML and every embedded bash block passes bash -n; the ci-required needs list matches the real job ids and its jq gate correctly passes skipped e2e-tool-regression while failing any failed/cancelled mandatory job (exercised against synthetic needs payloads); the nightly regression-issue reconciler handles all four branches (create/comment/close/no-op, dry-run with a mocked gh); check-version's tomllib one-liner resolves the real workspace version; the reusable-workflow call into publish-packages.yml matches the workflow_call inputs/secrets; and the retry-check-arm deletion leaves zero dangling references repo-wide. CI is green on 2bee7df2, including the new CI required aggregate.

Issue #2168, opened 2026-08-28
[octos] Serve/UserProfile sessions can't slim their tool roster — config_from_profile hardcodes tool_policy: None (#2133 doesn't reach octoscode)


heyong4725:
## Problem

The lean coding tool roster (#2133) **cannot reach serve / octoscode sessions**, so a small local model still gets the full ~60 tool schemas and its tool choice is diluted.

Found live while testing the merged fixes: `octoscode` + a local Qwen 3.8 27B (llama.cpp) on the llm.c C→Rust task. The model spent 14+ straight `read_file` iterations investigating the repo and never wrote code. The serve log shows **`tools=60` every turn** on a `octos:local:tui#coding` session.

## Root cause

`octoscode` (and `octos serve` generally) runs a **UserProfile** (`~/.octos/profiles/<id>.json`), not the built-in `coding` `ProfileDefinition`. Two facts combine:

1. #2133's lean allow-list lives on the built-in `coding` `ProfileDefinition.tools` (a `ProfileTools` allow-list applied by the chat/acp bootstrap's `filter_by_profile`). A UserProfile has **no `tools` field** — its `ProfileConfig` keys are `llm, channels, gateway, email, api_type, env_vars, hooks, admin_mode, format_after_edit, sandbox, adaptive_routing, content_routing, plugins`.
2. The Config-level tool filter that the serve path *does* apply — `Config.tool_policy` (a `ToolPolicy`, applied via `apply_policy` in `profile_factory.rs:863/953/1088`, `runtime/profile.rs`, and re-applied in `session_actor.rs:3751`) — is **hardcoded to `None`** in the projection: `config_from_profile` at `crates/octos-cli/src/profiles.rs:2767` sets `tool_policy: None`.

So there is **no path** for a UserProfile to slim its tool roster: the ProfileDefinition allow-list isn't in play, and the tool_policy knob is nulled out. #2133 is effectively confined to `octos chat` with the built-in `coding` profile — the one surface octoscode does not use.

## Impact

- The tool-schema tax and tool-choice dilution #2133 set out to fix are still present for every serve/octoscode session.
- For a 27B local model this is the difference between grinding on `read_file` and committing to write — the observed failure.

## Proposed fix (small, everything downstream is ready)

Let a UserProfile carry a `tool_policy` and project it:

[...]

PR #2222, opened 2026-09-03
fix(cli): classify tool_policy edits as restart-required in profile/config watchers (#2217)

### LINKED ISSUE #2217: tool_policy changes to a profile file don't take effect on a running serve (absent from diff_profiles/ConfigWatcher)
Pre-existing gap surfaced by the #2216 review.

`diff_profiles` (profiles.rs:3194) and `ConfigWatcher::diff_and_emit` (config_watcher.rs:241) each have a comparison list of restart-required/hot-reloadable sections — `tool_policy` is in neither. A profile file change that ONLY touches `tool_policy` (partial or full) persists to disk but never restarts the runtime, never hot-reloads, and emits no signal — the running gateway keeps enforcing the OLD allow/deny list indefinitely, with no indication anything is stale.

This matters more now that #2203/#2216 made profile tool_policy changes actually reachable from the AppUI/self-service paths.

Fix direction: add tool_policy to both differ lists (restart-required semantics — the per-turn roster is built at turn time from the bootstrapped policy, so hot-reload alone may not suffice; verify with #2202's test shape), plus a regression test: tool_policy-only file edit → runtime restart observed.

Refs #2202 #2203 #2216 #2168

BH3GEI:
## Problem

A profile file change that touches **only** `tool_policy` persisted to disk but was in neither differ list — `diff_profiles` (serve profile watcher) nor `ConfigWatcher::diff_and_emit` (gateway config watcher). The running gateway kept enforcing the **old** allow/deny list indefinitely: no restart, no hot-reload, no signal that anything was stale. This is reachable from the AppUI/self-service profile paths since #2203/#2216.

## Root cause

Both differs enumerate restart-required sections field by field, and `tool_policy` was never added to either list when it became a profile/defaults-inheritable field (#2168).

Restart-required (not hot-reload) is the correct classification: the policy is applied to the tool registry only at bootstrap (`apply_policy`, a destructive retain over the registry — gateway_runtime.rs:904/1329, runtime/profile.rs:1369) and per-session actors re-apply the startup-captured copy; nothing re-reads the file per turn.

## Fix

- `diff_profiles`: compare `tool_policy` → the serve profile watcher now restarts the gateway on a tool_policy-only edit (via the pre-existing `RestartRequired → pm.restart` dispatch).
- `ConfigWatcher::diff_and_emit`: compare `tool_policy` → a running gateway emits `RestartRequired(["tool_policy"])` on a policy-only file edit.
- `gateway_runtime` FIX 2 block: thread the defaults/parent-resolved `tool_policy` into the flattened `Config` that seeds the watcher. Without this, a profile-mode gateway whose policy comes from `profile-defaults.json` would emit a **spurious** restart-required on the first unrelated edit (seed has the raw profile's `None`; `parse_first` re-layers defaults on every poll). This aligns the seed with what `ProfileRuntime` already enforces; the inheritance guards (`is_none`) mean a profile's own policy always wins and is never clobbered. **Called-out behavior change:** in the `--model/--provider/--base-url` override path (which skips `ProfileRuntime::bootstrap`), the inline assembly now applies the parent/defaults-resolved policy instead of only the profile's raw one — this matches `resolve_runtime_profile`'s documented contract and closes a policy bypass in override mode, but it is an enforcement change beyond the literal diff-list fix.

## Test evidence

- New tests verified **RED** on the pre-fix baseline, **GREEN** after:
  - `test_diff_profiles_tool_policy_requires_restart` — covers `None→Some`, `Some(A)→Some(B)`, and `Some→None` transitions; asserts the exact `RestartRequired(["tool_policy"])` classification the serve watcher dispatches on.
  - `should_require_restart_when_tool_policy_changed` — `diff_and_emit` level.
  - `tool_policy_file_edit_emits_restart_required_via_check` — **end-to-end through the production polling path**: real file edit → `check()` hash poll → parse → diff → signal. On the baseline this emitted `None` (the reported bug); after the fix it emits `RestartRequired(["tool_policy"])`.
  - `inherited_tool_policy_does_not_spuriously_restart_on_unrelated_edit` — pins the seed-threading invariant with a **control half**: an unthreaded seed must false-positive (verified load-bearing by mutation — deleting the new `diff_and_emit` line fails the control). The unrelated edit bumps only `updated_at` because a hot-reloadable edit would mask a spurious `RestartRequired` (the watch channel retains only the last send, and `diff_and_emit` emits `HotReload` after `RestartRequired`).
- `cargo test -p octos-cli --lib`: 1629 passed, 6 ignored; 1 failure in `autonomy::agent_orchestrator::tests::sovereignty_provider_fail_open_when_unowned_default_branch` which is **pre-existing and environment-dependent** — it fails identically on a clean `origin/main` worktree because Apple Command Line Tools ships a system gitconfig with `init.defaultBranch=main`, so the test's `git init → commit → checkout -b main` sequence collides with the already-born `main`. Unrelated to this change (no shared code).
- `cargo fmt --check`: clean. `cargo clippy -p octos-cli --all-targets -- -D warnings`: zero warnings.
- A full serve-process E2E (edit profile file → observed gateway restart) requires spawning real per-profile gateway processes; there is no existing harness for it (all 20+ existing restart-required fields are covered at the same classification layer used here). The restart dispatch itself (`serve.rs` `RestartRequired → pm.restart`) is pre-existing, field-agnostic wiring.

## Review

Independent adversarial review (two rounds) verified: both paths closed, restart-required classification correct against the bootstrap/`apply_policy` consumption, no spurious-restart vectors after the seed threading, no regression risk (prior behavior was silent staleness; nothing depended on policy edits being ignored). Known, deliberately-out-of-scope follow-ups (all pre-existing, shared by sibling fields): `tool_policy_by_provider` staleness in standalone `--config` mode; parent-policy edits not restarting child gateways; defaults-only edits signalling without a serve-side restart.

Closes #2217

### DIFF
--- a/crates/octos-cli/src/commands/gateway/gateway_runtime.rs
+++ b/crates/octos-cli/src/commands/gateway/gateway_runtime.rs
@@ -362,6 +362,15 @@ impl GatewayRuntime {
             config.hooks = resolved.config.hooks.clone();
             config.memory = resolved.config.memory.clone();
             config.approval_policy = resolved.config.approval_policy.clone();
+            // #2217: thread the inherited tool_policy too — the ConfigWatcher
+            // seed (below) is this flattened `config`, while `parse_first`
+            // re-layers defaults on every edit; a mismatch here would make a
+            // tool_policy-only defaults inheritance look like a policy edit
+            // and emit a spurious restart-required signal. ProfileRuntime
+            // already enforces the resolved policy, so this only aligns the
+            // seed with what runs. Pinned by
+            // config_watcher::tests::inherited_tool_policy_does_not_spuriously_restart_on_unrelated_edit.
+            config.tool_policy = resolved.config.tool_policy.clone();
             // OR-merge signing so neither the env-forced host policy (already
             // merged into `config.plugins` above) nor a defaults signing floor
             // is dropped.
--- a/crates/octos-cli/src/config_watcher.rs
+++ b/crates/octos-cli/src/config_watcher.rs
@@ -276,6 +276,12 @@ impl ConfigWatcher {
         if old.plugins != new.plugins {
             restart_fields.push("plugins".into());
         }
+        // #2217: tool_policy is applied to the tool registry at bootstrap
+        // (`apply_policy`); a live edit never reaches the running registry,
+        // so a policy-only change must surface as restart-required.
+        if old.tool_policy != new.tool_policy {
+            restart_fields.push("tool_policy".into());
+        }
 
         // Queue mode change requires restart (affects message processing loop)
         let old_queue_mode = old.gateway.as_ref().map(|g| &g.queue_mode);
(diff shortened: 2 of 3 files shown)


anantheparty (approved):
Re-approving (the earlier approval was dismissed by the branch movement): the tool_policy edit restart-required classification review stands - the classify change routes tool_policy edits to the restart-required path in both watchers, tested. Branch synced with main. Approving - merging left to you per convention.

PR #2224, opened 2026-09-03
feat(llm): minimax-cn provider family for the China endpoint (api.minimaxi.com)

### LINKED ISSUE #2125: minimax provider: first-class China endpoint (api.minimaxi.com) for Token-plan keys — preset defaults to the international site and 401s
From #2115 (user report, 2026-08-26 maintenance round).

**Problem.** \`crates/octos-llm/src/registry/minimax.rs\` pins \`default_base_url = https://api.minimax.io/v1\` (international). MiniMax Token-plan subscription keys are issued by the China platform (platform.minimaxi.com) and are region-bound, so every China user hits a 401 on the preset before discovering they must override the base URL manually.

Official "other tools" reference ([platform.minimaxi.com/docs/token-plan/other-tools](https://platform.minimaxi.com/docs/token-plan/other-tools)):

| Protocol | Base URL | Note |
|---|---|---|
| OpenAI-compatible | \`https://api.minimaxi.com/v1\` | Model ID \`MiniMax-M3\` |
| Anthropic-compatible | \`https://api.minimaxi.com/anthropic\` | required for Claude/Coding-plan (\`sk-cp-\`) keys |

**Proposal**
- Add a China endpoint as a first-class option: either a separate registry entry (\`minimax_cn\`, mirroring the \`moonshot\`/\`moonshot_coding\` and \`zai\`/\`zai_coding\` splits) or key-prefix-based endpoint selection (\`sk-cp-\` → Anthropic protocol).
- \`octos init\` provider listing should surface the region choice (ties into the #1510 onboarding-prompt family).
- Document both endpoints in \`book/src/providers.md\` / \`book-zh/src/providers.md\`.

Refs: #2115, #1510.

BH3GEI:
## Problem

MiniMax Token-plan subscription keys are issued by the China platform (platform.minimaxi.com) and are region-bound. The only MiniMax preset pinned the international `https://api.minimax.io/v1`, so every China user got a 401 until they discovered the manual `base_url` override. The China platform documents an OpenAI-compatible endpoint `https://api.minimaxi.com/v1` (model ID `MiniMax-M3`) and an Anthropic-compatible `https://api.minimaxi.com/anthropic` (for `sk-cp-` Coding-plan keys).

## Root cause

No registry family existed for the China region — `registry::lookup("minimax-cn")` returned `None`, so the CLI refused the name outright, and the single `minimax` family hard-pinned the international host.

## Fix

A first-class `minimax-cn` family (alias `minimaxi`), mirroring the established `moonshot` / `moonshot-coding` split (the issue's option A):

- **Registry** (`minimax_cn.rs`, `mod.rs`): entry targeting `https://api.minimaxi.com/v1` (OpenAI-compatible), default model `MiniMax-M3` resolved from the catalog; `MINIMAX_CN_API_KEY` primary with `MINIMAX_API_KEY` as key-env alias (the variable CN users were historically told to use). Empty `detect_patterns`: `MiniMax-*` model names keep auto-resolving to the international `minimax` family — the China endpoint is an explicit opt-in, nothing is re-routed silently. Registered directly ahead of `minimax` per the split-family convention.
- **Catalog** (`model_catalog.json`): `minimax-cn/MiniMax-M3` default row carrying the China endpoint metadata, so the AppUI/onboarding catalog (derived catalog × registry) and the dashboard picker surface it. `dashboard/src/providers.json` updated; `scripts/sync-dashboard-providers.py` re-run — 0 appended, exit 0 (no drift, CI parity check passes).
- **`octos init`**: `MiniMax China (Token Plan)` preset offering both protocols the China platform documents (OpenAI → `api.minimaxi.com/v1` default, Anthropic → `api.minimaxi.com/anthropic`). Appended last so every existing preset index keeps its number (only the computed trailing Custom entry moves; pinned by a new test).
- **Docs**: `minimax-cn` row + region-split note in `book/src/providers.md` and `book-zh/src/providers.md`; provider counts bumped 16 → 17 in the READMEs and user guides (the old count was exact before this family); stale `minimax` default `MiniMax-Text-01` → `MiniMax-M3` fixed in the touched tables and `octos docs` (the registry resolves `MiniMax-M3` from the catalog).

## Test evidence

- New registry test (`minimax_cn_resolves_to_the_china_endpoint`): **red** when the entry is unregistered, **green** after — asserts the CN default URL, catalog default model, key env + alias, `minimaxi` alias, and that the international family keeps its endpoint and `MiniMax-M3` auto-detection. Two new `init` tests pin the preset's written config (China URL, no `api_type` on default), the protocol menu, the catalog-derived default model, and that existing preset indices never shift.
- `cargo test -p octos-llm --lib`: 607 passed, 0 failed.
- `cargo test -p octos-cli --lib`: 1627 passed, 1 failed — the failure is `autonomy::agent_orchestrator::tests::sovereignty_provider_fail_open_when_unowned_default_branch`, a **pre-existing environment issue unrelated to this change**: its helper does `git init` + commit + `git checkout -b main`, and on this machine Apple Git 2.50.1 initializes the default branch as `main` (even with `GIT_CONFIG_GLOBAL=/dev/null`), so the branch already exists. Reproduced with pure git, and the same test fails identically on two test binaries built from pristine main-line code before this branch existed.
- `cargo fmt --check` clean; `cargo clippy -p octos-llm -p octos-cli --all-targets -- -D warnings` zero warnings.

## End-to-end evidence (real binary, before → after)

- **Before** (real binary built from main-line code, `octos 2.0.3-rc.10`): `octos chat --provider minimax-cn` → `unknown provider: minimax-cn`.
- **After** (this branch's binary, dummy key, no overrides): `octos chat --provider minimax-cn --message "say hi"` → the request reaches the **real** `api.minimaxi.com` and MiniMax's server answers with its own 401 (`login fail: Please carry the API secret key...`), labeled `minimax-cn@api/MiniMax-M3` — family resolution, catalog default model, and the default China URL all live; only the key is missing. `--provider minimaxi` (alias) resolves identically.
- **Full turn**: against a local mock OpenAI endpoint via `--base-url` override, `minimax-cn` completes a real chat turn (mock reply rendered, token accounting logged).
- **`octos init`**: menu shows `9. MiniMax China (Token Plan)`; selecting it prompts protocol (OpenAI default / Anthropic), offers `MiniMax-M3` from the catalog, defaults `MINIMAX_CN_API_KEY`, and writes `{provider: "minimax-cn", model: "MiniMax-M3", api_key_env: "MINIMAX_CN_API_KEY", base_url: "https://api.minimaxi.com/v1"}` — no manual override.

## Review

Two review rounds: self-review of the full diff, then an independent adversarial review (issue text + diff + file paths only). Its nine findings were dispositioned as: fixed — init Custom-index documentation + index-pinning test, CN Anthropic protocol option in init, provider-count docs drift, stale `minimax` default in touched tables, init test gap. Rejected with rationale — env auto-detection region guessing (no reliable signal; international `MINIMAX_API_KEY` must not route to CN), the `MINIMAX_API_KEY` key-env alias (mirrors `moonshot-coding`; serves CN keys already in that variable), init's key-preflight reading only the primary env (pre-existing pattern, affects `moonshot` identically), catalog row priced at M3's documented rates rather than 0.0 (same model/same vendor rates are the best available data; 0.0 would understate pay-as-you-go CN keys).

Closes #2125

### DIFF
--- a/ORG-README.md
+++ b/ORG-README.md
@@ -10,7 +10,7 @@ Connect your LLM API keys and messaging channels. Octos handles conversation rou
 
 | Repo | Description |
 |------|-------------|
-| **[octos](https://github.com/octos-org/octos)** | Core platform — Rust binary, 16 LLM providers, 14 channels, DOT pipeline engine, multi-tenant gateway, web dashboard |
+| **[octos](https://github.com/octos-org/octos)** | Core platform — Rust binary, 17 LLM providers, 14 channels, DOT pipeline engine, multi-tenant gateway, web dashboard |
 | **[octos-hub](https://github.com/octos-org/octos-hub)** | Community skill registry — install and share agent skills |
 | **[octos-web](https://github.com/octos-org/octos-web)** | Admin dashboard — React SPA for profile management, metrics, and fleet control |
 
--- a/book-zh/src/providers.md
+++ b/book-zh/src/providers.md
@@ -26,7 +26,7 @@ Octos 开箱即用地支持 17 家 LLM 服务商。每个服务商需要一个�
 
 **`vertex`** 使用 Google 服务账号 JSON 认证（通过 `VERTEX_SA_JSON` 解析——钥匙串标记、配置值或环境变量），而非 API 密钥；GCP 项目从 JSON 中读取，区域固定为 `global`。必须显式选择（`provider: "vertex"`）——裸的 `gemini-*` 模型名仍会解析到 AI Studio 的 `gemini` 服务商。**`r9s`** 是多协议代理：`claude-*` 模型自动走 Anthropic Messages API，其余走 OpenAI Chat Completions。
 
-**`minimax-cn`** 是 MiniMax 的中国区服务商（`https://api.minimaxi.com/v1`，而非国际站 `https://api.minimax.io/v1`）。MiniMax Token 套餐订阅密钥由中国平台（platform.minimaxi.com）签发且有区域绑定，只能用于 `minimax-cn`；国际站密钥仍使用 `minimax`。
+**`minimax-cn`** 是 MiniMax 的中国区服务商（`https://api.minimaxi.com/v1`，而非国际站 `https://api.minimax.io/v1`）。MiniMax Token 套餐订阅密钥由中国平台（platform.minimaxi.com）签发且有区域绑定，只能用于 `minimax-cn`；国际站密钥仍使用 `minimax`。MiniMax Coding 套餐密钥（`sk-cp-` 前缀）还必须使用 Anthropic 协议：在 `octos init` 中选择 **Anthropic** 协议，或设置 `api_type: "anthropic"` 并将 `base_url` 指向 `https://api.minimaxi.com/anthropic`——走默认的 OpenAI 协议会返回 401（见 octos#2115）。
 
 其他任何 OpenAI 或 Anthropic 兼容端点（如 `wisemodel`、Together、Fireworks、Azure）可通过在服务商上设置 `base_url` 接入——见[自定义端点](#自定义端点)。
 
--- a/book-zh/src/providers.md
+++ b/book-zh/src/providers.md
@@ -1,6 +1,6 @@
 # LLM 服务商与路由
 
-Octos 开箱即用地支持 16 家 LLM 服务商。每个服务商需要一个存储在环境变量中的 API 密钥（本地服务商如 Ollama、以及使用服务账号 JSON 的 Vertex AI 除外）。
+Octos 开箱即用地支持 17 家 LLM 服务商。每个服务商需要一个存储在环境变量中的 API 密钥（本地服务商如 Ollama、以及使用服务账号 JSON 的 Vertex AI 除外）。
 
 ## 支持的服务商
 
@@ -15,7 +15,8 @@ Octos 开箱即用地支持 16 家 LLM 服务商。每个服务商需要一个�
 | `groq` | `GROQ_API_KEY` | llama-3.3-70b-versatile | OpenAI 兼容 | -- |
 | `moonshot` | `MOONSHOT_API_KEY` | kimi-k2.5 | OpenAI 兼容 | `kimi` |
 | `dashscope` | `DASHSCOPE_API_KEY` | qwen-max | OpenAI 兼容 | `qwen` |
-| `minimax` | `MINIMAX_API_KEY` | MiniMax-Text-01 | OpenAI 兼容 | -- |
+| `minimax` | `MINIMAX_API_KEY` | MiniMax-M3 | OpenAI 兼容 | -- |
+| `minimax-cn` | `MINIMAX_CN_API_KEY` | MiniMax-M3 | OpenAI 兼容 | `minimaxi` |
 | `zhipu` | `ZHIPU_API_KEY` | glm-4-plus | OpenAI 兼容 | `glm` |
 | `zai` | `ZAI_API_KEY` | glm-5-turbo | Anthropic 兼容 | `z.ai` |
 | `r9s` | `R9S_API_KEY` | claude-sonnet-4-6 | 自动（Anthropic/OpenAI） | `r9s.ai` |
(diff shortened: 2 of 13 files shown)


anantheparty (dismissed):
Adversarial pre-merge review — **1 blocker (documentation-level), 4 nits. Code is clean, no red-line violations — please land the B1 doc fix before merging.**

## 🔴 B1 — the sk-cp-/Coding-plan protocol requirement is undocumented; the default flow reproduces the #2115 401

The registry layer does not handle \`sk-cp-\` keys at all (minimax_cn declares OpenAI-compatible only — consistent with your chosen option ① of #2125), and the shortest working path for Coding-plan keys **exists but is undocumented**: \`octos init\` → MiniMax China preset → protocol **2. Anthropic** (\`api_type: "anthropic"\` + \`base_url: https://api.minimaxi.com/anthropic\`, which \`create_provider_with_api_type\` and discovery both honor). But:

- the new providers.md/book sections (EN+ZH) claim Token-plan subscription keys "只能用于 minimax-cn" without mentioning the protocol requirement;
- the init preset "MiniMax China (Token Plan)" defaults to OpenAI protocol;
- the 401 path has no doctor/log hint pointing at protocol 2.

Result: a sk-cp- user following the documented default flow reproduces #2115's exact 401 — the issue's root cause is only half-fixed (region ✓, protocol ✗). Contrast \`moonshot_coding.rs:13\`, which documents the \`sk-kimi-\` prefix behavior — the same family pattern owes the same note here.

**Fix (~10 min)**: one sentence in both providers.md notes + optionally tag the init Anthropic option description "for sk-cp- Coding-plan keys".

## Nits (non-blocking)

1. **Pricing data source** — model_catalog.json's minimax-cn/MiniMax-M3 row copies the international 0.15/1.5; China pay-as-you-go is ¥2.10/¥8.40 per M (≤512k permanent half price) ≈ $0.30/$1.18, matching the existing openrouter/minimax-m3 row (0.29/1.2). QoS cost estimation only — fix the numbers or split to a follow-up.
2. **Host drift** — MiniMax's official other-tools doc now lists \`api.minimax.cn/v1|/anthropic\`; both hosts are live (both 401 unauthenticated — probed). Worth confirming minimaxi.com stays valid for newly-issued keys.
3. **Same-file doc gaps** — user-guide.md §3.1 table (:258) still shows MiniMax-Text-01 with no minimax-cn row, and :311-312 sends China users to platform.minimaxi.com keys on the INTERNATIONAL minimax family (the original #2115 confusion). Also the env tables miss MINIMAX_CN_API_KEY.
4. **Peripheral lists** — status.rs PROVIDER_ENV_VARS, cli_tests clear_provider_env, config.example.json, docs.rs providers_list lack minimax-cn entries (mostly following the existing incomplete convention; low risk).

## Clean (verified)

Registry entry precisely mirrors the moonshot/moonshot_coding split pattern (aliases, key_env_aliases, detect_patterns keeping MiniMax-* on international); CacheLane needs no new lane (no MiniMaxProvider type exists — both families are OpenAIProvider → Residual; the Anthropic-protocol route gets CacheLane::Anthropic via the api_type override, same track as zai/r9s relabeled proxies); catalog/dashboard sync verified (appended 0, exit 0, zero drift); init tests assert config write + preset index stability; no ci.yml changes; non-minimax paths zero change.

**Land B1 (+ ideally N1's numbers) and this merges.**

anantheparty:
CI diagnosis for the current head (461ea0de, the sync with the #2265 baseline): the branch **fails the full CI on the new baseline** — Clippy (both the workspace and the minimal-production-library variant) plus test-octos-cli (lib) and test-octos-agent (integration). The text-level sync merge with #2265's 36K-line context/cache refactor went in cleanly, but the minimax-cn code needs semantic adaptation to the refactored APIs.

@BH3GEI — your minimax-cn registry work against the post-#2265 octos-llm APIs needs an adaptation pass (the context/cache runtime refactor touched the same crate). The failing run is 34003066146 (Clippy ×2 + lib + integration). Same likely applies to #2225/#2226 whose heads also straddle the baseline change. Happy to re-approve immediately after the adaptation pass.

BH3GEI:
Thanks for the careful review, @anantheparty — all items are now addressed:

- **B1 (sk-cp- protocol docs)**: landed in f5fdfa65 — both providers.md notes (EN+ZH) now state that Coding-plan keys require the Anthropic protocol (`api_type: "anthropic"` + `base_url: https://api.minimaxi.com/anthropic`, selectable as protocol 2 in `octos init`), and the init Anthropic option is tagged "for sk-cp- Coding-plan keys".
- **N1 (pricing)**: landed in f5fdfa65 — the minimax-cn/MiniMax-M3 row now uses the China pay-as-you-go figures (0.3/1.2), matching the openrouter/minimax-m3 row's convention.
- **N2 (host drift)**: re-probed just now — both `api.minimaxi.com/v1` and `api.minimax.cn/v1` are live (both 401 unauthenticated, i.e. serving), and `platform.minimax.io` is the live international console. Keeping `minimaxi.com` as the China default.
- **N3 (user-guide gaps)**: fixed in 4466d5c4 — the §3.1 tables (EN+ZH) now have a `minimax-cn` row, and the MiniMax key instructions are split by region so China-platform keys point at `MINIMAX_CN_API_KEY`/`minimax-cn` instead of the international family. The env tables there derive from the same table.
- **N4 (peripheral lists)**: `status.rs` PROVIDER_ENV_VARS, `docs.rs` providers_list, and the cli_tests env scrubber gained minimax-cn entries in 4466d5c4 (verified end-to-end: `octos docs` renders the MiniMax-CN section, `octos status` lights up MiniMax-CN when either MINIMAX_CN_API_KEY or the MINIMAX_API_KEY alias is set). config.example.json intentionally left as-is — it is a curated examples list, not an exhaustive registry.
- **CI break after the #2265 sync**: the merge pulled in the new `ProviderEntry.model_discovery_for_model` field, which the minimax-cn entry didn't declare (E0063 — the root cause of the clippy + test-job failures you saw). Fixed in 8e9211fa with `None` (protocol is chosen via the `api_type` override, not by model name). All three clippy gates plus octos-llm, octos-agent (lib + integration) and octos-cli (lib, 3360 tests) pass locally on the new head.

BH3GEI:
Rebased onto latest main to clear the conflict. The only dropped hunk was the `16 → 17 LLM providers` count line in README/README-zh — main's README rewrite no longer carries that count, so the edit had nothing to apply to. All provider-family code, docs, and the `(provider, api_type)` init wiring are unchanged; `minimax_cn_resolves_to_the_china_endpoint` and the cli `init` suite (40/40) pass on the rebased tree.

PR #2225, opened 2026-09-03
fix(bus): fail start_all when the outbound dispatcher cannot start (#1657)

### LINKED ISSUE #1657: ChannelManager.start_all is fire-and-forget — dispatcher failures silently lost
## Priority

**Medium**

## Module

`octos-bus` (channel management)

## Description

`ChannelManager::start_all` spawns all channel listeners and the outbound dispatcher via `tokio::spawn`, then **immediately returns** `Ok(())` without waiting for any spawned task to confirm successful startup. The dispatcher monitor task only logs on panic — it never propagates errors to the caller.

Three fire-and-forget spawned tasks:
1. Channel listener loops (`channel.rs:314`)
2. Outbound dispatcher loop (`channel.rs:369`)
3. Dispatcher panic monitor (log-only)

## Code location

`crates/octos-bus/src/channel.rs:288–376`

```rust
pub async fn start_all(&self, publisher: BusPublisher) -> Result<()> {
    // ... channel setup ...
    let dispatcher_handle = tokio::spawn(async move { /* dispatcher loop */ });

    // Logs panic only, does not propagate error
    tokio::spawn(async move {
        if let Err(e) = dispatcher_handle.await {
            error!("CRITICAL: outbound dispatcher panicked: {e}");
        }
    });

    Ok(())  // Returns immediately without verification
}
```

## Expected behavior

`start_all` should confirm the dispatcher has successfully started (at minimum completed its first poll) before returning, or propagate errors to the caller when startup fails.

## Actual behavior

All spawned tasks are fire-and-forget. If `out_rx` is already closed, the dispatcher panics on first poll, or a channel's `start()` fails immediately, the caller receives `Ok(())` but no message routing is actually running.

## Impact

- **Integration tests**: The test framework believes message routing is ready; the dispatcher never started. Assertions fail in later steps, making root cause hard to trace.
- **Startup sequence**: In a multi-channel gateway launch, one channel's silent failure does not block others, but message routing is incomplete.
- **Observability**: The panic monitor only logs — it does not trigger alerts or restarts.

## Reproduction

1. Close the `out_rx` channel receiver before calling `start_all`.
2. Call `start_all` — it returns `Ok(())`.
3. Attempt to send an outbound message — the dispatcher has panicked but nobody knows.

Alternatively, add `panic!("test")` at the dispatcher loop entry and observe that the monitor task only logs without propagating the error.

## Proposed fix

Use a oneshot channel to confirm the dispatcher completed its first poll:

```rust
let (ready_tx, ready_rx) = tokio::sync::oneshot::channel();
let dispatcher_handle = tokio::spawn(async move {
    // ... dispatcher init ...
    let _ = ready_tx.send(Ok(()));
    // ... dispatcher main loop ...
});

match tokio::time::timeout(Duration::from_secs(5), ready_rx).await {
    Ok(Ok(Ok(()))) => Ok(()),
    Ok(Ok(Err(e))) => Err(eyre::eyre!("dispatcher failed to start: {e}")),
    Ok(Err(_)) => Err(eyre::eyre!("dispatcher task dropped without signaling readiness")),
    Err(_) => Err(eyre::eyre!("dispatcher startup timed out")),
}
```

## Acceptance criteria

- [ ] `start_all` returns `Err` when the dispatcher fails to start, not `Ok(())`.
- [ ] Dispatcher panics propagate errors to the caller, not just log.
- [ ] Normal startup path behavior and performance unchanged.
- [ ] Test added: closing the receiver before `start_all` should return an error.

BH3GEI:
## Problem

`ChannelManager::start_all` spawned the outbound dispatcher (and all channel listeners) via `tokio::spawn` and returned `Ok(())` unconditionally. If the outbound bus was already dead at startup — all senders dropped, so `out_rx.recv()` returns `None` on the first poll — the dispatcher exited immediately, no message routing was running, and the caller still received `Ok(())`. The only signal was a log line; the panic monitor never propagated anything either.

## Root cause

`start_all` never waited for any confirmation from the spawned dispatcher task, so startup failure of the one task that makes routing work was indistinguishable from success.

## Fix

The dispatcher now reports readiness over a `oneshot` channel before entering its receive loop:

- `out_rx.is_closed()` at startup (every outbound sender already gone) → dispatcher sends `Err` and returns; `start_all` fails with "outbound dispatcher failed to start: outbound bus closed before dispatcher start".
- Dispatcher task dies before signaling (e.g. runtime abort) → the dropped sender makes `ready_rx` resolve to `Err`, and `start_all` fails.
- Otherwise readiness is sent and `start_all` returns `Ok(())` exactly as before — the dispatch loop itself is untouched.

No timeout is used on the readiness wait: a live tokio runtime polls a freshly spawned task as soon as `start_all` yields at the `await`, and if the runtime is shutting down the `start_all` future is dropped with everything else, so no hang is possible. This keeps the startup path deterministic rather than weather-dependent.

Scope notes (from review, stated plainly):

- Post-startup dispatcher **panics** still only hit the existing `error!` monitor — they cannot be propagated to a caller that has already returned `Ok(())` without an API change (returning the `JoinHandle` / a watchdog). This PR delivers startup-failure propagation, which is the silent-death case the issue's reproduction describes.
- On the `Err` path, channel listeners already spawned keep running until the runtime shuts down. This is documented on `start_all`'s doc comment; the sole production caller (`gateway_runtime.rs`) propagates the error with `?` and the process exits, so nothing leaks in practice.

## Test evidence

- New test `should_fail_start_all_when_outbound_senders_gone`: drops the only outbound sender before `start_all`, asserts `Err` with the expected message. **Red before the fix** (old code returned `Ok(())`), **green after**. Deterministic — no sleeps, no timing assumptions (verified `create_bus` mints exactly one sender per side, so `is_closed()` is deterministic).
- `cargo test -p octos-bus`: all 241 lib tests + 18 integration tests pass in one run, including the existing `test_dispatch_outbound` happy path (send → dispatch → clean shutdown on sender drop unchanged).
- `cargo fmt --check` clean; `cargo clippy -p octos-bus --all-targets -- -D warnings` zero warnings; `cargo check -p octos-cli` (the only consumer of this API) compiles.

End-to-end note: the failure scenario (outbound bus dead at startup) cannot be reached from the real `octos gateway` binary without patching internals — `create_bus` always returns a live pair and the gateway holds the agent handle past `start_all`. The strongest available end-to-end evidence is the red→green transition through the real public path `create_bus() → ChannelManager::start_all()`, which is what the test exercises; the happy path was additionally verified unchanged via `test_dispatch_outbound` and `cargo check` on the gateway caller.

## Review

Went through self-review of the full diff plus an independent adversarial review pass. Findings raised and resolved: (1) literal "panic propagation to caller" criterion is impossible post-return without an API change — scoped explicitly above; (2) listener-task leak on the `Err` path — documented on `start_all`, benign for the only caller; (3) `is_closed()` false-positive shape (buffered messages + dropped senders) — no such caller exists in the workspace; (4) hang/flakiness analysis of the readiness wait — none possible.

Closes #1657


anantheparty (approved):

Issue #2198, opened 2026-08-31
llm cache-cost follow-ups from #2194 review: cached-rate granularity, provider coverage, ledger migration


anantheparty:
Non-blocking notes from the #2194 pre-merge review (all pre-existing or conservative-direction), bundled for tracking.

1. **OpenAI/DeepSeek/Moonshot read side books 1.0x** (pricing.rs:1003-1008, third rate-card tier). Real values: gpt-4o ≈0.5x, gpt-4.1 ≈0.25x, DeepSeek ≈0.1x. Author's stance (never-understate without per-family constants) is self-consistent; the durable fix is a per-family cached-rate field in the catalog (the comment already points there) — OpenAI-heavy users' cost reports are currently meaningfully overstated.

2. **openai_responses.rs / openrouter.rs non-streaming don't extract cached_tokens** — cache hits there book at plain cost() (1.0x, never-understate). OpenRouter is also inconsistent with itself: streaming (shared openai SSE parse) normalizes, non-streaming doesn't.

3. **project_cost_usd not migrated** (cost_ledger.rs:190-194) — consumed by octos-pipeline/executor node costs and octos-swarm budget projection; on Anthropic it misses the 0.1x/1.25x premiums (understating). Needs the cache fields threaded through the signature.

4. **usage_status_json exposes read-side only** (ui_protocol_transport.rs:9883) — usage_ledger already accumulates cache_write_tokens (usage_ledger.rs:176); expose the write side too.

5. **Edge: provider-label substring matching** (pricing.rs:992-998, contains("anthropic")/etc. over free-text labels) — a relay literally named "my-anthropic-proxy" would get 0.1x reads = 5x understatement. Tighter: match on the parsed model family.

Refs #2194

PR #2226, opened 2026-09-03
fix(llm): extract cached_tokens in openrouter and openai-responses usage parsing

BH3GEI:
## Problem

Item 2 of #2198 (cache-cost follow-ups from the #2194 review): `openai_responses.rs` and `openrouter.rs` did not extract `cached_tokens` from usage responses. Cache hits on those providers were reported as plain `input_tokens`, violating `TokenUsage`'s documented disjoint cache accounting contract (total prompt = `input + cache_read + cache_write`). OpenRouter was also inconsistent with itself: its streaming path normalizes cached tokens via the shared openai SSE parse, while its non-streaming path did not — so the same request reported different usage depending on the stream flag.

## Root cause

Both providers report cached tokens INSIDE the prompt total on the wire (chat completions: `usage.prompt_tokens_details.cached_tokens`; Responses API: `usage.input_tokens_details.cached_tokens`). `openai.rs` already subtracts the cached share at its parse boundary; the openrouter non-streaming parser and both openai-responses parsers (non-streaming `parse_responses_api` and the streaming `response.completed` event) lacked the field and the normalization.

## Fix

- `openrouter.rs`: parse `prompt_tokens_details.cached_tokens` in the non-streaming `Usage` and normalize to the disjoint contract (`input_tokens = prompt_tokens - cached`, `cache_read_tokens = cached`), mirroring `openai.rs`. `saturating_sub` guards a pathological `cached > prompt_tokens` payload.
- `openai_responses.rs`: parse `input_tokens_details.cached_tokens` and apply the same normalization in both the non-streaming conversion and the streaming `response.completed` usage event. The streaming path had the identical gap (raw inclusive `input_tokens`); leaving it would have recreated, inside one file, the stream-flag-dependent inconsistency this PR removes for openrouter.

No pricing-behavior change today: both providers fall in the residual cache-rate bucket (reads at 1.0x), so booked cost is unchanged. What changes is that usage reporting is truthful and uniform across providers and across stream/non-stream, and the counts are ready for the per-family cached-rate work tracked in #2198 item 1. Scope is item 2 only; items 1/3/4/5 are untouched.

## Test evidence

- New tests are red on `origin/main`, green on this branch (verified by applying the test hunks to a clean `origin/main` worktree: 3 failures pre-fix, all pass post-fix).
- `cargo test -p octos-llm`: 611 passed, 0 failed.
- `cargo fmt --check`: clean. `cargo clippy -p octos-llm --all-targets -- -D warnings`: zero warnings.
- End-to-end (real HTTP path via a local wiremock server, provider `.chat()` → `TokenUsage` → `pricing::cost_with_cache_for_provider`), same canned response body (`prompt_tokens=100, cached_tokens=75` / `input_tokens=100, cached_tokens=75`):
  - Before: `openrouter` reported `input=100, cache_read=0`; `openai-responses` reported `input=100, cache_read=0`.
  - After: both report `input=25, cache_read=75` — matching what openrouter streaming already reports for the same payload.

## Review

Independently reviewed by a fresh adversarial pass over the full diff (issue text + diff + repo access, no author reasoning shared). Verdict: approve-with-nits. Both nits addressed: dropped an unused `Default` derive on the new openrouter detail struct, and added a boundary test locking in the `saturating_sub` clamp when `cached_tokens >= prompt_tokens`. Downstream consumers (`agent/budget.rs`, usage ledger, turn state) were checked and already honor the disjoint contract. Two pre-existing observations in `openai_responses.rs` (unhandled `response.incomplete`/`response.failed` terminal events; a dead `state.input_tokens` store) are out of scope here and left untouched.

Refs #2198


anantheparty (dismissed):

BH3GEI:
Synced with latest main and fixed the `check` failure: the sync brought in usage-parsing tests written against the old single-argument `openrouter_response_to_chat_response`, so the three new call sites now pass the model name like the rest. Verified locally: `cargo fmt --check` clean, `cargo clippy -p octos-llm --all-targets -- -D warnings` zero warnings, `cargo test -p octos-llm --lib openrouter` 20 passed / 0 failed.

BH3GEI:
Heads-up @anantheparty — your re-approval from Sep 5 is showing up on GitHub as a **dismissed** review (state `DISMISSED`, not `APPROVED`), so the approval is not registering for merge gating. Could you re-submit it as an approval? The branch is current with main and CI is green.

anantheparty (approved):
Re-approving (the Sep 5 approval had been dismissed by the branch movement — re-posting so it registers for merge gating). The tool_policy classify fix review stands: the restart-required classification for tool_policy edits is correct and tested. @BH3GEI thanks for the catch.

PR #2230, opened 2026-09-03
fix(agent): honor OCTOS_LLM_CALL_MAX_SECS=0 as disable (stop clamping to 1s)

### LINKED ISSUE #2228: OCTOS_LLM_CALL_MAX_SECS=0 does not disable the wall-clock cap (clamps to 1s, killing every stream)
## Problem

`OCTOS_LLM_CALL_MAX_SECS=0` is documented to **disable** the overall wall-clock stream cap, but it actually clamps to **1 second** and instantly kills every stream.

`crates/octos-agent/src/agent/streaming.rs:35` (doc): *"a value of `0` for `overall_max_secs` disables the wall-clock backstop (idle/TTFT guards still apply)"* — and the loop honors it: `let overall_deadline = (thresholds.overall_max_secs > 0).then(|| …)` (streaming.rs:176), so `0` ⇒ no deadline.

But the env parser floors it:

```rust
// crates/octos-agent/src/agent/mod.rs:166
fn env_secs_or(var: &str, default_secs: u64) -> std::time::Duration {
    let secs = std::env::var(var).ok()
        .and_then(|raw| raw.parse::<u64>().ok())
        .map(|v| v.clamp(1, 86_400))   // <-- 0 becomes 1
        .unwrap_or(default_secs);
    std::time::Duration::from_secs(secs)
}
```

`llm_call_max` is built via `env_secs_or("OCTOS_LLM_CALL_MAX_SECS", …)` (mod.rs:227), so `=0` → `1s`. Every stream then aborts after 1s with `stream idle timeout after 1s — provider stalled` (observed on a local model: the log shows `exceeded overall 1s wall-clock cap`). The documented disable path is unreachable.

## Root cause

`env_secs_or`'s `.clamp(1, 86_400)` is correct for the idle/TTFT knobs (they must be ≥1), but wrong for `OCTOS_LLM_CALL_MAX_SECS`, whose contract explicitly makes `0` a sentinel for "disabled".

## Fix

Allow `0` to pass through for the call-max knob (e.g. a `env_secs_allow_zero_or` variant clamping to `[0, 86_400]`, used only for `llm_call_max`). Idle/TTFT knobs keep the `≥1` floor.

## Acceptance

- `OCTOS_LLM_CALL_MAX_SECS=0` ⇒ `llm_call_max == 0` ⇒ no overall deadline (idle/TTFT still apply).
- `OCTOS_LLM_CALL_MAX_SECS` unset or a positive value ⇒ unchanged.
- Idle/TTFT env knobs still floor at 1.

heyong4725:
## What & why

Closes #2228. `OCTOS_LLM_CALL_MAX_SECS=0` is documented (`streaming.rs:35`) to **disable** the overall wall-clock stream cap, and the streaming loop honors `overall_max_secs == 0` as "no deadline" (`streaming.rs:176`). But the env reader `env_secs_or` clamps every value to `[1, 86_400]`, so `0` became **1s** — instantly aborting every stream with `stream idle timeout after 1s — provider stalled`. The documented disable path was unreachable; operators of slow local models could not turn the cap off.

## Change

- Extract the clamp into a pure, unit-testable helper `clamp_env_secs(parsed, default, min)`.
- Add `env_secs_allow_zero_or` (floor `0` instead of `1`) and use it **only** for `llm_call_max` (`OCTOS_LLM_CALL_MAX_SECS`), whose `0` is a disable sentinel.
- All other timeout knobs (idle, TTFT, voice, tool) keep the `≥1` floor via `env_secs_or` — a guard that must stay live cannot be disabled to `0`.

## Cloud safety

No default changes. `OCTOS_LLM_CALL_MAX_SECS` unset ⇒ `DEFAULT_LLM_CALL_MAX_SECS` (1200) as before; a positive value is unchanged; only the previously-broken `0` now does what the docs already promise. Every other guard is byte-for-byte unchanged.

## Tests

- `clamp_env_secs_floor_one_keeps_guard_live` — `0`→`1`, passthrough, `86_400` ceiling, `None`→default.
- `clamp_env_secs_floor_zero_allows_disable` — `0`→`0` (disabled), passthrough, ceiling, default.

`cargo test -p octos-agent`, fmt, clippy clean.

### DIFF
--- a/crates/octos-agent/src/agent/mod.rs
+++ b/crates/octos-agent/src/agent/mod.rs
@@ -159,17 +159,37 @@ pub const VOICE_STREAM_TTFT_SECS: u64 = 10;
 /// Tightened inter-chunk idle timeout for voice fail-fast turns (10s).
 pub const VOICE_STREAM_IDLE_SECS: u64 = 10;
 
+/// Pure clamp behind the `env_secs_*` readers: apply `[min, 86_400]` to a
+/// parsed value, or fall back to `default_secs` when the var was absent or
+/// unparseable. Extracted so the clamp is unit-testable without touching the
+/// process environment.
+fn clamp_env_secs(parsed: Option<u64>, default_secs: u64, min: u64) -> u64 {
+    parsed.map(|v| v.clamp(min, 86_400)).unwrap_or(default_secs)
+}
+
 /// Read an env-overridable seconds value, mirroring the convention in
 /// `octos-cli/src/session_actor.rs` (`std::env::var(...).parse()` with a clamp
 /// so a misconfigured value cannot disable the guard entirely). A parsed `0`
-/// is clamped up to `1` so the timeout is always live.
+/// is clamped up to `1` so the timeout is always live. Use
+/// [`env_secs_allow_zero_or`] for knobs whose contract makes `0` mean
+/// "disabled".
 fn env_secs_or(var: &str, default_secs: u64) -> std::time::Duration {
-    let secs = std::env::var(var)
+    let parsed = std::env::var(var)
         .ok()
-        .and_then(|raw| raw.parse::<u64>().ok())
-        .map(|v| v.clamp(1, 86_400))
-        .unwrap_or(default_secs);
-    std::time::Duration::from_secs(secs)
+        .and_then(|raw| raw.parse::<u64>().ok());
+    std::time::Duration::from_secs(clamp_env_secs(parsed, default_secs, 1))
+}
+
+/// Like [`env_secs_or`] but honors `0` as a disable sentinel (floor is `0`,
+/// not `1`). Used for `OCTOS_LLM_CALL_MAX_SECS`, whose `0` value disables the
+/// overall wall-clock backstop (see `streaming.rs`; the idle/TTFT guards stay
+/// live). Clamping `0` up to `1` here would instead abort every stream after
+/// 1s (#2228).
+fn env_secs_allow_zero_or(var: &str, default_secs: u64) -> std::time::Duration {
+    let parsed = std::env::var(var)
+        .ok()
+        .and_then(|raw| raw.parse::<u64>().ok());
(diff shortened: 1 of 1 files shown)


anantheparty (approved):
Adversarial pre-merge review passed — implements the #2228 fix plan exactly: clamp_env_secs extracted (unit-testable without touching the process env), env_secs_allow_zero_or with the [0, 86_400] floor used ONLY for llm_call_max (the documented 0=disable sentinel now reachable), env_secs_or keeps the ≥1 floor for idle/TTFT knobs. No other knob semantics change. Approving — merging left to you per convention.

BH3GEI (approved):
Verified locally at 34e9e7ee (detached worktree):

- Diff is exactly the #2228 fix, one file, no scope creep. clamp_env_secs is behavior-identical to the old env_secs_or body (parse failure/empty/out-of-range/default passthrough all match).
- Traced every consumer: llm_call_max flows only to streaming.rs:100 -> overall_max_secs, where streaming.rs:176 already maps 0 to no deadline; idle/TTFT knobs keep the >=1 floor (streaming.rs:173-174 + env_secs_or). Voice path untouched (voice_overall_deadline, floor 1).
- cargo build/clippy --all-targets -D warnings/fmt --check: clean.
- cargo test -p octos-agent: 2802 passed. 4 failures diagnosed as unrelated to this PR: 3 docker-sandbox tests fail identically on origin/main (no Docker backend on this machine), and exec_session_captures_all_output_when_process_exits_at_deadline is self-described timing-sensitive (5ms yield race) and passes in isolation on this head.
- End-to-end wiring check (throwaway integration test, since removed): OCTOS_LLM_CALL_MAX_SECS unset -> 1200s, =0 -> Duration::ZERO (cap disabled, the bug), =5 -> 5s, huge -> 86400s ceiling, unparseable -> default; OCTOS_LLM_STREAM_IDLE_SECS=0 still floors to 1s.

Fix is correct, minimal, and matches the documented contract.

PR #2231, opened 2026-09-03
fix(llm): resolve r9s model discovery per model — claude-* probes the Anthropic strategy + root

### LINKED ISSUE #2185: llm registry: r9s declares unconditional OPENAI model discovery but switches protocol per model (claude-* -> Anthropic)
Follow-up from the #2183 review.

`registry/r9s.rs:44-48` selects the provider by model name — `claude-*` constructs `AnthropicProvider` against a rewritten `{base}/anthropic` root, everything else uses OpenAI. But `r9s.rs:22` declares `OPENAI_MODELS` unconditionally, so after #2183 the discovery declaration for r9s+claude-* probes `{base}/models` with `Authorization: Bearer` instead of the Anthropic strategy.

Mitigations today (why this is a nit, not a regression): under the default root `https://api.r9s.ai/v1` the OpenAI list endpoint exists and returns a usable list, and a failed discovery is advisory only (manual model id / Test / Save are unaffected).

Suggested fix: either resolve the discovery declaration per model (like the provider switch does), or declare r9s `Unsupported` for discovery until a per-model strategy exists.

Refs #2165 #2183

BH3GEI:
## Problem

#2185 (follow-up from the #2183 review): r9s selects its wire protocol by **model name** — `claude-*` constructs an `AnthropicProvider` against a rewritten `{base}/anthropic` root, everything else uses OpenAI Chat Completions. But the registry entry declared `OPENAI_MODELS` discovery unconditionally, so a r9s + `claude-*` selection was probed at `{base}/models` with `Authorization: Bearer` instead of the Anthropic Messages listing (`{anthropic-root}/v1/models` with `x-api-key`).

## Root cause

`resolve_model_discovery` resolved the listing strategy from `(family, api_type)` only — it had no access to the selected model, so the family-wide declaration was the most specific answer it could give. The knowledge of the per-model protocol split lived only in r9s's `create` (and, for pricing, `prefers_anthropic`), unreachable from discovery.

## Fix

Per-model resolution, exactly as the issue's first suggestion:

- `ProviderEntry` gains `model_discovery_for_model: Option<ModelDiscoveryForModel>` — an optional per-model resolver consulted ahead of the family-wide `model_discovery` when a model is selected and no `api_type` override applies. The 18 other families declare `None` (behavior unchanged).
- r9s's resolver maps `claude-*` → `ANTHROPIC_MODELS` against `anthropic_root()`, the **same** root rewrite `create()` applies (extracted into one shared function), driven by the existing `prefers_anthropic` SSOT (raw, case-sensitive `claude-` prefix — identical to construction and pricing). The default root is now one shared const behind the ENTRY declaration and both fallbacks, so serving and probing can never drift onto different roots.
- `resolve_model_discovery(family, api_type, model, base_url)` returns a `DiscoveryRoute { discovery, base_url }`; `discover_models` resolves the probe root as route rewrite → caller `base_url` → family default. The base URL is normalized (trim + empty-filter) **before** the resolver sees it, so `base_url: ""` falls through to the family default instead of producing a relative `/anthropic` URL (caught in review).
- Both call surfaces (AppUI `profile/llm/fetch_models` RPC and admin REST `/api/my/provider-models`) pass the selected model through. The `api_type: "anthropic"` override still wins, mirroring the runtime bypass.

No model selected yet (listing before selection) and non-claude selections keep the family-wide OpenAI listing — unchanged.

## Test evidence

- New tests fail with the resolver disabled and pass with it (red/green verified):
  - `discovery::tests::should_resolve_r9s_discovery_per_model` — strategy + root per model, incl. empty/whitespace/padded `base_url`, mixed-case `Claude-*` (stays OpenAI), whitespace-padded model (stays OpenAI, matching runtime's raw match), no model, and the `api_type` override.
  - `discovery::tests::should_probe_anthropic_root_for_r9s_claude_selection_without_bearer` — loopback HTTP fixture asserts the probe path `/anthropic/v1/models`, `x-api-key` present, `Authorization` absent.
  - RPC level: `profile_llm_fetch_models_r9s_claude_selection_probes_the_anthropic_root` and `..._non_claude_selection_keeps_the_openai_listing` (AppUI), `admin_provider_models_r9s_claude_selection_probes_the_anthropic_root` (REST).
- `cargo test -p octos-llm --lib`: 608 passed, 0 failed.
- `cargo test -p octos-cli --lib --features api`: 3194 passed; 1 failure in `autonomy::agent_orchestrator::tests::sovereignty_provider_fail_open_when_unowned_default_branch` — pre-existing, environment-dependent (this machine's `git init` defaults to `main`, so the test's `git checkout -b main` fails with "branch already exists"); verified to fail identically on pristine `origin/main` in a module untouched by this diff.
- `cargo fmt --check` clean; `cargo clippy -p octos-llm / -p octos-cli --all-targets [--features api] -- -D warnings` clean.

**End-to-end** (real `octos serve` + loopback fixture server logging every request, `POST /api/my/provider-models` with `provider=r9s`, `base_url=http://127.0.0.1:…/v1`):

| request | before (origin/main binary) | after (this branch) |
|---|---|---|
| `model=claude-sonnet-4` | `GET /v1/models`, `Authorization: Bearer` — the mis-probe from the issue | `GET /anthropic/v1/models`, `x-api-key` + `anthropic-version: 2023-06-01`, no Bearer |
| `model=claude-sonnet-4`, padded base_url | — | same correct Anthropic probe (normalization) |
| `model=gpt-5` / no model | `GET /v1/models` + Bearer | unchanged: `GET /v1/models` + Bearer |

## Review

Independent adversarial review (given only the issue text + diff + file paths): initial REQUEST_CHANGES with one major finding (unnormalized `base_url` could regress `base_url: ""` to `endpoint_unreachable` on the REST surface) and four minor/nit findings (REST surface untested, model-trim divergence vs runtime, third hardcoded default root, comment accuracy). All five addressed; re-review verdict APPROVE.

Known limitation (out of scope): the shipped admin dashboard hardcodes `model: ''` when calling `/api/my/provider-models`, so the per-model path fires there only for API clients that send a model; the AppUI selection flow passes `model_id` through. Discovery remains advisory in all cases — a failed probe never blocks manual model-id entry, Test, or Save.

Closes #2185

### DIFF
--- a/crates/octos-cli/src/api/admin.rs
+++ b/crates/octos-cli/src/api/admin.rs
@@ -1209,12 +1209,17 @@ pub async fn provider_models(
     }
     // Protocol-aware discovery shared with the AppUI `profile/llm/
     // fetch_models` surface — the strategy resolves from the route (api_type
-    // override, then the family's declared protocol), never from the literal
-    // family id, so the two clients cannot drift.
-    let discovery =
-        octos_llm::discovery::resolve_model_discovery(Some(&req.provider), req.api_type.as_deref());
+    // override, then the family's declared protocol — per-model for families
+    // like r9s that pick the wire protocol by model name), never from the
+    // literal family id, so the two clients cannot drift.
+    let route = octos_llm::discovery::resolve_model_discovery(
+        Some(&req.provider),
+        req.api_type.as_deref(),
+        (!req.model.trim().is_empty()).then_some(req.model.trim()),
+        req.base_url.as_deref(),
+    );
     let outcome = octos_llm::discovery::discover_models(
-        discovery,
+        &route,
         &api_key,
         req.base_url.as_deref(),
         Some(&req.provider),
--- a/crates/octos-cli/src/api/ui_protocol_tests.rs
+++ b/crates/octos-cli/src/api/ui_protocol_tests.rs
@@ -9169,6 +9169,116 @@ async fn profile_llm_fetch_models_zai_never_gets_a_bearer_v1_models_probe() {
 }
 
 #[tokio::test]
+async fn profile_llm_fetch_models_r9s_claude_selection_probes_the_anthropic_root() {
+    let (root, captured) =
+        spawn_discovery_fixture("200 OK", r#"{"data":[{"id":"claude-sonnet-4"}]}"#).await;
+    let state = Arc::new(AppState::empty_for_tests());
+    // r9s serves claude-* over the Anthropic Messages protocol at a rewritten
+    // `{base}/anthropic` root — the probe must follow the SELECTED model
+    // (octos#2185), not the family-wide OpenAI declaration.
+    let request = RpcRequest::new(
+        "1",
+        APPUI_METHOD_PROFILE_LLM_FETCH_MODELS,
+        json!({
+            "selection": {
+                "family_id": "r9s",
+                "model_id": "claude-sonnet-4",
+                "route": {
+                    "route_id": "official",
+                    "base_url": format!("{root}/v1")
+                }
+            },
+            "api_key": "r9s-secret-key"
+        }),
+    );
+
+    let result = raw_profile_llm_fetch_models(&state, &request, Some("ada"))
+        .await
+        .expect("fetch_models result");
+
+    assert_eq!(result["status"], json!("discovered"));
+    assert_eq!(result["models"], json!(["claude-sonnet-4"]));
+    let requests = captured.lock().await;
+    assert_eq!(requests.len(), 1);
+    assert_eq!(requests[0].path, "/anthropic/v1/models");
+    assert!(
+        requests[0].authorization.is_none(),
+        "r9s claude-* speaks Anthropic Messages — never a Bearer probe"
+    );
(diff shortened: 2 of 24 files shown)


anantheparty (approved):
Adversarial pre-merge review passed (24 files, r9s per-model model discovery): claude-* now probes the Anthropic strategy at the rewritten {base}/anthropic root, driven by prefers_anthropic as the single source of truth shared with construction and pricing (serve/probe cannot drift); non-claude-* keeps the OpenAI family-wide behavior with no rewrite; three wire-level loopback tests pin the AppUI claude/non-claude and admin REST paths. All 19 other registry declarations migrated to the new DiscoveryRoute shape with zero behavior change; no old-signature residue; ci.yml untouched. Two advisory nits (responses+claude-* discovery/runtime mismatch — pre-existing since #2183; padded-model trim asymmetry) filed as a follow-up issue. Remaining checks (check/test-octos-cli/check-windows) pending at review time — approving on the review + current CI state; merging left to you per convention.

PR #2232, opened 2026-09-03
feat(agent): Pi-aligned defaults for local providers (server-sampled temperature + no wall-clock cap)

### LINKED ISSUE #2229: Local providers require config tweaks: forced greedy temperature + 1200s wall-clock cap should be Pi-friendly by default
## Problem

A fresh local-model user (e.g. llama.cpp via the `local` provider) must hand-tweak config to get usable behavior, because two global defaults are tuned for cloud and actively harm a slow local **reasoning** model:

1. **Forced greedy decoding.** `ChatConfig::default().temperature == Some(0.0)` (`crates/octos-llm/src/config.rs:139`) is sent on every request. On a local reasoning model this drives the repetition collapse that `#2172/#2173` documented — the user must set `llm_temperature` to escape it. Pi sends *no* temperature and lets the server sample.

2. **1200s overall wall-clock cap.** `DEFAULT_LLM_CALL_MAX_SECS = 1200` (`crates/octos-agent/src/agent/mod.rs:152`) aborts a **healthy, actively-streaming** reasoning turn — it fires even while tokens flow (streaming.rs:189 comment: *"catches a stream that trickles one token every `<idle>`s forever"*). A whole-repo C→Rust reasoning turn on a 27B local model legitimately exceeds 20 min, so it is guillotined as `provider stalled`. Pi has no wall-clock cap (idle-only). The user must set `OCTOS_LLM_CALL_MAX_SECS` to a large value.

Neither is a problem for cloud providers (greedy = reproducible; the cap backstops slow-trickle billing). But for `provider == "local"` both are wrong, and requiring per-user env/profile tweaks defeats out-of-the-box parity with Pi's `pi-llama` plugin (whose only special treatment is a 16384 output cap — which octos already defaults to for `local-default`).

## Fix

Make the defaults **provider-aware**, keyed on the already-available local-provider signal (`AgentConfig.provider_metadata` / provider name `"local"`), changing behavior **only** for local providers:

- **Temperature:** for a local provider, do not force `temperature = 0.0` when the operator hasn't set one — omit it (let the server sample, like Pi) or use a sane non-greedy default.
- **Wall-clock cap:** for a local provider, default the overall cap to **disabled** (0) — the inter-chunk idle + TTFT guards still catch a genuinely dead provider. Keep `DEFAULT_LLM_CALL_MAX_SECS = 1200` for cloud.

An explicit operator setting (profile `llm_temperature`, env `OCTOS_LLM_CALL_MAX_SECS`) always wins. Depends on #2228 so that a `0` cap actually disables.

## Cloud safety

Cloud/remote providers keep today's defaults byte-for-byte (greedy `0.0`, 1200s cap). Only `provider == "local"` (self-hosted, localhost) gets the Pi-aligned defaults.

## Acceptance

- Fresh `octos chat --profile <local>` with no `llm_temperature` and no `OCTOS_LLM_CALL_MAX_SECS`: request goes out with no forced `temperature=0.0`, and a >20-min streaming reasoning turn is not aborted by a wall-clock cap.
- A cloud profile is unchanged: `temperature=0.0`, 1200s cap.
- Explicit operator overrides still apply for both provider kinds.

heyong4725:
## What & why

Closes #2229. A fresh **local**-model user (llama.cpp via the `local` provider) had to hand-tweak config to get usable behavior, because two defaults tuned for cloud actively harm a slow local **reasoning** model:

1. **Forced greedy decoding** — `ChatConfig::default().temperature = Some(0.0)` is sent on every request; on a local reasoning model this drives the repetition collapse (`#2172/#2173`). Pi sends no temperature and lets the server sample.
2. **1200s overall wall-clock cap** — `DEFAULT_LLM_CALL_MAX_SECS = 1200` aborts a **healthy, actively-streaming** reasoning turn (it fires even while tokens flow). A whole-repo C→Rust reasoning turn on a 27B local model legitimately exceeds 20 min and got guillotined as `provider stalled`. Pi has no wall-clock cap.

Both are correct for cloud (greedy = reproducible; the cap backstops slow-trickle billing), so the fix makes them **provider-aware** — changing behavior only for `provider == "local"`.

## Change

- `Agent::is_local_provider()` — true when the active provider's metadata name is `"local"` (`self.llm.provider_metadata().provider`).
- **Temperature** (`build_chat_config`): gains a `local_provider` arg. An explicit `chat_temperature` still wins; otherwise, for local, temperature is left **unset** (`None`) so the request omits it (`skip_serializing_if`) and the server samples. Cloud keeps the built-in `0.0` default byte-for-byte.
- **Wall-clock cap** (`effective_llm_call_max_secs`): for local, defaults to **0 (disabled)** — the inter-chunk idle + TTFT guards still catch a genuinely dead provider — **unless** the operator explicitly set `OCTOS_LLM_CALL_MAX_SECS`. Cloud keeps 1200s.

Pairs with #2228 (so an explicit `OCTOS_LLM_CALL_MAX_SECS=0` also disables). Net effect: a fresh local user gets Pi-parity (server-sampled temperature, no wall-clock guillotine) with **zero config**, while cloud is unchanged.

## Cloud safety

Every cloud/remote path takes `local_provider = false` / `is_local_provider() == false`, so temperature stays `0.0` and the cap stays 1200s — byte-for-byte unchanged. Only `provider == "local"` gets the Pi-aligned defaults, and an explicit operator override wins for both knobs on either provider kind.

## Tests

- `build_chat_config_local_provider_unsets_temperature` — local + no override ⇒ `temperature == None`; same cfg on cloud ⇒ `Some(0.0)`.
- `build_chat_config_local_provider_respects_explicit_temperature` — explicit override wins on local.
- Existing `#2172` cloud-safety tests updated to pass `local_provider = false` and still assert the `0.0` default.

`cargo test -p octos-agent`, fmt, clippy clean.

### DIFF
--- a/crates/octos-agent/src/agent/loop_runner.rs
+++ b/crates/octos-agent/src/agent/loop_runner.rs
@@ -703,7 +703,16 @@ impl Agent {
     /// Build a `ChatConfig` with optional `chat_max_tokens` / `chat_temperature`
     /// overrides from `AgentConfig`. Delegates to [`build_chat_config`].
     fn chat_config(&self) -> ChatConfig {
-        build_chat_config(&self.config)
+        build_chat_config(&self.config, self.is_local_provider())
+    }
+
+    /// True when the active LLM provider is the built-in local/self-hosted
+    /// provider (metadata name `"local"`). Local reasoning models (e.g. Qwen
+    /// on llama.cpp) need Pi-aligned defaults (#2229) — server-sampled
+    /// temperature and no overall wall-clock cap — while cloud providers keep
+    /// the conservative defaults. An explicit operator override always wins.
+    pub(super) fn is_local_provider(&self) -> bool {
+        self.llm.provider_metadata().provider == "local"
     }
 
     /// Decide what to surface when the loop detector fires.
@@ -3954,16 +3963,24 @@ fn shell_retry_limit_message(content: &str) -> String {
 /// without constructing a full `Agent` — in particular the cloud-safety
 /// invariant (#2172): an unset `chat_temperature` must leave the built-in
 /// `0.0` default untouched, so cloud requests are byte-for-byte unchanged.
-fn build_chat_config(config: &crate::AgentConfig) -> ChatConfig {
+///
+/// `local_provider` (#2229): for the local/self-hosted provider, an unset
+/// `chat_temperature` leaves temperature UNSET rather than defaulting to `0.0`,
+/// so the server samples (Pi parity) — forced greedy degenerates local
+/// reasoning models. Cloud (`local_provider = false`) is unchanged.
+fn build_chat_config(config: &crate::AgentConfig, local_provider: bool) -> ChatConfig {
     let mut c = ChatConfig::default();
     if let Some(max) = config.chat_max_tokens {
         c.max_tokens = Some(max);
     }
-    // Temperature override. Unset → keep the built-in default (0.0), so cloud
-    // requests are unchanged; set → override, primarily to avoid forced-greedy
-    // repetition collapse on local models.
+    // Temperature. An explicit `chat_temperature` always wins. Otherwise: on a
+    // LOCAL provider leave it unset so the server samples (the request omits
+    // `temperature` via skip_serializing_if); on cloud keep the built-in `0.0`
+    // default so cloud requests are byte-for-byte unchanged.
     if let Some(temp) = config.chat_temperature {
         c.temperature = Some(temp);
+    } else if local_provider {
+        c.temperature = None;
     }
     // Extra sampler params (e.g. repeat_penalty) for OpenAI-compatible servers.
     // Unset → nothing added, cloud unchanged.
--- a/crates/octos-agent/src/agent/loop_runner_tests.rs
+++ b/crates/octos-agent/src/agent/loop_runner_tests.rs
@@ -6557,7 +6557,7 @@ fn build_chat_config_keeps_default_temperature_when_unset() {
         chat_temperature: None,
         ..AgentConfig::default()
     };
-    let chat = build_chat_config(&cfg);
+    let chat = build_chat_config(&cfg, false);
     assert_eq!(chat.temperature, ChatConfig::default().temperature);
     assert_eq!(chat.temperature, Some(0.0));
 }
@@ -6568,7 +6568,7 @@ fn build_chat_config_applies_temperature_override() {
         chat_temperature: Some(0.7),
         ..AgentConfig::default()
     };
-    let chat = build_chat_config(&cfg);
+    let chat = build_chat_config(&cfg, false);
     assert_eq!(chat.temperature, Some(0.7));
 }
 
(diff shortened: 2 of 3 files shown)


anantheparty (approved):
Adversarial pre-merge review passed — all four scoping requirements from the #2229 triage are met: (1) keyed on provider_metadata provider == "local" (the metadata-based signal that survives wrapper delegation post-#2194); (2) explicit operator settings always win (profile llm_temperature, env OCTOS_LLM_CALL_MAX_SECS); (3) pairs with #2230 — merge this AFTER #2230 so the local-default 0 cap is honored as disable rather than clamped to 1s; (4) temperature is OMITTED on local (server samples, Pi parity) rather than a invented non-greedy default, with cloud byte-for-byte unchanged (Some(0.0) sentinel kept). The effective_llm_call_max_secs three-way (local+env-unset → 0; env-set → value; non-local → config) is clean. Tests pin local-unset/cloud-unchanged/explicit-wins. **Merge order note: land #2230 first.** Approving — merging left to you per convention.

BH3GEI (approved):
Verified locally on 54ddd256 (isolated worktree):

- Scope: diff is exactly the 3 declared files; single production caller of build_chat_config (chat_config) threads is_local_provider(); voice path unaffected (uses voice_overall_deadline, streaming.rs:94).
- Temperature: local + unset -> None, and None is truly omitted on the wire (OpenAIRequest.temperature has skip_serializing_if, openai.rs:894-896; builder passes config.temperature through verbatim, openai.rs:515-518). Cloud keeps Some(0.0) byte-for-byte; explicit chat_temperature wins on both. The 2 new tests plus the 4 updated #2172 cloud-safety tests all pass.
- Wall-clock cap: overall_max_secs == 0 is genuinely 'disabled' (streaming.rs:192 .then()); TTFT/inter-chunk idle guards are untouched and still bound a dead provider. effective_llm_call_max_secs three-way (local+env-unset -> 0; env-set -> config value; non-local -> config) reads correctly; env presence is the only reader-proxy and no other code reads OCTOS_LLM_CALL_MAX_SECS.
- Provider identity: 'local' label set at registry/local.rs:60, survives LocalContextProbe delegation (local_context_probe.rs:733) and retry/fallback/failover/swappable wrappers post-#2194.
- Build/test/lint: cargo build -p octos-agent clean; cargo clippy --all-targets -D warnings clean; cargo fmt --check clean; octos-llm 605/605 pass; agent streaming tests 18/18 pass. Full octos-agent suite: 2793 pass; the 5 failures are pre-existing environment issues, all root-caused as unrelated to this diff: 3 Docker-sandbox tests fail identically on an origin/main worktree (no Docker backend on this machine), 1 plugin test passes in isolation, and exec_session_captures_all_output_when_process_exits_at_deadline is an inherent timing race (5ms yield vs instant exit over 40 trials, coding_tools_tests.rs:1453-1490 — flips pass/fail in isolation on both base and head).
- Merge order: as already noted, land #2230 first so an explicit OCTOS_LLM_CALL_MAX_SECS=0 on local means disable rather than clamp-to-1s.

Cloud behavior is byte-for-byte unchanged; the local path gets the Pi-aligned defaults the issue specifies. LGTM.

PR #2235, opened 2026-09-03
fix(cli): GC read-only chat runtime run dirs (skill-binary leak)

### LINKED ISSUE #1795: chat: GC for read-only per-run runtime dirs (skill-binary leak)
Follow-up to #1781 (merged 726ec099b1). `resolve_chat_runtime_paths` mints a fresh `data_dir/runtime/runs/<uuid>/` per read-only run, and skill bootstrap copies the full sibling skill-binary set (MB-scale × ~10 skills) into it. Nothing ever prunes `runtime/runs/`, so every `octos chat --sandbox read-only` invocation leaks a complete skill-binary set.

Options (from the #1781 review):
- keep-last-N / age-based GC of `runtime/runs/` at chat startup, or
- share a single bootstrap dir like `serve` does (sha256 staleness marker already handles refresh), trading the per-run isolation the PR chose for boundedness.

BH3GEI:
## Problem

Every `octos chat --sandbox read-only` mints a private run dir under `data_dir/runtime/runs/<uuid>/` (`resolve_chat_runtime_paths`), and skill bootstrap copies the full bundled skill-binary set (MB-scale × ~10 skills) into it. Nothing ever pruned `runtime/runs/`, so each read-only chat leaked one complete skill-binary copy.

## Root cause

The per-run isolation introduced for read-only chats (private `runtime/runs/<uuid>` subtree so concurrent sessions cannot race over generated skills) came with no eviction policy — dirs accumulated forever.

## Fix

At chat startup, prune `runtime/runs/` down to the newest 8 run dirs, with three guards so GC can never harm a live session or user data:

- **Liveness lock**: a read-only chat holds an exclusive flock on `<run_dir>/session.lock` for its whole lifetime; the pruner probes with a shared lock on a read-only fd and skips locked dirs. A long-lived session can no longer lose its skill binaries / spawn deliverables mid-run, even after newer runs push it out of the retention window. (First review round caught that a pure keep-last-N/age scheme could delete a live session's dir — this guard is the fix.)
- **Age from the run id, not mtime**: dir age is read from the embedded v7-UUID timestamp, which `cp -a`/`tar` cannot backdate and stray writes cannot refresh. Unknown age (non-v7 id, clock skew) → keep.
- **Ownership filter**: only canonical-UUID-named real directories (no symlinks) are prune candidates; anything else a user stashed there is left alone and never occupies a retention slot.

GC is best-effort: failures log a warning and never block chat startup. The retention floor trade-off is documented in the code (effective steady-state cap is `max(8, runs-per-24h)`).

## Test evidence

- 12 new unit tests in `commands::chat::tests`: retention window, current-run protection, age floor, embedded-timestamp-over-mtime, live-lock protection (incl. the exact regressed scenario: >24h-old live dir under the production min-age), dead-session leftover lock file removal, unknown-age (v4) keep, non-canonical UUID spelling ignored, symlink never followed, missing dir / stray file tolerance, lock only for read-only sessions.
- `cargo test -p octos-cli --lib`: 1638 passed; the single failure (`autonomy::agent_orchestrator::tests::sovereignty_provider_fail_open_when_unowned_default_branch`) is pre-existing on clean `origin/main` — it shells out to `git init` then `git checkout -b main`, which fails on machines whose git already defaults to `main` (`fatal: a branch named 'main' already exists`, reproduced on the unmodified baseline). Unrelated to this change.
- `cargo fmt --check` clean; `cargo clippy -p octos-cli --all-targets -- -D warnings` clean.

### End-to-end (real binary, `octos chat --sandbox read-only` against a seeded data dir)

- Baseline binary (origin/main): 12 stale run dirs → 12 remain (leak reproduced).
- Fixed binary: 12 stale dirs + new current dir → pruned to 9: the 4 oldest dead dirs removed, the newest 8 + current kept, and a stale dir **locked by a simulated live session survived** despite being the oldest prune candidate.

### Review

Two independent adversarial review rounds were run. Round 1 found the live-session eviction hole (major), stray-dir deletion/slot-squatting, and the mtime fragility — all fixed (liveness lock, UUID-canonicality filter, embedded-timestamp age). Round 2 verified those fixes and found 4 minor/nit issues — all addressed (shared-lock probe on a read-only fd so an unwritable lock file reads as "unknown → keep"; canonical-spelling check; doc note on the burst-retention trade-off; tests for old+live end-to-end, dead leftover lock, v4 unknown-age; failure logs raised to `warn!`).

One transparency note: during development, one new test (`lock_chat_run_dir_locks_only_read_only_sessions`) failed exactly once inside a full-suite run executed at peak machine load immediately after a full-workspace clippy build, without its panic message being captured. It has since passed 300 consecutive solo runs, ~27 full-suite runs (incl. 6 concurrent suites), and a 12,800-iteration adversarial stress harness driving the identical syscall sequence under a 32-thread flock storm — zero recurrences. The production code path is best-effort by design (any transient OS-level lock/fs failure degrades to the age floor with a warning), and the test asserts the intended semantics deterministically; the single failure is attributed to a transient OS-level resource error under extreme parallel load.

Closes #1795

### DIFF
--- a/crates/octos-cli/src/commands/chat.rs
+++ b/crates/octos-cli/src/commands/chat.rs
@@ -252,6 +252,142 @@ fn resolve_chat_runtime_paths(
     }
 }
 
+/// Upper bound on retained per-run dirs under `data_dir/runtime/runs/`.
+const RETAINED_CHAT_RUN_DIRS: usize = 8;
+
+/// A run dir younger than this is never pruned, so a burst of new chats
+/// within a day cannot evict a session's dir via the retention window alone.
+const MIN_CHAT_RUN_DIR_AGE: std::time::Duration = std::time::Duration::from_secs(24 * 60 * 60);
+
+/// Lock file a live read-only chat holds inside its run dir for the whole
+/// session; the pruner treats a held lock as "still in use" regardless of age
+/// or position in the retention window.
+const CHAT_RUN_LOCK_FILE: &str = "session.lock";
+
+/// Age of a run dir from its id's embedded v7-UUID timestamp (NOT the dir
+/// mtime, which `cp -a`/`tar` can backdate and stray writes can refresh).
+/// `None` — non-v7 id or a future timestamp from clock skew — means "age
+/// unknown": the dir is kept rather than guessed at.
+fn chat_run_dir_age(name: &str) -> Option<std::time::Duration> {
+    let (secs, _) = uuid::Uuid::parse_str(name).ok()?.get_timestamp()?.to_unix();
+    let created = std::time::UNIX_EPOCH + std::time::Duration::from_secs(secs);
+    Some(
+        std::time::SystemTime::now()
+            .duration_since(created)
+            .unwrap_or_default(),
+    )
+}
+
+/// Whether a live chat session holds `run_dir`'s [`CHAT_RUN_LOCK_FILE`].
+/// A missing lock file means no live session (only sessions create it); a
+/// lock released by process death acquires cleanly here and reads as dead.
+/// The probe is a SHARED lock on a read-only fd (an exclusive probe would
+/// need a writable fd, misreading an unwritable lock file as "dead"); any
+/// other open failure means "unknown" and errs toward keeping the dir.
+fn chat_run_dir_is_live(run_dir: &std::path::Path) -> bool {
+    match std::fs::OpenOptions::new()
+        .read(true)
+        .open(run_dir.join(CHAT_RUN_LOCK_FILE))
+    {
+        Ok(file) => fs2::FileExt::try_lock_shared(&file).is_err(),
+        Err(error) => error.kind() != std::io::ErrorKind::NotFound,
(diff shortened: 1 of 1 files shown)


anantheparty (approved):
Adversarial pre-merge review passed — exemplary fail-safe GC design. Every ambiguity errs toward keeping: liveness is a SHARED-lock probe on the session.lock a live chat holds (an exclusive probe would misread unwritable lock files as dead; process death releases cleanly), age comes from the v7-UUID embedded timestamp (dir mtime is backdatable via cp -a/tar), non-v7/future-timestamp/foreign-named dirs and open failures all mean keep, current_run_id is never a candidate, and prune failures log without blocking chat startup. The 24h age floor protects burst sessions from retention-window eviction. Tests pin the safety edges. This fixes the read-only chat skill-binary leak (MB-scale × ~10 skills per chat, never pruned). Approving — merging left to you per convention.

Issue #2237, opened 2026-09-03
goal_create refuses to start a new goal when the session's previous goal is `archived` — archived is terminal but the admission check only accepts `complete`


ZhangHanDong:
## Summary

`goal_create` admits a new goal only when no goal exists or the existing goal's status is
exactly `"complete"`. A goal that was archived (via `octos goal archive`, or the operator
transition RPC) has status `"archived"`, which the same orchestrator treats as terminal
everywhere else (`group_status_for_goal("archived") == Completed`). The admission check
therefore refuses with:

```
goal_create: cannot create a new goal because this session has an unfinished goal (status `archived`); complete or clear the existing goal first
```

The master cannot create a goal, `goal_update` rejects the stale card as well, and the
session is left driving peers without a goal envelope (no ledger, no `/goal stop`
close-out). The TUI header keeps showing the archived goal ("Goal: resume goal_01
(archived · …)") as if it were current.

## Environment

- Linux, `octos 2.0.3-rc.10 (917c02d4)` `--features api`, octoscode 0.3.0-rc.9
- Session `octos:local:tui#coding`, profile `octos`, instance `f182cc95436c22f7`

## Reproduction (observed 2026-09-04, OLP board entry #45)

1. Session had `goal_01` archived in an earlier campaign; supervisor view shows
   goal_01..04 as archived×2 / complete×2, none active.
2. New master turn is asked to `/goal` a fresh objective.
3. `goal_create` → error above. `goal_update` → rejected ("stale verdict").
4. Master reports honestly and proceeds with `peer_handoff` only; goal hygiene step
   `/goal stop` cannot be performed at close-out.

## Root cause

`crates/octos-cli/src/autonomy/agent_orchestrator.rs` L8992 (`model_create_goal`):

```rust
if existing.status != "complete" {
    return Err(format!("cannot create a new goal because this session has an unfinished goal (status `{}`) …", existing.status));
}
```

`archived` is a terminal status: L2250 maps the `archive` transition to `"archived"`,
L2318/L2336 treat it as a terminal prior state, and L14333 maps it to
`GroupStatus::Completed`. The admission predicate should be "no *unfinished* goal", but
it is implemented as "status is `complete`", so every terminal status other than
`complete` (at least `archived`; possibly `budget_limited` after archive) blocks creation.

## Expected behavior

- `goal_create` succeeds when the existing goal is in any terminal status
  (`complete`, `archived`); the "replace a finished goal mints a fresh goal id" path
[...]

PR #2238, opened 2026-09-03
docs(goal): goal_create description admits terminal archived goals (#2237)

BH3GEI:
The behavior fix landed via #2241 (goal_create treats archived as terminal). This PR is now narrowed to the remaining doc inaccuracy: GoalCreateTool's description still told the model a goal is replaced "only when it is already complete", which no longer matches the terminal-status predicate — archived goals are terminal too.

Refs #2237


anantheparty (approved):
The goal_tool.rs model-visible description sync from the #2238 supersession note — the description now matches the #2241-landed semantics (terminal = complete or archived). 2 lines, CI green. Approving — merging.

Issue #2236, opened 2026-09-03
peer_handoff worktree fence clones a bare tree with no build cache — Rust peers burn the 50-iteration budget on cold cargo builds


ZhangHanDong:
## Summary

When `peer_handoff` fences a peer (explicit `worktree=true` or the #20a smart-fence
auto-default), the fence is a fresh `git clone --no-hardlinks` under
`peers/<slug>/wt`. The clone has no `target/`, so the first `cargo build`/`cargo test`
inside the fence compiles the whole workspace from scratch. On this repository that is
a multi-minute build per invocation; a peer with the default 50-iteration turn budget
spends most of its iterations waiting on cold compiles and exits with
"did not complete within 50 iterations", leaving uncommitted work in the fence.

The fence-by-clone design is correct for isolation (documented rationale in
`peers/mod.rs`), but it silently discards the one thing a Rust peer needs most: a warm
build cache.

## Environment

- Linux, `octos 2.0.3-rc.10 (917c02d4)` built with `--features api`
- octoscode 0.3.0-rc.9 driving `octos serve --stdio --solo --danger-full-access`
- Repository: this one (`octos`), workspace `target/` ≈ 159 GB warm cache in the parent tree

## Reproduction (observed 2026-09-04, OLP board entry #45)

1. Master session on `fix/issue-2234-linux-auth-set-key` (main tree, warm `target/`).
2. `peer_handoff(name="fix-2234-linux-keystore", brief=<self-contained brief>)` with
   `worktree` unspecified → smart-fence chose FENCED (event `peer_staged` 2026-09-03T17:58:49Z).
3. Peer cwd = `~/.octos/instances/<i>/profiles/octos/data/peers/fix-2234-linux-keystore/wt`.
4. Peer edits `crates/octos-cli/src/auth/keychain.rs` (+403 lines) and runs
   `cargo test -p octos-cli --features api …`. A fresh 8.2 GB `wt/target/` is created;
   single cargo invocations observed at 147 s and 500 s.
5. `result-1.md`: `outcome: completed`, body "The agent did not complete within 50
   iterations". Working tree left modified, nothing committed, one compile error (E0596)
   still present.

## Root cause

`crates/octos-cli/src/peers/mod.rs` L1596–L1626: the fence is a `git clone
--no-hardlinks` of the workspace into `peers/<slug>/wt`. Nothing sets `CARGO_TARGET_DIR`
(or any build-cache hint) in the peer's environment, and the clone naturally excludes the
[...]

PR #2239, opened 2026-09-03
fix(auth): platform-neutral secret store, Linux file backend, --profile attaches route-declared keys (#2234)

### LINKED ISSUE #2234: auth set-key is unusable on Linux and --profile skips route-only API keys
## Summary

`octos auth set-key` is unusable on Linux because the implementation unconditionally shells out to macOS's `security` CLI. There is a second issue in the same flow: `--profile <id>` does not attach the stored key to a profile unless that profile already has the same name in `config.env_vars`, even when its primary LLM route declares that name as `api_key_env`.

Together these make it impossible to persist a key for a route-configured profile using the documented CLI on Linux. GUI/bar-launched processes that do not source the user's interactive shell then fail to bootstrap the profile.

## Environment

- Linux: Arch Linux, kernel 7.1.8
- `octos 2.0.3-rc.1 (7254517 2026-08-05)`
- x86_64

## Reproduction

Given an enabled profile whose primary route contains:

```json
{
  "llm": {
    "primary": {
      "family_id": "zai-coding",
      "route": {
        "api_key_env": "ZAI_API_KEY"
      }
    }
  },
  "env_vars": {}
}
```

Run:

```console
$ octos auth set-key --profile octos ZAI_API_KEY
Enter value for ZAI_API_KEY: …
Error:
   0: failed to run security add-generic-password
   1: No such file or directory (os error 2)

Location:
   crates/octos-cli/src/auth/keychain.rs:137
```

No secret is persisted. A subsequently launched `octos serve` that does not inherit `ZAI_API_KEY` fails profile bootstrap with:

```text
failed to create LLM provider for profile 'octos'
```

### 1. macOS-only backend is called on every platform

`crates/octos-cli/src/auth/keychain.rs::set_secret` invokes:

```rust
Command::new("security").args(["add-generic-password", ...])
```

without a `target_os = "macos"` gate or a Linux implementation. The module itself documents that it is macOS Keychain integration. `octos-cli/Cargo.toml` enables `keyring/apple-native` and `keyring/windows-native` target features, but does not enable `keyring/linux-native`; meanwhile `octos-agent` already has a Linux-native target dependency that may offer a useful precedent.

### 2. Explicit `--profile` does not add a missing marker

`crates/octos-cli/src/commands/auth.rs::set_key` updates a profile only inside:

```rust
if profile.config.env_vars.contains_key(name) {
    profile.config.env_vars.insert(name.to_string(), marker);
}
```

A profile whose primary/fallback route declares `api_key_env: "ZAI_API_KEY"` but whose `env_vars` map does not already contain the key is skipped. In the explicit `--profile octos` case, the command should recognize that route declaration as a reference and attach the marker (or clearly reject before storing anything).

## Expected behavior

On Linux:

```console
$ octos auth set-key --profile octos ZAI_API_KEY
Enter value for ZAI_API_KEY: …
OK Stored ZAI_API_KEY securely (1 profile updated)

$ octos auth keys --profile octos
ZAI_API_KEY: available
```

A fresh process with no ambient `ZAI_API_KEY` should then be able to bootstrap the selected profile.

## Suggested fix

1. Introduce a platform-neutral secret-store interface for `set_secret`, `get_secret`, `delete_secret`, and `unlock`.
2. Keep the current `security` implementation behind `#[cfg(target_os = "macos")]`.
3. On Linux, enable `keyring/linux-native` for `octos-cli` and use the Secret Service/keyring backend. Map unavailable Secret Service, locked collection, and headless-session failures to explicit actionable errors. Do not silently fall back to plaintext.
4. Use the existing Windows-native keyring feature for a Windows implementation, or fail early with a platform-specific unsupported error until implemented.
5. When `--profile ID` is supplied, treat matching primary/fallback `route.api_key_env` and sub-provider `api_key_env` declarations as references. Insert the marker into that profile's `config.env_vars` after the secret-store write succeeds. Preserve atomicity: if the profile save fails, do not leave an unreferenced secret behind.
6. Read interactive secrets without terminal echo rather than with `stdin.read_line`.

## Tests requested

- Linux backend test with an injected/fake secret store; it must never execute `security`.
- Explicit `--profile` adds a marker when the key is referenced only by the primary route.
- Primary, fallback, and sub-provider `api_key_env` references are recognized; unrelated names are rejected or reported as unreferenced.
- Secret-store failure leaves profile JSON byte-identical.
- Profile-save failure rolls back the newly stored secret.
- `auth keys` resolves and reports the platform backend without revealing the value.
- Interactive input does not echo the secret.

No credentials or real secret values are included in this report.

ZhangHanDong:
Fixes #2234

## Summary

`octos auth set-key` was unusable on Linux (unconditional shell-out to macOS `security`), and `--profile <id>` skipped keys that a profile referenced only through its LLM route (`route.api_key_env`). This PR lands the fix shape from the 2026-09-03 triage in three commits:

1. **45a — platform-neutral secret store** (`crates/octos-cli/src/auth/keychain.rs`): the public `set_secret` / `get_secret` / `delete_secret` / `unlock` API is unchanged; macOS keeps the existing `security` path behind `#[cfg(target_os = "macos")]`; Linux gets a file store under the octos home (`secrets/`, dir 0700, one 0600 file per key, permissions re-asserted after write, no plaintext fallback); other targets return an explicit "secret store unsupported" error. New `is_available()` / `backend_name()`; `relocate_keychain_backed_secrets` and `relocate_secret_to_keychain` now take the store-availability flag instead of `is_macos`, so the #2216 migration set applies on Linux too. `unlock` is a no-op on Linux.
2. **45b — `--profile` attaches route-declared keys** (`crates/octos-cli/src/commands/auth.rs`): a key counts as referenced if it is in `env_vars`, or is the `api_key_env` of the primary route, any fallback route, or any sub-provider. Explicit `--profile` with an unreferenced name is rejected before anything is stored. Store first, then save the profile; a save failure deletes the freshly stored secret.
3. **45c — no-echo input, backend report, rollback seam**: interactive secrets are read with `rpassword` (new dependency, `rpassword = "7"`; the workspace denies `unsafe_code`, so a hand-written termios path was not an option) with an injectable reader for tests; `auth keys` reports the platform backend and availability without printing values; the profile-save step is an injectable seam so the rollback path is unit-tested.

## Tests (as requested in the issue)

- Linux backend tests run against an injected temporary root and never execute `security` (24 keychain tests, 3 macOS-only ignored on Linux).
- `--profile` adds the marker when the key is referenced only by the primary route (zai-coding shape from the report); primary / fallback / sub-provider references recognised; unrelated names rejected; 17 `commands::auth` tests.
- Secret-store failure leaves the profile JSON byte-identical; profile-save failure rolls back the stored secret; interactive input does not echo (injected reader).

Gates on Linux: `cargo test -p octos-cli --features api --lib` (keychain 24/0, commands::auth 17/0), `cargo build --features api`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`. Independently re-verified in an isolated worktree by the outer loop.

## Notes

- Windows is not implemented here; it fails early with the unsupported error (tracked in the issue's suggested fix, item 4).
- Two harness defects surfaced while fixing this one and were filed separately: #2236 (fenced peer clone has no build cache) and #2237 (`goal_create` rejects `archived`).

https://claude.ai/code/session_016zbAVTB1tkMF49Zr52UB7Z

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4684,6 +4684,7 @@ dependencies = [
  "redb",
  "regex",
  "reqwest 0.12.28",
+ "rpassword",
  "rusqlite",
  "rust-embed",
  "rustix",
@@ -6144,6 +6145,27 @@ dependencies = [
  "quote",
  "serde_json",
  "syn",
+]
+
+[[package]]
+name = "rpassword"
+version = "7.5.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2da316a15f47e3d053de9cb2c439650bd8fa4aaeb9365f2e5f27f492ff73c196"
+dependencies = [
+ "libc",
+ "rtoolbox",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
+name = "rtoolbox"
+version = "0.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9a1efe12a1469752d0e6ff5ebec0b6ef4924cc5c4c71046b0ec730040535819d"
+dependencies = [
+ "libc",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,8 @@ repository = "https://github.com/octos-org/octos"
 [workspace.lints.rust]
 unsafe_code = "deny"
 [workspace.dependencies]
+# #2234/45c — echo-free interactive secret input (safe termios wrapper; the workspace denies unsafe_code, blocking a hand-rolled termios arm)
+rpassword = "7"
 # Internal crates
 octos-core = { path = "crates/octos-core" }
 octos-diagnostics = { path = "crates/octos-diagnostics" }
--- a/crates/octos-cli/Cargo.toml
+++ b/crates/octos-cli/Cargo.toml
@@ -122,6 +122,7 @@ octos-fleet-worker.workspace = true
 # `tokio::fs::read` followed symlinks between the canonical-descendant
 # check and the read, opening a TOCTOU window.
 libc = { workspace = true }
+rpassword = { workspace = true }
 # `rustix::fs::openat` for the component-by-component path walk in
 # `src/api/preview.rs::open_no_follow_walk`. Codex round-2 BLOCKING #1
 # (issue #996): the previous `symlink_metadata`-then-`open` ancestor
--- a/crates/octos-cli/src/api/admin.rs
+++ b/crates/octos-cli/src/api/admin.rs
@@ -291,7 +291,7 @@ pub(crate) fn relocate_keychain_backed_secrets(
             env_vars,
             &key,
             profile_id,
-            cfg!(target_os = "macos"),
+            crate::auth::keychain::is_available(),
             crate::auth::keychain::set_secret,
         )
         .map_err(|e| (StatusCode::BAD_REQUEST, e))?;
@@ -5536,12 +5536,27 @@ mod tests {
 
     #[test]
     fn relocate_keychain_backed_secrets_never_persists_raw_vertex_json_off_macos() {
-        // The shared helper used by every profile/sub-account save path must
-        // refuse a raw SA JSON on a non-macOS host (where keychain storage
-        // isn't available) rather than let it fall through to plaintext config.
-        // On macOS it would relocate to the keychain instead, so only assert on
-        // the non-macOS path (the one CI runs and the plaintext risk lives on).
-        if cfg!(target_os = "macos") {
+        // #2234/45a — the availability predicate is now `keychain::is_available()`
+        // (true on Linux: the file backend exists), NOT `cfg!(macos)`. The
+        // never-plaintext contract holds where NO backend exists (unsupported
+        // platforms); on Linux the raw JSON is legitimately relocated into the
+        // file store and the env slot becomes a marker.
+        if crate::auth::keychain::is_available() {
+            // Store-backed host (macOS keychain / linux file): relocation
+            // succeeds and the plaintext is replaced by a marker.
+            let mut env = std::collections::HashMap::new();
+            env.insert(
+                "VERTEX_SA_JSON".to_string(),
+                r#"{"type":"service_account","private_key":"x","project_id":"p"}"#.to_string(),
+            );
+            relocate_keychain_backed_secrets(&mut env, "sub-account-1")
+                .expect("store-backed host relocates raw SA JSON");
+            let stored = env.get("VERTEX_SA_JSON").expect("slot present");
+            assert!(
+                !stored.contains("private_key"),
+                "raw JSON must not persist as plaintext; got: {stored}"
+            );
+            assert!(stored.contains("keychain"), "marker present: {stored}");
             return;
         }
         let mut env = std::collections::HashMap::new();
@@ -5552,7 +5567,7 @@ mod tests {
         let res = relocate_keychain_backed_secrets(&mut env, "sub-account-1");
         assert!(
             res.is_err(),
-            "raw VERTEX_SA_JSON must be rejected off macOS, never saved as plaintext"
+            "raw VERTEX_SA_JSON must be rejected on hosts with no secret store"
         );
         // The raw value is left untouched (the caller bails before saving).
         assert!(env.get("VERTEX_SA_JSON").unwrap().starts_with('{'));
--- a/crates/octos-cli/src/api/admin.rs
+++ b/crates/octos-cli/src/api/admin.rs
@@ -5576,22 +5576,44 @@ mod tests {
     #[test]
     fn relocate_rejects_service_account_json_under_custom_env_name_off_macos() {
         // The dashboard "Custom" bypass: SA JSON pasted under VERTEX_API_KEY
-        // (not the whitelisted name) must still be caught by content detection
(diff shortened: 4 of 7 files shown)


anantheparty (approved):
Adversarial pre-merge review of the secret-store surface passed (the highest-stakes surface this week — reviewed accordingly):

- **Linux file backend**: 0600 applied atomically at creation (OpenOptions .mode()) with a set_permissions re-assert against umask quirks; secrets_root dir 0700; test root injection keeps the suite hermetic (never touches the real keychain).
- **Platform-neutral facade**: keychain::is_available/set/get/delete + backend_name; the availability predicate is now keychain::is_available() (true on Linux), NOT cfg!(macos) — and the never-plaintext contract still holds on hosts with NO backend (rejection kept).
- **Migration (#2216 interplay)**: on store-backed hosts the raw JSON relocates into the file store and the env slot becomes a keychain marker — the raw value never remains (marker assertions in the tests).
- **--profile route attach** (root cause 2): the key-name match now covers primary route api_key_env, fallback routes, AND sub_providers[] — the issue's zai-coding shape (env_vars empty + route declaring ZAI_API_KEY) attaches; unreferenced keys are REFUSED with an error listing every checked location (prevents mistyping-based orphan storage).
- mac keychain path preserved; tests comprehensive.

Approving — merging left to you per convention.

BH3GEI (commented):
Thanks for tackling #2234 — the overall shape (platform-neutral facade, Linux file store with 0700/0600, route-declared key references, store-then-save ordering) is right, and the predicate change from `cfg!(macos)` to `is_available()` is the correct contract. However, verification on this head (6d3a596) turned up one blocking build break and several issues that need another round.

**1. Blocking: the crate no longer compiles on macOS.** `crates/octos-cli/src/auth/keychain.rs:16` narrowed `use eyre::{Result, WrapErr};` to `use eyre::Result;`, but the macOS arms still call `.wrap_err(...)` — `macos_unlock` (keychain.rs:384) and `macos_set_secret` (keychain.rs:428). Reproduced locally on macOS (Darwin 25.5, aarch64): `cargo check -p octos-cli --features api` fails with E0599 at both sites (the `WrapErr` import inside `linux_file` is cfg-gated away on macOS). Linux CI can't see this. Fix: keep `WrapErr` in the top-level import (it's used by macOS and Linux arms alike).

**2. `cargo test` doesn't compile off-Linux.** `test_override_secrets_root` is exported only under `#[cfg(all(test, target_os = "linux"))]` (keychain.rs:371), but two call sites are not cfg-gated: `crates/octos-cli/src/api/admin.rs:5589` and `crates/octos-cli/src/api/auth_handlers.rs:6038`. Same local run: E0425 at both. Either gate those tests to Linux or provide a no-op stub on other platforms (the doc comment at keychain.rs:369 already promises "no-op elsewhere" — the stub would make that true).

**3. The relocation test is no longer hermetic.** `relocate_keychain_backed_secrets_never_persists_raw_vertex_json_off_macos` (admin.rs:5538): the `is_available()` arm relocates **without** injecting a test root (unlike its twin at 5589). On Linux this writes a real `~/.octos/secrets/VERTEX_SA_JSON::sub-account-1` on the dev/CI host with no cleanup; on macOS (once #1/#2 are fixed) it writes into the developer's real login keychain — exactly what the old `if cfg!(macos) { return; }` skip existed to avoid. It also goes flaky where `$HOME` is read-only (`is_available()` true, `set_secret` fails, `.expect()` panics).

**4. Rollback can destroy a pre-existing or shared secret.** `set_key_with_save` (commands/auth.rs:389-394) answers any profile-save failure with `delete_secret(&account)`. Two loss shapes: (a) key rotation — re-running `set-key` overwrites the existing secret, and if the save then fails, the rollback deletes the account entirely while the on-disk profile still holds the marker, so the previous good credential is permanently lost; (b) shared (non-scoped) key without `--profile` — profile A saves with the bare marker, profile B's save fails, and the rollback deletes the *shared* account that A (and any pre-existing bare-marker profiles) now resolve against. The current tests only cover the fresh-scoped-account case where deletion happens to be correct. Consider only deleting when the account didn't exist before this call, or snapshotting and restoring the prior value.

**5. Linux backend write is non-atomic and follows symlinks** (`linux_file::set_secret`, keychain.rs:318-336): (a) no `O_NOFOLLOW` on `root.join(name)` — a same-UID process that plants `~/.octos/secrets/FOO` as a symlink gets the target truncated, the secret written through it, and the target chmod'd 0600 (this repo already established the no-follow precedent with the `openat` walk in preview.rs); (b) truncate-then-write means a crash/interrupt mid-write leaves a zero-byte/partial secret that `get_secret` then returns as `Ok(Some(""))` with no error, silently injecting an empty credential at runtime — write to a temp file + rename keeps the old value durable until the new one is; (c) for a pre-existing file with loose permissions, `.mode(0600)` is creation-only, so the secret is written while world-readable and only chmod'd after — chmod before write instead.

Minor: on Linux, `octos auth unlock` now reports "Keychain is already unlocked" (and `is_accessible()` mkdir/chmods `~/.octos/secrets` as a side effect of what reads as a probe, keychain.rs:309-311), while the help text and fallback prompt still say "macOS login password" (commands/auth.rs:626-635).

What I verified: full diff against origin/main (b47b6671); the macOS compile failures above reproduced locally; the rollback and hermeticity items traced through `keychain_target`/`should_scope` and the marker-resolution paths. The `--profile` reference guard, route/sub-provider matching, and store-then-save ordering all read correctly.

Issue #2233, opened 2026-09-03
llm discovery: api_type=responses + claude-* probes the wrong strategy/root (pre-existing gap made explicit by per-model resolution)


anantheparty:
Follow-up nits from the #2231 review (r9s per-model model discovery), split out per the review.

1. **`api_type:"responses"` + r9s + `claude-*` probe/serve mismatch**: `resolve_model_discovery` only special-cases `api_type=="anthropic"` — `"responses"` falls into the per-model branch, so r9s claude-* probes `{base}/anthropic/v1/models` while the runtime `create_provider_with_api_type` bypasses the registry for "responses" and hits `{base}` via `OpenAIResponsesProvider`. Discovery and runtime fully disagree for this combination. Pre-existing since #2183 (discovery never modeled "responses"); #2231 made it explicit ("wrong" instead of "approximate"). Advisory-only, rare combination. Fix: discovery returns OPENAI without rewrite for "responses".

2. **Padded-model trim asymmetry (edge)**: both discovery entry points trim the model, but the runtime `create` matches raw — a hand-edited `" claude-x"` (with whitespace) classifies Anthropic in discovery but serves OpenAI. Unreachable via AppUI (save path trims); record only.

Refs #2231 #2185 #2183

BH3GEI:
Both items are resolved on current main (ab09eae5):

- **Item 1** (`api_type: "responses"` + claude-* probe/serve mismatch): fixed by #2243 (merged as 38eca094). `resolve_model_discovery` now returns the OpenAI-compatible listing for `api_type == Some("responses")` regardless of the family's declared strategy (`crates/octos-llm/src/discovery.rs:198`), and the pin test `should_use_openai_strategy_when_api_type_is_the_responses_override` passes on main (`cargo test -p octos-llm --lib`: ok, 1 passed).
- **Item 2** (padded-model trim asymmetry) was marked "record only" in the issue body — unreachable via AppUI since the save path trims.

Closing as completed.

PR #2243, opened 2026-09-04
fix(llm): probe the OpenAI-compatible listing for api_type=responses routes

BH3GEI:
## Problem

`resolve_model_discovery` only special-cased `api_type: "anthropic"`. A route with `api_type: "responses"` on a family whose declared discovery protocol is Anthropic Messages (zai, zai-coding, anthropic) probed `{root}/v1/models` with `x-api-key`/`anthropic-version` — while the runtime's `create_provider_with_api_type` bypasses the registry for `"responses"` and serves the OpenAI Responses API at `{base}` via `OpenAIResponsesProvider`. Discovery and the runtime fully disagreed for that combination (pre-existing since #2183; made explicit by the per-model resolution review in #2231).

This PR covers item 1 of #2233 only. Item 2 (padded-model trim asymmetry) is marked "record only" in the issue and is intentionally untouched.

## Root cause

`resolve_model_discovery` (`crates/octos-llm/src/discovery.rs`) mirrors the runtime's api_type precedence but only modeled the `"anthropic"` registry bypass, not the `"responses"` one, so `"responses"` fell through to the family's declared strategy.

## Fix

`api_type == Some("responses")` now resolves to the OpenAI-compatible `GET {base}/models` listing regardless of the family's declared strategy — mirroring the runtime's registry bypass. One exception: a family that declares discovery `Unsupported` (vertex) stays manual-only, since no protocol override conjures up a listing endpoint it does not have (and probing would send the Vertex service-account JSON as a Bearer token to the configured host). Both call surfaces (admin REST `/api/my/provider-models` and AppUI `profile/llm/fetch_models`) share this one resolver, so both are fixed.

## Test evidence

- New test `should_use_openai_strategy_when_api_type_is_the_responses_override` fails on the base commit (`zai + responses` → `AnthropicMessages`) and passes after; it also pins `gemini + responses` → OpenAI listing and `vertex + responses` → unchanged `Unsupported`.
- `cargo test -p octos-llm --lib`: 607 passed, 0 failed, first run.
- `cargo fmt --check` clean; `cargo clippy -p octos-llm --all-targets -- -D warnings` zero warnings.

## End-to-end evidence

Real `octos serve` (api feature) + real admin REST `POST /api/my/provider-models` against a loopback listing fixture that records path + auth headers, same request body `{provider: "zai", api_key, base_url, api_type: "responses"}`:

- **Before** (binary built from `origin/main`): fixture recorded `GET /api/anthropic/v1/models` with `x-api-key` + `anthropic-version: 2023-06-01` — the Anthropic Messages listing, disagreeing with the Responses runtime.
- **After**: fixture recorded `GET /api/anthropic/models` with `Authorization: Bearer` — the OpenAI-compatible listing, matching `OpenAIResponsesProvider`'s `{base}` surface.
- Regression probes on the fixed binary: zai without `api_type` still probes the Anthropic listing unchanged; `vertex + responses` returns `[]` advisory with **no** outbound probe (verified: fixture log unchanged).

Known residual (pre-existing, out of scope): the runtime's `responses` branch applies only an explicit `base_url` and never falls back to the family default root, while discovery root resolution does — so a `responses` route without `base_url` on a non-OpenAI family can still disagree on root. Strategy selection (this fix) is what #2233 asked for.

## Review

One independent review pass (adversarial maintainer perspective): no blockers/majors. Three minors accepted and fixed (dropped a misleading `custom` test assertion — the runtime bails on custom+responses; preserved `Unsupported` for vertex; added gemini/vertex pins), one nit rejected as pre-existing/out-of-scope (root-fallback divergence above). All verification re-run green after those changes.

Refs #2233


anantheparty (approved):
This closes the #2233 follow-up: discovery now special-cases api_type="responses" to probe the OpenAI-compatible listing instead of the family-declared Anthropic strategy — the probe/serve mismatch for responses routes (e.g. zai with a responses override) is resolved per the nit's suggested fix. Scoped to the one function, tests included. Approving — merging left to you per convention.

Issue #2251, opened 2026-09-04
test: sovereignty provider tests fail on macOS — Apple Git system config sets init.defaultBranch=main


BH3GEI:
## Summary

`autonomy::agent_orchestrator::tests::sovereignty_provider_fail_open_when_unowned_default_branch` (and likely its siblings using the same helper) fail deterministically on macOS:

```
thread '...sovereignty_provider_fail_open_when_unowned_default_branch' panicked at crates/octos-cli/src/autonomy/agent_orchestrator.rs:33632:13:
git ["checkout", "-q", "-b", "main"] failed
```

## Root cause

`init_repo_on_branch` (`crates/octos-cli/src/autonomy/agent_orchestrator.rs` ~33627) runs `git init` then `git checkout -b <branch>` with `branch = "main"`. Apple Git (2.50.1, Apple Git-155) ships a **system** gitconfig at `/Library/Developer/CommandLineTools/usr/share/git-core/gitconfig` containing `init.defaultbranch=main`, so the initial commit already creates `main` and `checkout -b main` exits 128 (`fatal: a branch named 'main' already exists`). On CI (ubuntu, default `master`) it passes — the test silently assumes the platform default branch is not `main`.

## Suggested fix

Make the helper independent of the platform default, e.g. `git init -q -b master` (or `git init -q -b <anything-but->`), or use `git checkout -q -B <branch>`.

Reproduced on unmodified `origin/main` (e08c2715) with `cargo test -p octos-cli --lib --features api`.

BH3GEI:
Fixed on main by d51601df (#2265): `init_repo_on_branch` now seeds the fixture repo with `git init -q -b fixture-seed` and switches to the requested branch with a create-or-switch checkout, so a platform `init.defaultBranch=main` no longer collides with `checkout -b main`.

Verified on macOS with Apple Git-155 (system gitconfig sets `init.defaultbranch=main`): on main @ 1db86bc1, `cargo test -p octos-cli sovereignty_provider` passes both tests that use the helper (`sovereignty_provider_fail_open_when_unowned_default_branch`, `dual_goal_sovereignty_provider_scan_claim_and_denial`). Closing.

Issue #2053, opened 2026-08-17
check-windows flakes on fixed-duration timeouts (rustfmt 5s, background spawn, speculative overflow)


ymote:
`check-windows` fails intermittently on timing-based tests. Three occurrences today alone, on PRs whose diffs cannot affect them.

## Observations

| run | failures |
|---|---|
| #2043 attempt 1 | `edit_file::…format_after_edit_enabled`, `write_file::…format_after_edit_enabled`, `spawn::test_background_spawn_persists_workflow_phase_transitions` |
| #2043 attempt 1 (in-job retry) | `edit_file::…format_after_edit_enabled` only |
| #2043 attempt 1 (octos-cli) | `session_actor::tests::test_speculative_overflow_concurrent` |
| #2052 | `edit_file::…format_after_edit_enabled` only |
| #2043 re-run | all pass |

The retry inside a single job produced a *different* failure set (3 → 1), and a plain re-run went fully green. That is flakiness, not breakage.

## Cause

All of them are fixed-duration waits that a slow Windows runner misses:

```
output must state the file was reformatted: Successfully edited code.rs
Note: post-edit formatting with rustfmt timed out after 5s and was killed;
      the edit itself was applied unchanged.
```

```
background spawn task did not complete in time
```

`edit_file.rs:1312` and `write_file.rs:674` assert the output says the file was reformatted, but the 5s rustfmt budget elapses first on Windows, so the tool reports the un-formatted path and the assertion fires. On macOS/Linux rustfmt returns well inside 5s, which is why this is Windows-only.

## Why this matters more than the individual failures

The workaround is to re-run until green. That is the same route-around-a-red-check pattern that let #1923 sit unnoticed — it just costs 45 minutes each time instead of hiding entirely. It also trains readers to assume a red `check-windows` is noise, which is exactly when a real Windows regression slips through.

## Suggested fix

Scale the timeout by platform rather than re-running: a larger rustfmt budget under `cfg(windows)` (or driven by an env var CI sets), and the same for the background-spawn and speculative-overflow waits. The assertions stay meaningful; only the patience changes.

Found while merging #2043/#2048/#2041 — not caused by any of them.

ymote:
Fourth occurrence today, on #2049 (`7a524ee86`):

```
thread 'tools::edit_file::tests::should_return_formatted_content_when_format_after_edit_enabled'
  panicked at crates\octos-agent\src\tools\edit_file.rs:1312:9
[...]

PR #2252, opened 2026-09-04
fix(test): master_continuation_tick_reenters_actor_loop — condition-based wait instead of fixed 10-yield spin

### LINKED ISSUE #2011: flaky: master_continuation_tick_reenters_actor_loop fails ~25% on main — fixed 10-yield spin races real scheduling (has caused 3 false attributions)

BH3GEI:
## Problem

`session_actor::tests::master_continuation_tick_reenters_actor_loop` failed ~25% of the time on unmodified `main` under machine load (#2011). It waited for the tick → drain → `process_inbound` → persist chain with a fixed 10-iteration `advance(250ms)` + `yield_now` budget. `tokio::time::advance` moves **virtual** time while the chain also needs **real** scheduling (blocking I/O threads), so under load the budget expired before the work landed and the assertion fired with `: []` — producing false attributions on unrelated branches.

## Root cause

A fixed iteration budget on virtual time races real-world scheduling. Ten cooperative yields grant the chain essentially zero wall-clock time; on a loaded machine the persistence write (blocking pool thread) has not run when the assertion fires.

## Fix

Test-only change, confined to the one test. Both waits are now bounded by a 10s **wall-clock** deadline (`std::time::Instant` — tokio's clock is paused in this `start_paused` test), and each virtual `advance` is paced with a real 10ms `std::thread::sleep` so starved OS threads get CPU. On timeout the loops fall through to the **original assertions unchanged**, so a genuine regression still fails with the same diagnostic messages (including the full message dump).

The comment also records what actually keeps the actor alive during the wait (verified against `session_actor.rs`): the pre-loop `upsert_agent` is synchronous, the first `interval` tick drains it immediately and resets `idle_sleep`, and `idle_sleep` cannot fire mid-turn — so the deadline only has to cover [actor start → first drain].

The sweep for the same pattern elsewhere is deliberately **not** in this PR: the `format_after_edit` rustfmt-budget cases are tracked by #2053, and no other `for _ in 0..N { yield_now }`-before-assert instance of this shape exists in this test file.

## Test evidence

Reproduced and verified on an 8-core macOS host under artificial load (16 CPU spinners, load average ~60-90), running the compiled test binary directly:

- **Before** (unmodified `origin/main` @ e08c2715): **4/24 runs failed**, with the exact signature from the issue — `panicked at crates/octos-cli/src/session_actor_tests.rs:2525:5: master continuation should persist the model-generated progress summary: []`.
- **After**: **0/30 runs failed**, and 0/30 again in a second load run after the review-driven comment edit (60/60 total) — exceeds the issue's 20/20-under-load acceptance bar.
- `cargo test -p octos-cli --lib --features api`: **3202 passed, 0 failed** (2 filtered out — see below).
- `cargo test -p octos-cli --lib` (default features; this test is `api`-gated and not compiled): 1637 passed, 1 failed — the failure is `sovereignty_provider_fail_open_when_unowned_default_branch`, a **pre-existing, unrelated** environmental failure: Apple Git's system gitconfig sets `init.defaultbranch=main`, so that test's fixture `git checkout -b main` collides with the already-existing branch. Root-caused and filed separately as #2251; its code path is untouched by this diff.
- `cargo fmt --check`: clean. `cargo clippy -p octos-cli --all-targets --features api -- -D warnings`: zero warnings (run twice, before and after the comment edit).

## End-to-end note

The fix target is itself an integration-level test (real `SessionActor\::run loop, real `SessionHandle` file persistence, mock LLM provider), so the load-loop runs above **are** the end-to-end evidence of the behavior difference: same binary path, before = 4/24 red with the issue's exact signature, after = 60/60 green under the same load.

## Review

- Round 1 self-review: full `git diff origin/main` read line-by-line; no debug residue, no unrelated changes, failure paths keep the original assertions.
- Round 2 independent review (fresh-eyes subagent given only the issue text + diff + file paths): no blockers; one NIT — the original comment misattributed the idle-timeout safety to pacing. Corrected to credit the actual invariants (synchronous pre-loop upsert, reset-on-drain, no mid-turn idle fire).
- Round 3: after that comment edit, reran the 30x load loop (0/30), the full `--features api` lib suite (3202/3202), and clippy (clean).

Closes #2011

### DIFF
--- a/crates/octos-cli/src/session_actor_tests.rs
+++ b/crates/octos-cli/src/session_actor_tests.rs
@@ -2498,11 +2498,21 @@ async fn master_continuation_tick_reenters_actor_loop() {
         })
         .unwrap();
 
-    for _ in 0..10 {
+    // #2011: `advance` moves VIRTUAL time, but the tick → drain →
+    // process_inbound → persist chain also needs REAL scheduling (blocking
+    // I/O threads), so a fixed iteration budget races machine load and
+    // flaked ~25% under high load. Bound the wait by a wall-clock deadline
+    // (std Instant — tokio's clock is paused in this test) and pace each
+    // virtual advance with a real sleep so starved OS threads get CPU.
+    // Pacing is not what keeps the actor alive: the pre-loop upsert is
+    // synchronous, the first interval tick drains it immediately (which
+    // resets idle_sleep), and idle_sleep cannot fire mid-turn — the
+    // deadline only has to cover [actor start → first drain].
+    let deadline = std::time::Instant::now() + Duration::from_secs(10);
+    while provider.call_count.load(Ordering::Relaxed) == 0 && std::time::Instant::now() < deadline {
         tokio::time::advance(Duration::from_millis(250)).await;
-        if provider.call_count.load(Ordering::Relaxed) > 0 {
-            break;
-        }
+        tokio::task::yield_now().await;
+        std::thread::sleep(Duration::from_millis(10));
     }
 
     assert!(
@@ -2510,24 +2520,19 @@ async fn master_continuation_tick_reenters_actor_loop() {
         "periodic actor tick must drain queued child completion into process_inbound"
     );
 
-    // `process_inbound` persists through the real spawn-blocking JSONL path.
-    // This test uses a paused Tokio clock, so advancing virtual time alone
-    // cannot guarantee that the blocking-pool completion has been observed.
-    // Poll in bounded real-time slices, matching the goal-continuation test
-    // below, instead of racing the durable append.
-    for _ in 0..500 {
-        tokio::task::spawn_blocking(|| {
-            std::thread::sleep(Duration::from_millis(2));
-        })
-        .await
-        .unwrap();
+    let deadline = std::time::Instant::now() + Duration::from_secs(10);
+    loop {
         let session_handle = SessionHandle::open(dir.path(), &session_id);
         if session_handle.session().messages.iter().any(|message| {
             message.role == MessageRole::Assistant
                 && message.content.contains("child progress summary")
-        }) {
+        }) || std::time::Instant::now() >= deadline
+        {
             break;
         }
+        tokio::time::advance(Duration::from_millis(250)).await;
+        tokio::task::yield_now().await;
+        std::thread::sleep(Duration::from_millis(10));
     }
     let session_handle = SessionHandle::open(dir.path(), &session_id);
     let session = session_handle.session();


BH3GEI:
Rebased onto current main (1db86bc1) to clear the conflict in `session_actor_tests.rs`: #2265 independently reworked this test's second wait (the persistence poll). The rebase keeps this PR's deadline-based loops for both waits — the first wait's fixed 10-advance spin (the actual #2011 flake) is still unfixed on main, and the deadline form subsumes the second wait's fixed 500-iteration budget with a stronger 10s wall-clock bound while keeping both waits in one consistent style.

Verified locally: `master_continuation_tick_reenters_actor_loop` passes 20/20 consecutive runs on the rebased branch.

anantheparty (approved):
Re-approving the updated head - the condition-based wait with the wall-clock deadline review stands (the deadline scoping comment covers start-to-first-drain). Test-only. Approving - merging left to you per convention.

PR #2256, opened 2026-09-04
fix(skills): resolve ./-prefixed MCP server args against the skill dir

### LINKED ISSUE #2255: skills: manifest MCP server args with ./ are not resolved against the skill dir
`plugins/extras.rs::resolve_mcp_server` resolves only the `command` field against the skill dir; `args` pass through verbatim, and `connect_stdio` (`mcp.rs`) sets no `current_dir`. A skill declaring `{"command": "node", "args": ["./server.js"]\}` therefore resolves the script against the daemon cwd and fails to start — the hook-side twin of this bug was #2245 (fixed in #2254).

Suggested: apply the same `./`/`../` → `skill_dir` join to every `args` element (a shared `resolve_skill_path(arg, skill_dir)` helper would now serve both call sites), or document that only `command` is resolved.

Found during review of #2254.

BH3GEI:
## Problem

A skill manifest declaring an MCP server with a skill-relative script in `args` — e.g. `{"command": "node", "args": ["./server.js"]}` — failed to start: `resolve_mcp_server` resolved only the `command` field against the skill dir while `args` passed through verbatim, and `connect_stdio` (`mcp.rs`) sets no `current_dir`, so the child process resolved `./server.js` against the daemon's cwd. This is the MCP-side twin of the hook bug fixed in #2254.

## Root cause

`plugins/extras.rs::resolve_mcp_server` applied the `./`/`../` → `skill_dir` join to `command` only; `args: srv.args.clone()` went out unchanged.

## Fix

Extract the existing resolution predicate into a shared `resolve_skill_path(arg, skill_dir)` helper (as suggested in the issue) and apply it to `command` and to every `args` element. Bare commands (`node`), flags (`--port`), flag values with embedded `./` (`--config=./other.cfg`), absolute paths, and plain arguments pass through verbatim — identical predicate to the pre-existing `command` behavior and to the hook-side resolution, so manifest semantics stay uniform. `resolve_hook` is deliberately untouched: open PR #2254 is rewriting that function, and whichever lands second should unify it onto the helper.

Known limitation (pre-existing, shared with `command` and the hook side): only `/`-spelled `./`/`../` prefixes resolve; Windows-native `.`/`..` spellings do not. Portable manifests use `/`, which Windows accepts.

## Test evidence

- New unit test `test_resolve_mcp_relative_args`: **red before the fix** (args came back verbatim), green after. Pins `./` and `../` resolution plus negative cases (`--port`, `--config=./other.cfg`, bare `..`, `/abs/path.cfg`, `plain-arg`).
- `cargo test -p octos-agent`: 2808 passed; the only 3 failures (`docker_sandbox_fails_command_validator_closed` and two sandbox-threading tests) fail identically on unmodified `origin/main` — they require a Docker sandbox backend this macOS host doesn't provide. Unrelated to this change.
- `cargo fmt --check` clean; `cargo clippy -p octos-agent --all-targets -- -D warnings` zero warnings.

## End-to-end evidence

Drove the real production path (`PluginLoader::load_into` → `resolve_extras` → `McpClient::start` → `connect_stdio` spawn) with a real skill dir containing `manifest.json` (`{"command": "python3", "args": ["./server.py"]}`) and a stub MCP stdio server that writes marker files on spawn and on answering `initialize`:

- **Before the fix**: resolved args `["./server.py"]`; the child died (`python3` can't open the file relative to the daemon cwd) — no marker files, server silently skipped.
- **After the fix**: resolved args `["<skill_dir>/./server.py"]`; spawn marker shows the script executed with the daemon's cwd, and the MCP `initialize` handshake completed.

## Review

Multi-round review: self-review of the full diff, then an independent reviewer subagent (issue text + diff + file paths only, no author reasoning). Its findings and dispositions: Windows `.` spelling gap (kept the conventional predicate for consistency with `command`/#2254; disclosed above), cwd-relative-arg semantic change (accepted — it's the requested behavior and docs already promised skill-relative resolution), test gaps (fixed: added embedded-`./` and bare-`..` negative cases), `resolve_hook` duplication (deferred to avoid conflicting with #2254). Full suite, fmt, clippy, and the end-to-end run were all repeated green after the review-driven test changes.

Closes #2255

### DIFF
--- a/crates/octos-agent/src/plugins/extras.rs
+++ b/crates/octos-agent/src/plugins/extras.rs
@@ -51,8 +51,8 @@ pub struct SkillExtras {
 
 /// Resolve manifest extras against the skill directory.
 ///
-/// - MCP: resolves relative commands against `skill_dir`, looks up env var names
-///   from the process environment.
+/// - MCP: resolves relative commands and `./`/`../` arg elements against
+///   `skill_dir`, looks up env var names from the process environment.
 /// - Hooks: parses event strings into `HookEvent`, resolves relative command paths.
 /// - Discovery: renders a short 5-line skill card (PR-F) preceded by a
 ///   generic exploration preamble emitted once per call.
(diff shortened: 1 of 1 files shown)


anantheparty (approved):
Re-approving the updated head - the MCP server args resolution fix review stands (./-prefixed args resolve against the skill dir). CI green. Approving - merging left to you per convention.

PR #2260, opened 2026-09-04
fix(auth): restore macOS-gated WrapErr import in keychain.rs (#2258)

ZhangHanDong:
## Summary

Fixes #2258. #2239 (b5db3121) removed `WrapErr` from the `eyre` import because it became unused on Linux, but the `#[cfg(target_os = "macos")]` helpers `macos_unlock` / `macos_set_secret` still call `.wrap_err(...)`, so `octos-cli` stopped compiling on macOS (seen on #2257's macos-14 `e2e-tool-regression` job).

One change: a macOS-gated `use eyre::WrapErr;` next to the existing `use eyre::Result;`, so Linux clippy (`unused_imports`, `-D warnings`) and macOS rustc agree.

## Verification

- Linux: `cargo check -p octos-cli` and `cargo clippy -p octos-cli -- -D warnings` clean.
- macOS: not buildable locally; the macOS `e2e-tool-regression` job on this PR (if `e2e-changes` fires) or a maintainer's local macOS build is the proof. #2259 tracks making a macOS compile a required check.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016zbAVTB1tkMF49Zr52UB7Z

Issue #2259, opened 2026-09-04
CI: required checks never compile the macOS target (blind spot that let #2239 break main on macOS)


ZhangHanDong:
## Problem

The required checks on `main` (`check`, `check-matrix`, `check-windows`, `test-octos-*`) do not compile for macOS. The only macOS build is `e2e-tool-regression` (macos-14, `--all-features`), which runs only when `e2e-changes` fires. Result: #2239 merged green while breaking `octos-cli` on macOS (see the companion compile-error issue).

## Proposal

Add a macOS `cargo check -p octos-cli` (default features, or `--all-features` if the runner has the deps) to `check-matrix` or as its own required job, so a cfg(target_os = "macos")-only compile error blocks merge. Cost: one macos runner per PR for a check, not a full test run.

Operator-tier (ruleset required-checks list changes with it).

Found by the OctoLoop evolution loop (octoscode LEP-003).

anantheparty:
Maintainer evidence addendum: I attempted **local cross-platform checks** (the pre-CI alternative) and both are infeasible from a Linux box — which strengthens this issue's premise that **CI gates are the only viable path**:

- `cargo check --target aarch64-apple-darwin --workspace` → dies at `lzma-sys` (cc-rs needs a macOS C toolchain; no osxcross).
- `cargo check --target x86_64-pc-windows-msvc --workspace --all-targets` with the pinned 1.97.1 → dies at the same class of C build scripts (no MSVC-compatible cc from Linux; GNU-target warnings then build-script failure).

So: no local gate can catch the #2258/#2263/#2267-Windows family. The three escapes to date all share one property — **the broken code was cfg-gated (macos/windows/test-only), invisible to the Linux-only required checks**. A `cargo check --all-targets` on macos-14 and windows-latest runners (even without running tests — compile-only catches the E0435/E0425/E0599 class) would have caught all three. Public repo, free runners.

Suggested minimal gate (compile-only, no test execution cost):
```yaml
check-darwin:  cargo check --workspace --all-targets (macos-14)
check-win-test: cargo test --workspace --all-targets --no-run (windows-latest)
```
aggregated into `CI required`.

PR #2261, opened 2026-09-05
fix(auth): preserve credentials and harden secret storage after #2239

ymote:
## Summary

Follow-up to the five confirmed findings in #2239:

- Restore the macOS-gated `eyre::WrapErr` import so ordinary macOS builds compile.
- Snapshot and restore existing credentials on failed profile saves; remove only newly created, uncommitted accounts. When an earlier profile has already saved a shared marker, retain that shared credential and report partial completion explicitly. Report rollback failures rather than claiming rollback succeeded.
- Make the test-store override available on every host and actually redirect operations away from the real credential store. On Unix it exercises the file backend; on Windows it uses encoded temporary fixture files. Guards are thread-local, non-Send, and restore nested overrides.
- Isolate all three changed API relocation tests and keep their temporary directories scoped for cleanup; add API auth/relocation test steps to the existing Windows CI job.
- Anchor file operations to no-follow directory descriptors. Write through an exclusive private temporary file, restrict its opened descriptor before writing, sync, and atomically rename within the same directory. Account symlinks/hardlinks cannot redirect writes; symlinked roots/parents and non-regular reads are refused. Failed replacement cleans up the temporary file.

This is per-account, single-command recovery, **not** an atomic multi-profile or concurrent-writer transaction. The restrictive-umask regression covers file permissions inside an already-owned existing root, not arbitrary ancestor initialization.

## Related work / merge coordination

Main was rechecked at `e64b531f16501f953d2d6a963833c7c3ef2785e7` before publication. #2260 is still open and independently contains the guarded `WrapErr` import; this branch temporarily includes that same semantic fix because it is needed to compile locally. If #2260 merges first, rebase/reconcile that import before merging this PR. No changes or comments were posted to #2260. The broader macOS CI gate tracked by #2259 is intentionally out of scope here.

## Validation

On macOS / aarch64, with fake fixture values only:

- TDD RED: API-enabled CLI auth run had 64 passing tests and 6 intended failures, proving existing/shared credential loss and symlink/hardlink escape.
- Final `cargo test -p octos-cli --features api --lib auth`: **221 passed, 3 ignored**.
- Final `cargo test -p octos-cli --features api --lib relocate`: **7 passed**.
- Exact-source Unix file-store harness: **26 passed, 3 ignored**. Additional child-only umask regression reproduced mode `000` before the descriptor permission fix and passed afterward; the parent test process's umask is never changed.
- `cargo check -p octos-cli`: passed (ordinary macOS production build).
- `cargo clippy -p octos-cli --features api --all-targets -- -D warnings`: passed.
- `cargo fmt --all -- --check` and `git diff --check`: passed.

No real Keychain access, credentials, providers, or installed service processes were used for these fixtures. No binaries were installed. Native Linux/Windows execution is not claimed locally; the exact Unix backend runs under the macOS test adapter, and Windows API test coverage is added to CI.

### Full-suite limits (not a blanket all-targets pass)

The first full API-enabled parallel library run had **3,243 passed / 1 failed / 10 ignored**. Its untouched sovereignty test does `git init` followed by `git checkout -b main`, conflicting with this host's global `init.defaultBranch=main`. Focused RED and a process-only `init.defaultBranch=master` GREEN control confirmed that issue; no global Git configuration was changed.

A controlled full parallel run then had **3,242 passed / 2 failed / 10 ignored** in unchanged runtime tests: `closed_peer_park_produces_neither_wake_nor_escalation_row` (temporary ledger ENOENT) and `master_continuation_tick_reenters_actor_loop` (completion not yet persisted). The former passed focused; the latter also failed focused. These tests do not participate in the auth changes and are not modified or waived by this PR.

One bounded serial full audit completed with **3,241 passed / 3 failed / 10 ignored** (519 seconds): the unchanged `should_delete_primary_promote_fallback_and_reload_dynamic_profile_runtime`, `should_upsert_endpoint_edit_reload_dynamic_profile_runtime_for_next_turn`, and `master_continuation_tick_reenters_actor_loop` runtime tests failed. These full commands stopped at the library failure, so subsequent integration targets were not executed by them. The full audits preceded only the final descriptor-fchmod hardening; focused tests and strict clippy were rerun against the final commit. No blanket all-targets test pass is claimed.

All raw failures and successful focused logs are retained in the isolated local evidence directory. No unrelated OUP/runtime fixes or shared-worktree changes are included.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -769,6 +769,16 @@ jobs:
       - name: Build (octos-cli + all features)
         run: cargo build -p octos-cli --features ${{ env.FEATURES }}
 
+      # #2239: default-feature tests never compile the API's test-only secret
+      # hooks. Exercise those callers on Windows too, using isolated fixtures.
+      - name: Test API auth and secret-store fixtures
+        if: ${{ !cancelled() }}
+        run: cargo test -p octos-cli --features api --lib auth
+
+      - name: Test API credential relocation fixtures
+        if: ${{ !cancelled() }}
+        run: cargo test -p octos-cli --features api --lib relocate
+
       # Sharded per-crate tests — same rationale as the `check` job.
       #
       # Every test step runs even after an earlier one fails. GitHub aborts a
--- a/crates/octos-cli/src/api/admin.rs
+++ b/crates/octos-cli/src/api/admin.rs
@@ -5536,6 +5536,9 @@ mod tests {
 
     #[test]
     fn relocate_keychain_backed_secrets_never_persists_raw_vertex_json_off_macos() {
+        let secrets = tempfile::tempdir().unwrap();
+        let _secrets_root =
+            crate::auth::keychain::test_override_secrets_root(secrets.path().to_path_buf());
         // #2234/45a — the availability predicate is now `keychain::is_available()`
         // (true on Linux: the file backend exists), NOT `cfg!(macos)`. The
         // never-plaintext contract holds where NO backend exists (unsupported
@@ -5585,8 +5588,9 @@ mod tests {
         // INJECTED temp root) the JSON is legitimately relocated: Ok, the
         // slot becomes a keychain marker, the raw value never remains.
         // Hosts with NO backend keep the rejection.
+        let secrets = tempfile::tempdir().unwrap();
         let _secrets_root =
-            crate::auth::keychain::test_override_secrets_root(tempfile::tempdir().unwrap().keep());
+            crate::auth::keychain::test_override_secrets_root(secrets.path().to_path_buf());
         let mut env = std::collections::HashMap::new();
         env.insert(
             "VERTEX_API_KEY".to_string(),
--- a/crates/octos-cli/src/api/auth_handlers.rs
+++ b/crates/octos-cli/src/api/auth_handlers.rs
@@ -4573,20 +4573,43 @@ mod tests {
         std::fs::create_dir_all(&clone_dir).unwrap();
         let clone_path = clone_dir.join("mine.wav");
         std::fs::write(&clone_path, b"fake").unwrap();
-        let json = format!(
-            r#"{{
-              "default_voice": "doubao",
-              "models_base_path": "{base}",
-              "voices": {{
-                "doubao": {{ "ref_audio": "ref_audios/doubao_ref.wav", "ref_text": "x", "aliases": ["vivian"] }},
-                "ghost":  {{ "ref_audio": "ref_audios/ghost_ref.wav", "ref_text": "y", "aliases": [] }},
-                "other-clone": {{ "ref_audio": "{clone}", "ref_text": "z", "aliases": [] }}
-              }}
-            }}"#,
-            base = dir.to_string_lossy(),
-            clone = clone_path.to_string_lossy()
-        );
+        let json = readiness_registry_json(&dir.to_string_lossy(), &clone_path.to_string_lossy());
         octos_llm::ominix::VoicesRegistry::parse(&json).unwrap()
+    }
+
+    fn readiness_registry_json(base: &str, clone: &str) -> String {
+        serde_json::json!({
+            "default_voice": "doubao",
+            "models_base_path": base,
+            "voices": {
+                "doubao": { "ref_audio": "ref_audios/doubao_ref.wav", "ref_text": "x", "aliases": ["vivian"] },
+                "ghost": { "ref_audio": "ref_audios/ghost_ref.wav", "ref_text": "y", "aliases": [] },
+                "other-clone": { "ref_audio": clone, "ref_text": "z", "aliases": [] }
+            }
+        })
+        .to_string()
+    }
+
+    #[test]
+    fn should_preserve_paths_when_serializing_readiness_registry_fixture() {
+        // These are serialization inputs, not directories to create: Windows
+        // paths and JSON-special characters are exercised on every test host.
+        for (base, clone) in [
+            (r"C:\Users\ci\models", r"C:\Users\ci\profiles\mine.wav"),
+            (r"\\server\voices\models", r"\\server\voices\mine.wav"),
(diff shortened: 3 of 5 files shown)


anantheparty (dismissed):
Adversarial pre-merge review passed — no blocker. The five #2239-review findings are all real and all correctly fixed here: (1) WrapErr import (superseded by #2260 — drop this PR's hunk on rebase, keep main's annotated version); (2) the rollback-destroys-existing-credentials hazard is properly fixed with snapshot-and-restore + shared-key preservation + honest rollback-failure reporting (4 tests); (3) relocation test hermeticity restored with injected roots and TempDir guards; (4) the cross-platform test_store with thread-local nesting and Windows hex-encoded names is strictly better than #2264's no-op stub — #2264 should be closed as superseded; (5) the Linux backend hardening (openat/O_NOFOLLOW/EXCL/fchmod-before-write/fsync/renameat + fstat non-regular + symlink-root rejection + umask-777 subprocess regression) is complete and preserves the #2239 contract.

Rebase recipe (merge-tree verified, 3 files 4 hunks, all same-point): keep main's keychain.rs:19-25 (#2260 version); take the PR side on the admin.rs/auth_handlers.rs cfg-gate hunks (main's linux-only gates defeat the purpose); ci.yml auto-merges. Also: squash 6aae774d (the voice-readiness test commit) to your noreply identity, and update the body's merge-coordination paragraph (#2260 landed; #2264 superseded). Minor leftovers for a follow-up: unlock help-text wording, is_accessible() side-effect mkdir, delete_secret dir-fsync, get_secret empty-file check. Approving — rebase and merge.

anantheparty (approved):
Re-acknowledging after the main merge: the conflict resolution in the auth test-store (adopting main's Linux-gated `test_override_secrets_root(.keep())` form for the off-macos relocation tests) is orthogonal to the five #2239-review fixes my earlier adversarial review verified — those all stand. Checks green on the updated head. Good to merge from my side.

BH3GEI:
Heads-up from this PR's check-windows run (job 103802523153): the only remaining failure is `api::auth_handlers::tests::my_profile_rejects_service_account_json_under_custom_env_off_macos`.

Verified against main: that test's store-root injection is `#[cfg(target_os = "linux")]`, so on Windows — no backend, no injection — the relocation fails before the profile save, and the `VERTEX_API_KEY` slot the assertions read never exists. The two new `auth`/`relocate` Windows steps added here also match it by name filter, so the lane stays red even after the voice fixture fix.

I've opened #2332 for main's broader check-windows break (the other 23 deterministic failures); it gates this one test to linux for now — the Windows reject-path coverage is preserved by the `admin.rs` twin. Once the portable test store in this PR lands, the gate can be relaxed and the injection enabled on Windows too.

The `serde_json::json!` fixture rewrite is exactly the right fix for the 3 voice-readiness failures — nice catch.

BH3GEI (approved):
Thanks for the thorough follow-up — the snapshot-and-restore semantics and the fd-anchored store rewrite both check out.

Verified on this side:

- Full diff review plus two independent adversarial passes (filesystem security / rollback state machine): no blockers. The rollback matrix (shared vs scoped × prior secret present vs not × first vs later profile failure × restore ok vs failed) behaves correctly and reports honestly on every path; the early return that retains a shared secret once an earlier profile committed a marker is the right call and is pinned by tests.
- Merged current main (including #2332) into this head locally: clean merge, net diff unchanged (4 files). The linux gate on `my_profile_rejects_service_account_json_under_custom_env_off_macos` clears the last check-windows failure — in the head's run that lane's auth step was 207 passed / 1 failed with that test as the only failure, and relocate was 7/7, so a rebase should go green.
- Local macOS run: `cargo fmt --all -- --check` clean; `cargo test -p octos-cli --features api --lib auth` → 226 passed / 0 failed / 3 ignored; `--lib relocate` → 7 passed; the keychain hardening suite passes (symlink/hardlink/fifo escapes refused, temp-file cleanup, restrictive-umask child regression).
- Load-bearing check: temporarily reintroducing the old `format!` fixture builder, and separately removing the retained-early-return, each make the corresponding new test fail — the coverage is real, not incidental.
- clippy note: the only `-D warnings` hit locally is `result_large_err` at handlers.rs:2024, which comes from main's #2296 (this PR doesn't touch that file) — pre-existing.

Two non-blocking notes for a possible follow-up, both fine to land as-is:

1. In `open_root`'s create path, a umask that strips owner-read (e.g. 0400) on a brand-new root makes `mkdirat` yield 0300, and the subsequent RDONLY `openat` hits EACCES before `fchmod` can restore 0700. The old path-based chmod survived umask 0400, so this is the one spot where the hardening ends up slightly stricter than before. As the body notes, ancestor initialization is out of scope here — just flagging it for a follow-up (an `fchmodat` right after a fresh `mkdirat` would close it).
2. In `set_secret`'s failure path, the `let _ = unlinkat(...)` swallows cleanup errors, so a leftover `.secret-<uuid>.tmp` holding the new secret could go unmentioned if the unlink itself fails. Folding that into the returned error would be a nice touch.

Approving — thanks again for the careful work on this.

BH3GEI:
Main's check-windows is now red on exactly the three voice fixture tests this PR fixes. Run 34931713823 (commit ab09eae5, both attempts): `empty_selection_falls_back_to_any_visible_synthesizable_voice`, `selected_voice_must_itself_be_synthesizable_not_just_any_voice`, and `selected_voice_owned_by_another_tenant_is_not_ready` all panic with `failed to parse voices.json: invalid escape at line 3 column 39` — the `readiness_registry` format!-built JSON embeds unescaped Windows backslash paths, which this PR's `readiness_registry_json` (serde_json::json!) resolves.

For what it's worth, this branch's own check-windows failure from 09-13 (`my_profile_rejects_service_account_json_under_custom_env_off_macos`) was gated to non-Windows on main by #2332, so an update-branch should turn this PR's CI green — and merging it would restore main's required lane.

Issue #618, opened 2026-04-27
Live regression: tool_progress SSE events not surfacing during plugin execution (no in-bubble progress timeline)


ymote:
## Symptom
While a plugin tool runs (e.g. \`news_fetch\` ~10s, \`deep_search\` ~5min, \`run_pipeline\` ~minutes), the chat UI shows **nothing** to indicate the tool is in flight:
- No tool-call bubble appears with progress phases
- No \`runtime-status timeline\` populates into the assistant bubble
- User sees a blank assistant placeholder, then suddenly the final answer (or, when overflow kicks in per #616, nothing at all)

The user reports this as \"the search wasn't kicked\" — even though server logs confirm the plugin spawned and completed.

## Test evidence
\`tests/live-tool-progress.spec.ts:56\` — \"progress lands in the original tool-call bubble\" — fails on mini1 (dspfac.crew.ominix.io) post-deploy d191c2c7:

\`\`\`
Error: expect(received).toBeGreaterThan(expected)
  Expected: > 0
  Received:   0
  Call Log: Timeout 60000ms exceeded while waiting on the predicate

  > 83 |     await expect
       |     ^
    84 |       .poll(() => page.locator(TOOL_CALL_BUBBLE_SELECTOR).count(), {
    85 |         timeout: 60_000,
    86 |         intervals: [2_000, 3_000, 5_000],
\`\`\`

Selector under test: \`[data-testid='tool-call-bubble']\` — never appears within 60s of sending a deep-research prompt.

## Real-user repro on mini3
Live demo session \`web-1777267327685-db0etz\` on mini3 (2026-04-27 ~05:23 UTC):
- User: \"今日头条新闻查一下\"
- Server log: \`spawning plugin process plugin=news tool=news_fetch pid=29770 timeout_secs=600 args_size=119\`
- Server log 10s later: \`plugin process completed exit_code=0 stdout_len=137081\`
- UI during those 10s: assistant bubble shows \"我来查一下今日头条新闻\" then nothing else — no progress indicator, no \"News fetching…\" status

Server log line that should be reaching the UI but isn't:
\`\`\`
INFO executing tool batch parallel_tools=1 tool_names=news_fetch dispatch=\"parallel\" profile=dspfac
INFO spawning plugin process plugin=news tool=news_fetch executable=… timeout_secs=600
INFO plugin process spawned plugin=news tool=news_fetch pid=29770
[...]

PR #2265, opened 2026-09-05
feat(oup): unified context/cache runtime and rc.11 preparation

ymote:
## Summary

Prepare Octos 2.0.3-rc.11 with the reviewed OUP semantic context/cache implementation and chat/ACP convergence. This also integrates runtime termination/partial-answer integrity, durable ledger recovery, provider cache diagnostics and receipt ownership fixes. Root README edits and the proposed implicit goal-budget removal are deliberately excluded.

The original dirty worktree was preserved. The changes were transplanted onto current main without importing the old local ancestry; upstream cache-aware pricing, sampler defaults, lazy recovery and stdio scope handling are retained. Integration-specific regressions cover sampler override protection and shared chat/OUP configuration.

## Validation

- Core implementation all-target tests: **8,354 passed**, 62 ignored, 94 suites (CLI, agent, core, LLM, bus, services).
- Strict workspace/all-target clippy: passed, including a final rerun after the sampler guard fix.
- LLM final regression run: **684 passed**, three ignored; reserved-key negative control failed before the fix.
- Additional minimal-feature, release-feature and ABI checks are running; required CI must pass before merge.
- Prior Fable review and real local/mini3 30-round + peer soak passed on the pre-integration candidate. **These are not claimed as proof of the new integrated binaries.** Final artifact validation will be tracked separately.
- Staged known mini-credential checks found no password/host matches. Gitleaks' eleven backend hits were manually inspected synthetic secret-redaction fixtures, not production credentials. No broad test-file exclusion was used.

Implementation and detailed acceptance: `docs/adr/oup-semantic-boundary-context-cache.md`.

## Release sequencing

Merge only with required checks satisfied, then publish tag `v2.0.3-rc.11` from the merged commit. The companion OctosCode RC will pin that exact backend revision. This PR does not change stable release channels.


anantheparty:
Review scope note (the diff exceeds the API diff limit at 36K+/132 files, so this is a partitioned adversarial review — four parallel passes: octos-llm (context/cache core), octos-agent (termination integrity + ledger recovery), octos-cli (chat/ACP convergence + receipts + auth interplay), and the periphery (bus/core/ffi/docs/specs + version). Results land as they complete.

Two questions while those run:

1. **Audit chain**: the body describes the implementation as "reviewed" — where do those reviews live? The #618 precedent (reviews claimed off-platform, trail not verifiable at the stated location) is why I'm asking up front: if the trail is on-platform or accessible, link it; otherwise these four partition reviews become the review of record.

2. **Red lines confirmed so far**: ci.yml untouched ✓ (the only manifest change is the rc.11 version bump). The falsified-approaches memo stays respected per the nightly job note.

@ymote #622 (the rc.10 counterpart) is draft — same partitioned treatment at flip.

ymote:
Additional local release checks passed: 1,682 minimal-feature tests; strict combined channel/release-feature clippy; strict minimal production-library clippy; FFI/UniFFI tests; C header declaration check; real localhost C/Python incomplete-answer and success controls; eight peer-recovery/stdio regressions. Isolated FFI compilation exposed an API-gating integration bug, now fixed with a required minimal-production-library CI gate. The latest follow-up is formatting-only. Preparing a fresh isolated mini3 run of the integrated candidate while CI runs; old cloud proof is not being relabeled as new.

anantheparty:
CI failure triage for the hosted run (test-octos-cli: 3361 passed / **2 failed** — both look like the same root cause, and your local minimal-feature pass wouldn't have covered them):

Both failures are in `api::ui_protocol_transport::tests` — `should_keep_spawn_only_sent_file_identity_and_one_hydrated_attachment` and `..._and_each_hydrated_attachment_once` — panicking at `ui_protocol_tests.rs:25195` (`.unwrap()` on the episode-store open):

```
failed to open redb database at /tmp/.tmpwkfcpV/episodes.redb:
Database already open. Cannot acquire lock. Another octos process
(typically an `octos serve` daemon) already owns this data directory.
  at crates/octos-memory/src/store.rs:344
```

**Root cause**: redb's single-owner file lock — the test spawns a process (daemon/serve) that opens the episode store first and holds the lock; the test body's own second open against the same temp data dir then fails. Not a logic bug in the attachment-identity code under test — it's test-harness isolation: the two handles (spawned process + test body) must not open the same `episodes.redb`.

**Fix direction**: either give the test body's store its own `--instance-data-dir`-equivalent (the error message's own remedy), or route the assertion through the spawned process's handle instead of opening a second store. Worth checking whether the paste-editing/mixed-projection dedup commits changed the store-open timing here (the tests passed in your local minimal-feature pass, so this is likely a full-suite concurrency exposure — the two tests may be racing each other's temp dirs under parallel test execution rather than with their own spawned process).

Everything else in the suite (3361) is green, and the FFI gating fix + new minimal-production-library CI gate from your local pass look right. Flagging the two failures so the mini3 run includes them.

ymote:
Fresh mini3 integration soak PASS: 30 main rounds + 4 peer rounds; all seven independent acceptance gates pass. 80/80 cache manifest/usage pairs, 76 nonzero cache reads, 1,556,224 cached tokens, six privacy checks pass. Credentials cleaned up and existing remote profile unchanged; exported evidence also passes known-provider-secret scan. Ubuntu CLI CI exposed two test cold-reopen races: asynchronous cache Drop had not yet released the old DB. A weak-reference assertion reproduced both locally; the fixture now waits for actual store release with a bounded failure deadline before cold opening. Targeted tests and strict clippy pass; new CI run required. Separately, classic main protection requires one non-author APPROVED review before merge.

anantheparty:
Progress confirmed: your Drop-release fixture fix worked — both attachment-identity tests pass now (3362 passed / 1 failed on this run).

The remaining failure is **different and looks like a real bug, not harness flake**: `should_delete_primary_promote_fallback_and_reload_dynamic_profile_runtime` (ui_protocol_tests.rs:2157) —

```
assertion `left != right` failed: primary deletion must rebuild the cached ProfileRuntime
  left: 140118543649920
 right: 140118543649920
```

The cached ProfileRuntime pointer is **identical before and after the primary deletion** — the deletion-promote-reload sequence did not rebuild the runtime. That is exactly the #2184-family invalidation semantics (config-changing operations must invalidate the cached runtime so the running gateway doesn't keep serving via stale provider state); a stale runtime after a primary deletion would keep the old provider wired in until process restart.

Suspects worth a look: the deletion path returning before the cache-invalidations step, the rebuild keyed on something that didn't change (profile id only, not content/generation), or the new context/cache runtime holding a second reference that satisfies the equality. The test is doing its job — the fix belongs in the invalidation path, not the assertion.

ymote:
The final CI rerun passed both cold-media reload regressions, then exposed a different pre-existing fixture bug: runtime reload compared a freed Arc address against a new allocation, so allocator address reuse caused a false failure. Both reload tests now retain Weak allocation identity (without retaining the single-writer DB). Targeted tests, strict all-target clippy and fmt pass; production code is unchanged. Operator has explicitly authorized admin merge once the remaining validation is resolved.

anantheparty:
Correction accepted — my "stale runtime" read was wrong: comparing a freed Arc's address against a new allocation is unsound (allocator reuse makes the identity check fail spuriously), and the Weak-allocation-identity pattern (without retaining the single-writer DB) is the right fixture fix. Production code unchanged confirms the rebuild semantics were correct — good catch on the fixture, and the #2184-family invalidation concern is retracted.

Merge path acknowledged: partitioned review (4 areas) is in flight; once CI settles green and the partitions report, this merges with the authorized admin path.

ymote:
Merged with explicit operator authorization for admin squash merge after the entire latest CI run (34000572993, cfecf7d0) completed successfully, including macOS security sandbox and tool end-to-end checks. The merged tree exactly matches the validated PR head. Release tag v2.0.3-rc.11 targets d51601dff70cb27318ac7b3bd95e1c018132b756; release workflow 34002062418 is running. The paired client PR remains gated on its exact backend pin, final checks, and publication.

PR #2272, opened 2026-09-09
fix(peer-status): report current execution and outcomes from trusted evidence

ZhangHanDong:
Peer 状态以前可能把上一轮的完成结果当成当前状态，也难以区分“任务还存在”“正在执行”和“最终成功/失败”。这次将三者分开显示，并核对当前轮次和所属身份；旧结果、外来结果或缺少可靠证据时显示 `unknown`。

- 统一 CLI 表格、JSON 和 UI 协议的状态来源，使用只读 lifetime 投影；显式 profile/data-dir 路由保持一致。
- 保留真实的 failed、errored、interrupted、rate_limited 结果，避免把终止状态一律显示成成功。
- 行为合约：`specs/task-evo-peer-turn-status.spec.md`（24 个场景绑定真实测试）；接口定义：`docs/peer-status-interface.json`。
- 明确边界：closed 分支的身份字段为 null；begin 后崩溃留下的 Running 投影没有自动降级，可能持续显示 running。契约已按真实 terminal 写入顺序说明崩溃窗口及 last_outcome 可能为旧值或 null 的情况。

验证：运行实现 `49c84eaa` 已经 GLM-5.3 / K3-256K 互审和 Codex 独立外层审查，包含实际 CLI、路由及渲染验证。与 goal verifier 修复 `59972429` 组合后的本地验证提交 `8b4da851` 通过：

```sh
cargo fmt --all -- --check
cargo clippy --all-targets --features api -- -D warnings
cargo test --all-targets --features api --no-fail-fast -- --test-threads=8
```

全量集成验证：172 套件，10,044 通过、0 失败、100 忽略。该数字对应两项修复的组合树。运行实现 `49c84eaa` 的 GitHub CI 也已通过。

当前 HEAD `e8f77422ab250d30a198ef600577229af9cb0162` 根据审查修正了注释和契约文字，未改运行行为或测试。该增量通过 GLM/K3 最终核对、Codex 文档复核、fmt、JSON 解析和 agent-spec parse/lint；字段集合及 24 个场景绑定未变。保留 closed 身份的行为建议仍属后续改进，本次没有实现该行为变更。

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,11 @@ test_a2.md
 !specs/task-approval-post-tool-continuation.spec.md
 !specs/kv-cache-friendly-compaction.spec.md
 !specs/task-compaction-instruction-priority.spec.md
+# peer-status SDD contract (task-evo-peer-turn-status) + consumer field
+# contract — consumed by tests and outer-loop verification; plan under
+# docs/superpowers stays local by convention.
+!specs/task-evo-peer-turn-status.spec.md
+!docs/peer-status-interface.json
 # User-facing docs only (internal analysis/audits stay local)
 !docs/user-guide.md
 !docs/user-guide-zh.md
--- a/.gitignore
+++ b/.gitignore
@@ -78,11 +78,6 @@ test_a2.md
 !specs/task-approval-post-tool-continuation.spec.md
 !specs/kv-cache-friendly-compaction.spec.md
 !specs/task-compaction-instruction-priority.spec.md
-# peer-status SDD contract (task-evo-peer-turn-status) + consumer field
-# contract — consumed by tests and outer-loop verification; plan under
-# docs/superpowers stays local by convention.
-!specs/task-evo-peer-turn-status.spec.md
-!docs/peer-status-interface.json
 # User-facing docs only (internal analysis/audits stay local)
 !docs/user-guide.md
 !docs/user-guide-zh.md
--- a/crates/octos-cli/src/api/ui_protocol_tests.rs
+++ b/crates/octos-cli/src/api/ui_protocol_tests.rs
@@ -34517,7 +34517,7 @@ fn peer_gather_callback_composes_done_and_running_rows() {
     std::fs::write(gamma.join("result.md"), "partial result before close\n").unwrap();
     std::fs::write(gamma.join("closed"), "tenant-a:api:master\n1700000000\n").unwrap();
 
-    let callback = build_peer_gather_callback(peers_root.clone());
+    let callback = build_peer_gather_callback(peers_root.clone(), "octos".to_owned());
     let text = callback(None).expect("gather composes");
     assert!(text.contains("## peer alpha (done)"), "{text}");
     assert!(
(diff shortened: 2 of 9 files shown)


anantheparty:
Adversarial pre-merge review passed — **0 blocker, 7 nits, mergeable.** Verified by reading the full diff AND compiling/running the tests on the PR ref (244 passed / 0 failed on the peer/outer_peer suites).

## Clean (the load-bearing parts, all verified)

- **Trust chain**: four fail-closed layers — typed lifetime deserialization, exact registry-key binding (`peer_wire_key(profile, slug)` — cross-runtime/cross-profile same-slug isolation is real, tested both directions), writer-shape turn binding (Pending's None+Some accepted per the real writers' shape — two verbatim ported probes guard against the reverse bug), and Idle's SHA256 digest binding to the current result.md bytes.
- **The old bug is genuinely fixed**: the old code derived status from file existence (round1-completed + round2-queued displayed as finished); the new current-execution authority is the closed marker + the trusted projection, with regression tests binding exactly that.
- **Real failures preserved**: the outcome whitelist keeps failed/errored/interrupted/rate_limited as-is (tests for done+failed, closed+errored, rate_limited).
- **Single source**: one reader feeds CLI/serve-JSON/UI-protocol (all four production call sites pass real profile ids); profile routing tested through real clap parsing.
- **SDD**: 24/24 spec `Filter:` entries resolve to real passing tests; the interface contract is two-way locked to the serialized key set; 4 extra tests beyond spec.
- Read-only claim holds on every path; the writers are untouched.

## Two one-line contract-wording fixes requested before merge (SDD hygiene)

1. **Closed rows return identity=None even when the lifetime is trusted** (peers/mod.rs:4326-4336) — but `PeerListRow.master_session_id`'s doc and interface.json's identity_fields_note both say "null when the projection is untrusted". The behavior is conservatively defensible; the documented CONTRACT is wrong. One clarifying line in both docs.
2. **The begin→crash Running window is not in known_limitations** — cross-restart adjudication handles closed→retire and Idle→rebind, but a crash orphaned Running has no degradation path and reads as running indefinitely (conservative direction, writer semantics declared out-of-scope — so list it). Also the crash_window note's "may present running/queued" wording is imprecise (post-finish can only be idle/queued/failed).

## Follow-up nits (non-gating)

spec:132 says generation=2 vs fixture generation=1 (assertion semantics identical); interface.json idle_no_turns round=0 unreachable; interface.json cli_command missing --profile; version tag drift (json "version": 2 vs code "v3.1 frozen contract"); the serve peer_list spec wording vs the (correct) implementation superset; the ~3-4 extra fd reads per peer per list on the serve hot path (negligible, noted).

**Land the two one-liners and this merges.**

anantheparty (approved):
Final delta review — the two one-line contract fixes landed: the closed-row identity=None doc clarification now covers the closed case explicitly, and the crash-orphaned Running window is documented in known_limitations. The trust-chain/three-source/failure-preservation/SDD verdicts stand. Approving — merging left to you per convention.

PR #2268, opened 2026-09-06
feat(build-cache): per-peer target slot pool with sandbox grants, space gate, GC and outer-loop verify slots

ZhangHanDong:
> 分支 `feat/build-cache-pool`,基线 main 9c157101,已合并 origin/main 25ff2734(tip 51641c3c,已与合入 #2266 后的 main 同步,并含维护者 ymote 的 313e23d6:acquire 先写 last_used 再发布 holder,失败不遗留活 claim);两轮审查(外部 anantheparty + gpt-6-astra)的 6 Blocker / 6 Should-fix / 4 Nit 已在 `#10-*` 提交中修复。

## 问题

多个 peer(以及外环的验证构建)在同一台机器上并发 `cargo build/test` 时:
- 各自的 clone 带各自的 `target/`,同一仓库的依赖被重复编译数十次,一次磁盘清理前 43 个 target 目录占了 300 GB 以上;
- 若改为共享 workspace 的 `target/`(#2236 的做法),两个 peer 同时写同一 target 会有锁竞争与半成品产物风险,且 workspace target 不受任何回收管理;
- 磁盘写满会直接导致 herdr / 网关崩溃。

## 方案

- **槽池**:`<data_dir>/build-cache/<repo-key>/slot-N`(repo-key = 主仓库 canonical 路径的 sha256 前 12 hex),默认 2 个 peer 槽 + 1 个 verify 槽,`[build_cache]` 配置段(`peer_slots`、`verify_slots`、`min_free_gb`、`stale_hours`)。分配用 `flock(EX|NB)` + 原子写 `holder.json`;释放幂等;`reclaim_stale` 按 pid 判活(ESRCH=死 / EPERM=活 / 其他保守视为活),绝不按 mtime,绝不删 `.lock`。
- **peer 接入**:`stage_peer` 首轮 acquire 并记录到 `peers/<slug>/build-cache`;serve boot 每轮 take/adopt;主释放在 per-turn 终态,中断 / close / fleet 回滚 / 早期 Errored 为幂等兜底。工具调用按 call 注入 `CARGO_TARGET_DIR=<slot>/target` 与 `CARGO_INCREMENTAL=0`(`ToolContext.build_cache_slot`,无 `set_var`)。
- **沙箱**:macOS SBPL 只放行 peer 自己槽的 read/write subpath,元字符 fail-closed;其他槽不可写(真实内核测试验证)。bwrap/docker 暂为已知降级。
[...]

PR #2274, opened 2026-09-09
test: gate Unix-only tests to fix the check-windows build break on main

BH3GEI:
## Problem

`check-windows` is red on main (run 34397691827 on 946b4e49, also failing since 09-06). Two distinct breaks, all in test code:

1. **octos-cli test targets fail to compile on Windows** —
   - `peers/mod.rs`: `workspace_scope_encoding_is_lossless_and_distinct` uses `std::os::unix::ffi::OsStrExt` / `OsStr::from_bytes` (E0433/E0599); `peer_list_symlinked_turns_index_is_ignored` uses `std::os::unix::fs::symlink` (E0433).
   - `build_cache/pool.rs`: `report_only_preserves_dead_holder_and_all_slot_files`, `dead_holder_slot_is_reclaimable`, `unreadable_holder_json_is_skipped_not_stale` call `spawn_dead_pid()`, which is itself `#[cfg(unix)]` (E0425 ×3).
2. **octos-agent Windows runtime failure** — `sandbox::docker::tests::build_cache_target_is_mounted_and_env_reaches_container` panics at docker.rs:192: on Windows `std::fs::canonicalize` returns a `\\?\C:` verbatim path whose drive colon trips the `-v` mount-injection guard in `wrap_with_slot`, so the command comes back as the refusal stub without the asserted `--env` args.

## Root cause

- The pool tests exercise dead-holder reclamation, which hinges on `pid_alive`'s `kill(pid, 0)` semantics. On non-Unix hosts `pid_alive` is a conservative always-alive stub (pool.rs:506-511), so the behavior under test does not exist on Windows by design — these tests can never be meaningful there, only fail to compile.
- The peers tests construct non-UTF-8 `OsStr`s and symlinks via Unix-only extension traits.
- The docker test asserts Unix path semantics; the Docker backend is structurally Unix-shaped in this code path (the workspace cwd mount guard also refuses drive-letter paths), so the mount command it expects cannot be produced on Windows. Making the mount work on Windows (dunce + drive-letter-aware parsing) would be a feature, not a build-break fix; ci.yml's own note on this job says to prefer "fixing it or narrowing the failing test to `#[cfg(unix)]`".

## Fix

Add `#[cfg(unix)]` (plus a one-line rationale comment, matching the file's existing gated tests such as `pid_alive_distinguishes_eperm_from_esrch`) to the five test functions. No production code touched; 14 insertions across 3 files.

## Test evidence

- **Windows compile (the broken path)**: `cargo check --workspace --all-targets --target x86_64-pc-windows-gnu` on this branch — zero errors (covers every lib-test and integration-test target of octos-cli/octos-agent under `cfg(windows)`; `windows-gnu` resolves E0433/E0599/E0425 name-resolution identically to CI's `windows-msvc`). Also clean with octos-cli's full CI feature set (`api,telegram,discord,dingtalk,whatsapp,feishu,twilio,wecom,wecom-bot,audio_mp3`).
- **Unix regression**: the five gated tests still exist and pass on Unix (5/5), plus `sandbox::docker` tests 15/15.
- **fmt/clippy**: `cargo fmt --all -- --check` clean; `cargo clippy -p octos-cli -p octos-agent --all-targets -- -D warnings` clean.
- Three local `peers::build_cache_peer_tests` failures observed during the run are environmental (the machine had 40 GB free, below the 50 GB space gate added in #2268) and unrelated to this diff.
- Independent review pass confirmed each CI error maps to exactly one gate, that the remaining `spawn_dead_pid` call sites were already gated on main, and that no other ungated Unix-only API use exists in these crates' test code.

## End-to-end note

The failing environment is Windows CI itself; `check-windows` runs its test steps only on main/manual, so the post-merge main run is the final confirmation. The pre-merge evidence above reproduces the exact failing compile surface (`cfg(windows)` test targets, all targets) locally and shows it clean.

Refs #2267

### DIFF
--- a/crates/octos-agent/src/sandbox/docker.rs
+++ b/crates/octos-agent/src/sandbox/docker.rs
@@ -171,7 +171,11 @@ fn cache_mount_refusal() -> Command {
 mod tests {
     use super::*;
 
+    // Unix-only: on Windows `std::fs::canonicalize` returns a `\\?\C:`
+    // verbatim path whose drive colon trips the `-v` injection guard, and
+    // the Docker backend refuses drive-letter cwd mounts there anyway.
     #[test]
+    #[cfg(unix)]
     fn build_cache_target_is_mounted_and_env_reaches_container() {
         let sb = DockerSandbox {
             config: DockerConfig::default(),
--- a/crates/octos-cli/src/build_cache/pool.rs
+++ b/crates/octos-cli/src/build_cache/pool.rs
@@ -1972,7 +1972,10 @@ mod tests {
         assert_eq!(pid_alive(i32::MAX as u32 + 1), Some(false));
     }
 
+    // Dead-holder reclamation hinges on `pid_alive`'s kill(pid, 0)
+    // semantics, which only exist on Unix; `spawn_dead_pid` is cfg(unix).
     #[test]
+    #[cfg(unix)]
     fn report_only_preserves_dead_holder_and_all_slot_files() {
         let tmp = tempfile::tempdir().unwrap();
         let root = tmp.path().join("pool");
@@ -2165,6 +2168,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(unix)]
     fn dead_holder_slot_is_reclaimable() {
         let tmp = tempfile::tempdir().unwrap();
         let root = tmp.path().join("pool");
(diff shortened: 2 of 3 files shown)


anantheparty (approved):
Adversarial pre-merge review passed — gates the two Unix-only test compilation breaks (peers/mod.rs OsStrExt + the auth test family) behind cfg so check-windows compiles again. This is the #2259 blind-spot family's fourth escape (macOS WrapErr, test helper, deep-search fixtures, now Windows test compilation) — each fix is correct, but the aggregate case for the compile-only darwin/windows gates is now airtight. 14 lines, test-only, CI green. Approving — merging immediately to unbreak main's windows lane.

PR #2154, opened 2026-08-27
OctoLoop: dual-loop reliability hardening + model-adaptation layer (16-branch integration)

ZhangHanDong:
## OctoLoop — reliability & dual-loop hardening (16-branch integration)

This PR integrates the outer-loop-verified battle chain from the dual-loop (OLP) operation. Every branch was independently re-verified in an isolated worktree by the outer loop (Claude), then adversarially cross-reviewed by a second outer agent (Codex) — two BLOCKERs it raised are fixed and re-verified in this tree.

### Reliability fixes (peer/goal runtime)
- **Provider quota auto-degrade** (#27b): quota 402/429 reclassified `Quota → SwitchProvider → RotateAndRetry` — the ProviderChain now actually advances to `llm.fallbacks` on quota exhaustion. 401 auth keeps its pinned FailFast contract.
- **Orphan re-attach** (#27c): peer_handoff children orphaned by a serve restart become recoverable `Parked` (production-verified during rotation) instead of terminal `failed`.
- **Malformed tool-call self-correction** (#27d): diagnostics fed back to the model (3/turn) instead of instant turn death; stream-layer non-retryable invariant untouched.
- **Budget-exhaustion checkpoint** (#27e + #27h/#27h-r1): dirty worktree auto-checkpointed with a staged result; respects the peer's `.result-owner` sidecar (single shared owner predicate, cross-pinned contract tests in both crates).
- **result.md single-writer** (#27f): ownership sidecar prevents the runtime summary from clobbering a peer's authoritative result.
- **Goal exits** (#25/#26a/#26a-r1): `octos goal reopen/archive` CLI, tolerant event replay, goal-scoped view with the #25 cross-session ambiguity contract restored (composite fold key).
- **Turn continuation** (#26b): zero-delay turn chaining for active goals.
- **Serve shutdown** (#21): real broken-pipe cleanup selector (adjudicated and pushed by the Codex outer lane).

### Model-adaptation layer (#28 series)
- **File-change receipts** (28a/28c/28c-r1): every bash/exec tool result carries `files_changed: N` + scope tag, on both the agent lane (ShellTool) and the interactive coding lane; snapshot root resolves a literal leading `cd` — kills silent "phantom edits" by cheap coding models.
- **Per-family write policy** (28b/28d): `tool_policy.bash_file_writes = allow|warn|deny` (default `allow` = zero behavior change), wired into both lanes, with an explicit `# octos:allow-write` escape hatch.

### Also included
[...]

PR #2275, opened 2026-09-10
test: name serve_broken_pipe e2e module after its target so the CI skip works

BH3GEI:
## Problem

The `test-octos-cli` job has a broad integration step named "Test octos-cli (integration, excluding serve process e2e)" that runs `cargo test -p octos-cli --tests -- --skip serve_broken_pipe`, with a dedicated serial step right below for the process-spawning BrokenPipe e2e. The exclusion is silently broken: the four process e2e run in the fully-parallel broad step anyway and flake there — the 60s cleanup-marker poll in `imp::serve_shutdown_broken_pipe_cleanup_marker_observed` timed out on two unrelated PR CI runs ([33997597638](https://github.com/octos-org/octos/actions/runs/33997597638) on Sep 5, [34415043855](https://github.com/octos-org/octos/actions/runs/34415043855) on Sep 9), failing both PRs' required checks.

## Root cause

`--skip` matches **test names**, not the target name. The tests live in `mod imp` (added in cd65eb68 / #2154), so their names — `imp::serve_shutdown_broken_pipe_cleanup_marker_observed`, `imp::serve_shutdown_order_preserved`, `imp::serve_startup_broken_pipe_no_panic`, `imp::subprocess_panic_stderr_broken_pipe_no_abort` — have never contained the substring `serve_broken_pipe`. The filter has matched nothing since introduction: `cargo test -p octos-cli --tests -- --skip serve_broken_pipe --list` still lists 3843 of 3843 tests.

## Fix

Rename the module `imp` → `serve_broken_pipe`. Every test name now carries the target name, so the existing CI filter excludes exactly these 4 tests — and any future test added to the file. One-word change plus an explanatory comment; no workflow edit. The `#[cfg(unix)]` gating, the SERIAL mutex, and every test body are untouched.

## Test evidence

- Broad step filter behavior, before → after (same binary, `--list`): 3843 → 3839 tests; the only four excluded are the renamed e2e, zero leakage. The two unrelated lib unit tests with `broken_pipe` in their names (`serve_console_write_line_broken_pipe_returns_ok`, `write_panic_report_broken_pipe_no_panic`) still run.
- Dedicated serial step unchanged and green for real: `cargo test -p octos-cli --features api --test serve_broken_pipe -- --test-threads=1` → 4 passed, 0 failed (64.97s).
- The spec selectors in `specs/task-s-broken-pipe-shutdown.spec.md` use bare names, which remain substring-matched under the new module prefix.
- End-to-end note: this is a CI/test-selection change, so the "product" is the test list itself — the before/after `--list` diff above is the end-to-end demonstration. The full parallel integration suite itself is unchanged in content (only these four tests move out of it, back to their intended lane).

## Review

Two-round review: self-review of the full diff, then an independent adversarial review (twice — once for an earlier ci.yml four-substring variant, which it approved with wording notes I adopted, and once for this final module-rename variant). The reviewer verified repo-wide that nothing references the old `imp::…` names, that no tooling parses these test names, that Windows compilation shape is unchanged, and that the rename closes the future-addition leak. Verdict: approve.

No tracking issue — filed directly as a CI build-break-class fix after the flake failed required checks on two unrelated PRs.

### DIFF
--- a/crates/octos-cli/tests/serve_broken_pipe.rs
+++ b/crates/octos-cli/tests/serve_broken_pipe.rs
@@ -15,14 +15,20 @@
 //! `cargo test -p octos-cli --features api --test serve_broken_pipe -- --test-threads=1`
 
 // Process-control tests require libc pipe/kill/setsid — unsafe is inherent.
-// #39 — the whole `imp` module is unix-only BY ORIGINAL DESIGN (libc pipes,
+// #39 — the test module is unix-only BY ORIGINAL DESIGN (libc pipes,
 // raw-fd Stdio, SIGINT/SIGKILL): the #37 hardening broke the cfg structure
 // by leaving the imports and several `libc::` call sites outside any gate
 // (E0432/E0433 on Windows). Gate the entire module so the file compiles on
-// Windows with the SAME existence shape as before #37: no tests, no imp.
+// Windows with the SAME existence shape as before #37: no tests, no module.
+// The module is named `serve_broken_pipe` so every test name carries the
+// target name: the broad CI integration step excludes these
+// process-spawning e2e with `--skip serve_broken_pipe` (they run in the
+// dedicated serial step below it), and --skip matches test names, not
+// target names — under the old `imp` name the filter silently matched
+// nothing and the e2e leaked into the parallel step, flaking under load.
 #[cfg(unix)]
 #[allow(unsafe_code)]
-mod imp {
+mod serve_broken_pipe {
     use std::os::fd::FromRawFd;
     use std::process::{Command, Stdio};
     use std::sync::Mutex;


anantheparty (approved):
Test-infra hygiene: the e2e module was renamed at some point, so the CI skip pattern (--skip serve_broken_pipe) stopped matching and the known-flaky family leaked back into the main integration step. Renaming the module after its target restores the skip contract. 9 lines, test-only, CI green. Approving — merging.

anantheparty (approved):
Test-infra hygiene verified on the synced head: the renamed e2e module restores the CI skip contract (--skip serve_broken_pipe), keeping the known-flaky family out of the main integration step. Test-only, CI green. Approving - merging.

PR #2278, opened 2026-09-10
fix(cli): relocate the goal operator-control socket when the data dir exceeds sun_path (#2248)

### LINKED ISSUE #2248: serve: startup fails with `path must be shorter than SUN_LEN` on long `--data-dir` paths
`octos serve` creates `<data_dir>/.octos-goal-control.sock`; when the data dir path is long (~100+ bytes) startup fails with `path must be shorter than SUN_LEN` from `commands/goal.rs`. The error does not say which path or how to fix it.

Suggested: fall back to a socket in `$TMPDIR` (or `XDG_RUNTIME_DIR`) when the data-dir path is too long, and name the path in the error.

---

BH3GEI:
## Problem

`octos serve` binds its goal operator-control RPC at `<data_dir>/.octos-goal-control.sock`. Unix socket paths are bounded by the fixed-size `sun_path` array (104 bytes on macOS/BSD, 108 on Linux), so a data-dir path of ~80+ bytes made the bind fail with `path must be shorter than SUN_LEN` — and since `spawn_goal_operator_control` is fatal to startup (serve.rs), the entire serve refused to start. The error named neither the path nor a remedy.

## Root cause

The socket path is `data_dir.join(".octos-goal-control.sock")` with no length handling; `sockaddr_un.sun_path` cannot hold the result, so Rust's std returns `InvalidInput` before the syscall.

## Fix

- New `goal_control_bind_path`: canonicalize the data dir (so a short-symlink spelling cannot split the two sides' decisions), use the canonical in-data-dir socket when it fits, else relocate to a deterministic FNV-1a-hashed name under `` — falling back to the temp dir, nested in a per-user 0700 subdir when the temp dir is shared (1777), so no other local user can pre-create the predictable name and block startup. A relocated path that still exceeds `sun_path` (pathological `TMPDIR`) is a hard error naming the path and the fix.
- The serve side publishes the real socket path in a `.octos-goal-control.sock.path` pointer file inside the data dir (atomic write+rename, 0600; stale pointer removed when canonical). The connect side (`octos goal reopen/archive`) reads the pointer and otherwise uses the canonical path — it needs no environment knowledge, so a systemd-launched serve and an interactive shell resolve the same endpoint even with different `XDG_RUNTIME_DIR`/`TMPDIR`. Pointer absent + serve lock held = old serve, which the connect side already fails closed against.
- Bind failure now names the socket path; relocation is logged at info.

## Test evidence

- 10 new tests in `tests_2248_sun_len` (all `#[cfg(unix)]`): canonical location for short dirs, relocation determinism and per-dir uniqueness, threshold-straddling symlink spelling resolving to the same endpoint, a bind→publish→connect round trip through the pointer file, connect defaulting to canonical without a pointer, stale-pointer removal on canonical publish, fallback-base selection table (absolute/relative XDG, shared vs per-user temp), the overlong-fallback named error, and a golden FNV vector pinning the cross-release hash contract.
- `cargo test -p octos-cli --features api --lib commands::goal` — 22/22 pass. Full `--lib` suite: 3683 pass; the only 3 failures (`peers::build_cache_peer_tests` flock slot tests) reproduce identically on unmodified origin/main (c8e6ba15) on this machine and are green in CI — pre-existing local-environment issue, not this diff.
- `cargo clippy -p octos-cli --features api --lib --tests` clean; `cargo fmt --check` clean.

## End-to-end

Built `octos` from origin/main and from this branch (`--features api`), ran `serve --data-dir <120-byte path> --port 0`:

- **Before**: startup aborts with `failed to start local goal operator-control RPC: path must be shorter than SUN_LEN` (goal.rs:120) — the exact issue report.
- **After**: serve starts, `/health` returns `{"status":"healthy"}`; the socket lives at the hashed per-user-temp path (mode 0600) and the pointer file inside the data dir names it. Short data dir regression-checked: canonical socket, no pointer file.
- The CLI connect path itself is covered by the pointer round-trip test (real bind + real connect); a full `octos goal reopen` against a live serve needs a seeded goal in the supervisor store, which the unit-level round trip exercises identically at the socket layer.

## Review

Three rounds: self-review of the full diff, then two independent adversarial reviews. The first found the spelled-vs-canonical decision split and the shared-/tmp pre-creation DoS (both fixed). The second found the environment-dependent fallback breaking bind/connect agreement — motivating the pointer-file design — plus the fixed-name subdir DoS (now per-uid via `rustix::process::geteuid`), unvalidated relative `XDG_RUNTIME_DIR` (now rejected), and a silent canonicalize failure (now debug-logged). All adopted findings re-verified by the full test battery and the end-to-end runs above.

Closes #2248

### DIFF
--- a/crates/octos-cli/src/commands/goal.rs
+++ b/crates/octos-cli/src/commands/goal.rs
@@ -18,7 +18,9 @@
 use std::path::PathBuf;
 
 use clap::{Args, Subcommand};
-use eyre::{Result, WrapErr, bail, eyre};
+#[cfg(all(unix, feature = "api"))]
+use eyre::WrapErr;
+use eyre::{Result, bail, eyre};
 
 use super::Executable;
 use crate::autonomy::supervisor_store::{
@@ -107,10 +109,14 @@ const GOAL_OPERATOR_TRANSITION_METHOD: &str = "session/goal/operator_transition"
 /// `sun_path` capacity for a bound unix socket. The terminating NUL is not
 /// representable in the path, so the limit is exclusive — a path of exactly
 /// this many bytes still fails to bind.
-#[cfg(any(target_os = "linux", target_os = "android"))]
+#[cfg(all(feature = "api", any(target_os = "linux", target_os = "android")))]
 const SUN_PATH_CAPACITY: usize = 108;
 /// Conservative value for every other unix target (macOS/BSD).
-#[cfg(all(unix, not(any(target_os = "linux", target_os = "android"))))]
+#[cfg(all(
+    feature = "api",
+    unix,
+    not(any(target_os = "linux", target_os = "android"))
+))]
 const SUN_PATH_CAPACITY: usize = 104;
 
 /// Marker file a relocated serve publishes inside the data dir, carrying the
@@ -119,7 +125,11 @@ const SUN_PATH_CAPACITY: usize = 104;
 const GOAL_OPERATOR_CONTROL_SOCKET_POINTER: &str = ".octos-goal-control.sock.path";
 
 /// Bind-side resolution of the goal operator-control socket (#2248).
-#[cfg(unix)]
+///
+/// The bind side only exists under the `api` feature (serve), so these
+/// helpers are gated to match — the minimal production library build
+/// (`--no-default-features`) would otherwise see them as dead code.
+#[cfg(all(unix, feature = "api"))]
 struct GoalControlSocketPath {
     path: PathBuf,
     /// True when the canonical in-data-dir path did not fit `sun_path` and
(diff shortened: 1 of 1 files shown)


BH3GEI:
CI fix pushed (af7dd730): the `check-matrix` minimal production library job (`cargo clippy -p octos-cli --lib --no-default-features -- -D warnings`) failed because the goal-control bind-side helpers were gated on `cfg(unix)` while their only caller, `spawn_goal_operator_control`, is `cfg(all(unix, feature = "api"))` — under a no-`api` build they were dead code (8 items: `SUN_PATH_CAPACITY`, `GoalControlSocketPath`, `goal_control_bind_path`, `publish_goal_control_socket`, `fallback_base_dir`, `fallback_socket_base_dir`, `hashed_fallback_socket`, `fnv1a64`). They are now gated to match their caller, the 2248 test module gates on `feature = "api"` too, and `WrapErr` (only used by the bind side) moved behind the same gate.

Verified locally: the exact failing command now passes, plus the matrix-feature clippy variant, default `--all-targets` clippy, `cargo fmt --check`, and all 11 `tests_2248_sun_len` tests.

anantheparty (approved):
Re-approving the updated head - the sun_path relocation fix review stands (threshold-gated deterministic hashed relocation + the in-data-dir pointer file). CI green. Approving - merging left to you per convention.

Issue #2279, opened 2026-09-10
test(peers): three build_cache_peer_tests read real disk free space — fail on any machine with <50 GB free


BH3GEI:
## Symptom

On a machine with less than 50 GB of free disk, `cargo test -p octos-cli --lib` on current main (c8e6ba15) fails 3 tests deterministically (not a timing flake — 0.07 s, 3/3 in isolation):

```
peers::build_cache_peer_tests::two_staged_peers_acquire_distinct_slots_and_release_on_terminal
peers::build_cache_peer_tests::adopt_rule_first_boot_does_not_double_acquire
peers::build_cache_peer_tests::registry_release_is_idempotent_across_terminal_and_close
```

Panic (peers/mod.rs:2081):

```
peer A acquires: RpcError { code: -32603, message: "build-cache slot for peer 'slug-a': free space 5.44 GB (5845180416 bytes) is below the 50 GB gate — refusing to allocate a build-cache slot ..." }
```

CI stays green because the runners have >50 GB free, so this only shows on disk-constrained dev machines.

## Cause

#2268 (098dfa28) added the real free-space gate to the build-cache slot pool (`crates/octos-cli/src/build_cache/pool.rs`, `fs2::available_space`, default `min_free_gb: 50`). These three tests call the production acquire path with `BuildCacheConfig::default()`, i.e. the real 50 GB gate against the real disk, while the file's other tests already neutralize the gate with `min_free_gb: 0` (peers/mod.rs:1984, :2022, :4754, :5160). What the three tests assert (slot namespaces, distinct paths, idempotent release) has nothing to do with free space.

## Suggested fix

Set `min_free_gb: 0` in the three tests' `BuildCacheConfig`, matching the file's own convention — gate behavior itself is already covered by the pool's own tests.

Refs #2268

anantheparty:
Maintenance triage: confirmed — deterministic (0.07s, 3/3 in isolation) means it is a test-isolation defect, not flakiness: the space gate in the #2268 slot pool reads the REAL disk free space inside tests, so the pass/fail depends on the host's disk, not the code. Exactly the #2264 precedent class (environment-dependent test behavior needing an injected source).

**Fix direction**: inject the free-space source into the pool/space-gate (a  trait or a test override returning a fixed value), with the production impl reading real disk — the same test_override_secrets_root pattern the auth family standardized. The three tests then pin gate behavior at chosen thresholds instead of the host's luck.

[...]

PR #2280, opened 2026-09-10
fix(cli): populate hook payload session_id/profile_id on chat and serve --stdio (#2246)

### LINKED ISSUE #2246: hooks: `session_id`/`profile_id` missing from LLM/tool hook payloads on chat and serve
`HookContext` is only populated on some paths. On rc.10 both `octos chat` and `octos serve --stdio --solo` deliver `before_llm_call` / `after_llm_call` / `after_tool_call` payloads without `session_id` or `profile_id` (verified by a dump hook), while the serve log line for the same call does print `session=…`. Any per-session policy (budget, rate limit, audit) has to fall back to the parent pid, which under `serve` means one bucket for the whole server.

Suggested: set the hook context when the session actor builds the agent, on every path.

---

BH3GEI:
## Problem

On `octos chat` and `octos serve --stdio --solo`, `before_llm_call` / `after_llm_call` / `after_tool_call` hook payloads carry no `session_id` / `profile_id`, so per-session hook policy (budget, rate limit, audit) is stateless-blind on exactly the two transports where operators run hooks. The gateway path already populates them via `ActorFactory::hook_context_template`; the OUP transport never did.

## Root cause

Both OUP agent construction sites attach the hook executor without a context:

- `SessionRuntime::bootstrap` (runtime/session.rs) chains `.with_hooks()` but never `.with_hook_context()`;
- `run_standalone_turn` (api/ui_protocol_transport.rs) rebuilds the turn agent from a fresh `Agent::new_shared` every turn, so even a bootstrap-time context could not carry over — and it only re-applied `.with_hooks()`.

## Fix

Per the maintainer's triage guidance (populate at session construction, not per-emission-site):

- `SessionRuntime::bootstrap` now sets `HookContext { session_id: session_key.to_string(), profile_id: profile.profile_id }` on the session agent — the same string forms the gateway template produces.
- `run_standalone_turn` re-applies the same context onto the per-turn rebuilt agent, mirroring the spawn-tool wiring already present in the same function.
- `Agent` gains a `hook_context()` accessor (mirroring the existing `hooks()` accessor, which exists for exactly this kind of runtime-layer assertion) so the regression test can observe the wiring.

Side effect, verified and intended: harness error events on these transports now carry the real session id instead of `"unknown"` (`loop_runner::harness_error_context` falls back to `hook_ctx().session_id`). The context is populated whether or not hooks are configured; with no hooks configured nothing reads it except that fallback.

Known follow-up, deliberately out of scope: gateway sessions without a named profile emit `profile_id: null` while chat/serve now emit `"_main"` for the same ambient profile; unifying that is a gateway-template change.

## Tests

- New `runtime::session::tests::session_runtime_agent_carries_hook_context_ids`: fails before the fix (`hook_context()` is `None` — verified by reverting the wiring), passes after.
- New structural guard `standalone_turn_reapplies_hook_context` (same source-scan idiom as `turn_loop.rs::standalone_turn_loop_breaks_on_interrupt_step`) pins the per-turn re-application in `run_standalone_turn`, the path the reporter actually observed; verified to fail when the wiring is removed.
- `cargo test -p octos-cli --lib`: 3674 passed, 3 failed — the 3 are `peers::build_cache_peer_tests` failing on this machine's real free disk (<50 GB) via #2268's production space gate, unrelated to this diff (no shared code; deterministic 0.07 s failures with an explicit free-space error; filed as #2279).
- `cargo test -p octos-agent --lib hooks`: 65 passed.
- `cargo fmt --check`, `cargo clippy -p octos-cli --all-targets -- -D warnings`, the CI minimal (`--lib --no-default-features`) and matrix-feature clippy variants, and `cargo clippy -p octos-agent --all-targets -- -D warnings` all pass.

## End-to-end evidence

Real binary, real hook, on the reported path (`octos chat --message hi` with a `before_llm_call` dump hook appending stdin to a file; provider pointed at a dead port so the hook payload is the observation):

- Before: `{"event": "before_llm_call"}` — no ids, reproducing the issue.
- After: `{"event": "before_llm_call", "session_id": "_main:cli:01a0899f-…", "profile_id": "_main"}`.

## Review

One round of independent adversarial review (given only the issue, the diff, and file paths): approve, no blockers. Its one actionable finding — the per-turn site was guarded only by the bootstrap test — is addressed by the structural guard above. Two observations recorded for the record: the unconditional (not hook-gated) population is intentional and noted above; the gateway "_main" vs null divergence is the follow-up noted above.

Closes #2246

### DIFF
--- a/crates/octos-agent/src/agent/mod.rs
+++ b/crates/octos-agent/src/agent/mod.rs
@@ -1200,6 +1200,13 @@ impl Agent {
         self.hooks.clone()
     }
 
+    /// Returns the session-level context injected into hook payloads, if
+    /// any. Counterpart to [`Self::hooks`] — the runtime layer uses it to
+    /// assert the context survives session construction (#2246).
+    pub fn hook_context(&self) -> Option<HookContext> {
+        self.hook_ctx()
+    }
+
     /// Set session-level context for hook payloads.
     pub fn with_hook_context(self, ctx: HookContext) -> Self {
         *self.hook_context.lock().unwrap_or_else(|e| e.into_inner()) = Some(ctx);
--- a/crates/octos-cli/src/api/ui_protocol_tests.rs
+++ b/crates/octos-cli/src/api/ui_protocol_tests.rs
@@ -43268,3 +43268,29 @@ async fn ui_transport_autonomous_consumer_emits_verifier_warning_wire_shape() {
         "an unverified claim never completes the goal"
     );
 }
+
+/// #2246 — structural guard: `run_standalone_turn` rebuilds the turn agent
+/// from `Agent::new_shared`, so the bootstrap agent's hook context does not
+/// carry over; the re-application must stay wired (this is the path
+/// `octos chat` / `serve --stdio` actually serve turns on, and the one the
+/// issue reporter observed empty `session_id`/`profile_id` through).
+#[test]
+fn standalone_turn_reapplies_hook_context() {
+    let path =
+        std::path::Path::new(env!("CARGO_MANIFEST_DIR")).join("src/api/ui_protocol_transport.rs");
+    let text = std::fs::read_to_string(&path).expect("read ui_protocol_transport.rs");
+    let start = text
+        .find("async fn run_standalone_turn")
+        .expect("run_standalone_turn exists");
+    let body = &text[start..];
+    let hooks_at = body
+        .find("request_agent = request_agent.with_hooks(hooks);")
+        .expect("run_standalone_turn attaches the profile hook executor");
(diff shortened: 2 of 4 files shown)


anantheparty (approved):
Adversarial pre-merge review passed - implements the #2246 triage prescription exactly: HookContext populated at session construction (the single point where the harness knows both ids), not per-emission-site - session_id from the session key, profile_id from the session runtime/profile, on both chat and serve --stdio --solo. The per-session hook policy surface (budget, rate limit, audit) is no longer stateless-blind. CI green. Approving - merging left to you per convention.

PR #2282, opened 2026-09-10
fix(peers): retain validated identity for closed peers

ZhangHanDong:
Closing a peer used to discard its identity even when its lifetime record was valid, preventing correlation with terminal events. Retain the identity only after the same full trust validation used for open peers; missing, foreign, malformed, and digest-mismatched records keep null identity. Closed status and the last execution outcome remain independent.

Follow-up to #2272. Align the interface and spec for closed identity, generation, idle rounds, profile selection, schema/spec versioning, and serve text. Orphan-running recovery and unmeasured read-performance changes remain deferred.

Validation: closed cases 6/6 and peer-list cases 35/35; GLM-5.3 and K3-256k independent reviews plus mutual review passed. On updated main e0af1634, the combined repair tree passes cargo fmt, cargo clippy --all-targets -- -D warnings, and 12 selected regression tests run through cargo test --all-targets (including both upstream hook-context tests).

The earlier unfiltered workspace run completed 172 suites: 10,061 passed, 2 failed, 100 existing ignored. The fixed-port test failure (3103 instead of 3101) reproduces on unmodified upstream main because another local service owns port 3101. The scheduler test passes in isolation on both upstream and the repair binary. No tests were weakened or newly ignored; the full local run is not reported as green. GitHub CI is still running.

### DIFF
--- a/crates/octos-cli/src/commands/peer.rs
+++ b/crates/octos-cli/src/commands/peer.rs
@@ -69,9 +69,10 @@ pub(crate) struct PeerListRow {
     pub name: Option<String>,
     pub model_lane: Option<String>,
     /// Trusted-lifetime identity (anti cross-runtime/same-slug fencing);
-    /// null when the projection is untrusted — and also for the `closed`
-    /// branch, which intentionally reports identity-free (the closed marker
-    /// takes precedence over the projection; conservative contract).
+    /// null when the projection is untrusted. The `closed` branch runs the
+    /// SAME full trust check (not bypassed by the marker): a trusted
+    /// lifetime keeps its identity for terminal-event correlation; a
+    /// missing/forged/foreign/corrupt one stays null.
     pub master_session_id: Option<String>,
     pub task_id: Option<String>,
     pub generation: Option<u64>,
--- a/crates/octos-cli/src/peers/mod.rs
+++ b/crates/octos-cli/src/peers/mod.rs
@@ -4324,15 +4324,33 @@ pub(crate) fn derive_peer_execution_facet(
     let terminal = read_last_terminal_evidence(peer_dir, slug);
     let last_outcome = terminal.as_ref().map(|(_, outcome)| outcome.clone());
     if closed {
+        // The closed marker wins the EXECUTION label (lifecycle terminal),
+        // but it does NOT bypass the trust check: identity is retained only
+        // when the SAME full projection validation as the open branch passes
+        // (registry_key, writer shape, non-empty identity, Idle digest
+        // re-bind). Missing/forged/foreign/corrupt ⇒ all-null, exactly like
+        // an untrusted open peer (merged-review audit 2026-09-10, PR #2272).
+        // Destructure ONCE (ROOT review 2: successive `identity.map` calls
+        // move the Option after the first consumption — compile error).
+        let (master, task_id, generation, turn_id) =
+            match trusted_lifetime_projection(peer_dir, profile_id, slug) {
+                Some(p) => (
+                    Some(p.master),
+                    Some(p.task_id),
+                    Some(p.generation),
+                    p.turn_id,
+                ),
+                None => (None, None, None, None),
+            };
         return PeerExecutionFacet {
             execution: "closed",
             last_outcome,
             round: rounds_delivered,
             rounds_delivered,
-            master_session_id: None,
-            task_id: None,
-            generation: None,
-            turn_id: None,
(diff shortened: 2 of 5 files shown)


anantheparty (dismissed):
Delta review passed — nit-1 fixed correctly: closed peers now walk the SAME trusted_lifetime_projection as open peers (retain identity on pass, all-null on failure), making the behavior consistent with the untrusted-to-null contract instead of contradicting it; docs synced in three places (peer.rs header, interface.json identity_fields_note, spec); fencing authorization reads recovery directly so the identity retention does not spill into security decisions; 36 peer_list tests green including 5 new closed-row tests + the contract drift lock. The crash-orphan Running known_limitations retention is as-documented. Optional: the plan doc carries a personal local path (/Users/zhangalex/...) — worth scrubbing. Approving — merging left to you per convention.

BH3GEI (approved):
Thanks for putting this together — the fix is minimal and the contract story is consistent across code, interface doc, and spec.

Verified locally at 36c2a3b0 (isolated worktree):

- The closed branch (crates/octos-cli/src/peers/mod.rs:4336) now runs the exact same trusted_lifetime_projection as the open branch (recovery.rs:97) — same registry_key check, writer-shape binding, non-empty identity gate, and Idle digest re-bind; failure degrades to all-null, matching the untrusted-open contract. Destructure-once is correct.
- The retained identity is correlation-only: every consumer of execution_facet.master_session_id/task_id/generation/turn_id is display/serialization (ui_protocol_transport.rs:15940, commands/peer.rs:139); fencing and authorization read the durable files/registry directly, so nothing security-relevant changes behavior.
- last_outcome stays independent of the projection (strict turns.txt × highest result-N cross-check), as the spec claims.
- Doc claims spot-checked against code: the serve suffix superset matches compose_peer_list_text (all rows except closed-without-outcome), the idle_no_turns round>=1 reasoning holds (trusted Idle requires a digest-bound result.md), --profile exists on the CLI, and the spec generation=1 scenario now matches the test fixture (mod.rs:6406).
- Tests: the modified closed test is a genuine RED pin against the old hardcoded-null behavior; the four negative cases are honest regression pins. Local runs: cargo fmt --all --check clean; cargo clippy -p octos-cli --features api --all-targets -- -D warnings clean; 6/6 new closed-peer tests, 36 peer_list tests, 96 peers module tests, and the full octos-cli lib suite 3682 passed / 0 failed / 10 ignored. GitHub CI all green (incl. check-windows).

Approving; merge left to the maintainers per convention.

BH3GEI (approved):
Thanks for refreshing this against latest main. Re-verified at the new merge head 8a49d3b9: the PR-side diff is byte-identical to the previously approved 36c2a3b0 patch and none of the five touched files changed on main in between, so the only thing left to check was interaction with newly landed main.

- The merged ui-protocol refactor (e7b57f2c, replay send loop) has no overlap with peer facet code; the identity-field consumers are still display/serialization-only — serve JSON at api/ui_protocol_transport.rs:15936-15943 and the CLI row copy at commands/peer.rs:139-142. No authorization or fencing path reads the retained identity, so closed-row retention still cannot leak into security decisions.
- Confirmed on the merged tree that the closed branch (crates/octos-cli/src/peers/mod.rs:4346) runs the identical trusted_lifetime_projection as the open branch (crates/octos-cli/src/peers/recovery.rs:97-142): registry_key binding, writer-shape check, non-empty master/task_id, and the Idle result.md re-read + SHA-256 re-bind. Every negative test scenario (foreign registry key, malformed JSON, digest mismatch) maps to a real rejection branch in that function.
- The serve_text_note claim still matches compose_peer_list_text: the exec=/outcome= suffixes apply to all rows except closed-without-outcome (crates/octos-cli/src/peers/mod.rs:4668).
- Fresh verification in an isolated worktree at 8a49d3b9: cargo fmt --all --check clean; cargo clippy -p octos-cli --features api --all-targets -- -D warnings clean; peer_list tests 35/35, peers module 96/96, and the full octos-cli lib suite 3683 passed / 0 failed / 10 ignored. GitHub CI on the new head is fully green (check, check-matrix, check-windows, test-octos-cli, e2e-changes, typos).

An independent pass over the diff surfaced only the read-amplification for Idle peers (result.md read + hashed per listing); that is explicitly deferred with a wontdo disposition in the plan doc and predates this change for open peers, so it does not block. Approving at the new head; merge left to the maintainers per convention.

PR #2273, opened 2026-09-09
fix(goals): bound completion verification and isolate durable evidence by session

ZhangHanDong:
Goal 请求完成时，验证器可能只返回空错误，重复消耗验证费用，或复用不属于当前会话的完成证据。本 PR 明确失败原因，将单轮验证调用限制为最多两次，并核对费用、会话身份和持久记录。

- 严格解析 DONE，区分四类失败，仅对暂时性故障重试；调用费用先入账，再检查预算。
- 完整快照身份、按存储目录/session 隔离的 v3 账本及并发合并共同阻止旧证据串用；存储不可用时在调用模型前拒绝。
- 不存在的相对目录在创建前后保持相同身份，覆盖 `.`、`..` 和软链接组合，避免同一证据重复调用模型。
- 对用户可见的调用错误按 Unicode 字符限制长度；持久化无效答复的有界原因，重启重放保留诊断细节，兼容缺少新字段的旧 v3 记录。
- 将结构化原因传到 goal 工具、session actor 和 UI 通知；规格解析器识别 40 个验收场景，均绑定真实测试。

当前补丁 `38130f80` 已通过 GLM-5.3 / K3-256K 增量互审及 Codex 独立检查。三个审查问题均有真实 RED→GREEN 回归；相关回归 65 通过、0 失败、1 忽略，最后的测试矩阵调整后 32 项 verifier 回归通过。八种路径的独立实际目录创建探针全部通过。格式检查、`cargo clippy -p octos-cli --features api --all-targets -- -D warnings`、规格 parse/lint 均通过。

与 #2272 `e8f77422` 组合后的提交 `142ca5a2` 已在普通干净检出中完成下面的全量检查，源码前后未变：fmt 和 clippy 均成功；172 套件共 10,047 通过、1 失败、100 忽略，all-targets 退出码 101。唯一失败是未改动的 `stdio_connection_stops_dispatch_after_writer_failure`：同一编译产物单独运行连续三次通过，与 `embedded_client_uses_real_oup_negotiation_and_rpc_errors` 两测试并行时重现全局计数器串扰（实际计数 3，断言期望 1）。两个测试及 transport 实现相对原已验证组合 `8b4da851` 均无改动。此项作为已有测试隔离缺陷保留，不将本轮全量记为通过；修复提交 `38130f80` 的 GitHub CI 已全部通过，外环增量复审也已重新 APPROVED。

```sh
cargo fmt --all -- --check
cargo clippy --all-targets --features api -- -D warnings
[...]

PR #2281, opened 2026-09-10
test(peers): neutralize the 50 GB free-space gate in three build_cache_peer_tests (#2279)

BH3GEI:
## Problem

On any machine with less than 50 GB of free disk, `cargo test -p octos-cli --lib` fails three tests deterministically (0.07 s, 3/3 in isolation — not a timing flake):

- `two_staged_peers_acquire_distinct_slots_and_release_on_terminal`
- `adopt_rule_first_boot_does_not_double_acquire`
- `registry_release_is_idempotent_across_terminal_and_close`

CI stays green only because the runners have >50 GB free, so every disk-constrained dev machine hits this.

## Root cause

#2268 added the real free-space gate to the build-cache slot pool (`fs2::available_space`, default `min_free_gb: 50`). These three tests call the production acquire path with `BuildCacheConfig::default()` — the real 50 GB gate against the real disk — while asserting things that have nothing to do with free space (slot namespaces, distinct paths per peer, idempotent release). The file's other tests already neutralize the gate with `min_free_gb: 0` (peers/mod.rs:1984, :2022, :4754, :5160); these three were missed.

## Fix

Set `min_free_gb: 0` in the three tests' `BuildCacheConfig` (one `config` binding per test, shared by `set_build_cache_config` and every `acquire_for_staging` call), matching the file's own convention. Gate behavior itself remains covered by the pool's own deterministic tests (`space_gate_low_space_is_typed_error_not_panic`, `space_gate_zero_disables_the_gate`, `space_gate_unknown_fails_closed`), so no coverage is lost. Test-only change; no production code touched.

## Test evidence

On a machine with 32.85 GB free (below the gate):

- Before: the 3 tests fail with `free space 32.85 GB ... is below the 50 GB gate` (3 passed / 3 failed in the module).
- After: `cargo test -p octos-cli --lib` — **3677 passed, 0 failed** (full suite, first run).
- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p octos-cli --lib --no-default-features -- -D warnings` — clean.
- `cargo clippy -p octos-cli --all-targets --no-default-features --features api,telegram,discord,dingtalk,slack,whatsapp,feishu,email,matrix,audio_mp3 -- -D warnings` — clean.

## End-to-end evidence

The issue scenario *is* the test suite on a disk-constrained machine, so the before/after above is the real-path demonstration: same host, same disk (32.85 GB free), the exact command from the issue goes from 3 deterministic failures to fully green.

## Review

[...]

PR #2283, opened 2026-09-10
fix(goals): route verifier warnings and deduplicate replay notes

ZhangHanDong:
Verifier failure warnings now use the client wire session ID while goal lookups retain their internal cwd scope. Failure notes recover correctly when the verifier ledger was saved but the transcript write failed or the actor restarted before appending the note.

Persist notes by a stable goal-and-evidence identity under the existing per-session lock. A replay repairs a missing durable note, returns an existing note without duplication, and updates RAM only after persistence succeeds. Preserve the original note content/timestamp, reject malformed or unreadable transcripts, and recover a zero-byte file left before metadata initialization.

Follow-up to #2273 and the review on this PR. Add 10 persistence tests and 3 actor recovery tests, including real filesystem failure and a truly stale RAM mirror; bind recovery and rejection cases in the behavior contract. Charging, verifier TTL and ledger retention are unchanged.

Merged upstream main into the review branch in `15652e88`. The only content conflict was both branches appending tests to `session_tests.rs`; resolution retains all 10 idempotent-note tests and all 5 upstream torn-tail tests, with their original assertions. The strict note helper and actor recovery logic remain unchanged. After #2282 landed, merge `dd79aa0a` brought in current main `f4b0afcd` to clear the latest-base requirement. Its five added files are identical to main; all nine reviewed verifier files and the complete PR-side diff remain byte-identical to the reviewed candidate.

Validation on `15652e88`:

- `cargo fmt` and `cargo clippy --all-targets -- -D warnings` passed.
- `cargo test --all-targets --no-fail-fast` completed 172 suites: **10,081 passed, 1 failed, 100 ignored**. Both conflicted test groups passed, along with the actor recovery and warning regressions.
- The sole failure remains `should_allocate_bridge_ports_as_consecutive_pair`: an existing local Node service owns port 3101, so allocation returns 3103. The test and production port allocator are unchanged from main. No service was stopped and no assertion was relaxed or newly ignored.
- Independent GLM and Codex source reviews and cross-reviews found no code blockers; root verified the nine source hashes and runtime review receipts.

Validation on final sync `dd79aa0a`: `cargo fmt` and `cargo clippy --all-targets -- -D warnings` passed. `cargo test --all-targets peer_list` compiled all 172 target suites and ran **40 matching tests, all passing**. The full unfiltered run above was on `15652e88`; the verifier source and PR-side diff did not change in this sync.

Non-blocking follow-up: diagnostic line counting skips blank lines in a corrupted transcript, so the reported error location can be too small. Writes still fail closed; this is recorded for a separate correction.

GitHub CI is rerunning on the new merge commit; a fresh reviewer approval is still needed.

### DIFF
--- a/crates/octos-bus/src/session.rs
+++ b/crates/octos-bus/src/session.rs
@@ -2436,6 +2436,150 @@ pub async fn persist_message_through_canonical_path(
     handle.add_message_with_seq_unlocked(message).await
 }
 
+/// Idempotently persist ONE system note to a session through the canonical
+/// per-key persist path, returning the durable message.
+///
+/// Semantics (PR #2283 follow-up): the existence check and the append run
+/// inside the SAME `persist_lock_for` critical section as every other
+/// canonical write, so two concurrent callers for the same `note_id` commit
+/// exactly one row and both receive that row.
+///
+/// STRICT canonical read (this helper does NOT relax the legacy loader):
+/// - the file's size is bounded by `MAX_SESSION_FILE_SIZE`;
+/// - the first line must be a valid `SessionMeta` with a supported schema;
+/// - every non-empty subsequent line must parse as a `Message` OR a
+///   `SessionControlRecord` — an unparsable line, a missing trailing
+///   newline on a non-empty final line, or an unreadable-but-existing file
+///   are `Err` (fail-closed: never treated as "absent", never appended past
+///   a corrupt tail);
+/// - visibility follows rollback semantics: the note is "present" only if
+///   `assemble_session_messages` (the same fold every reload uses) still
+///   shows it — a rolled-back note is legitimately absent and is re-added.
+///
+/// On success returns the durable `Message` — the row this call appended,
+/// or the pre-existing row with its ORIGINAL timestamp and content (the
+/// note is never re-stamped with a fresh timestamp).
+///
+/// The id rides `client_message_id`; `role` is forced to `System` and
+/// `thread_id` to `None` (no thread semantics implied).
+pub async fn persist_system_note_once_through_canonical_path(
+    data_dir: &Path,
+    key: &SessionKey,
+    note: Message,
+    note_id: &str,
+) -> Result<Message> {
+    let lock = persist_lock_for(key);
+    let _guard = lock.lock().await;
+    let mut handle = SessionHandle::open(data_dir, key);
+
+    let canonical = handle.session_path();
+    let file_meta = match std::fs::metadata(&canonical) {
+        // Genuinely absent: a new session — the append below writes the
+        // meta line itself. Only NotFound counts as absent; permission or
+        // any other stat error is fail-closed (never "treated as absent").
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => None,
+        Err(error) => {
+            return Err(eyre::eyre!(
+                "session transcript stat failed ({}): {error}",
+                canonical.display()
+            ));
+        }
+        Ok(meta) => Some(meta),
+    };
+    // A zero-byte regular file can remain after creation but before the
+    // metadata write; the canonical append initializes it on retry.
+    if let Some(meta) = file_meta.filter(|meta| !meta.is_file() || meta.len() != 0) {
+        if meta.len() > MAX_SESSION_FILE_SIZE {
+            return Err(eyre::eyre!(
+                "session transcript exceeds size limit ({}: {} > {})",
+                canonical.display(),
+                meta.len(),
+                MAX_SESSION_FILE_SIZE
+            ));
+        }
+        let bytes = std::fs::read(&canonical).map_err(|error| {
+            eyre::eyre!(
+                "session transcript exists but cannot be read ({}): {error}",
+                canonical.display()
+            )
+        })?;
+        let text = String::from_utf8(bytes.clone()).map_err(|error| {
+            eyre::eyre!(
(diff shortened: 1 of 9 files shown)


anantheparty (dismissed):
Delta review passed — both nits addressed correctly: the verifier warning routing normalizes the session key before the wire (no NUL/cwd- prefix leakage into UI warnings; the autonomy station's un-stripped passthrough is intentional and comment-verified), the three consumer points share one canonical Display source, and the replay dedup keeps warn-always while durable/in-memory notes append only on non-replayed. The independent-diagnostic-kind suggestion's wontdo disposition is evidence-backed (the diagnostic field already distinguishes at every consumer point). Zero regression surface on the approved DONE ladder/charging/v3-isolation/single-flight/preflight. On merge, flip the spec status line to merged. Approving — merging left to you per convention.

BH3GEI (commented):
Thanks for the quick follow-up — this closes out all four items from my #2273 review, and the care shows. I verified locally on this head (9916b0e8): `cargo fmt --all -- --check` clean; `cargo clippy -p octos-cli --features api --all-targets -- -D warnings` clean; the four new regression tests pass; the goal (260), sentinel (13), wire_key (5), verifier (47), and replay (56) suites all pass; a full `cargo test -p octos-cli --features api --lib` run is 3678 passed / 3 failed, where the three failures are the known `peers::build_cache_peer_tests` 50 GB free-space gate on my machine (24 GB free) in a module this PR doesn't touch — the same environmental failures I saw on #2273, since neutralized upstream by #2281. GitHub CI is green across the board.

Confirming each prior item, verified against the code:

1. **Warning wire key** — fixed at the right boundary. `goal_verifier_failure_warning` (ui_protocol_transport.rs:32300) now normalizes through `wire_key_from_goal_key` (agent_orchestrator.rs:1025), so the interactive call site (:36739) keeps passing the turn-pinned scoped key for goal lookups while the `WarningEvent.session_id` that hits the wire is the plain id. The autonomous station (:37625) correctly keeps using `goal_verifier_warning_event` with the plain `params.session_id` — I traced `session_id` through `run_standalone_turn` and the GoalContinue runner to confirm it is never the scoped key. The new test drives the real constructor over a real `WsConnection` send and asserts the exact wire id — nice.
2. **Spec boundary** — the two test files are now in Allowed Changes, and all four `Filter:` names match the test functions byte-for-byte.
3. **Doc drift** — historical banner on the 2026-09-09 plan, status flipped, personal path removed; the Decision 2(f) delimiter guard and Decision 5 `reason` column text match the merged implementation (goal_loop_runtime.rs:517-528, agent_orchestrator.rs:16534-16536).
4. **Replay note dedup** — implemented as suggested: warn-always, note only on `!outcome.replayed`. The digest-stability reasoning behind the tests checks out: the appended note is a System message and never enters `assistant_tail` (session_actor.rs:5389-5399), so call 2 is a deterministic gate replay.

One concern before merge, in the new dedup logic:

- **The replay suppression turns a self-healing write into a silent single-shot** — `crates/octos-cli/src/session_actor.rs:5486-5491`. The durable note append is still best-effort (`let _ = …`), but its failure mode changed. The wrapper's ledger append is durable and happens before the outcome is returned, so if `persist_message_through_canonical_path` fails here — or the process dies in the window between the ledger append and this write — the next turn (or post-restart turn) replays the stored verdict with `replayed=true` (agent_orchestrator.rs:4920-4933) and the new guard at :5478 skips the note. The in-memory copy pushed at :5482-5485 doesn't survive a restart either, so the session permanently loses the user/model-facing "goal completion not verified" note for that evidence digest, with nothing logged. Pre-PR this hole self-healed because every replayed turn re-appended the note. Could we at least log the persist error (it's currently discarded silently), and ideally make the suppression robust to it — e.g. treat a failed durable write as a reason to retry the note on the next replay rather than relying on the ledger alone?

Non-blocking nit: the new comment at ui_protocol_transport.rs:32308 cites "interactive :36723", but the call site in this revision is :36739 — the reference is already stale within the same file.

Everything else looks solid; happy to re-review quickly once the note-durability question is addressed.

ZhangHanDong:
Resolved the main-branch conflict in 15652e88. Both EOF test groups are preserved byte-for-byte: 10 idempotent-note tests from this PR and 5 upstream torn-tail tests. Follow-up merge dd79aa0a now includes current main f4b0afcd (merged #2282), clearing the latest-base requirement. The five added files are identical to main, and all nine reviewed verifier files plus the complete PR-side diff are byte-identical to the reviewed candidate.

The earlier note-durability concern remains fixed by 41660658: every failed-verification replay retries a canonical, idempotent append; errors are logged, and RAM is updated only after persistence succeeds. Merging main did not change that helper or actor recovery logic. The stale call-site line-number comment was also removed in that repair.

Full validation on 15652e88: fmt and clippy passed; all-targets completed 10,081 passed / 1 failed / 100 ignored across 172 suites. The only failure is the unchanged bridge-port test because a pre-existing local service owns port 3101. All conflict-related tests and recovery regressions passed. Independent GLM/Codex source review and cross-review found no code blockers. On final sync dd79aa0a, fmt and clippy passed again; cargo test --all-targets peer_list compiled all 172 target suites and ran 40 matching tests, all passing. This is additional validation of the newly merged upstream increment; the full unfiltered result above belongs to 15652e88. GitHub CI is rerunning on dd79aa0a; a fresh reviewer approval is still needed.

PR #2284, opened 2026-09-10
fix(cli): fire the documented on_turn_end hook on the OUP turn path

### LINKED ISSUE #2244: hooks: `on_turn_end` is documented but never fired
`book/src/advanced.md` lists `on_turn_end` among the ten lifecycle events and `HookPayload::on_turn_end` exists, but no runtime path calls `hooks.run(HookEvent::OnTurnEnd, ..)` (grep over `crates/` at rc.10 and `main`). A hook registered on it never runs and there is no warning.

Repro: register `{"event":"on_turn_end","command":["sh","-c","echo hi >> /tmp/turn_end"]}` in a profile, run any turn, `/tmp/turn_end` stays absent.

Suggested: fire it at the end of `run_conversation` with `turn_summary`, or drop it from the docs until it exists.

Related: #435 ("Add generic harness lifecycle hooks for turn-end, resume, and spawn completion") is closed, but the turn-end event has no call site.

---

BH3GEI:
## Problem

`book/src/advanced.md` documents `on_turn_end` as one of the ten lifecycle events, and the SessionActor (gateway/channel) path fires it — but the OUP turn path (`octos chat`, `octos serve-stdio`, WS `turn/start`) never did: `run_standalone_turn` in `ui_protocol_transport.rs` had no emission, so a hook registered on the event silently never ran for every CLI/local turn. The issue's grep-based claim that *no* runtime path fires it was too broad (the SessionActor emission goes through the `emit_hook_payload` indirection), but the repro is real on the OUP path.

Reproduced on `main` (b54e54e3) with a real binary + mock OpenAI server: an `on_turn_end` hook's output file stays absent after `octos chat -m`, while a `user_prompt_submit` hook in the same config fires — proving hooks are loaded and only this event is missing.

## Root cause

`run_standalone_turn` drives the whole OUP turn and terminates it through `try_emit_terminal` (Completed/Errored/Interrupted) without ever running `HookEvent::OnTurnEnd`.

## Fix

Fire `HookPayload::on_turn_end` at the turn's three outcome points inside `run_standalone_turn`, with the same whitespace-normalized `turn_summary` + `session_id`/`profile_id` context the SessionActor path emits:

- **done / error arms**: the fire is awaited BEFORE the terminal frame, so a client that observes the turn's end can rely on the hook having run — in particular a one-shot `octos chat -m`, whose process exits right after the terminal. In the done arm the fire is placed before the `active_turns_registry()` lock is taken, so the (up to `timeout_ms`) hook wait never happens under that lock.
- **interrupt arm**: the fire follows the terminal frame instead, because the interrupt handler is blocked on that frame within the 5s `INTERRUPT_ACK_TIMEOUT` — an up-to-`timeout_ms` hook wait must not sit between the state flip and the ack.

Short-circuit paths that never ran the agent (slash-command dispatch, voice no-speech, pre-dispatch runtime errors) intentionally do not fire, matching "when an agent turn finishes". Master-continuation turns (goal/loop/peer drains) do fire, matching SessionActor parity.

The observe-only result handling (warn on Deny/Modified/Feedback/Error) is shared with SessionActor via the extracted, behavior-identical `emit_lifecycle_hook_payload`; `git_turn_summary` becomes `pub(crate)`.

## Testing

- New `local_oup` tests drive a full in-process OUP turn (mock LLM) with an `on_turn_end` capture hook registered:
  - completed turn → payload contains `event`, whitespace-normalized `turn_summary`, `session_id`, `profile_id`, and exactly one fire per turn;
  - errored turn (non-retryable auth failure) → still exactly one fire.
  - Both verified RED without the transport/session_actor changes and GREEN with them.
- End-to-end with the real binary + mock OpenAI server on the exact issue repro: `octos chat -m` now emits `{"event":"on_turn_end","session_id":"_main:cli:…","profile_id":"_main","turn_summary":"final e2e check"}` (absent before the fix); a hook that sleeps 2s still completes before process exit, validating the before-terminal ordering.
- `cargo test -p octos-cli --lib`: 3679 passed / 0 failed. `cargo fmt --check` clean. Clippy `-D warnings` clean in all three CI variants (workspace `--all-targets`, minimal `--lib --no-default-features`, `--all-targets` with api+all channel features).
- The interrupt path's fire is covered by code inspection + the shared call shape with the two e2e-verified arms; scripting a mid-turn interrupt against the local harness was not practical (turns complete before the interrupt RPC lands).

## Review

Independent adversarial review (given only the issue, the diff, and file paths) confirmed exactly-once semantics (done/error arms break with `interrupt_observed == false`; the interrupt section is guarded), correct snapshot placement (after STT/voice prompt rewrites, before `prompt` moves into the agent task), and a behavior-identical helper extraction. Its one blocker — the hook wait could blow the 5s interrupt-ack deadline — is fixed by the interrupt-arm reorder described above; its test-coverage nits (exactly-once assertion, error-path test) are adopted. Rejected: firing on pre-agent short-circuits (no agent turn ran) and doc changes (the existing "when an agent turn finishes" wording already covers autonomous turns, matching SessionActor behavior).

Closes #2244

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol_transport.rs
+++ b/crates/octos-cli/src/api/ui_protocol_transport.rs
@@ -32688,6 +32688,19 @@ async fn run_standalone_turn(
             return;
         }
     };
+    // #2244 — `on_turn_end` fires at this turn's terminal below (completed,
+    // errored, or interrupted). On the done/error arms it fires BEFORE the
+    // terminal frame is emitted, so a client that observes the turn's end —
+    // including a one-shot `octos chat -m` whose process exits right after —
+    // can rely on the hook having run; the interrupt arm fires right after
+    // its frame instead, to stay inside the 5s interrupt-ack deadline.
+    // Resolved once here from the same profile the turn's agent hooks come
+    // from; `None` (no hooks configured) makes each fire a no-op.
+    let turn_end_hooks = session_runtime.profile.hook_executor.clone();
+    let turn_end_hook_ctx = octos_agent::HookContext {
+        session_id: Some(session_id.to_string()),
+        profile_id: Some(session_runtime.profile.profile_id.clone()),
+    };
     // Outer-loop #4 (§4.2): this turn's peers root — `Some` ONLY when this
     // session is a peer (topic `peer-<slug>`) running under the profile's
     // data dir. The interrupted-terminal release below keys the slot registry
@@ -35297,6 +35310,10 @@ async fn run_standalone_turn(
     // `session`/`turn` from this span (postfix `.instrument` keeps the block
     // itself untouched).
     let turn_span = crate::turn_trace::turn_span(&session_id, &turn_id);
+    // #2244 — snapshot the turn summary now that every prompt rewrite above
+    // (STT transcription merge, voice-mode suffix) has landed; `prompt`
+    // itself moves into the agent task below.
+    let turn_end_summary = crate::session_actor::git_turn_summary(&prompt);
     let agent_task = tokio::spawn(async move {
         let start = std::time::Instant::now();
         // RFC-3 (#1292): wrap the agent.process_message future in the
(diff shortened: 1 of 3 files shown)


anantheparty (approved):
Re-approving the updated head - the on_turn_end OUP-path wiring review stands (the documented lifecycle event now fires on the chat/serve-stdio/WS turn path, matching the gateway path). CI green. Approving - merging left to you per convention.

anantheparty:
Heads-up from tonight's nightly triage (octos#2267, Night-6): after this landed, the protocol-e2e lane went from 1 known failure (the -32601/-32004 ruling) to **15 failures with a single signature — `timeout waiting for notification turn/completed`** — turn/start happy path, both interrupt-idempotency specs, progress/updated, tool correlation, fault-injection replay, and the web-client WS suite (https://github.com/octos-org/octos/actions/runs/34751342318).

This change fires `HookPayload::on_turn_end` at exactly the three outcome points those tests wait on, so it's the prime bisect suspect (window = the 12 commits since last night's nightly). Not confirmed yet — if you can check whether a hook resolution/execution delay or error on the OUP path can defer or swallow the terminal notification, that would settle it fast; otherwise consider reverting until the nightly is green so the semantic ruling on family 3 stays diagnosable.

PR #2287, opened 2026-09-10
fix(runtime): create session skill output lazily

### LINKED ISSUE #2286: Avoid eagerly creating skill-output/ in the project root
Starting a writable coding session creates a `skill-output/` directory directly under the selected project root, even when no skill has been invoked and no artifacts have been generated.

This adds an unexpected tool-managed directory to the user's source tree. Once populated, it can also introduce untracked files unless the user adds an ignore rule.

## Current behavior

In `crates/octos-cli/src/runtime/session.rs`, session bootstrap selects `workspace_root.join("skill-output")` for writable sessions and immediately calls `create_dir_all`.

Verified in upstream main at b54e54e302cfd19774adf4ac814af2a4a8158e6f:
https://github.com/octos-org/octos/blob/b54e54e302cfd19774adf4ac814af2a4a8158e6f/crates/octos-cli/src/runtime/session.rs#L307-L331

PR #2265 moved read-only session output outside the selected project, but writable sessions still use the top-level directory. A search of related commits, PRs, and issues did not find a fix specifically addressing this behavior for writable sessions.

## Steps to reproduce

1. Open a writable coding session in a project without a `skill-output/` directory.
2. Allow the session to initialize without invoking any skills.
3. Observe that `<project>/skill-output/` has been created.

## Expected behavior

Session initialization should not create an unused top-level artifact directory.

Consider using `<project>/.octos/skill-output/` for default tool-managed output and creating it only when an operation actually needs it. Explicit user-selected output paths should continue to be respected.

## Implementation considerations

The output location is referenced by plugin path resolution, output hints, pipeline tools, artifact discovery, and workspace contracts. These should use a shared path resolver rather than independently hardcoding `skill-output/`.

Existing artifacts and persisted paths should remain accessible; migration should not silently delete or overwrite user files.

## Acceptance criteria

- Starting a writable session does not create `<project>/skill-output/`.
- Default skill output uses the designated tool-managed directory, created on demand.
- Plugin execution, artifact validation, discovery, and delivery continue to work.
- Existing output paths remain supported.
- Read-only sessions continue to keep generated runtime output outside the project.

---
Drafted and submitted by Codex (GPT-6), at the user's request.

chrislearn:
Starting a session currently creates an unused `skill-output/` directory before any plugin runs. Resolve the session output path without creating it and rely on the existing plugin execution path to create its working directory before argument rewriting and process startup.

This follows the maintainer's requested scope: keep writable and read-only output locations unchanged, preserve existing artifacts, and change only directory creation timing.

Fixes #2286

## Validation

- Session runtime tests: 33 passed, including writable-session lazy creation, preservation of existing artifacts, default-workspace lazy creation, and read-only workspace isolation. One existing test fails on Windows because it invokes `printf`, which is unavailable in the Windows command shell; that command is also present on main.
- Plugin execution regression: passed on Windows. Verifies that binding leaves the output directory absent, the first execution creates it and writes an artifact, and the registry-bound directory takes precedence over the fallback session scope. The fixture now supports both Unix and Windows; the Unix run was blocked locally by WSL's missing `cc` linker.
- `cargo fmt --all -- --check` and `git diff --check`: passed.


anantheparty (approved):
Adversarial pre-merge review passed - the #2286 lazy-materialization fix: the session output path is resolved without creation, and the existing plugin execution path creates the working directory where it is actually needed (before argument rewriting and process startup). A clean session leaves the tree exactly as found. Test-only-adjacent surface, CI green. Approving - merging left to you per convention.

BH3GEI (approved):
Thanks for putting this together — the lazy-creation approach matches the scope requested on #2286 and keeps both writable and read-only output locations unchanged.

Verified locally on the PR head (70c5a4c):

- `cargo test -p octos-cli --lib runtime::session`: 34/34 pass, including the new `writable_session_bootstrap_keeps_skill_output_lazy_and_preserves_artifacts` (bootstrap leaves `skill-output/` absent; reopening preserves existing artifacts at the same path) and the updated read-only assertion at `crates/octos-cli/src/runtime/session.rs:1211`.
- `cargo test -p octos-agent --lib plugins::tool::tests`: 123/123 pass, and `plugin_prefers_registry_rebound_work_dir_over_scope` passes individually — binding leaves the output dir absent and the first execution creates it plus the artifact.
- `cargo fmt --all -- --check` and `cargo clippy -p octos-cli -p octos-agent --lib --tests`: clean.
- Read through every consumer of `plugin_work_dir` / `skill-output` / `output_dir_hint` across crates: nothing canonicalizes, serves, or git-inits the directory at bootstrap. `PluginTool::execute` creates the effective work dir at `crates/octos-agent/src/plugins/tool.rs:2796` before argument rewriting (`:2833`) and spawn, and `/api/files/list` guards on `ws_dir.exists()` (`crates/octos-cli/src/api/handlers.rs:3455`), so a lazily-created dir is safe end to end.
- The Windows rewrite of the registry-rebound test uses valid batch syntax and the inline `ctx_with_scope` expansion is verbatim-equivalent to the helper.

CI is green across the board, including check-windows. Approving — merge left to the maintainers.

Issue #2276, opened 2026-09-10
flaky: api::router::tests::http_trace_uses_route_templates_and_omits_credentials failed once in CI (test-octos-cli lib), not reproducible locally


BH3GEI:
Observed once: [run 34405701140](https://github.com/octos-org/octos/actions/runs/34405701140) (PR #2225, head 6bb2a06a, 2026-09-09), `Test octos-cli (lib)`:

```
thread 'api::router::tests::http_trace_uses_route_templates_and_omits_credentials' panicked at crates/octos-cli/src/api/router.rs:1515:9:
assertion failed: logs.contains("path=/api/ui-protocol/ws")
test result: FAILED. 3635 passed; 1 failed
```

The same job passed on rerun with an identical tree, and the failing PR touches only `crates/octos-bus/src/channel.rs` — no plausible connection to the router trace test.

Reproduction attempts (all clean):
- The test in isolation: pass.
- Full `cargo test -p octos-cli --lib` suite (the CI command): 6 consecutive local runs, all pass.
- A targeted stress test (8 threads continuously registering INFO-capped `tracing_subscriber::fmt` subscribers while the trace-capture flow runs 2000 iterations against a real `TraceLayer` + `make_http_trace_span` router): 2000/2000 pass.

Mechanism notes for whoever picks this up:
- The obvious suspect — cross-test poisoning of the shared `debug_span!` callsite interest cache by concurrently registered INFO-capped subscribers — does **not** hold: tracing-core's `Interest::and` maps differing subscriber opinions to `sometimes` (re-ask the current dispatcher), so a concurrent INFO subscriber cannot hard-disable the victim's DEBUG span. Verified in tracing-core 0.1.36 source (`subscriber.rs` `Interest::and`, `callsite.rs` `rebuild_callsite_interest`) and empirically by the stress test above.
- Concurrent tests in the same binary that install subscribers: `turn_trace::tests::capture` (INFO) and `rejected_auth_logs_omit_query_and_token_material` (fmt default = INFO). Timeline of the failed run shows `turn_trace` tests completing in the same second as the failure — suspicious adjacency, but no proven mechanism.
- The span context reaches the captured log via tower-http 0.6.8 `ResponseFuture::poll` entering the request span before `on_response` emits "finished processing request"; all on a current-thread runtime inside `with_default`, so no cross-thread dispatch is expected.

[...]

PR #2288, opened 2026-09-10
test(agent): give the structured-metadata plugin test the shared 120s budget

### LINKED ISSUE #2093: flaky: plugins::tool::tests::execute_preserves_plugin_structured_metadata fails ~1-in-3 under full-parallel octos-agent --lib

BH3GEI:
## Problem

`plugins::tool::tests::execute_preserves_plugin_structured_metadata` flakes under full-parallel `cargo test -p octos-agent --lib` (#2093) — ~1-in-3 on the reporter's macOS arm64 host, always green in isolation.

## Root cause

The test was the last `plugins::tool` case still on a hardcoded `Duration::from_secs(5)` timeout; every other non-timeout test in the file uses `TEST_PLUGIN_TIMEOUT` (120s), whose doc comment already records this exact failure class ("a loaded host can exceed [5s] just by scheduling the subprocess"). On a saturated host the plugin child cannot get scheduled within 5s, `PluginTool::execute` returns `Err("plugin ... timed out after 5s")` (tool.rs wait-timeout path), and the test panics at `.expect("execute should succeed")`.

## Reproduction (mechanism confirmed, not just hypothesized)

On this repo's `main`, with the host CPU-saturated (14 `yes` burners on 10 cores, 3 suites concurrently) — 3/3 runs failed with the exact signature:

```
thread '...execute_preserves_plugin_structured_metadata' panicked at crates/octos-agent/src/plugins/tool_tests.rs:1456:10:
execute should succeed: plugin 'test-plugin' tool 'metadata_tool' timed out after 5s
````

Without induced load: 4/4 full-suite runs green (consistent with the issue's intermittent, load-dependent character).

## Fix

One line: `.with_timeout(Duration::from_secs(5))` → `.with_timeout(TEST_PLUGIN_TIMEOUT)`. The test asserts `structured_metadata` parsing, not timeout behavior; the one test that genuinely exercises the timeout path (`execute_timeout_returns_error`) keeps its own 1s value and is untouched. The same 120s now also covers the stdin-write timeout (both paths share `self.timeout`).

## Test evidence

- Fixed test in isolation: green.
- Full suite `cargo test -p octos-agent --lib` after the fix: 2889 passed, 0 failed.
- End-to-end before/after under the same saturation that failed 3/3 on main: **3/3 green** after the fix (2889 passed each run).
- `cargo fmt --all -- --check` and `cargo clippy -p octos-agent --all-targets -- -D warnings`: green.

## Review

Independent adversarial review (fresh agent, given only the issue, the diff, and the file paths): APPROVE — verified the timeout raise does not weaken the test's assertions, that both timeout paths (stdin-write + wait) are covered, that no `from_secs(5)` remains in the file, and that the change matches existing convention.

Closes #2093

### DIFF
--- a/crates/octos-agent/src/plugins/tool_tests.rs
+++ b/crates/octos-agent/src/plugins/tool_tests.rs
@@ -1448,7 +1448,7 @@ async fn execute_preserves_plugin_structured_metadata() {
         make_tool_def("metadata_tool", "returns structured metadata"),
         script_path,
     )
-    .with_timeout(Duration::from_secs(5));
+    .with_timeout(TEST_PLUGIN_TIMEOUT);
 
     let result = tool
         .execute(&json!({}))


anantheparty (approved):
One-line flake fix verified: the last hardcoded-budget plugins::tool test joins the shared 120s budget, closing the ~1-in-3 full-parallel flake (#2093). The #2276 flake-filing discipline (single observation + root cause + minimal fix) working as intended. Test-only, CI green. Approving - merging left to you per convention.

PR #2289, opened 2026-09-10
fix(bus): seal torn session JSONL tail on append and fsync session rewrite

### LINKED ISSUE #2006: session log: no torn-tail seal — a partial line eats the NEXT message, can resurrect a rolled-back turn, and breaks the documented post-fsync guarantee
## Summary

The session JSONL log has no torn-tail seal on append and no repair on read. A single partial line after an abrupt exit therefore (1) silently destroys the **next** message written, (2) can **resurrect a rolled-back turn**, and (3) breaks a durability guarantee the code explicitly documents to the UI.

octos already implements exactly the right pattern elsewhere — this is an inconsistently applied discipline, not a missing capability.

## 1. A torn tail eats the NEXT message too

`crates/octos-bus/src/session.rs`, `append_to_disk` (~:1839-1894) opens with `OpenOptions::new().create(true).append(true)` and `writeln!`s. If the file does not end in `\n` (process killed mid-write), the leftover partial bytes and the new row **fuse into one line**.

On read, `parse_session_timeline` (~:416-432) does `serde_json::from_str(line).ok()` and silently drops anything unparseable. So the fused line takes **two** messages with it: the truncated one *and* a complete, correctly-written one. No warning, no counter, no metric.

**octos already has the fix, in production**, at `crates/octos-cli/src/autonomy/supervisor_store.rs:1126-1153` (`write_row_sealed_locked`):

```rust
let len = file.metadata()?.len();
if len > 0 {
    file.seek(SeekFrom::Start(len - 1))?;
    let mut last = [0_u8; 1];
    file.read_exact(&mut last)?;
    if last[0] != b'\n' { payload.push(b'\n'); }   // seal the torn tail
}
```

~8 lines to mirror.

## 2. A torn tail can resurrect a rolled-back turn

`append_rollback_marker` (~:2215-2219) appends through the same unsealed path. If that marker is the row that gets fused/dropped, the rollback is lost on reload and **`/undo` un-does itself** — the turn the user removed comes back.

This is a correctness bug, not a durability one, and it is the reason to treat this as more than housekeeping.

## 3. `session.rs:1363` documents a guarantee the code does not provide

The comment states (UPCR-2026-012) that the observer fan-out is **post-fsync**, so a `message/persisted` notification reflects a row that is *"durably visible"*. `append_to_disk` never calls `sync_all`/`sync_data` — the notification is post-`write(2)` only. Either the comment or the code should change; right now the protocol contract is documented as stronger than it is.

Related: `rewrite` (~:1946-1955) does `File::create` → `write_all` → `file.flush()` → `fs::rename`. **`flush()` on a `std::fs::File` is a no-op** (there is no userspace buffer), and there is no `sync_all` before the rename or directory fsync after. The nearby doc ("atomic write-then-rename to avoid corruption on crash") is true about *atomicity* and false about *durability*.

## 4. The supervisor store seals on write but refuses on read

`supervisor_store.rs:1370 read_ledger_rows` returns `InvalidData` on any unparseable line (:1386-1395), which fails the whole boot restore in `configure_supervisor_store` (`agent_orchestrator.rs:1347-1363`). So the write path heals a torn tail but the read path rejects it — a genuinely half-written final line is only repaired if a *write* happens before the next read.

## Recommended scope

Do:
- torn-tail seal in `session.rs` append paths (~8 lines, mirroring `supervisor_store.rs:1136-1143`) — fixes (1) and (2)
- `sync_all` + directory fsync on `rewrite`'s rename, using the helper pattern already in `supervisor_store.rs:1513-1516` / `octos-memory/src/memory_store.rs:164,175`
- make `read_ledger_rows` torn-tolerant on the FINAL line only (strict elsewhere), so a torn tail cannot block boot restore
- fix or downgrade the `session.rs:1363` durability comment so it matches reality

Do **not** add per-append fsync to the session log: it defends only against hard power loss and would reverse two deliberate, ADR-documented tradeoffs (`ui_protocol_ledger.rs:2189-2197` + `docs/M9-LEDGER-DURABILITY-ADR.md`, and `supervisor_store.rs:907-921` "DURABILITY (KNOWN LIMITATION, accepted)").

## Worth stealing separately

Every JSONL reader in octos (`session.rs`, the UI ledger, cron) uses `.ok()`-and-skip, so **mid-file** corruption is invisible everywhere — only tail damage is even theoretically detectable. Sequence-gap validation on read (detect and refuse, rather than silently skip) would make corruption observable across all of them. Filed here as context; probably its own issue.

## Provenance

Found while adversarially fact-checking an external comparison report. The report claimed octos lacks torn-tail handling entirely — that is false (see `supervisor_store.rs` above) — but the session log specifically does lack it, and the knock-on effects (2) and (3) are worse than what the report described.

BH3GEI:
## Problem

A crash or kill mid-write leaves the session JSONL with a torn tail — a partial final line with no newline. Three consequences, all reproduced on `main`:

1. **A torn tail eats the NEXT message.** The append paths (`SessionManager::append_to_disk`, `SessionHandle::append_to_disk`) open with `append(true)` and `writeln!` straight onto the torn bytes; the fused line is unparseable, and `parse_session_timeline` silently drops it — destroying the complete new row along with the torn one.
2. **A torn tail can resurrect a rolled-back turn.** `append_rollback_marker` uses the same unsealed path; when the marker row is the one fused and dropped, `/undo` un-does itself on the next load.
3. **The documented durability guarantee is stronger than the code.** The UPCR-2026-012 comment called the `message/persisted` observer fan-out "post-fsync" / "durably visible", but no fsync exists on the append path (deliberate tradeoff); and `rewrite` did `File::create` → `write_all` → `flush()` (a no-op on `std::fs::File`) → `rename` with no `sync_all` and no directory fsync.

## Root cause

The session log never adopted the torn-tail discipline octos already has in production in `supervisor_store.rs::write_rows_sealed_locked` — an inconsistently applied pattern, not a missing capability.

## Fix

- New `seal_torn_tail` helper (~15 lines, mirroring `write_rows_sealed_locked`): if the file is non-empty and does not end in `\n`, write the missing terminator first. Torn bytes are preserved, never truncated. Wired into all three append paths; `append_rollback_marker` seals for both flat and per-user targets.
- `rewrite` (all three variants): `sync_all()` the temp file before the rename, `fsync_dir` the parent after it — reusing the helpers already in this file — so the rewrite is durable, not just atomic. The tmp file is now also cleaned up if the newly-real `sync_all` fails (previously only `create`/`write_all` could fail).
- The UPCR-2026-012 comment now says what the code does (post-commit, write(2) returned, per-append fsync deliberately not done). No per-append fsync was added, per the issue's scope.

The supervisor-store read path the issue mentions (`read_ledger_rows`) was already made tolerant by the later #26a change — verified, nothing needed there.

## Test evidence

Five new tests, each **red on `main`, green with the fix** (verified by stashing only `session.rs`):

- `torn_tail_does_not_eat_next_appended_message` — manager append path
- `torn_tail_does_not_eat_handle_appended_message` — SessionHandle append path
- `torn_tail_does_not_eat_rollback_marker` — flat layout
- `torn_tail_does_not_eat_rollback_marker_in_per_user_layout` — per-user (production-canonical) layout
- `torn_tail_complete_row_is_recovered` — torn bytes that are a complete row missing only its terminator are preserved and recovered

`cargo test -p octos-bus`: 247+16+2 passed, 0 failed. `cargo fmt --all -- --check` clean. `cargo clippy -p octos-bus --all-targets -- -D warnings` clean.

**End-to-end:** a throwaway example binary drove the real public API (`SessionManager::open` → `add_message` → simulated crash mid-write → restart → `add_message` → cold reload). Pre-fix: reload yields `["before torn"]` (the post-crash message lost). Post-fix: `["before torn", "after torn"]`. The example was removed after the run and is not in this diff.

**Independent review:** an adversarial maintainer-perspective pass found no blockers/majors; its minor findings are addressed here (per-user rollback test, complete-row test, tmp cleanup on `sync_all` failure, seal docstring scoping note, one stale comment). The `.read(true)` access-mode widening matches the supervisor-store precedent and was judged acceptable.

Closes #2006

### DIFF
--- a/crates/octos-bus/src/session.rs
+++ b/crates/octos-bus/src/session.rs
@@ -17,7 +17,7 @@ const CURRENT_SESSION_SCHEMA: u32 = 1;
 
 /// Observer callback invoked AFTER a successful durable commit by
 /// [`SessionManager::add_message_with_seq`] (and the equivalent
-/// `SessionHandle` path). Implements the post-fsync hook UPCR-2026-012's
+/// `SessionHandle` path). Implements the post-commit hook UPCR-2026-012's
 /// `message/persisted` notification dispatches through.
 ///
 /// Strict-ordering invariant: `add_message_with_seq` calls observers
@@ -149,6 +149,32 @@ fn rewrite_tmp_path(target: &Path) -> PathBuf {
     let pid = std::process::id();
     let seq = REWRITE_TMP_COUNTER.fetch_add(1, Ordering::Relaxed);
     target.with_extension(format!("jsonl.{pid}-{seq}.tmp"))
+}
+
+/// Seal a torn tail before appending: if the file is non-empty and does
+/// not end in a newline (a crash split an earlier write), write the
+/// missing terminator first so the next row can never fuse with the torn
+/// bytes into one unparseable line. The torn bytes are preserved, never
+/// truncated — they may be a complete row that merely lost its
+/// terminator; the read path skips what it cannot parse. Mirrors
+/// `supervisor_store::write_rows_sealed_locked`.
+///
+/// Callers serialize appends per session key (the persist lock), like the
+/// supervisor store's append lock; the seal does not defend against
+/// unsynchronized cross-process writers to the same file.
+///
+/// O_APPEND: the sealing write lands at EOF regardless of the read seek.
+fn seal_torn_tail(file: &mut std::fs::File, file_len: u64) -> std::io::Result<()> {
+    use std::io::{Read, Seek, SeekFrom, Write};
+    if file_len > 0 {
+        file.seek(SeekFrom::Start(file_len - 1))?;
+        let mut last = [0_u8; 1];
+        file.read_exact(&mut last)?;
+        if last[0] != b'\n' {
+            file.write_all(b"\n")?;
+        }
+    }
+    Ok(())
 }
 
 /// FNV-1a 64-bit hash — deterministic across Rust versions (unlike DefaultHasher).
@@ -1396,12 +1422,13 @@ impl SessionManager {
         session.updated_at = Utc::now();
         record_session_persist("committed");
         let committed_seq = session.messages.len().saturating_sub(1);
-        // UPCR-2026-012: post-fsync observer fan-out. Fires AFTER the
+        // UPCR-2026-012: post-commit observer fan-out. Fires AFTER the
         // append_to_disk above succeeded and the in-memory mirror is
         // updated, so a `message/persisted` notification reflects a row
-        // that is durably visible. A failed disk write returns above
-        // before this point, so the observer never sees a row that did
-        // not commit.
+        // the OS has accepted (the append write returned; per-append
+        // fsync is deliberately not done). A failed disk write returns
+        // above before this point, so the observer never sees a row
+        // that did not commit.
         if let Some(committed) = session.messages.last() {
             notify_message_commit(&observer_root, key, committed, committed_seq);
         }
@@ -1876,6 +1903,7 @@ impl SessionManager {
             use std::io::Write;
 
             let mut file = std::fs::OpenOptions::new()
+                .read(true)
                 .create(true)
                 .append(true)
                 .open(&path)?;
@@ -1921,6 +1949,8 @@ impl SessionManager {
                     updated_at: Utc::now(),
                 };
                 writeln!(file, "{}", serde_json::to_string(&meta)?)?;
+            } else {
+                seal_torn_tail(&mut file, file_len)?;
             }
 
             writeln!(file, "{msg_json}")?;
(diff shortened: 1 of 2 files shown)


anantheparty (approved):
Adversarial pre-merge review passed - the #2273 nit-4 convention implemented, and better than prescribed: seal_torn_tail writes the missing terminator while PRESERVING the torn bytes (forensics-friendly, rather than discarding the partial line), the session rewrite fsyncs the temp file before rename and the parent dir after (durable, not just atomic), and five tests pin every fusion variant (next message / rollback marker / handle append / complete-row recovery / per-user layout). The O_APPEND EOF-landing caveat is documented. CI green. Approving - merging left to you per convention.

PR #2290, opened 2026-09-10
fix(services): frame the services compaction summary as demoted background (#1912)

### LINKED ISSUE #1912: compaction: octos-services summary channel lacks the background/instruction-priority framing
Follow-up from #1892 (review item 3).

#1892 fixed the instruction-priority drift on two of the three compaction channels (spec `task-compaction-instruction-priority`):

- `crates/octos-agent/src/compaction.rs` — `LLM_COMPACTION_SYSTEM_PROMPT` now demotes summarized history to BACKGROUND and anchors the current task to the newest user message.
- `crates/octos-cli/src/api/context_manager.rs` — the rendered `[Conversation summary]` row gets an explicit `BACKGROUND ONLY` header + precedence footer.

The third channel — `crates/octos-services/src/compaction.rs` (~:155 session-manager path and ~:286 session-handle path) — still inserts the summary bare:

```rust
content: format!("[Conversation summary]\n{summary}"),
```

with no background framing and no precedence footer, so a post-compaction model on this channel can still mistake restated historical goals/plans for the current instruction (the live drift observed 2026-08-02).

**Suggested fix:** apply the same render-side framing as context_manager.rs (keep the exact `[Conversation summary]` first-line sentinel — `message_repair` matches on it), or better, extract one shared `format_compaction_summary(summary: &str) -> String` helper so all three channels render identically and can't drift apart again.

BH3GEI:
## Problem

#1892 fixed the instruction-priority drift on two of the three compaction channels (spec `task-compaction-instruction-priority`), but the third — `crates/octos-services/src/compaction.rs` (both the `SessionManager` path and the `SessionHandle` path) — still installed the summary bare:

```rust
content: format!("[Conversation summary]\n{summary}"),
```

A post-compaction model on this channel could still mistake restated historical goals/plans for the current instruction (the live drift observed 2026-08-02).

## Root cause

The services channel never got the render-side framing the other two channels received: no BACKGROUND demotion header, no precedence footer.

## Fix

- New `format_compaction_summary` helper in `octos-services/src/compaction.rs` renders the exact same framing as `octos-cli/src/api/context_manager.rs` (verified byte-identical, accounting for Rust string-continuation whitespace): `[BACKGROUND ONLY — …]` header + `[End of background. The newest user message … takes precedence …]` footer.
- Both call sites (`maybe_compact_with_config`, `maybe_compact_handle`) route through the helper.
- Line 1 keeps the exact legacy `[Conversation summary]` sentinel, so `message_repair`'s `starts_with` match (and every other sentinel consumer) is unaffected — verified by reading all match sites (`octos-agent/src/agent/message_repair.rs:121`, context_manager projections, session_actor tests).
- Cross-reference comments in both copies so the duplicated literal cannot silently drift (a single shared helper would have to live in a lower crate; octos-services does not depend on octos-cli).

## Test evidence

- Two new behavior tests drive the real public entry points (stub summarizer, real disk-backed `SessionManager`/`SessionHandle`): red before the fix (installed content was the bare two-line form), green after. They assert exact full-string equality with the reference framing, verbatim summary-body preservation, and the retained recent tail.
- A third test pins the empty-summary edge.
- `cargo test -p octos-services`: 55/55. `cargo test -p octos-cli --no-default-features --features api --lib context_manager`: 95/95. `cargo fmt --all -- --check` clean. Clippy clean for `-p octos-services --all-targets` and the CI api-feature gate (`-p octos-cli --all-targets --no-default-features --features api,telegram,discord,dingtalk,slack,whatsapp,feishu,email,matrix,audio_mp3`), both with `-D warnings`.
- End-to-end with a real LLM (DeepSeek `deepseek-chat`, 46-message session through `maybe_compact_handle`): before the fix the installed row was the bare `[Conversation summary]\n<body>`; after, it carries the BACKGROUND header + precedence footer with the body verbatim, and the newest user instruction survives as the last row.
- Independent adversarial review: no blockers/majors; its Minor findings (drift-prone duplication → cross-reference comments; substring assertions → exact-match; empty-body edge) are all addressed in this revision.

Note: this channel has no in-workspace production callers (it is re-exported via `octos-cli` for downstream consumers), so the new tests are the in-repo exercise of the path.

Closes #1912

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4961,6 +4961,7 @@ dependencies = [
 name = "octos-services"
 version = "2.0.3-rc.11"
 dependencies = [
+ "async-trait",
  "chrono",
  "dirs",
  "eyre",
--- a/crates/octos-cli/src/api/context_manager.rs
+++ b/crates/octos-cli/src/api/context_manager.rs
@@ -3185,6 +3185,9 @@ impl ContextManager {
                             // agent loop's message_repair matches
                             // `starts_with("[Conversation summary]")` to keep
                             // summary rows out of the system prompt.
+                            // Keep byte-identical to the sibling copy in
+                            // `octos-services/src/compaction.rs`
+                            // (`format_compaction_summary`).
                             format!(
                                 "[Conversation summary]\n\
                                  [BACKGROUND ONLY — everything in this summary is \
--- a/crates/octos-services/Cargo.toml
+++ b/crates/octos-services/Cargo.toml
@@ -30,6 +30,7 @@ futures = { workspace = true }
 test-util = []
 
 [dev-dependencies]
+async-trait = { workspace = true }
 tempfile = { workspace = true }
 
 [lints]
--- a/crates/octos-services/src/compaction.rs
+++ b/crates/octos-services/src/compaction.rs
@@ -13,6 +13,29 @@ const DEFAULT_THRESHOLD: usize = 40;
 /// Default number of recent messages to keep intact (not summarized).
 const DEFAULT_KEEP_RECENT: usize = 10;
 
+/// Render a compaction summary for installation into the live message list.
+///
+/// Spec task-compaction-instruction-priority: frame the summary as demoted
+/// BACKGROUND with an explicit precedence footer, so a post-compaction model
+/// cannot mistake restated historical goals/plans for the current
+/// instruction. Line 1 keeps the EXACT legacy sentinel: the agent loop's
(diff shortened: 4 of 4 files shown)


anantheparty (approved):
Adversarial pre-merge review passed - completes the #1892 instruction-priority family: the third compaction channel (octos-services, both SessionManager and SessionHandle paths) now frames the summary as demoted background, matching the spec contract the first two channels established. The drift family is fully closed across all three channels. CI green. Approving - merging left to you per convention.

Issue #2263, opened 2026-09-05
octos-cli api tests do not compile on Windows/macOS: keychain::test_override_secrets_root is linux-only


ZhangHanDong:
## Symptom

`cargo test -p octos-cli --features api --lib` fails to compile on non-Linux targets:

```
error[E0425]: cannot find function `test_override_secrets_root` in module `crate::auth::keychain`
  --> crates/octos-cli/src/api/admin.rs:5589:36
  --> crates/octos-cli/src/api/auth_handlers.rs:6038:36
note: found an item that was configured out
  --> crates/octos-cli/src/auth/keychain.rs:371:55
```

Seen on #2257's `check-windows` job once its new gate tests ran the api-feature lib tests on Windows (https://github.com/octos-org/octos/actions/runs/33968563613/job/101312921120).

## Cause

#2239 (b5db3121) exports the test hook as `#[cfg(all(test, target_os = "linux"))] pub(crate) use linux_file::override_root_for_tests as test_override_secrets_root;` with the comment "no-op elsewhere", but no non-Linux definition exists, while the two api test modules call it unconditionally. Required CI never compiles the api-feature tests off Linux, so it merged green.

## Fix

Provide the documented no-op on non-Linux test builds:

```rust
#[cfg(all(test, not(target_os = "linux")))]
pub(crate) fn test_override_secrets_root(_root: std::path::PathBuf) {}
```

PR follows. Found by the OctoLoop evolution loop (octoscode LEP-003) via #2257's feature-carrying gate tests.

BH3GEI:
Verified fixed on main (c9abaa54): the exact symptom command from this report, `cargo test -p octos-cli --features api --lib`, now compiles clean on macOS (Apple Silicon, rustc 1.96.0).

The fix landed in d51601df (#2265) rather than via the no-op shim proposed here: all three call sites of `keychain::test_override_secrets_root` in `api/admin.rs` and `api/auth_handlers.rs` are now individually `#[cfg(target_os = "linux")]`-gated, so the linux-only export is never referenced on other targets. PR #2264 (the no-op variant) was closed unmerged and is superseded.

Regression coverage: #2291's `check-darwin` job runs `cargo check -p octos-cli --all-targets --features api,...` on macos-14 for every PR and passed on exactly this code, so this class of break is now gated in CI.

Issue #2258, opened 2026-09-04
main fails to compile on macOS: keychain.rs lost the WrapErr import in #2239


ZhangHanDong:
## Symptom

Since e64b531f (merge of #2239), `octos-cli` does not compile on macOS:

```
error[E0599]: no method named `wrap_err` found for enum `Result<T, E>` in the current scope
   --> crates/octos-cli/src/auth/keychain.rs:384:10
   --> crates/octos-cli/src/auth/keychain.rs:428:10
```

Seen on PR #2257's `e2e-tool-regression` job (macos-14, "Build octos with all features"): https://github.com/octos-org/octos/actions/runs/33912697008/job/101157812996. That PR does not touch `keychain.rs`; the same job passed on PRs based before #2239.

## Cause

b5db3121 (#2239, 45a "平台中立 secret store") changed `use eyre::{Result, WrapErr};` to `use eyre::Result;` because the trait was unused on Linux after the refactor, but `macos_unlock` / `macos_set_secret` (`#[cfg(target_os = "macos")]`) still call `.wrap_err(...)`.

## Fix

Restore the import for the macOS-gated code (`#[cfg(target_os = "macos")] use eyre::WrapErr;`). PR follows.

## Why CI missed it

None of the required checks compile the macOS target; the only macOS build (`e2e-tool-regression`) runs conditionally on `e2e-changes`. Tracked separately as a CI-matrix issue.

Found by the OctoLoop evolution loop (octoscode LEP-003) while verifying #2257.

PR #2291, opened 2026-09-10
ci: add check-darwin so required checks compile the macOS target

### LINKED ISSUE #2259: CI: required checks never compile the macOS target (blind spot that let #2239 break main on macOS)

BH3GEI:
## Problem

None of the required checks on `main` compile for macOS. Every PR job runs on Linux or Windows; the only two macOS jobs are conditional — `e2e-tool-regression` (macos-14) fires only on e2e path changes, and the `security` suite compiles a single `octos-agent` test binary only on security-path changes. An ordinary PR gets no macOS compile signal at all, so a `cfg(target_os = "macos")`-only compile error sails through green.

That blind spot already burned main twice from a single PR: #2239 broke the macOS default build (#2258, E0599) and the api-feature test code (#2263, E0425) while every required check stayed green.

## Root cause

The workspace has 89 `cfg(target_os = "macos")` sites that the Linux runners every other PR job uses can never compile. Compile coverage for macOS existed only behind path filters, never as a gate.

## Fix

Add a compile-only `check-darwin` job (macos-14) to the PR matrix and wire it into the `ci-required` aggregate (both the `needs:` list and the required-jobs loop):

- `cargo check --workspace --all-targets`
- `cargo check -p octos-cli --all-targets --features ${{ env.FEATURES }}`

Both steps pass `--all-targets` so `cfg(test)` code is compiled too — #2263's break lived in feature-gated test modules, which a lib-only check would never see. Compile-only (no test execution), so the cost is one macos-14 runner for the duration of a cargo check.

## Test evidence

Verified on an Apple Silicon Mac (macOS 26.5, cargo 1.96.0) against `origin/main` (11db2902):

- `cargo check --workspace --all-targets` — **green** (1m43s)
- `cargo check -p octos-cli --all-targets --features "api,telegram,discord,dingtalk,whatsapp,feishu,twilio,wecom,wecom-bot,audio_mp3"` — **green** (1m42s). This also confirms the #2263 call sites have since been cfg-gated on main, so the full feature + test-target combination compiles today.
- **End-to-end negative proof**: temporarily reverting the #2258 fix (dropping the macOS-gated `WrapErr` import in `crates/octos-cli/src/auth/keychain.rs`) makes the gate's command fail with the exact E0599 signature from #2258 at keychain.rs:384/428; restoring it returns to green. The gate catches the failure class it exists for.
- `actionlint` on the modified workflow: no new findings (the single `ubuntu-24.04-riscv` label warning pre-exists on main and is the intentional manual-only job).
- YAML parse verified; `ci-required` fails closed on a skipped/cancelled `check-darwin` (result != `success`), so the new job cannot silently report green without running — the failure mode this repo has hit before.

The job's own execution on this PR is the final end-to-end check: `check-darwin` should appear and go green below.

## Review

Independent adversarial review (given only the issue, the diff, and the repo paths) returned REQUEST_CHANGES with one Major and two Minors; all addressed:

- **Major**: the feature step originally omitted `--all-targets`, leaving the #2263 failure class ungated. Local verification showed the combination compiles on main today, so `--all-targets` was added to step 2 — the class is now covered outright, no deferred follow-up.
- **Minor**: corrected an over-broad claim about `e2e-tool-regression`'s feature set (it builds api+telegram, not all features).
- **Minor**: clarified that #2239 produced both breaks from one merge.

Note for the operator: `ci-required` already aggregates this job, so no ruleset change is needed for it to gate — the existing branch-protection status simply turns red if `check-darwin` fails.

Closes #2259

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -833,6 +833,41 @@ jobs:
         if: ${{ github.event_name != 'pull_request' && !cancelled() }}
         run: cargo test --workspace --doc
 
+  check-darwin:
+    # Runs on PRs, like check-windows: the workspace has 89
+    # `cfg(target_os = "macos")` sites that NEVER compile on the Linux runners
+    # every other PR job uses, so a macOS-only break was invisible until after
+    # merge — #2239 alone produced two such breaks (#2258's E0599 in the
+    # default build, #2263's E0425 in api-feature test code) while every
+    # required check stayed green. The existing macOS jobs do not close this:
+    # e2e-tool-regression builds octos-cli (api+telegram features) but fires
+    # only on e2e path changes, and the security suite compiles one
+    # octos-agent test binary only on security-path changes — so an ordinary
+    # PR gets no macOS compile signal at all.
+    #
+    # Compile-only, so the cost is one macos-14 runner for a cargo check, not
+    # a test run. Both steps pass --all-targets so cfg(test) code is compiled
+    # too — #2263's break lived in feature-gated test modules, which a
+    # lib-only check would never see.
+    runs-on: macos-14
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: darwin
+
+      - name: Check workspace (all targets)
+        run: cargo check --workspace --all-targets
+
+      - name: Check octos-cli + all features (all targets)
+        run: cargo check -p octos-cli --all-targets --features ${{ env.FEATURES }}
+
   # ── Release artifact builds (main only) ──────────────────────────
   build-macos-arm:
     if: github.event_name == 'push'
@@ -1254,6 +1289,7 @@ jobs:
       - test-octos-cli
       - check-matrix
       - check-windows
+      - check-darwin
       - e2e-changes
       - e2e-tool-regression
       - security
@@ -1270,7 +1306,7 @@ jobs:
           for job in dashboard swarm-app typos ux-gate-report \
             windows-deploy-script web-app-base-url-windows check \
             test-octos-agent test-octos-cli check-matrix check-windows \
-            e2e-changes security; do
+            check-darwin e2e-changes security; do
             result=$(echo "$RESULTS" | jq -r --arg job "$job" '.[$job].result')
             if [ "$result" != success ]; then
               echo "::error::required job $job finished with $result"


anantheparty (changes requested):
Request changes on one line — **the new job is not wired into the `CI required` aggregate.**

The check-darwin design itself is right (compile-only on macos-14, `--all-targets` so cfg(test) code compiles, the feature-gated octos-cli check; no test-execution cost). But the required-checks ruleset enforces exactly one context — `CI required` — and the diff does not add check-darwin to the aggregate's `needs`. Without that line, a darwin compile failure reports a red check-darwin while `CI required` stays green and the merge proceeds — the job is decorative and the #2258/#2263 escape class continues.

**The fix**: add `check-darwin` to the aggregate job's `needs` (alongside fmt/test/clippy/check-windows). One line. Then this does what the PR title promises.

BH3GEI:
Thanks for the careful review — the aggregate wiring matters exactly as you describe, so I double-checked the diff at 43dc2316.

I'd respectfully disagree that it is missing: the commit already adds `check-darwin` to both halves of the `ci-required` aggregate:

- the `needs:` list: https://github.com/octos-org/octos/blob/43dc2316329cd55a25d7d25b0c32a219d14b9deb/.github/workflows/ci.yml#L1292
- the shell gate loop (`for job in ... check-darwin e2e-changes security; do`): https://github.com/octos-org/octos/blob/43dc2316329cd55a25d7d25b0c32a219d14b9deb/.github/workflows/ci.yml#L1309

Both lines appear in the PR diff under `.github/workflows/ci.yml` (hunks `@@ -1254` and `@@ -1270`). With those in place, a red check-darwin makes `CI required` fail. Possibly the review was taken against a stale render of the diff — happy to restructure if you see it differently.

anantheparty (approved):
Re-review after the update: both wiring points from my change request are in —

- `check-darwin` added to the `CI required` aggregate `needs`
- `check-darwin` added to the aggregate verification script's job list (beyond what I asked — the jq loop would have let it pass unverified)

The lane itself is green on this PR and the compile-only/--all-targets design is unchanged. Good to merge from my side; leaving the merge to you per convention.

PR #2293, opened 2026-09-10
fix(matrix): stay silent when a group message mentions another user

### LINKED ISSUE #1547: Bots answer group-room messages that explicitly @-mention a different bot
## Summary

In a room with multiple octos-family bots, a human message that explicitly @-mentions ONE specific bot is answered by ALL of them. Reply gating in both channels is sender/room-scoped only and never consults the message's mention set.

Missing rule: **in a group room, if a message contains explicit mentions and none of them is me, stay silent.**

Related but distinct from #1546: that issue covers **bot-triggered** replies (fixed by `m.notice` msgtype marking). This one covers **human-triggered** replies, which remain `m.text` even after #1546 — mention-aware suppression is an independent mechanism, and both are needed for usable multi-agent rooms.

## Observed

- bob sends `@myagent 你是谁` (proper Matrix mention pill targeting the user-channel agent)
- `myagent` replies — correct (mentioned)
- the appservice bot `@octos` ALSO replies with its own self-introduction — it was not mentioned
- Mirrored case exists too: `@octos <question>` draws an unsolicited reply from `myagent` when `require_mention: false`
- Cost: every addressed message in an N-bot room triggers N LLM calls (observed 77.5K + 90.1K input tokens for one question)

## Root cause

- `matrix_channel.rs` (appservice): replies to any human `m.text` in bound rooms; `allowed_senders` maps to `admin_allowed_senders` (`:618-624`) and gates admin commands only — **no reply-gating config exists at all**
- `matrix_user_channel.rs`: gates on `room_allowed` (`:840`), `sender_allowed` (`:848`), `require_mention` (`:932`) — `require_mention` checks "was I mentioned", but there is no "was someone ELSE explicitly mentioned" check
- Client-side explicit targeting only routes within the appservice's own child-bot namespace; independent user-channel agents are invisible to the appservice bot's reply decision

## Proposed rule

- [ ] Mention-aware suppression, default ON in group rooms, implemented in BOTH channels: if inbound `m.mentions` / mention pills exist and none targets this bot's MXID → do not reply
- [ ] DM rooms exempt (1:1 keeps answering everything)
- [ ] Unaddressed messages unchanged (room-level policy decides)
- [ ] Config escape hatch (e.g. `mention_policy: "strict" | "open"`) for rooms where a bot should answer regardless

## Environment

octos @ cd97d127, palpo homeserver, appservice bot + matrix user channel agent in one room (invited via Robrix2 `/invitebot`), DeepSeek backend. Observed 2026-07-08.

BH3GEI:
## Problem

In a room with several octos-family bots, a human message that explicitly @-mentions ONE bot was answered by every user-account channel running with `require_mention: false` — reply gating was sender/room-scoped only and never consulted the message's mention set (#1547; observed cost: 77.5K + 90.1K input tokens for a single question). The appservice channel already covers this via its `mention_only` gate (default on); the user-account channel had no equivalent.

## Root cause

`matrix_user_channel.rs` parsed only `mentioned_self`; the forward gate consulted `require_mention` and the allowlists, nothing else. With `require_mention: false` every text message in an allowed room was forwarded to the agent, including messages explicitly addressed to someone else.

## Fix

Mention-aware suppression in the user channel (`crates/octos-bus/src/matrix_user_channel.rs`):

- Parsing now records `mentions_other`: the message carries explicit mentions (structured `m.mentions`, a matrix.to pill in `formatted_body`, or a hand-typed `@user:server`) and at least one targets a user other than this account. Rich-reply context is stripped first — the replied-to sender (auto-included in `m.mentions` by clients), the `<mx-reply>` pill block, and the `> `-quoted body fallback — so a reply aimed at the bot by name is still answered.
- The forward gate drops such messages in group rooms, even with `require_mention: false`. Slash commands and messages that also mention this account are unaffected, as are unaddressed messages (room-level policy still decides those).
- DM exemption: a room counts as a 1:1 when listed in the account's `m.direct` account data (tracked from sync, kept fresh by incremental syncs) **or** when a cached `joined_members` probe finds at most one other joined user — the probe is what makes the exemption work in practice, since nobody ever writes the bot's own `m.direct`. Probe results are cached 60s (failures 10s); on probe failure the gate falls back to the pre-#1547 behaviour (answer) rather than extending its reach on a homeserver error.
- Escape hatch: new channel setting `mention_policy` (`strict` default / `open` restores the old behaviour), with camelCase `mentionPolicy` accepted like the neighbouring keys. Documented in book/book-zh channels.md.

## Testing

- Reproduction first: a test driving the real HTTP sync + forward path against a local homeserver stub (issue scenario: `require_mention: false`, group room, `@otherbot:example.org 你是谁` with `m.mentions`) **failed on unmodified main** (message forwarded) and passes with the fix.
- 28 `matrix_user_channel` tests cover: suppression in a group room; exemption via `m.direct` (pinned so only account data can exempt); exemption via membership probe (no `m.direct`); fail-open on probe error; `mention_policy: open`; mention of self+other; unaddressed group messages still forwarded; `require_mention: true` ordering; `m.direct` parse/preserve/clear; reply shapes (auto-included sender ignored, extra mentions count, pill after `</mx-reply>` counts, typed MXID after fallback counts); MXID edge shapes (`+` localpart, case-insensitivity, `:port`, trailing dot, email non-mentions). Settings plumbing test (default strict / `open` / camelCase) added in `gateway/mod.rs`.
- Full CI-equivalent run, all green on the first pass: `cargo fmt --all -- --check`; `cargo test -p octos-bus` (247) and `--features matrix` (417+16+2+10); `cargo clippy --workspace --all-targets -- -D warnings` (0 warnings); `cargo clippy -p octos-cli --all-targets --no-default-features --features api,telegram,discord,dingtalk,slack,whatsapp,feishu,email,matrix,audio_mp3 -- -D warnings` (0 warnings); `cargo test -p octos-cli --features api --lib` (3678 passed, 0 failed).

## End-to-end evidence

A live two-bot Matrix deployment was not available here, so the end-to-end demonstration drives the same code path over real HTTP: a local axum homeserver stub serves `/sync` (timeline event + `m.direct` account data) and `/joined_members`, and the channel's own reqwest client runs sync → parse → gate → bus-forward. Before the fix the issue scenario's message reached the bus; after the fix it is dropped in a 3-member room, and still delivered in a 1:1 (via `m.direct` or the probe). The reviewed trade-off: on homeservers where the probe always fails, suppression never engages (pre-#1547 behaviour) — noted as a deliberate fail-open.

## Review

Two rounds of independent adversarial review: round 1 returned REQUEST_CHANGES (DM exemption relied on `m.direct` only — dead in real deployments; rich replies to another person were silenced; plus three minors) — all fixed with tests; round 2 APPROVE, and its five minor polish items (probe ordering after `require_mention`, replied-to exclusion in the text path, case-insensitive replied-to compare, bare `>` fallback lines, negative probe caching) are all incorporated. Known deferred gap: percent-encoded matrix.to identifiers (`matrix.to/#/%40user%3Aserver`) are not detected.

Closes #1547

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4628,6 +4628,7 @@ dependencies = [
  "tokio-tungstenite 0.26.2",
  "tower",
  "tracing",
+ "tracing-subscriber",
  "uuid",
  "webpki-roots 0.26.11",
 ]
--- a/book-zh/src/channels.md
+++ b/book-zh/src/channels.md
@@ -419,7 +419,7 @@ Matrix 是一等公民频道，也是本章多处提到的「人工审批」与�
 }
 ```
 
-使用 `matrix` 特性标志编译。在 **appservice / 管理机器人**房间中，Matrix 会为[人工审批规则](./configuration.md)渲染原生的 批准/拒绝 卡片（通过 Robrix），并处理 `/schedule`、`/schedules`、`/unschedule`、`/allbots` 聊天命令（见下文[定时任务](#定时任务)）。而普通的 `mode: "user"` 账户渠道本身并不解释这些管理命令；它会把消息文本转发给 agent，但**默认仅在机器人被 @ 提及时**转发（`require_mention` 默认为 `true`——在其 settings 中设置 `"require_mention": false` 可转发每一条消息）。
+使用 `matrix` 特性标志编译。在 **appservice / 管理机器人**房间中，Matrix 会为[人工审批规则](./configuration.md)渲染原生的 批准/拒绝 卡片（通过 Robrix），并处理 `/schedule`、`/schedules`、`/unschedule`、`/allbots` 聊天命令（见下文[定时任务](#定时任务)）。而普通的 `mode: "user"` 账户渠道本身并不解释这些管理命令；它会把消息文本转发给 agent，但**默认仅在机器人被 @ 提及时**转发（`require_mention` 默认为 `true`——在其 settings 中设置 `"require_mention": false` 可转发每一条消息）。即使关闭了 `require_mention`，群房里显式提及他人的消息（`m.mentions` 条目、matrix.to pill、或手打的 `@user:server`）也会被视为发给对方而不予回复，这样多个机器人共处一室时不会对每条指明对象的消息全体抢答；私聊（DM）不受此限，设置 `"mention_policy": "open"` 可恢复"不论提及谁都回复"的旧行为。
 
 ---
 
--- a/book/src/channels.md
+++ b/book/src/channels.md
@@ -419,7 +419,7 @@ Log in with an `access_token` (preferred), or with `user_id` + `password` (both 
 }
 ```
 
-Build with the `matrix` feature flag. In **appservice / management-bot** rooms, Matrix renders native Approve/Deny cards for [Human Approval Rules](./configuration.md) (via Robrix) and handles the `/schedule`, `/schedules`, `/unschedule`, and `/allbots` chat commands (see [Cron Jobs](#cron-jobs) below). A plain `mode: "user"` account channel does not interpret those management commands; it forwards message text to the agent, but **by default only when the bot is mentioned** (`require_mention` defaults to `true` — set `"require_mention": false` in its settings to forward every message).
+Build with the `matrix` feature flag. In **appservice / management-bot** rooms, Matrix renders native Approve/Deny cards for [Human Approval Rules](./configuration.md) (via Robrix) and handles the `/schedule`, `/schedules`, `/unschedule`, and `/allbots` chat commands (see [Cron Jobs](#cron-jobs) below). A plain `mode: "user"` account channel does not interpret those management commands; it forwards message text to the agent, but **by default only when the bot is mentioned** (`require_mention` defaults to `true` — set `"require_mention": false` in its settings to forward every message). Even with `require_mention` off, a group-room message that explicitly mentions someone else (an `m.mentions` entry, a matrix.to pill, or a hand-typed `@user:server`) is treated as directed at them and stays unanswered, so several bots can share a room without all replying to every addressed message; direct chats are exempt, and `"mention_policy": "open"` restores answering regardless of mentions.
 
 ---
 
--- a/crates/octos-bus/Cargo.toml
+++ b/crates/octos-bus/Cargo.toml
@@ -71,3 +71,4 @@ tokio-test = { workspace = true }
 tempfile = { workspace = true }
 tower = { workspace = true }
 proptest = { workspace = true }
+tracing-subscriber = { workspace = true }
--- a/crates/octos-bus/src/lib.rs
+++ b/crates/octos-bus/src/lib.rs
@@ -88,7 +88,8 @@ pub use matrix_channel::{
 };
 #[cfg(feature = "matrix")]
 pub use matrix_user_channel::{
-    MatrixAutoJoin, MatrixGroupPolicy, MatrixInviteStore, MatrixPendingInvite, MatrixUserChannel,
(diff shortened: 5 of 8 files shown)


anantheparty (changes requested):
Request changes on two small items — the design and semantics are sound (three-path mention extraction, gating order, DM triple-immunity, the policy escape hatch), but the PR **introduces a new hang surface** on the hot path:

1. **The DM probe has no timeout** (probe_room_is_dm reuses matrix_http_client(), which has none): the forward hot path now makes a synchronous HTTP call per candidate message — a hung homeserver blocks the entire forward_messages loop (sync has the server's 30s backstop; the probe has nothing on either side). **Fix: a dedicated ~5s timeout on the probe request.** ~2 lines.

2. **First-suppression observability**: the default flip is a silently-breaking change for existing require_mention:false deployments — a suppression that used to answer now goes silent with only a debug log. **Fix: log the first suppression per room at info** (with the policy and the mention set), so operators discover the behavior change from their logs instead of from silence.

Nit: mentioned_self vs mentions_other parse asymmetry (legacy clients without m.mentions + display-name bodies → the mentioned bot gets suppressed while others answer — contradicts "contains self → respond"; low probability, follow-up fine). Unknown policy values silently → Strict: a one-time warn would catch typos. @room unrecognized: pre-existing, defensible.

Everything else verified: three-path extraction (m.mentions / pills with loop-advance / hand-typed MXID with port/trailing-dot/email exclusions), the gating order (room → sender → require_mention → new gate → dedup, dropped messages never probe), DM triple-immunity (m.direct / 60s-cached joined_members probe / fail-open on probe failure — 4 e2e tests), the config matrix including the camelCase escape hatch, and 142/142 matrix tests passing (the Windows-flaky :923 untouched).

BH3GEI:
Thanks for the careful review — both items are addressed in 70105525:

1. **Probe timeout**: `probe_room_is_dm` now runs the whole exchange (connect → request → headers → body read) under a single 5s deadline (`DM_PROBE_TIMEOUT`). A hung homeserver fails open like any other probe failure instead of stalling `forward_messages`. Pinned by a paused-clock e2e test whose stub accepts the probe and never answers: it fails open at exactly the 5s mark (and hangs forever without the fix — verified by reverting).

2. **First-suppression observability**: the first suppression per room is logged once at info with the policy and the mention set; repeats stay at debug. To log the mention set, `content_mentions_other_user`/`text_mentions_other_user` now return the mentioned users (`Vec<String>`) instead of a bare bool — same three extraction paths and reply-context stripping, verified by the existing test matrix plus a log-capture test asserting exactly one info notice per room.

Also picked up the nit: unrecognized `mention_policy` values now warn at startup (`unrecognized matrix mention_policy; falling back to strict`), with a settings test covering typo → warn + Strict fallback and legitimate values staying silent. The mentioned_self/mentions_other parse asymmetry is left for a follow-up as you suggested.

Verification: octos-bus 419 + CLI matrix suites green, `cargo fmt --check`, and both CI clippy variants (minimal lib and `--features …,matrix` with `-D warnings`) clean.

anantheparty (approved):
Both requested items landed and verified: the DM probe now carries a dedicated 5s timeout (DM_PROBE_TIMEOUT - no more unbounded homeserver hang on the forward hot path), and the first mention-gate suppression per room announces at info with the policy and mention set (operators discover the behavior change from logs, not silence). The three-path extraction, gating order, DM triple-immunity, and 142/142 matrix verdicts stand. Approving - merging left to you per convention.

anantheparty (approved):
Fix round verified on the updated head: the 5s DM-probe timeout and the first-suppression info-log both landed as requested. Approving - merging left to you per convention.

PR #2294, opened 2026-09-10
test(pipeline): widen the codergen worker-timeout promptness ceiling to 5s (#1932)

### LINKED ISSUE #1932: codergen_worker_timeout promptness assert (elapsed < 1750ms) is load-flaky on Windows
`codergen_timeout_secs_cancels_stalled_worker` (crates/octos-pipeline/tests/codergen_worker_timeout.rs) asserts the 1s worker timeout trips within 750ms of margin:

```rust
assert!(
    elapsed < Duration::from_millis(1_750),
    "worker timeout should trip promptly, got {elapsed:?}"
);
```

## Evidence it flakes under load, on real hardware

Windows Server 2019 (AMD64), same box, same commits:

| run | result |
|---|---|
| full `cargo test -p octos-pipeline` (#1931 branch) | **FAILED** at the promptness assert |
| same test **solo**, same commit, ×3 | ok — 1.62s / 1.63s / 1.73s |
| same test solo on `main` | ok — 1.64s |

Identical code passes solo at identical timing and fails only when the rest of the suite loads the machine. Note 1.73s is already within 20ms of the 1750ms ceiling **solo** — on a loaded runner the margin is effectively zero.

It has already cost one red `check-windows` (when the #1931 commits sat on the reaper branch), and it will recur on slow or shared runners.

## Suggested fix

Keep the functional assertions (outcome is `Error`, message contains "timed out after 1s") and either:

- widen the promptness ceiling substantially (e.g. 5s — still proves the timeout is not wedged, which is the regression the test exists to catch), or
- measure promptness relative to a same-process baseline rather than wall-clock absolute.

A timing assert this tight in an integration test is a check that fails on machine weather rather than on code.

BH3GEI:
## Problem

`codergen_timeout_secs_cancels_stalled_worker` asserts the 1s worker timeout trips within a 750ms wall-clock margin (`elapsed < 1_750ms`). #1932 documents it flaking under load on real hardware: on the Windows Server runner the identical test passes **solo** at 1.62s / 1.63s / 1.73s — fixed per-run overhead of 620–730ms on top of the 1s timeout — so the ceiling has a ~20ms effective margin that any suite-level load consumes. It already cost one red `check-windows`.

## Root cause

The assert measures wall-clock elapsed of the whole `handler.execute` call — provider resolution, tool-registry construction, prompt assembly, agent setup, sqlite-backed episode recall — against a fixed 750ms margin. That overhead is machine weather, not code behavior: 13ms measured on a fast dev host (1.013s internal elapsed), 620–730ms solo on the reporting runner. A fixed ceiling this tight asserts on machine speed.

## Fix

Widen the promptness ceiling from 1_750ms to 5s, per the issue's own first suggestion. What the ceiling protects is unchanged in kind:

- **Wedged timeout** (the regression this test exists for): the run would hang on `HangingProvider`'s 3_600s sleep — traced the path: nothing else bounds the blocked `chat()` call (the agent's own activity budget is only checked at loop-iteration boundaries, tool-batch timeouts wrap tool calls, not the LLM call). 5s catches it with three orders of magnitude of headroom.
- **Unit-scale regression** (secs↔minutes = 60s): still caught.
- The functional assertions (`OutcomeStatus::Error`, "timed out after 1s") are untouched, so the test still proves the outer `timeout_secs` branch — not an internal fallback — is what fired.

Acknowledged trade-off (also in the code comment): a 2–4x wrong-scale timeout is no longer caught on fast hosts. Any ceiling above ~2.7s accepts that — and on the runner that flaked, 1_750ms never had that coverage either (2s + 730ms overhead already exceeds it).

## Test evidence

- Solo before change: 15/15 pass on a fast host, internal elapsed 1.013s (measured via temporary instrumentation, removed before commit) — confirming the ~13ms fast-host overhead and that the margin is weather-dependent.
- Under CPU saturation (load average 33–44 on a 10-core host): 15/15 pass; this host is too fast to reach the 750ms ceiling — the red evidence is the issue's CI run plus the margin arithmetic above.
- After change: target test 6/6 green; full `cargo test -p octos-pipeline` green; `cargo fmt --check` clean; `cargo clippy -p octos-pipeline --tests` no warnings.

## End-to-end

The test itself is the end-to-end path: a real `CodergenHandler` with a real 1s `tokio::time::timeout` over a provider that hangs for 3_600s. Verified it still fires at ~1s, returns the `Error` outcome with the configured-duration message, and that the only behavior change is the ceiling.

## Review

One independent adversarial review (given only the issue, the diff, and the file): no Blockers/Majors; its three Minors (comment should name the wrong-scale trade-off, attribute the 13ms figure as measured, `from_secs(5)` over `from_millis(5_000)`) are all applied in the final diff, after which fmt/test/clippy were re-run green.

Closes #1932

### DIFF
--- a/crates/octos-pipeline/tests/codergen_worker_timeout.rs
+++ b/crates/octos-pipeline/tests/codergen_worker_timeout.rs
@@ -89,8 +89,19 @@ async fn codergen_timeout_secs_cancels_stalled_worker() {
         "unexpected outcome: {}",
         outcome.content
     );
+    // #1932 — the ceiling's job is to prove the timeout is not WEDGED (a
+    // regression there hangs on the provider's 3_600s sleep, three orders of
+    // magnitude out). The fixed overhead on top of the 1s timeout is machine
+    // weather: 13ms measured on a fast dev host, but 620–730ms SOLO on the
+    // Windows Server runner that reported the flake (1.62s/1.63s/1.73s solo
+    // against the old 1_750ms ceiling — a ~20ms margin any suite-level load
+    // consumed). 5s still catches a unit-scale regression (secs↔minutes =
+    // 60s); what it no longer catches on fast hosts is a 2–4x wrong-scale
+    // timeout, a trade-off any ceiling above ~2.7s accepts — and on the
+    // runner that flaked, 1_750ms never had that coverage either (2s + 730ms
+    // overhead already exceeds it).
     assert!(
-        elapsed < Duration::from_millis(1_750),
+        elapsed < Duration::from_secs(5),
         "worker timeout should trip promptly, got {elapsed:?}"
     );
 }


anantheparty (approved):
Adversarial pre-merge review passed - the timing-assertion tolerance fix: the 1s worker-timeout BEHAVIOR is unchanged; only the wall-clock margin bounding the assertion widens to 5s (the 750ms ceiling was already exceeded solo on the Windows Server runner at 1.62s per #1932, so any load guaranteed the flake). The same load-tolerant-margin discipline as #2288. Test-only, CI green. Approving - merging left to you per convention.

PR #2295, opened 2026-09-10
refactor(ui-protocol): drop the dead profile-scope arm from the replay send loop

### LINKED ISSUE #2078: ui-protocol: the session/open replay-send profile check is dead — same predicate, same vector, same scope as the retain filter
## Summary

`ledger_event_matches_profile_scope` is called twice over the *same* vector with the *same* profile id, so the second call cannot change any outcome. Deleting it entirely leaves the whole profile-scope test suite green — the check is unfalsifiable as written.

## Detail

Boundary 1 — `open_session_result` filters the replay vector in place:

```rust
replay.retain(|event| {
    ledger_event_matches_topic_scope(&event.event, topic_scope.as_deref())
        && ledger_event_matches_profile_scope(&event.event, &ledger_profile_id)
});
```

Boundary 2 — `handle_session_open` then re-filters `outcome.replay`, which *is* that already-retained vector, with `outcome.profile_id`, which *is* that same `ledger_profile_id` (`SessionOpenOutcome { .., profile_id: ledger_profile_id }`):

```rust
for event in outcome.replay {
    if !ledger_event_matches_profile_scope(&event.event, &outcome.profile_id) {
        continue;
    }
```

Same predicate, same input, same scope value ⇒ the boundary-2 arm is dead by construction. Confirmed empirically during #2067: removing it leaves every profile-scope test passing.

## Why it matters

It reads like defence-in-depth but provides none, and it costs real safety: a future change that weakens boundary 1 would not be caught by boundary 2, and a reader auditing the delivery path will count two independent gates where only one exists. Mutation testing would flag this arm immediately.

## Options

1. Make it mutation-sensitive — have boundary 2 filter against a value that can genuinely diverge from boundary 1 (e.g. the per-connection scope at send time rather than the open-time outcome), which would give it independent meaning.
2. Delete it and rely on boundary 1, with a comment at the send loop pointing at where the filtering actually happened.
3. Keep it but document explicitly that it is a redundant assertion, not a gate.

Option 2 or 3 is probably right; option 1 needs care after #2067, which deliberately made the *live forwarder* scope per-forwarder and immutable.

## Provenance

Found by codex during adversarial review of #2075 (fix for #2067), hypothesis H5. Not fixed there to keep that PR's scope to the leak and the forwarder-scope regression.

BH3GEI:
## Problem

#2078: `handle_session_open`'s replay send loop re-filters `outcome.replay` with `ledger_event_matches_profile_scope` against `outcome.profile_scope` — but `open_session_result` already retained that exact vector against that exact scope value. The second arm is dead by construction: it reads as defence-in-depth while providing none, and an auditor counting gates sees two where only one exists.

## Root cause

Same predicate, same input, same scope value at both boundaries:

- Boundary 1 (`open_session_result`): `replay.retain(|| ... && ledger_event_matches_profile_scope(&event.event, profile_scope.as_deref()))`, then `replay` is moved unmutated into `SessionOpenOutcome { replay, ..., profile_scope }` (the only constructor).
- Boundary 2 (`handle_session_open`): the send loop skipped events failing `ledger_event_matches_profile_scope(&event.event, outcome.profile_scope.as_deref())` — the same value, moved through the struct.

Traced on current main: no mutation of `replay` between retain and construction, no rebinding of `profile_scope`, single constructor, single consumer.

## Fix

Option 2 from the issue: delete the dead arm, point the send-loop comment at where filtering actually happens, and correct the two stale "three gates" attributions the arm left behind (the filter's doc comment and the wire-level test's boundary numbering). Option 1 (a genuinely independent send-time scope) is deliberately not applicable: #2067 made the per-connection scope frozen for the connection's lifetime, so `outcome.profile_scope` *is* the per-connection scope.

## Test evidence

- `cargo test -p octos-cli --features api --lib`: 3678 passed, 0 failed.
- The surviving guard is falsifiable at the wire level: `should_drop_cross_profile_goal_frames_when_connection_scopes_another_profile` drives `handle_session_open` end-to-end and asserts beta's frames never reach the alpha connection — after this PR that test actually pins boundary 1 (before, boundary 2 would have masked a boundary-1 regression). `should_retain_only_same_profile_goal_events_when_open_session_result_replays` pins the retain directly.
- `cargo fmt --all -- --check` clean; clippy with `-D warnings` clean on both the `api`-feature lib and the full feature matrix (`api,telegram,discord,dingtalk,slack,whatsapp,feishu,email,matrix,audio_mp3`, all-targets).
- End-to-end behaviour difference: none by construction (provably dead arm; same predicate/vector/scope, no side effects beyond `continue`). The wire-level suite above is the substitute evidence, and an independent review pass re-traced the dataflow and confirmed the deletion cannot change behaviour.

Closes #2078

### DIFF
--- a/crates/octos-cli/src/api/ui_protocol_tests.rs
+++ b/crates/octos-cli/src/api/ui_protocol_tests.rs
@@ -14562,8 +14562,9 @@ async fn should_retain_only_same_profile_goal_events_when_open_session_result_re
     );
 }
 
-/// #2067 boundaries 2 and 3 (session/open replay send loop + live forwarder):
-/// neither may put profile-b's durable goal frames on a profile-a connection.
+/// #2067 boundaries 1 and 3 (the `open_session_result` retain + live
+/// forwarder), observed on the wire: neither may put profile-b's durable goal
+/// frames on a profile-a connection.
 #[tokio::test]
 async fn should_drop_cross_profile_goal_frames_when_connection_scopes_another_profile() {
     let temp = tempfile::tempdir().expect("tempdir");
@@ -14626,7 +14627,7 @@ async fn should_drop_cross_profile_goal_frames_when_connection_scopes_another_pr
     .await;
     assert!(opened, "alpha must be able to open the shared session");
 
-    // Boundary 2 — the replay send loop.
+    // Boundary 1's retain, observed through the replay send loop.
     let frames = drain_session_open_frames(&mut rx).await;
     assert_eq!(
         replayed_goal_objectives(&frames),
--- a/crates/octos-cli/src/api/ui_protocol_transport.rs
+++ b/crates/octos-cli/src/api/ui_protocol_transport.rs
@@ -19055,10 +19055,12 @@ async fn handle_session_open(
     // so dropping them is not lossy from their perspective.
     //
     // Reusing the helper keeps replay and live capability behavior in lockstep.
+    //
+    // No profile-scope arm here: `open_session_result` already retained this
+    // exact vector against this exact scope (`outcome.profile_scope`), so a
+    // second filter could never fire — and reading it as an independent gate
+    // would overstate the delivery path's defences.
     for event in outcome.replay {
-        if !ledger_event_matches_profile_scope(&event.event, outcome.profile_scope.as_deref()) {
-            continue;
-        }
         let projected = features
             .projection_envelope_v2
             .then(|| project_v2_ledger_event(ledger, &event.event, &event.cursor))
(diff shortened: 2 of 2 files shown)


anantheparty (approved):
Adversarial pre-merge review passed - the construction-level reachability proof holds: open_session_result retains outcome.replay already filtered by ledger_event_matches_profile_scope against the SAME outcome.profile_scope value, so the re-filter in the send loop is a no-op by construction. Removing it is correct - a redundant gate that reads as defence-in-depth but filters nothing is worse than one gate (auditors count two and trust two). 12/-9, tests green, CI green. Approving - merging left to you per convention.

Issue #341, opened 2026-04-11
Expose session background task supervisor state to the agent


ymote:
## Problem

The web/runtime path now has a real background task supervisor and `/api/sessions/:id/tasks`, but the agent still narrates background task state by guessing from chat context. That creates a split-brain architecture:

- UI/runtime task state comes from the supervisor.
- Agent status narration comes from model inference.
- `session_actor` then rewrites premature or incorrect completion claims after the fact.

This is why background tasks like `fm_tts`, `voice_transcribe`, and `podcast_generate` can produce contradictory assistant text such as "already generated" while the task is still running.

## Goal

Give the agent a read-only, session-scoped way to query the background task supervisor so it can ground status answers on actual task state instead of guessing.

## Proposed scope

1. Add an agent-accessible tool, e.g. `check_background_tasks`, limited to the current session.
2. Return structured task state from the supervisor only:
   - task id
   - tool / task kind
   - status (`spawned`, `running`, `completed`, `failed`)
   - started/updated/completed timestamps
   - output artifacts if available
   - error summary if failed
3. Update prompts/tooling policy so direct-delivery background tasks do not freehand completion claims.
4. When the user asks "is it done yet?" or equivalent, the agent should query the supervisor instead of inferring from message history.
5. Longer term, reduce `session_actor` text-rewrite policy once the agent has a grounded status path.

## Non-goals

- Exposing cross-session or cross-user task state.
- Replacing the web `/tasks` API.
- Making assistant message history the lifecycle source of truth.

## Acceptance criteria

- The agent can answer status questions about the current session's background tasks using real supervisor state.
- `fm_tts`, `voice_transcribe`, and `podcast_generate` no longer rely on guessed completion wording.
- Premature "completed/file delivered" assistant text can be removed or materially reduced.

ymote:
Fixed on main by 77c99d3 (session-scoped background task supervisor visibility via check_background_tasks).

Issue #337, opened 2026-04-10
Do not persist transient upload routing hints into session history


ymote:
Regression:

Temporary system hints used to route uploaded audio/files are being written into persisted session history instead of being injected only for the current turn.

Symptoms:
- image-only uploads in an old session can get polluted by stale uploaded-audio context
- users may see hidden routing behavior carry over across unrelated later turns
- absolute workspace paths can leak into persisted history if workspace attachment hints are stored

Scope:
- `[Uploaded files copied into workspace]`
- uploaded-audio clarification/routing hints
- `[Most recent uploaded audio file]` follow-up binding hints

Expected behavior:
- assistant outputs like OCR / image analysis / transcription results remain in history
- internal routing hints stay transient and are not written to disk

Fix direction:
- build these hints into the per-turn agent context only
- never add them via `SessionHandle::add_message(...)`
- add regression tests for upload->clarify->follow-up flows

ymote:
Fixed in ddc39c2: attachment routing hints now stay in per-turn context instead of being persisted into session history. File-only uploads persist as a generic placeholder without leaking filenames or transient [Attached files] summaries.

Issue #336, opened 2026-04-10
【feature】Add support for  DingTalk channel


lijingrs:
Implement support for the DingTalk channel integration.
Enable message sending, event receiving, and bot interaction via DingTalk.

ymote:
Non-closing current-main refresh attempt for PR #1285:

- Local refreshed DingTalk head `db1224af7020abf85edc9abfd0d4c79bcff85b85` was validated against current GitHub `main` `4adbe05fff212cb85d1f0a6d6225da0713446016` in isolated checkout `/private/tmp/octos-1285-current.JaP3B7/octos`.
- Focused DingTalk Rust tests, CLI feature checks, dashboard typecheck/build, touched-file typos, formatting, bus clippy with `-D warnings`, and default workspace clippy with `-D warnings` passed. Full command list is on PR #1285: https://github.com/octos-org/octos/pull/1285#issuecomment-4599007571
- The PR branch could not be updated from Codex because the existing PR changes `.github/workflows/ci.yml` and this OAuth token lacks `workflow` scope. GitHub rejected the exact-lease push with: `refusing to allow an OAuth App to create or update workflow .github/workflows/ci.yml without workflow scope`.

Decision: keeping #336 open. The refreshed code validates locally, but a workflow-scope maintainer/actor needs to push the refreshed head before CI/review/merge can close this through PR #1285.

ymote:
Non-closing current-main refresh for PR #1285 / #336:

- Validated PR #1285 head `7d93fafc405d0f6796874b69242ccbcd02d50c00` by cleanly merging it into current GitHub `main` `f6936c452389c5172496ee0c3e13393956086d92` in isolated checkout `/private/tmp/octos-336-f693.xAOp94/octos`.
- The merge had no conflicts. `git status --short` after validation showed only the expected PR file changes and no untracked generated files.
- Passed validation commands on the merged tree:
  - `git diff --check origin/main`
  - `cargo fmt --all -- --check`
  - `CARGO_TARGET_DIR=/private/tmp/octos-336-f693-target cargo test -p octos-bus --features dingtalk dingtalk -- --nocapture` (5 DingTalk tests passed)
  - `CARGO_TARGET_DIR=/private/tmp/octos-336-f693-target cargo test -p octos-cli --no-default-features --features api,dingtalk profiles::tests::test_channel_serde_roundtrip -- --nocapture` (round-trip test passed)
[...]

PR #2296, opened 2026-09-10
fix(api): repair web persistence, profile startup, and live hydration

### LINKED ISSUE #2298: Return retained canonical projection sequences in session hydration
A fresh live chat on mini3 receives a real DeepSeek answer, but its user prompt remains duplicated as an optimistic row, shows a 1969 timestamp, and completion can be lost during gap hydration. Captured `session/hydrate` replies provide transcript rows and tool/background subsets; the frontend resequences those rows and replaces live canonical state, so a later canonical terminal no longer follows a contiguous sequence. The server does not supply the full retained v2 stream, user timestamps are dropped, and a settled successful GhostBubble still renders its text. Preserve canonical event coordinates through hydration, merge older transcript-only history without replacing retained streams, preserve available persisted timestamps, and hide confirmed optimistic rows. Verify actual chat, queued turns, mid-turn reconnect and fresh-browser history on mini3; add protocol regressions for the recorded frame shapes.

Reproduced with web 1fae296 and core caf603c8 in the isolated mini3 review instance.

### LINKED ISSUE #2299: Serialize cold profile runtime bootstrap to avoid same-process database lock failures
A cold real browser login on the isolated mini3 server starts several requests concurrently. With only one `octos serve` process and all standalone gateways stopped, the first chat fails with `data_dir_locked`. Later attempts succeed after one request caches the runtime. `ensure_session_profile_runtime` checks the runtime map and asynchronously bootstraps without serializing creation for the same profile; two requests can both observe a cache miss and try opening the same redb store. This is distinct from the original test setup gateway conflict.

Expected: concurrent requests for the same profile share one runtime bootstrap. Keep generation checks for configuration changes, and do not serialize unrelated profiles. Add a concurrent real-bootstrap regression and verify the first browser chat after a clean server restart.

### LINKED ISSUE #2300: Allow authenticated file listing without a standalone profile gateway
In a valid standalone `octos serve` deployment, authenticated profile chat works without any standalone gateway process. `GET /api/files/list` still calls `resolve_profile_data_dir`, which requires a live profile gateway API port and returns 503 `no gateway`. Slide creation writes `script.js`, `memory.md` and `changelog.md` successfully, but the browser cannot discover them and reports scaffold failure. Reproduced repeatedly on mini3 with actual browser requests, core caf603c8 plus live hydration fixes.

Use the same authenticated data-directory resolver as file reads and mutations, preserving profile/host authorization. Test file listing and returned-handle reading with a real profile store and no process manager, including rejection of another account's profile header. Rerun actual slide creation and fresh-browser discovery.

### LINKED ISSUE #2301: Large live tool turns lose transcript text and continuation coordinates during hydration
An actual DeepSeek tool turn that builds a small React/Vite site causes `session/hydrate` to replace ordinary user/assistant text with `[oversized field omitted]`. A browser reload still receives omitted text. The browser also briefly reports `Send not confirmed within 30s` despite the agent executing tools.

Reproduced on isolated mini3 with core b2d3c08 and web 3ae0b08. The final delivered hydrate contained 49 messages (~12.6 KB after rewriting), 26 replayed tool events (~9.4 KB), and 1,993 retained projection envelopes (~561 KB). It contained 39 stub markers in the message section, 20 in tool replay and 3 in full projection replay. The retained projection starts partway through a thread whose event sequence exceeds 5,000, so the web adapter falls back to re-sequenced transcript rows.

Two interacting limitations need regression coverage: duplicate/bulky replay drives the aggregate 1 MiB frame truncator to erase small durable chat text; a thread longer than the replay retention window cannot restore its live sequence from the transcript alone. Bound or compact hydration before the generic frame truncator, preserve user-visible transcript and per-thread continuation coordinates, and verify a real tool-rich turn through reload and terminal delivery. Do not raise the transport cap or disable gap detection as a workaround.

This is outstanding live correctness work; earlier short-text chat passes do not certify this flow. Redacted size summaries and failed screenshots are retained with the review evidence.

### LINKED ISSUE #2303: Slides prompts instruct the model to call an intentionally hidden generation tool
The first-party slides system prompt says to call `mofa_slides` directly and declares generation unavailable if that name is absent from the visible tool list. Since RFC-1, that implementation is intentionally internal-hidden: the visible interface is `mofa_make({content_type: "slides", args: ...})` with `mofa_describe_content_type` for the schema. `keep_tool_in_slides_session` explicitly retains that dispatcher pair.

A live mini3 slide capability check encountered this surface mismatch. The scaffold script and system prompt still instruct the obsolete entry point; the model must rediscover the dispatcher or may incorrectly report generation unavailable. Installed third-party skill instructions also reference the legacy name.

Update the first-party scaffold and slides prompt to use the visible dispatcher, inspect its current schema, distinguish a missing slides content type from a hidden implementation, and retain the no-shell-fallback and artifact verification requirements. Add a regression against the generated prompt and existing registry visibility coverage.

This correction does not supply a missing image-provider key or establish full deck rendering acceptance. Refs #2296 and octos-org/octos-web#341.

### LINKED ISSUE #2304: Background slide tools incorrectly promise audio output
An actual `mofa_make` slides invocation on the isolated mini3 emits a canonical tool result saying “Running in background — audio will be sent when ready.” The spawn-only execution path uses this same audio-specific text for every tool, including slide/site/research work.

The UI should describe pending results without promising audio for non-audio work. Replace the generic spawn acknowledgement with “Running in background — results will be delivered when ready.” This changes only a display string, preserving task handles, routing and terminal state.

Reproduced on core `55cf887`, web `3b4fa70`, using a synthetic slide import probe. Fix belongs in paired review PR #2296; full renderer acceptance remains separately blocked by the missing image provider key.

### LINKED ISSUE #2306: web_fetch text extraction includes scripts and CSS, crowding out page evidence
Actual weather lookup records from mini3 show `web_fetch(extract_mode="text")` returning CSS/JavaScript in the first 3,000 bytes instead of the forecast body. The weather.com.cn result begins with `.xyn-weather-box { width: ... }`; tianqi.com begins with `var pagetype = ...` and CSS. The model then abandoned those sources and performed more lookups.

Cause: `extract_text()` is a character-level angle-bracket stripper. It removes tags but retains script/style bodies, mishandles quoted `>` in attributes, and leaves HTML entities undecoded. Large non-visible blocks consume the response cap before useful evidence reaches the model.

Replace the scanner with the existing HTML parser's DOM, skip non-content subtrees, preserve readable block boundaries and inline text, and test style/script exclusion, malformed/quoted markup, entities, and truncation after extraction. Keep URL provenance, SSRF/redirect/allowlist behavior, body limits and tool-output caps intact. Verify against a real public weather page as well as deterministic fixtures.

This is one concrete grounding-input defect discovered while following up #2305; repairing extraction alone does not certify model citation or weather-fact accuracy.

ymote:
Generated site previews could execute with application credentials, file/slide edits lacked durable APIs, cold profile requests could conflict on redb, and compacted chat hydration could erase text or strand completion. This repairs those paths for the paired web review and improves the web evidence used in answers.

Preview responses enforce sandbox CSP for iframe and direct navigation, with opaque-origin module CORS limited to signed URLs. Authenticated file rename/delete use profile-scoped resolution and anchored no-follow operations on Unix. Slide edits persist atomically with revision conflicts; the client waits for matching rendered revisions and fresh PNG/PPTX outputs.

Concurrent cold requests share one profile bootstrap. File listing works with the in-process serve runtime. Hydration preserves canonical coordinates and compacts redundant streaming replay before frame truncation, returning terminal state and per-thread continuation checkpoints. Durable transcript text remains the source for compacted content. Slide instructions name the visible dispatcher, and generic background acknowledgements no longer incorrectly promise audio.

Plain-text `web_fetch` now traverses the HTML5 DOM instead of stripping tag delimiters, preventing scripts/styles from exhausting the output budget before page evidence. It omits non-content/hidden subtrees, decodes entities and separates blocks without splitting inline words. Markdown extraction preserves links and excludes scripts/styles/templates/noscript. The parser dependencies were already pinned transitively through the existing converter; network/SSRF/redirect/allowlist and output caps are unchanged.

Gateway instructions keep short follow-ups tied to the actual topic, avoid repeated report templates/unrelated comparisons, require direct links for web-based claims, distinguish observation/forecast times and require separate evidence for heat causes or active warnings. Existing custom prompts are preserved. Model adherence and source freshness still require live acceptance.

**Paired client:** https://github.com/octos-org/octos-web/pull/350. Deploy the API first. Both are deployed only to the isolated mini3 instance; the designated public canary is not certified.

Validation:
- All backend CI checks pass on `b931bd2` ([run 34573793760](https://github.com/octos-org/octos/actions/runs/34573793760)). The embedded debug build with the production `/app/` bundle passes.
- 19 web-fetch tests pass, including network restrictions; four new plaintext cases fail on the original extraction. All 15 gateway prompt contracts pass. Earlier validation includes CLI library Clippy, 48 review-filtered tests, canonical hydrate/core serde checks, 17 project-template tests and a 5,000-event frame-compaction regression.
- Mini3 candidate browser tests pass actual text extraction and two natural weather turns with actual lookups and matching inline links. Manual comparison of inspected dates/measurements against the exact returned tool text supports those candidate answers. Three fresh weather cases pass again on the final isolated runtime.
- Final web `39bbb1a` / core `b931bd2` passes 5/5 remote real core tests in 16.7s and original-history reads/reloads in Chromium and WebKit. Source-link punctuation exposed by a post-deploy test is fixed in the paired client; the source site's own Chromium redirect remains recorded separately.
- A new 38-cycle **30m 33s** remote soak passes exact full history order/reload/navigation/reconnect with zero retries, page errors or observed HTTP failures on **core `7590b73` / web `ea8ce95`**. It is not attributed to the later extraction/prompt changes. Earlier evidence includes actual generated file and React/Vite preview isolation tests, a 36-second tool turn and an intact 3,335-chunk/16,603-character reply.

Slide generation/edit rendering/PPTX export, ASR/TTS and full learning workflows remain outside live acceptance because the isolated lane lacks required capabilities. Weather freshness, heat-cause and active-warning acceptance remains open under #2305. See the [latest validation/evidence record](https://github.com/octos-org/octos-web/blob/fix/review-correctness-gaps-20260910/docs/reviews/2026-09-11-live-followup.md).

Refs octos-org/octos-web#336, octos-org/octos-web#341, octos-org/octos-web#343.
Fixes #2298, fixes #2299, fixes #2300, fixes #2301, fixes #2303, fixes #2304, fixes #2306.
Refs #2305.


anantheparty (dismissed):
Adversarial pre-merge review passed - 0 blocker, 3 nits (performance/documentation-class), the security-critical parts all verified: (1) the #336 server half - CSP sandbox headers converge on ALL preview routes (site-preview/preview/preview-signed, direct-new-tab too), generated JS lands in an opaque origin with the credential-delivery surface narrowed to none, while the signed-URL chain (256-bit token, 600s TTL, three-layer rate limits, post-consume bearer re-resolution) stays intact; (2) the #341/#343 server half - OCC with in-lock revision reads, temp+fsync+rename atomic writes, openat NOFOLLOW throughout, NOREPLACE renames, fail-closed profile auth (forged X-Profile-Id 403s with files untouched, matching the #629 precedent); (3) no #337/#342-style auth leaks on the new routes (unified user_auth_middleware). Cross-repo pairing verified field-by-field against octos-web#350 (closed loop). 8 new tests, CI green.

Merge order: #2296 first (additive - the preview going opaque-origin immediately IS the #336 fix intent), then #350 (which hard-depends on /api/files/mutate and /api/slides/edits). Approving - merging left to you per convention.

BH3GEI (commented):
Thanks for putting this together — the durable file/slide mutation APIs and the preview sandboxing are a clear step up, and the profile-scoped `openat`/`NOREPLACE` work held up under an adversarial pass (I could not construct any traversal, symlink, or clobber scenario on the Unix paths, and the signed-URL chain is intact).

I verified locally: `cargo fmt --all -- --check` clean, `cargo test -p octos-core --lib` 400/400, `-p octos-agent --lib` 2894/2894 (3 ignored), `-p octos-cli --lib` 3692/3692 (10 ignored). However, I found one build regression and one behavioral concern that I think need addressing before merge:

**1. Clippy regression — `slide_edit_location` misses the crate's `result_large_err` allowance (must-fix)**

`crates/octos-cli/src/api/handlers.rs:2024` returns `Result<(PathBuf, PathBuf), Response>`. On stable 1.98.1, `cargo clippy -p octos-cli --all-targets -- -D warnings` (the CI command's per-crate equivalent) fails with `clippy::result_large_err` ("the `Err`-variant is at least 128 bytes"). I ran the identical command with the identical toolchain against `origin/main` (e7b57f2c) and it is clean, so this is introduced by the PR — the current green `check` job appears to be a toolchain-version coincidence that will flip red when Actions' stable moves. Every other `Result<_, Response>` helper in this file already carries the allowance (e.g. `resolve_file_access_data_dir` at handlers.rs:1453); adding the same `#[allow(clippy::result_large_err)]` here would match the established pattern.

**2. `after`-cursor incremental hydrate is compacted to terminals only (behavioral)**

`compact_hydrate_projection_replay` (`crates/octos-cli/src/api/ui_protocol_transport.rs:25016-25019`) classifies a thread as "complete" only when its projected sequence starts at `seq == 1` and is contiguous. But `session/hydrate` with `after` (documented at `crates/octos-core/src/ui_protocol.rs:2838-2841` as the reconnect path) feeds `snapshot_with_cursor` output (`ui_protocol_transport.rs:25090`) into that function, and an incremental window's sequences necessarily start above 1. So every `after` reconnect with `include: ["messages"]` gets its window reduced to `TurnTerminal` events only — user messages, deltas, and tool events replayed from the missed window are all dropped. Since durable transcript rows lag streaming output mid-turn (the PR's own notes position `messages` as the source for compacted content), a reconnect during a long turn loses the not-yet-durable streamed text — which is the exact failure mode this PR set out to fix. I could not find any test combining `after` with the new channel (the cli tests at `ui_protocol_tests.rs` and the e2e reconnect probe use full hydrate). One caveat: whether the paired web client reconnects this way is not verifiable from this repo, but the server-side drop is fully traceable. Could the completeness check be made relative to the requested window (or compaction skipped when `after` is present)?

Two smaller notes while you're in these files:

- `crates/octos-cli/src/prompts/slides_default.txt:40` demotes the inline `slides` array to a conditional fallback ("use inline slides only when the installed schema and skill explicitly support that mode"), but line 41 still directs "trim the inline `slides` array" as the primary partial-regeneration mechanism. The old text always used the inline array precisely because `input=` needs Node.js on the host; as written, a host without Node now ends at "explain the prerequisite" where it previously succeeded.
- Pre-existing, but adjacent to this PR's sandbox claim: the preview error branch (`site_preview_html`, reached from the `resolve_preview_asset_path` miss around handlers.rs:3236) serves `text/html` with no CSP headers and interpolates the percent-decoded `request_path` unescaped, so `…/preview-signed/<token>/%3Cscript%3E…` reflects active script into a same-origin page. Since the PR's stated goal is "preview responses enforce sandbox CSP", the error responses arguably deserve the same headers (plus escaping).

Happy to re-review once these are addressed — the rest of the change looked solid under line-by-line and adversarial review.

anantheparty:
Cross-link for the hydration half of this repair: the browser client shipped its companion fix for the same symptom family — after a gap recovery, the client was re-applying recovery-buffered deltas that the hydrate snapshot already contained (duplicating assistant text). octos-org/octoscode-web#51 guards the buffered drain with the hydrate cursor (`fromRecoveryBuffer` + `cursor.seq <= hydrate cursor → ignore stale`), with regression tests on both the projection and runtime levels.

So this PR (server-side: compacted chat hydration must not erase text or strand completion) and the client guard (client-side: don't replay what the snapshot covers) close both halves. Also noting: the octoscode-web contract/core-integration suites pin the UI-protocol wire shape, so they'll validate compatibility of any protocol-adjacent changes here when this lands.

anantheparty:
FYI from the octoscode-web maintenance side: ran an SSH reachability audit across all configured targets (39 total, 23 reachable / 17 unreachable). GitHub SSH (github.com:22) is reachable — this machine pushes to octoscode-web via SSH. The frp/osaka tunnels (osaka.ananthe.party on various ports) are stable. Local network targets (192.168.x.x) are unreachable from this machine's network segment as expected.

No action needed — just noting the infra baseline for reference.

PR #2307, opened 2026-09-12
fix(release): ship .sha256 checksum sidecars so installers can verify bundle downloads

### LINKED ISSUE #1929: Server bundles ship without .sha256, so octoscode's auto-install silently skips verification
Found by running a real from-scratch auto-provision on a clean Linux box (Ubuntu 26.04, aarch64) — the path no test exercises.

## What happens

```
octoscode: octos backend not found; downloading the octos server bundle v2.0.3-rc.1 ...
octoscode: could not fetch the bundle checksum (HTTP status client error (404 Not Found) for url
  (https://github.com/octos-org/octos/releases/download/v2.0.3-rc.1/octos-bundle-aarch64-unknown-linux-gnu.tar.gz.sha256));
  skipping verification
octoscode: installed the octos server to /home/ubuntu/.octos/bin
```

A **209 MB server binary** was downloaded, unverified, and installed — then spawned as a child process. The client did the right thing by *trying*; the artifact it needs has never been published.

## This has never worked, on any platform

```
v2.0.3-rc.1   0 bundle .sha256 files
v2.0.2        0
v2.0.2-rc.13  0
```

Release assets are exactly:

```
octos-bundle-aarch64-apple-darwin.tar.gz
octos-bundle-aarch64-unknown-linux-gnu.tar.gz
octos-bundle-x86_64-pc-windows-msvc.zip
octos-bundle-x86_64-unknown-linux-gnu.tar.gz
```

No `.sha256` alongside any of them. So the verification branch in `backend_ensure` has been dead since it was written, and **every** auto-provision on **every** platform installs an unverified binary.

## Why nobody noticed

It degrades to a **warning** and continues. Nothing fails, nothing is red, and the message only appears on a first-launch auto-install — the one flow a developer with octos already installed never sees.

This is the same shape as #1923 (e2e green with every step skipped) and #1925 (UX detectors matching chrome that no longer renders): **a check that cannot fail guards nothing.** Third instance this week, and the first with a security consequence.

## Fix

**Here (octos):** publish a `.sha256` next to each bundle in the release workflow. That alone makes the existing client-side check start working — no octoscode change needed.

**In octoscode, worth deciding separately:** whether a missing checksum should be fatal. Right now "cannot verify" and "verified OK" are the same outcome, which means an attacker who can serve a 404 for the `.sha256` also turns verification off. Failing closed once checksums exist would be the stronger posture; failing closed *before* they exist would break every install, so the order matters — publish first, then tighten.

## Repro

Any machine with no `octos` on PATH and no `~/.octos/bin/octos`:

```
octoscode --stdio-command "octos serve --stdio --solo"
```

(The TUI then dies on `No such device or address` without a TTY — irrelevant here; the install has already happened by that point.)

BH3GEI:
## Problem

#1929: no release has ever shipped `<bundle>.sha256` sidecars. A from-scratch octoscode auto-provision on a clean Linux box logs a 404 for `…/octos-bundle-aarch64-unknown-linux-gnu.tar.gz.sha256`, **silently skips verification**, and installs the 209 MB binary unverified.

## Root cause

`scripts/bundle-release.sh` (the single source of truth for what ships) produces only the archive. Nothing in `release.yml` generates, uploads, or checks a checksum — the release pipeline had no checksum concept at all.

## Fix

- **`scripts/bundle-release.sh`** — after creating the archive, write `<archive>.sha256` next to it in standard `sha256sum -c` format (`"<hex>  <filename>"`, relative filename). Uses `sha256sum`, falling back to `shasum -a 256` (macOS runners); both emit byte-identical lines, verified below. The hash is computed on the build runner against the exact built bytes.
- **`release.yml`** — each platform build uploads its sidecar alongside the bundle, with `if-no-files-found: error` so a missing file fails the build job instead of surfacing three jobs later. The publish job's existing `find artifacts -name 'octos-bundle-*'` picks the sidecars up unchanged. `verify-release` now expects all four sidecars and round-trips every published sidecar against the published bytes with `sha256sum -c ./*.sha256` — the exact check octoscode's auto-install performs (Darwin sidecars are produced via `shasum`, the others via `sha256sum`, so checking all four covers both producer paths).
- **`scripts/test-bundle-release.sh`** (new, same style as the other `scripts/test-*.sh`) — builds a fake `target/release` tree, runs the bundle script for the tar.gz and zip cases, and asserts each sidecar exists and verifies with `sha256sum -c` itself. Wired into a new lightweight `bundle-release-check` job in `ci.yml` so the behavior can't regress silently (Linux runners exercise the tar.gz case; the zip case runs where `7z` exists and reports a distinct skip message otherwise).

octoscode needs no changes: it already fetches `<asset>.sha256` and parses the first whitespace-delimited token (64 hex, case-insensitive) — exactly the line format above. Older releases can't be back-filled retroactively; octoscode's existing warn-and-skip path keeps covering them.

## Testing evidence

- **Red → green**: on baseline `origin/main`, `scripts/test-bundle-release.sh` fails with `FAIL: missing checksum sidecar: …tar.gz.sha256` (issue reproduced); on this branch it passes for both archive formats.
- **Cross-tool**: the `shasum -a 256` fallback branch was exercised explicitly (restricted `PATH` on macOS): output is byte-identical to `sha256sum` output and verifies under `sha256sum -c`.
- **End-to-end (local replay of the full release path)**: built all four platform bundles from a fake release tree → assembled the 12-asset release directory (4 bundles + 4 sidecars + 4 install scripts) → replayed the updated `verify-release` run block verbatim with the two `gh` calls localized: expected-assets gate passes, all four sidecars verify (`OK ×4`), `tar -tzf` passes. Negative check: a corrupted sidecar line makes `sha256sum -c` fail with exit 1.
- **Workflow lint**: `actionlint` clean on both touched workflows (the `ubuntu-24.04-riscv` warning in `ci.yml` predates this change — verified identical on `origin/main`).
- A true tag-push release can't be exercised from a PR; the replay above walks the same bytes through the same script bodies, and the first real release will additionally run the new `verify-release` checks in CI.

## Review

One independent maintainer-perspective review pass (issue text + diff + workflows only): no blockers; its findings — wiring the test into CI, round-tripping all four sidecars (not just one), fail-loud artifact uploads, skip-message accuracy, and a sync note on the fake-binary list — are all addressed in this change.

Closes #1929

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -211,6 +211,14 @@ jobs:
         shell: bash
         run: '"$BASH" scripts/tests/test-build-web-app-base-url.sh'
 
+  bundle-release-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Test bundle checksum sidecar
+        run: scripts/test-bundle-release.sh
+
   check:
     runs-on: ubuntu-latest
     env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -96,7 +96,12 @@ jobs:
       - uses: actions/upload-artifact@v7
         with:
           name: macos-arm
-          path: octos-bundle-aarch64-apple-darwin.tar.gz
+          # Fail loudly when the bundle or its sidecar is missing instead of
+          # letting verify-release fail 3 jobs later with a cryptic gap.
+          if-no-files-found: error
+          path: |
+            octos-bundle-aarch64-apple-darwin.tar.gz
+            octos-bundle-aarch64-apple-darwin.tar.gz.sha256
 
   build-linux-x86:
     needs: check-version
@@ -144,7 +149,10 @@ jobs:
       - uses: actions/upload-artifact@v7
         with:
           name: linux-x86
-          path: octos-bundle-x86_64-unknown-linux-gnu.tar.gz
+          if-no-files-found: error
+          path: |
+            octos-bundle-x86_64-unknown-linux-gnu.tar.gz
+            octos-bundle-x86_64-unknown-linux-gnu.tar.gz.sha256
 
   build-linux-arm:
     needs: check-version
@@ -194,7 +202,10 @@ jobs:
       - uses: actions/upload-artifact@v7
         with:
           name: linux-arm
-          path: octos-bundle-aarch64-unknown-linux-gnu.tar.gz
+          if-no-files-found: error
+          path: |
+            octos-bundle-aarch64-unknown-linux-gnu.tar.gz
+            octos-bundle-aarch64-unknown-linux-gnu.tar.gz.sha256
 
   build-windows:
     needs: check-version
@@ -246,7 +257,10 @@ jobs:
       - uses: actions/upload-artifact@v7
         with:
           name: windows
-          path: octos-bundle-x86_64-pc-windows-msvc.zip
+          if-no-files-found: error
+          path: |
+            octos-bundle-x86_64-pc-windows-msvc.zip
+            octos-bundle-x86_64-pc-windows-msvc.zip.sha256
 
   # TODO: uncomment after Dockerfile is fixed
   # build-docker:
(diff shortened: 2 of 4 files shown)


anantheparty (approved):
Adversarial pre-merge review passed — no blocker, the release trust chain closes correctly: sidecar generation through the single bundle-release.sh source (standard `sha256sum -c`-compatible format with the shasum fallback for macos-14, all four bundle variants), the pipeline three-point consistency (generate → explicit sidecar uploads with if-no-files-found → verify-release downloading all 8 assets and running `sha256sum -c` for real, with package-publishers gated behind verify-release), and the elegant cross-repo activation: octoscode's `bundle_urls()` + `verify_sha256()` (mismatch → hard "refusing to install" error) needed ZERO changes — the dead code branch comes alive the moment the sidecars exist. Old releases keep the 404→warn+skip behavior (no regression).

Nits: (1) the release.yml:98-100 comment overstates `if-no-files-found: error` (union semantics — a bundle-without-sidecar still uploads; the early red only fires on double-missing; fix the comment or split the upload steps); (2) the first-party installers (install.sh/install.ps1) still download without verification — worth a follow-up ledger entry since the sidecars are now consumable; (3) the security-posture boundary deserves documenting: sidecars same-origin same-pipeline guard against transport corruption/truncation, not pipeline compromise (no minisign/sigstore), and the 404→skip window for pre-sidecar historical releases is permanent. Approving — merging left to you per convention.

PR #2308, opened 2026-09-12
fix(pipeline): assign only router-resolvable models from the catalog (#1901)

### LINKED ISSUE #1901: deep_research silently downgrades to the default provider when model assignment picks an unregistered model

BH3GEI:
## Problem

`model_assignment.rs` rewrites unset lane models using the on-disk catalog, but the persisted `model_catalog.json` is a deliberate superset (93 researched entries, most never registered by the profile's `ProviderRouter`). So assignment keeps handing nodes unresolvable model keys — and since #1902, the run no longer crashes but **silently executes those nodes on the default provider**, with a `warn!` firing on every run. That's the remaining Layer 1 + Layer 4 work scoped in the issue.

## Root cause

`build_pools` filtered candidates by health (stability) only. It never consulted the router, so the strong/fast pools routinely contained models only `router.resolve` could reject. The executor's `resolve_provider` fallback (the #1902 stop-gap) then fired as the *normal* path instead of a safety net.

## Fix

- **Layer 1** — `assign_from_catalog_dir` now takes `Option<&ProviderRouter>` (threaded from `ExecutorConfig.provider_router` at the existing call site) and `build_pools` retains only candidates the router can actually `resolve`. The predicate is deliberately the router's own resolution semantics (exact → case-insensitive → prefix split → suffix alias), **not** `keys()` membership — a literal membership check is the #1904 defect class, because the router resolves lane keys and aliases that a literal set rejects. With no router, the filter fails open: assignment behaves exactly as before.
- **Layer 4** — the original lane key is physically lost by the time the router is consulted (assignment already rewrote it), so `router.rs:154` cannot report it without a signature change through the whole resolution chain. Instead the fix closes the gap at the assignment boundary, where the lane still exists:
  - `build_pools` warns once with the dropped count and the sorted registered keys,
  - the assignment `info!` now logs the concrete `strong_pool` / `fast_pool` contents, so `no provider registered for key 'X'` traces back to the lane that produced `X` in one grep,
  - both fallback warns (`executor.rs` free fn, `CodergenHandler`) now carry `node=` so the gap is pinned to a node, not just a model key.

Constraint check (from the issue discussion): the filter lives **only** in `model_assignment.rs` candidate selection — `known_model_keys_from_catalog_dir` and the pre-execution validator are untouched. The #1902 fallback behavior and `resolve_provider_falls_back_to_default_when_key_absent` are preserved verbatim (only a `node` parameter was added for attribution); it is now a safety net that stays silent in normal operation.

## Testing evidence

- 4 new unit tests in `model_assignment.rs`, each verified RED against the pre-fix code by no-op'ing the retain: superset catalog + lane-only router assigns nothing and injects no planner model; the filter passes a compound-key catalog entry via the suffix-alias arm while dropping an unresolvable one (pins the resolution-semantics-not-membership rule); `None` router fails open; the `dynamic_parallel` comma pool joins only resolvable entries.
- Full suite: `cargo test -p octos-pipeline` — 464 passed, 0 failed; `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean (CI parity).
- Real-LLM e2e (temporary harness, DeepSeek, not part of the diff): a 3-node DOT pipeline, no `model=` attrs, catalog seeded with `qwen3-max` (best score, unresolvable) + `deepseek/deepseek-reasoner` (resolvable), router registering `cheap`/`strong`/`deepseek/deepseek-reasoner`, default provider = `deepseek-chat`.
  - **Before (main):** nodes got `qwen3-max`, each logged `pipeline node model not in provider router; using default provider`, and actually ran on the default `deepseek-chat` — the exact silent-downgrade from the issue.
  - **After (this branch):** one `skipped healthy catalog models … dropped=1` warn, all three nodes assigned `deepseek-reasoner`, all three ran on `openai/deepseek-reasoner` via the router, zero fallback warns.

## Review

Self-reviewed full diff, then an independent reviewer (issue text + diff + related sources only) returned SHIP with no blocker/major findings; the actionable notes (sorted `router_keys`, a stale "Returns Ok(())" doc line, an executor comment that still described the now-fixed rewrite as current behavior) are addressed in this final state. One pre-existing remnant is knowingly left out of scope: for a `dynamic_parallel` node with an explicit `model=` but no `planner_model=`, assignment can still inject a planner model that outranks the explicit one (`planner_model > model` in the executor). That behavior predates this fix and is only narrowed by it (the injected model is now guaranteed resolvable); flagging it here in case you want it tracked separately.

Happy to take an adversarial pass on the Layer 4 placement — the alternative (threading the lane key down to `router.rs`) would need a new resolve entry point used only by pipelines, which seemed worse than logging at the boundary where the information still exists.

Closes #1901

### DIFF
--- a/crates/octos-pipeline/src/executor.rs
+++ b/crates/octos-pipeline/src/executor.rs
@@ -1117,17 +1117,22 @@ fn spawn_pipeline_heartbeat(
 }
 
 /// Resolve an LLM provider from a model key using an optional router.
+///
+/// `node` names the pipeline node the key came from so the fallback warn
+/// pins WHICH node has the router gap (#1901 Layer 4) instead of leaving
+/// the operator to correlate the bare model key against the graph by hand.
 fn resolve_provider(
     default: &Arc<dyn LlmProvider>,
     router: Option<&Arc<ProviderRouter>>,
     model_key: Option<&str>,
+    node: &str,
 ) -> Result<Arc<dyn LlmProvider>> {
     match (model_key, router) {
         // A missing/unknown model key must degrade to the default provider,
-        // matching CodergenHandler::resolve_provider (handler.rs) — the
-        // model-assignment pass may rewrite lane keys (e.g. `strong`) to
-        // catalog models this profile's router never registered, and that
-        // must not fail the whole pipeline.
+        // matching CodergenHandler::resolve_provider (handler.rs) — an
+        // explicit `model=` the router never registered (e.g. a stale
+        // catalog or a hand-written lane key on a narrower profile) must
+        // not fail the whole pipeline.
         (Some(key), Some(r)) => match r.resolve(key) {
             Ok(provider) => Ok(provider),
             // Keep the router's error: it names the keys that ARE registered
@@ -1136,6 +1141,7 @@ fn resolve_provider(
             // but not what would have worked.
             Err(err) => {
                 warn!(
+                    node,
                     model = key,
                     %err,
                     "pipeline model not in provider router; using default provider"
@@ -1145,6 +1151,7 @@ fn resolve_provider(
         },
         (Some(key), None) => {
             warn!(
+                node,
                 model = key,
                 "model override but no provider router; using default"
             );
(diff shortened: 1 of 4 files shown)


anantheparty (approved):
Re-approving the updated head - the router-resolvable assignment fix review stands (the catalog superset no longer feeds unresolvable model keys to pipeline nodes). CI green. Approving - merging left to you per convention.

PR #2240, opened 2026-09-03
feat(peers): fenced peers reuse the workspace build cache via wt/.cargo/config.toml (#2236)

ZhangHanDong:
Fixes #2236

## Summary

A fenced peer (`peer_handoff` with `worktree=true`, or the #20a smart-fence default) is a fresh `git clone --no-hardlinks` under `peers/<slug>/wt` and therefore has no `target/`; on this workspace the first `cargo test` compiled from scratch (147 s / 500 s per invocation) and the peer burned its 50-iteration budget waiting on builds (OLP #45, 2026-09-04).

This PR lets a fenced peer reuse the workspace's warm build cache without touching the fence's `.git` isolation or the sandbox policy:

- When the fence is created and the workspace root has a `Cargo.toml`, write `wt/.cargo/config.toml` with an absolute `[build] target-dir = "<workspace_root>/target"` and add it to `wt/.git/info/exclude`, so the fence's `git status` stays clean and it can never be committed.
- If the repository ships its own `.cargo/config.toml`, leave it untouched and say so.
- Non-Cargo workspaces: no file written, no note.
- The decision is auditable: `model_note` gains a `build cache: …` line, and the `peer_staged` event detail becomes `peer staged (build cache: shared|none|repo-config)`; unfenced peers keep `peer staged`.
- No environment variables are injected; no new dependencies.

## Tests

Six contract-bound tests in `peers::tests` (`fenced_peer_gets_shared_target_dir_config`, `fenced_peer_without_cargo_toml_writes_nothing`, `fenced_peer_keeps_repo_cargo_config`, `unfenced_peer_untouched_by_build_cache`, `peer_staged_detail_reports_build_cache`, `fenced_peer_git_status_clean_after_config`) against real temporary git repositories, no cargo execution. Gates: `cargo test -p octos-cli --features api --lib peers` 38/0, clippy `-D warnings`, fmt; `agent-spec lifecycle` 6/6. Independently re-verified in an isolated worktree by the outer loop.

Contract: `specs/task-issue-2236-fenced-peer-build-cache.spec.md` (un-ignored in `.gitignore` per repo convention).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016zbAVTB1tkMF49Zr52UB7Z

PR #2310, opened 2026-09-12
fix: avoid redundant CLI rebuilds in Git worktrees

ZhangHanDong:
In a linked worktree, .git is a pointer file. Watching nonexistent .git/HEAD paths made Cargo rebuild the CLI even with no source changes. Cargo colors in CI also broke the regression tests' log detectors, allowing the old linked-worktree bug to pass undetected.

Resolve existing Git metadata paths through Git and keep the embedded hash fresh after commits and branch/detached switches. When a packed branch lacks a loose ref, watch its nearest existing parent; this avoids remote-ref fetch rebuilds when refs/heads or a narrower directory exists and still catches new branch-ref creation without relying on reflogs. Archives nested in another repository retain an empty hash.

Fixture Cargo commands force --color never. Freshness tests positively check the first compile, and a deliberately broken watcher must produce a detected Dirty/recompile. Real local-fetch tests cover ordinary and linked worktrees, disabled/removed reflogs, and missing nested branch directories. No global Cargo configuration or cache wrapper is changed.

Validation: fmt and cargo clippy -p octos-cli --features api --all-targets -- -D warnings passed. All 15 build regressions passed under outer CARGO_TERM_COLOR=always and never. Replacing build.rs with the original main version makes the corrected linked-worktree regression fail under both settings, demonstrating that it detects the original bug. Independent source review found no blocker in the delta.

Full cargo test -p octos-cli --features api --all-targets --no-fail-fast ran all 25 suites: 3,859 passed, 1 failed, 13 ignored. The only failure is the unchanged should_allocate_bridge_ports_as_consecutive_pair test: the existing Node service occupies port 3101. Its baseline failure on unmodified main f4e44b19 was recorded previously, and the service still owns that port. The previously flaky stdio test passed this run. Source hashes stayed unchanged during validation. The full-run failure is not presented as green.

Draft: GLM/K3 final cross-review of the updated head remains incomplete; the previous K3 attempt hit weekly quota HTTP 403. The automated fixtures and Codex source review are not substitutes for live dual-model review receipts.

### DIFF
--- a/crates/octos-cli/build.rs
+++ b/crates/octos-cli/build.rs
@@ -1,14 +1,70 @@
+use std::path::Path;
 use std::process::Command;
 
-fn main() {
-    // Git short hash
-    let hash = Command::new("git")
-        .args(["rev-parse", "--short", "HEAD"])
+/// Run `git -C <dir> <args>`; Ok(stdout) only when the command exited 0.
+fn git_out(dir: &Path, args: &[&str]) -> Option<String> {
+    Command::new("git")
+        .arg("-C")
+        .arg(dir)
+        .args(args)
         .output()
         .ok()
         .filter(|o| o.status.success())
         .map(|o| String::from_utf8_lossy(&o.stdout).trim().to_string())
-        .unwrap_or_default();
+}
+
+/// Resolve an ABSOLUTE, actually-existing git metadata path via
+/// `git rev-parse --path-format=absolute --git-path <name>`.
+/// Returns None when git is unavailable, the repo doesn't exist, or the
+/// resolved path doesn't exist on disk (e.g. packed refs have no loose
+/// ref file) — a rerun-if-changed on a nonexistent path makes cargo mark
+/// the unit dirty on EVERY build (the linked-worktree rebuild bug).
+fn existing_git_path(dir: &Path, name: &str) -> Option<std::path::PathBuf> {
+    let out = git_out(
+        dir,
+        &["rev-parse", "--path-format=absolute", "--git-path", name],
+    )?;
+    let path = Path::new(&out);
+    (path.is_file() || path.is_dir()).then(|| path.to_path_buf())
+}
+
+/// The current branch name from HEAD (`ref: refs/heads/<b>`), or None for
+/// detached HEAD / unreadable HEAD.
+fn head_branch(dir: &Path) -> Option<String> {
+    let head = existing_git_path(dir, "HEAD")?;
+    let text = std::fs::read_to_string(head).ok()?;
+    let branch = text.trim().strip_prefix("ref: ")?;
+    if branch == "HEAD" || !branch.starts_with("refs/heads/") {
+        return None; // detached
+    }
+    Some(branch.to_string())
+}
+
+/// This package belongs to `<project>/crates/octos-cli`. An archive nested
+/// in another repository must not inherit that outer repository's version.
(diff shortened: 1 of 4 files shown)


anantheparty (approved):
Adversarial pre-merge review passed — 0 blocker, 2 nits (non-gating). Verified beyond the diff: independent test rerun in a scratch package (15/15) plus the negative control (the old build.rs fails the linked-worktree test with cargo's own "Dirty probe: .git/HEAD is missing" error — the root-cause mechanism doubly confirmed), the packed-refs/prune/submodule/archive edge matrix (git 2.43 live simulations), and zero interaction with #2240/#2268 (different layers — runtime CARGO_TARGET_DIR injection vs build.rs watch paths).

Nits (follow-up): (1) build.rs:68 bakes CARGO_MANIFEST_DIR at compile time — a checkout+target mv silently empties the hash until clean/touch; the runtime env var is sturdier; (2) git <2.31 lacks --path-format=absolute → silent degrade to build.rs-only watching (2021-era git; a TODO suffices). Approving — merging left to you per convention.

BH3GEI (approved):
Thanks for the careful fix and the thorough regression suite.

What I independently verified (isolated worktree at 2a7c3b31):

- **Mechanism review**: worktree/detached/packed/pruned-ref watch resolution, the owning-repo guard (incl. the tracked-archive case), the nearest-existing-parent fallback, and the never-empty rerun set (build.rs self-watch) all check out, including unborn-branch and gc/pack-refs transitions.
- **Local tests**: `cargo test -p octos-cli --features api --test build_git_metadata` — 15/15, run twice; `cargo fmt --all -- --check` clean; `cargo clippy -p octos-cli --features api --all-targets -- -D warnings` clean.
- **Negative control**: swapping in main's original build.rs makes `build_git_watch_linked_worktree_second_build_fresh` fail (the nonexistent `../../.git/HEAD` watch visible in the fixture log); restoring this PR's script makes it green again — the regression genuinely detects the original bug.
- **Real-repo e2e**: in this repo's own linked worktree, a second no-change `cargo check -p octos-cli --features api` finishes in 0.3s with zero rustc invocations for the crate — the redundant-rebuild symptom is gone in the exact scenario the PR describes.
- **CI red triage**: `test-octos-cli` failed on `build_cache::pool::tests::transient_flock::inherited_flock_live_holder_and_wrong_token_remain_excluded` — the `<180ms` elapsed assertion at pool.rs:1629 on this head. The PR does not touch that module (OCTOS_GIT_HASH has no consumer in build_cache), and the test passes 21/21 consecutive local runs (0.02–0.04s each), so it reads as a runner-load timing flake rather than something introduced here; `CI required` is just the aggregation job.

Non-blocking follow-ups, no action needed in this PR:
1. `run_binary_hash` hardcodes `fixture-target/debug/probe` — an inherited `CARGO_BUILD_TARGET` (or `[build] target` in a developer config) would relocate the binary and trip the `.expect` in ~9 tests; an explicit `--target` would make it robust.
2. The fixture copy of build.rs resolves `rev-parse` with the developer's real git config while the expected hash comes from isolated-config git — a global `core.abbrev` would desync them; passing `GIT_CONFIG_*=/dev/null` through `cargo_vv` would close that.
3. The suite is not Windows-safe (`.exe` suffix, read-only git objects vs `remove_dir_all`) — fine while the Windows test lane stays opt-in, worth a note before enabling it.

PR #2311, opened 2026-09-12
test(autonomy): accept either tiebreak order when restart stamps share a millisecond (#2309)

### LINKED ISSUE #2309: flaky test: multi_epoch_scatter_persist_failure_then_restart_reemits_latest_join fails when the epoch-0 join and the ChildCompleted persist in the same millisecond
## Evidence

Two CI failures today with the identical signature, on two different commits:

- main `b5987df3` (run [34677520292](https://github.com/octos-org/octos/actions/runs/34677520292), job `test-octos-cli`): `left: ChildCompleted`, `right: ScatterJoinComplete` at `crates/octos-cli/src/autonomy/agent_orchestrator.rs:47099`
- PR #2308 branch (run [34678601505](https://github.com/octos-org/octos/actions/runs/34678601505), job `test-octos-cli`): same assertion, same values

Note `b5987df3` only touches release bundling scripts (`bundle-release.sh`, `release.yml`, `ci.yml`) — nothing near the autonomy orchestrator, and the same commit also has CI runs where the test passes, so this is not a deterministic regression.

## Root cause

`configure_supervisor_store` re-enqueues still-`Queued` records in the SF1 deterministic order sorted by `(join_epoch, queued_at_ms, continuation_id)`.

In this test, the store at restart holds exactly two `Queued` records:

- the epoch-0 `scatter_join/...` control row (persisted when agent-2 completed), stamp T1
- agent-3's `child/...` ChildCompleted (persisted when agent-3 completed), stamp T2

Both parse as `join_epoch = 0` (the epoch only ranks scatter records among themselves). T1 and T2 are taken microseconds apart by `now_ms_u64()`; on a fast host both stamps land in the **same millisecond**, so the sort falls through to the `continuation_id` tiebreak — `child/...` < `scatter_join/...` (`'c'` < `'s'`), which deterministically puts the ChildCompleted in turn 0. The test pinned the millisecond-separated order (scatter first) only.

Local instrumentation on an M3 Max shows T2 − T1 is consistently 1–3 ms (which is why the test is hard to reproduce locally in isolation), while CI Linux runners can persist the whole sequence in under a millisecond.

The SF1 determinism contract itself is intact — two fresh orchestrators over the same store still re-enqueue identically (the tiebreak is total and durable). Only the test's assumed causal order breaks.

BH3GEI:
## Problem

`multi_epoch_scatter_persist_failure_then_restart_reemits_latest_join` fails on CI with `left: ChildCompleted, right: ScatterJoinComplete` at the turn-0 assertion — observed twice with the identical signature on unrelated commits (main `b5987df3` run 34677520292 and the #2308 branch run 34678601505, both in `test-octos-cli`). Full analysis in #2309.

## Root cause

At restart the store holds exactly two `Queued` records: the epoch-0 `scatter_join/…` control row (stamped T1) and agent-3's `child/…` ChildCompleted (stamped T2), microseconds apart. Both parse as `join_epoch = 0`, and when T1/T2 land in the same millisecond the SF1 restart sort (`configure_supervisor_store`, key `(join_epoch, queued_at_ms, continuation_id)`) falls through to the `continuation_id` tiebreak, which deterministically orders `child/…` before `scatter_join/…` — flipping the order the test had pinned. The SF1 determinism contract itself is intact (the tiebreak is total and durable); only the test's assumed causal order breaks. Production code is unchanged by this PR.

## Fix

Two test-module changes in `crates/octos-cli/src/autonomy/agent_orchestrator.rs`:

1. The flaky test now accepts **either** delivery order for the two pre-crash pending rows across turns 0–1, while still asserting each turn is single-item, that the epoch-0 control row is among the first two turns with epoch 0, and the unchanged turn-3 assertions (epoch-1 re-emit, `terminal_children == 3`, exactly three turns). This is the "assert on the set of emitted outcomes + their per-record fields, not the emission order" direction from the triage.
2. New `restart_reenqueue_orders_equal_queued_at_ms_by_continuation_id` pins the previously untested contract: two `Queued` records with identical durable `queued_at_ms` (persisted out of delivery order — scatter first, child second) drain as `[ChildCompleted, ScatterJoinComplete]`, and a second fresh orchestrator over the same store re-enqueues identically.

## Tests

- New test green; fixture fields (incl. `priority`) match what production `queued_continuation_record` writes.
- Relaxed test green, plus 100/100 process-level reruns (each run re-rolls the T1/T2 millisecond boundary).
- Full suite: `cargo test -p octos-cli --lib` → 3693 passed, 0 failed (single run, no reruns needed); `cargo clippy -p octos-cli --all-targets -- -D warnings` and `cargo fmt --all -- --check` clean.
- End-to-end: test-only change, so the real-path evidence is the tests themselves — the new test drives the full disk-backed path (`SupervisorStore` append → `load_state` replay → `configure_supervisor_store` sort → scheduler drain), and an independent review pass additionally forced the same-millisecond case directly (temporarily quantizing the clock so both stamps share a millisecond): the child-first branch is real, and under it the turn-3 assertions still hold (3/3 runs), then the tree was restored bit-exact and re-verified.
- Independent review (full diff, adversarial pass): no blockers or majors; its one actionable note — the fixture omitted the `priority` metadata field that production records carry — is included above.

Closes #2309

### DIFF
--- a/crates/octos-cli/src/autonomy/agent_orchestrator.rs
+++ b/crates/octos-cli/src/autonomy/agent_orchestrator.rs
@@ -47088,28 +47088,45 @@ mod tests {
                 .expect("epoch segment")
                 .to_owned()
         };
-        // Turn 1: the restored epoch-0 scatter record re-enqueues (it
+        // Turns 1-2: the restored epoch-0 scatter record re-enqueues (it
         // persisted `Queued`; A never tombstoned it — only its MARKS were
-        // seeded) and delivers as a SUPERSEDED-epoch control row. This is
-        // the already-accepted SF1/Nit2 behavior (`restart_reenqueues_in_
-        // deterministic_epoch_order` drains persisted epochs 0/1/2; the Nit2
-        // dual-sink test allows one superseded epoch-0 control row).
+        // seeded) and delivers as a SUPERSEDED-epoch control row (#2309)
+        // beside agent-3's ChildCompleted (persisted Queued, never delivered
+        // pre-crash). Their relative order is stamp-resolution dependent: A
+        // persisted them microseconds apart, and when both `queued_at_ms`
+        // stamps land in the SAME millisecond the SF1 restart sort's final
+        // `continuation_id` tiebreak deterministically orders `child/…`
+        // before `scatter_join/…` ('c' < 's' — pinned by
+        // `restart_reenqueue_orders_equal_queued_at_ms_by_continuation_id`),
+        // so either delivery order satisfies the restored-queue contract.
         let turn0 = &turns[0];
         assert_eq!(turn0.len(), 1);
-        assert_eq!(
-            turn0[0].reason,
-            MasterContinuationReason::ScatterJoinComplete
-        );
-        assert_eq!(
-            epoch_of(&turn0[0]),
-            0,
-            "the superseded epoch-0 control row drains first (SF1 epoch order)"
-        );
-        // Turn 2: agent-3's ChildCompleted (persisted Queued, never
-        // delivered pre-crash) delivers.
         let turn1 = &turns[1];
         assert_eq!(turn1.len(), 1);
-        assert_eq!(turn1[0].reason, MasterContinuationReason::ChildCompleted);
+        assert!(
+            matches!(
+                (&turn0[0].reason, &turn1[0].reason),
+                (
+                    MasterContinuationReason::ScatterJoinComplete,
+                    MasterContinuationReason::ChildCompleted,
+                ) | (
+                    MasterContinuationReason::ChildCompleted,
+                    MasterContinuationReason::ScatterJoinComplete,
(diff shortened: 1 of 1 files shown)


anantheparty (approved):
Adversarial pre-merge review passed — the root cause is precisely identified and the fix matches the #2309 triage direction: when both stamps land in the same millisecond, the SF1 restart sort's continuation_id tiebreak deterministically orders child/ before scatter_join/ ('c' < 's') — so the CI failure is the deterministic sort hitting the millisecond collision, not random flake. The turn-0 assertion now accepts either tiebreak order (the recovery semantics — re-enqueue, superseded-epoch control row, ChildCompleted delivery — do not depend on the order), and the deterministic tiebreak itself keeps its own pinning test. Test-only. Approving — merging left to you per convention.

PR #2312, opened 2026-09-12
test(api): rebuild callsite interest cache in the http_trace test

anantheparty:
The deterministic fix for #2276 (root cause fully pinned in the issue thread — the four-step tracing-core/tracing-subscriber source verification).

**Root cause**: the shared debug_span! callsite is lazily registered on first execution; DefaultCallsite::register's JustOne branch only asks the CURRENT thread's default dispatcher, so a no-subscriber sibling test executing the router builder first caches Interest::never for the shared callsite. The victim test's with_default does not re-ask (the macro short-circuits on the cached never) → the trace span dies → the credential-omission assertion fails under full-parallel cargo test --lib. Solo runs cache correctly → always pass.

**Fix**: force rebuild_interest_cache() inside the victim's with_default closure — the current thread default = the victim subscriber, so both the JustOne and multi-dispatcher rebuilt paths deterministically re-ask. Plus the self-documenting assertion from the #2276 triage (the next occurrence, if any, prints the full captured logs).

Test-only, one hunk + one assertion upgrade. Refs #2276.

### DIFF
--- a/crates/octos-cli/Cargo.toml
+++ b/crates/octos-cli/Cargo.toml
@@ -50,6 +50,7 @@ eyre = { workspace = true }
 uuid = { workspace = true }
 color-eyre = { workspace = true }
 tracing = { workspace = true }
+tracing-core = "0.1.36"
 tracing-subscriber = { workspace = true }
 tracing-appender = { workspace = true }
 rustyline = { workspace = true }
--- a/crates/octos-cli/src/api/router.rs
+++ b/crates/octos-cli/src/api/router.rs
@@ -1477,6 +1477,12 @@ mod tests {
             .unwrap();
 
         tracing::subscriber::with_default(subscriber, || {
+            // #2276: this shared debug_span! callsite may have been lazily
+            // registered by a no-subscriber sibling test first (the JustOne
+            // rebuilder only asks the current thread's default), leaving a
+            // stale NEVER in the interest cache. Force a rebuild so the
+            // victim's DEBUG span is re-asked under this subscriber.
+            tracing_core::callsite::rebuild_interest_cache();
             runtime.block_on(async {
                 for uri in [
                     "/api/ui-protocol/ws?token=synthetic-query-marker%21&feature=chat",


anantheparty:
The tracing-core direct dependency is added (E0433 fixed - the router test calls tracing_core::callsite::rebuild_interest_cache() which needs the explicit dep). @ZhangHanDong @BH3GEI a quick approving review would unblock the merge - this is the #2276 flake root-cause fix.

BH3GEI (commented):
Thanks for putting this together — I verified the mechanism end-to-end against the lockfile-pinned tracing-core 0.1.36 source and it holds up, including a deterministic reproduction of the #2276 race (first-executing the shared `debug_span!` callsite from a subscriber-less thread after `with_default` entry kills the span; the rebuild inside the closure recovers it).

Two gaps I would want fixed before this lands, both quick:

1. **Missing Cargo.lock update.** `crates/octos-cli/Cargo.toml:53` adds `tracing-core = "0.1.36"`, but the lockfile entry for `octos-cli` does not list `tracing-core`. On a clean checkout of a05c52b6, `cargo metadata --locked` fails with "cannot update the lock file … because --locked was passed" (verified locally), which breaks every `--locked` flow including the nightly `cargo test --workspace --locked`. Since the lock already pins tracing-core 0.1.36, the fix is just committing the regenerated one-line addition to the `octos-cli` package entry (any local `cargo test -p octos-cli` produces it).

2. **The description promises an assertion upgrade that is not in the diff.** The body says "one hunk + one assertion upgrade … the next occurrence, if any, prints the full captured logs", but the victim assertions at `crates/octos-cli/src/api/router.rs:1521-1523` are still bare `assert!(logs.contains(...))` with no failure message (no `logs:\n{logs}` anywhere in the file). Given the #2276 recurrence protocol relies on capturing the full logs on the next occurrence, could we either add the `assert!(…, "logs:\n{logs}")` upgrade or amend the description to match the diff?

Smaller suggestion while you are in there: the only `tracing_core` usage is test code (`router.rs:1485`, inside `mod tests`), so the dependency would fit the crate's curated `[dev-dependencies]` section better than `[dependencies]` — as written it is a production dependency that only `cargo test` exercises. (Alternative worth knowing: `tracing` re-exports the module as `tracing::callsite`, though the re-export is `#[doc(hidden)]`, so the explicit dev-dependency is the more defensible API choice.)

What I verified on a05c52b6 in an isolated worktree: mechanism per tracing-core 0.1.36 source (`NoSubscriber::register_callsite` → `Interest::never()`, subscriber.rs:676-678; the `JustOne` rebuilder asks only the current thread default, callsite.rs:562-567; `Interest::and` yields `sometimes` rather than `never` on disagreement, subscriber.rs:658-664, so the multi-dispatcher path re-asks land on the victim subscriber at runtime); a deterministic A/B reproduction (poisoned callsite → span dead without the rebuild, recovered with it, run in separate processes); `cargo fmt --all -- --check` clean; the victim test passes solo; `cargo test -p octos-cli --lib` green in two clean runs; CI check/clippy and test-octos-cli jobs green on this SHA. One nit on the body wording: `with_default` does re-ask callsites already registered at entry (`Dispatch::new` → `register_dispatch` rebuilds them), so the surviving window is specifically a first execution of the callsite landing on a subscriber-less thread inside the closure — exactly the window the rebuild closes.

BH3GEI (commented):
Thanks for putting this together — I verified the mechanism end-to-end against the lockfile-pinned tracing-core 0.1.36 source and it holds up, including a deterministic reproduction of the #2276 race (first-executing the shared `debug_span!` callsite from a subscriber-less thread after `with_default` entry kills the span; the rebuild inside the closure recovers it).

Two gaps I would want fixed before this lands, both quick:

1. **Missing Cargo.lock update.** `crates/octos-cli/Cargo.toml:53` adds `tracing-core = "0.1.36"`, but the lockfile entry for `octos-cli` does not list `tracing-core`. On a clean checkout of a05c52b6, `cargo metadata --locked` fails with "cannot update the lock file … because --locked was passed" (verified locally), which breaks every `--locked` flow including the nightly `cargo test --workspace --locked`. Since the lock already pins tracing-core 0.1.36, the fix is just committing the regenerated one-line addition to the `octos-cli` package entry (any local `cargo test -p octos-cli` produces it).

2. **The description promises an assertion upgrade that is not in the diff.** The body says "one hunk + one assertion upgrade … the next occurrence, if any, prints the full captured logs", but the victim assertions at `crates/octos-cli/src/api/router.rs:1521-1523` are still bare `assert!(logs.contains(...))` with no failure message (no `logs:\n{logs}` anywhere in the file). Given the #2276 recurrence protocol relies on capturing the full logs on the next occurrence, could we either add the `assert!(…, "logs:\n{logs}")` upgrade or amend the description to match the diff?

Smaller suggestion while you are in there: the only `tracing_core` usage is test code (`router.rs:1485`, inside `mod tests`), so the dependency would fit the crate's curated `[dev-dependencies]` section better than `[dependencies]` — as written it is a production dependency that only `cargo test` exercises. (Alternative worth knowing: `tracing` re-exports the module as `tracing::callsite`, though the re-export is `#[doc(hidden)]`, so the explicit dev-dependency is the more defensible API choice.)

What I verified on a05c52b6 in an isolated worktree: mechanism per tracing-core 0.1.36 source (`NoSubscriber::register_callsite` → `Interest::never()`, subscriber.rs:676-678; the `JustOne` rebuilder asks only the current thread default, callsite.rs:562-567; `Interest::and` yields `sometimes` rather than `never` on disagreement, subscriber.rs:658-664, so the multi-dispatcher path re-asks land on the victim subscriber at runtime); a deterministic A/B reproduction (poisoned callsite → span dead without the rebuild, recovered with it, run in separate processes); `cargo fmt --all -- --check` clean; the victim test passes solo; `cargo test -p octos-cli --lib` green in two clean runs; CI check/clippy and test-octos-cli jobs green on this SHA. One nit on the body wording: `with_default` does re-ask callsites already registered at entry (`Dispatch::new` → `register_dispatch` rebuilds them), so the surviving window is specifically a first execution of the callsite landing on a subscriber-less thread inside the closure — exactly the window the rebuild closes.

anantheparty:
CI green on the current head (the tracing-core direct dep + the lock update + the assertion upgrade). @ZhangHanDong @BH3GEI - an approving review unblocks the merge (the #2276 flake root-cause fix).

Issue #2086, opened 2026-08-20
serve: all HTTP hangs (not just WS upgrades) when ui-protocol ledger data is present


alanpoon:
## Symptom

`octos serve` intermittently stops completing WebSocket upgrades on `/api/ui-protocol/ws`. The process stays healthy by every external sign — still `LISTEN`ing, still accepting TCP — but the HTTP upgrade never gets a response:

```
$ curl -sv --http1.1 -m 8 \
    -H 'Connection: Upgrade' -H 'Upgrade: websocket' \
    -H 'Sec-WebSocket-Version: 13' -H 'Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==' \
    -H 'authorization: Bearer local-dev-token' -H 'x-profile-id: alan' \
    http://127.0.0.1:3334/api/ui-protocol/ws

* Connected to 127.0.0.1 (127.0.0.1) port 3334
* Operation timed out after 8005 milliseconds with 0 bytes received
```

Only a restart clears it. Version: `octos 2.0.2 (7cec5e8c 2026-07-30)`, macOS 24.1.0, launched as
`octos serve --auth-token local-dev-token --port 3334 --data-dir ~/.octos-tui-data`.

## Two properties that make it awkward to spot

**An already-connected client keeps working.** A running octoscode TUI completed a full multi-tool turn (write a file, run bash, report output — 8 actions, 36s) while *new* connections could not get past the upgrade. So the server looks fine from the one place you are likely to be watching.

**It can also wedge mid-session.** Observed once: `session/open` returned normally, then `turn/start` on the same connection got no reply within 30s and never did.

## Not reliably reproducible

I could not find the trigger. Ruled out by direct test, each on a freshly restarted server:

| Hypothesis | Result |
| --- | --- |
| `session/open` wedges it | 6/6 handshakes fine afterwards; existing and brand-new session keys both reply in 0.1s |
| Connection churn | 24/24 sequential connect→close handshakes clean |
| Uptime | healthy at 15/30/60/90/120/150s idle |
| A concurrent `claude -p` subprocess | handshake ok immediately before and after |
| A connected TUI blocking other clients | still failed after killing the TUI |
| My Python client | `curl` times out identically, so it is not the client |

It appears within seconds sometimes (a restart wedged ~20s later with only a TUI reconnecting) and not at all other times (150s idle, then 24 clean handshakes).

## Why it matters

[...]

Issue #21, opened 2026-03-17
/queue command response preempted by speculative mode


ymote:
## Bug

From Telegram bot test report (scenes 03, 04): `/queue` returns long conversation content instead of the mode name.

### Symptoms
- Q-15: `/queue` should reply "Followup" but returns a long analysis
- X-04: `/queue` should reply "Speculative" but returns HTTP/2 vs HTTP/3 analysis

### Root Cause
In speculative mode, the previous message spawns a concurrent LLM call. This speculative response may arrive and be sent to the user between the `/queue` command handling and the `/queue` reply delivery. The outbound channel is FIFO — so the speculative response appears to "overwrite" the command reply.

`try_handle_command()` (session_actor.rs:712) runs before speculative processing, but the **previously spawned speculative task** is still running and may send its response at any time.

### Fix
When a slash command is detected and handled via `try_handle_command()`:
1. Cancel any in-flight speculative task
2. Or: tag command replies with priority so the outbound forwarder delivers them immediately

PR #2313, opened 2026-09-12
fix(cli): reap serve's gateway children on SIGTERM with a bounded drain (#2086)

BH3GEI:
## Problem

#2086 (comment "Separate finding"): killing `octos serve` with SIGTERM — what `pkill`, `systemctl stop`, and every supervisor sends — left every auto-started `octos gateway` child running. The reporter accumulated **38 orphaned gateways over ~15 restart cycles** (5 per start), each still holding its profile data dir, which both leaks resources unboundedly and mimics the issue's other failure mode (orphan contention makes the server listen but answer nothing).

Baseline on `main` (27ad221a), one enabled profile + `kill -TERM`:

- serve dies by signal (`signal=Some(15)`, exit code `None`) — the shutdown path never runs
- no `stopping all gateway child processes` in the tracing log — `stop_all()` is never reached
- the gateway child keeps running after the parent is gone

## Root cause

serve's graceful-shutdown future only caught SIGINT:

```rust
.with_graceful_shutdown(async {
    let _ = tokio::signal::ctrl_c().await;   // SIGINT only
    ...
})
```

SIGTERM fell through to the OS default disposition. Two secondary defects sat in the same path:

1. The handler was registered only when axum's shutdown future was first **polled** — i.e. after the gateway auto-start loop (each profile's startup health check holds that loop ~2s) — so even SIGINT could still hit the default disposition during boot, exactly when a restarting supervisor's kill lands.
2. `axum::serve(...).await?` returns early on a serve-loop error — past `stop_all()` — orphaning the gateways. And graceful shutdown waits for in-flight connections indefinitely: one open SSE stream (`GET /api/events/harness` never ends on its own) hangs a `systemctl stop` until `TimeoutStopSec` escalates to SIGKILL, which skips `stop_all()` the same way the signal death does.

## Fix (`crates/octos-cli/src/commands/serve.rs`)

- New `spawn_serve_shutdown_signal_watcher()` installs the SIGTERM listener **synchronously** (tokio only installs the handler when `signal()` runs, and a spawned task isn't polled until the runtime yields) and catches SIGINT-or-SIGTERM into a `watch` channel. It is called **before** the gateway auto-start loop, so the handler is in place before any child exists. The channel is stateful: a signal arriving before `axum::serve` polls its shutdown future is observed on the first poll, not lost.
- The auto-start loop checks the latch each iteration — once the stop signal is latched it stops **starting** new gateways (the in-flight `start()` completes; it cannot be interrupted mid-start).
- The serve future races a bounded drain: after the signal latches, `SERVE_SHUTDOWN_GRACE` (10s) for in-flight connections, then proceed to `stop_all()` + exit anyway rather than let a supervisor SIGKILL the process with the gateways still running.
- Both arms (clean shutdown and grace expiry) funnel into `stop_all()`. A serve-loop error no longer `?`-returns past it: it is logged, the children are reaped, and the exit code reflects the failure.
- Second-signal semantics: tokio keeps the handler installed for the process lifetime, so a second SIGTERM during the drain is captured-but-unobserved — it cannot kill serve before `stop_all()` runs (verified empirically: double SIGTERM 1s apart → shutdown completes, exit 0).

## Tests

New `crates/octos-cli/tests/serve_sigterm.rs` — unix-gated e2e driving the **real** octos binary, serial-only alongside `serve_broken_pipe` per the #21 outer-loop ruling:

1. `serve_sigterm_reaps_gateway_children` — steady state (HTTP serving, one gateway up): SIGTERM → parent exits **0, not by signal**, gateway gone and stays gone (3s settle window outlasts the 2s auto-restart), `stop_all` marker present in the tracing log.
2. `serve_sigterm_during_startup_reaps_gateways` — SIGTERM inside the auto-start window (first of four gateways up, second still in its startup check): pins handler-before-auto-start ordering (a handler installed only inside axum's shutdown future would still orphan here) and the latch break (at most 2 of 4 `auto-starting gateway` lines may appear).
3. `serve_sigterm_with_open_sse_stream_still_reaps_gateways` — open SSE stream + a 12s signal-free age floor (pins that the drain clock starts at the **signal**, not at boot — a plain `tokio::time::timeout` around `axum::serve` would kill a healthy idle serve): SIGTERM → exit 0 within the cap, gateway reaped.

Baseline red on pristine `main`: `serve must run its shutdown path under SIGTERM (exit 0), not die by signal (signal=Some(15))`.

CI wiring: the serial step becomes `--test serve_broken_pipe --test serve_sigterm`; every broad `cargo test` lane that runs the suite unskipped (test-octos-cli PR lane, check-arm, check-riscv, nightly check-arm, self-hosted fast-linux and linux-arm) adds `--skip serve_sigterm`. nightly check-windows and the self-hosted windows job are unchanged — the suite is unix-gated and compiles to zero tests there.

Validation (macOS aarch64, rustc 1.98.1, CI-exact commands):

- `cargo fmt --all -- --check` ✓
- `cargo clippy --workspace --all-targets -- -D warnings` ✓, `cargo clippy -p octos-cli --lib --no-default-features -- -D warnings` ✓, all-features all-targets matrix variant ✓
- `cargo test -p octos-cli --features api --test serve_broken_pipe --test serve_sigterm -- --test-threads=1` → **7/7** ✓
- `cargo test -p octos-cli --lib` → **3693/3693** ✓
- `cargo test -p octos-cli --tests -- --skip serve_broken_pipe --skip serve_sigterm` → all targets ✓
- Manual e2e beyond the suite: grace-expiry path with an open SSE stream logs `WARN shutdown grace period expired ... grace_secs=10` → `stopping all gateway child processes` → `gateways stopped count=1`, exit 0; double-SIGTERM 1s apart survives with exit 0.

## Review

An independent review pass over the full diff led to five applied changes: synchronous SIGTERM registration (before the watcher task spawns, closing a registration window the task-spawn form would leave), the auto-start latch break plus a bounding assertion in the startup-window test, the serve-loop error no longer bypassing `stop_all()` (exit code now reflects it), skip coverage extended to the nightly/self-hosted lanes, and a stale comment fix in `serve_broken_pipe`. One mechanism claim from the review (a second signal during drain killing the process) was rebutted with the manual double-SIGTERM e2e above; all review-driven changes were re-validated with the full command set.

Notes / explicitly out of scope:

- The 10s grace sits under typical supervisor budgets (systemd `TimeoutStopSec` default 90s, `docker stop` 10s) so the serve cuts an immortal stream itself instead of being SIGKILLed mid-drain.
- Pre-existing narrow race between `stop_all()` and the 2s auto-restart monitor for a gateway that crashes exactly during shutdown — predates this change, untouched here.
- The ledger-hang defect from the same issue (serve hangs with `ui-protocol/` content present) is **not** addressed by this PR.

Refs #2086 (the gateway-orphaning finding only).

### DIFF
--- a/.github/workflows/ci-self-hosted.yml
+++ b/.github/workflows/ci-self-hosted.yml
@@ -44,8 +44,11 @@ jobs:
       - name: Clippy
         run: cargo clippy --workspace --all-targets -- -D warnings
 
+      # The serve_sigterm e2e spawns real `octos serve` children that starve
+      # under default test parallelism; ci.yml runs it in a dedicated serial
+      # step (see the #21 outer-loop ruling there).
       - name: Test
-        run: cargo test --workspace
+        run: cargo test --workspace -- --skip serve_sigterm
 
       - name: API feature tests
         run: cargo test -p octos-cli --features api api::auth_handlers
@@ -83,7 +86,9 @@ jobs:
           cargo build --workspace
           cargo build -p octos-cli --features ${{ env.FEATURES }}
           cargo test --workspace --no-run
-          cargo test --workspace
+          # serve_sigterm e2e: serial-only (see the ci.yml #21 ruling); the
+          # broad run skips it.
+          cargo test --workspace -- --skip serve_sigterm
 
   windows:
     name: windows
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -444,16 +444,17 @@ jobs:
         run: cargo test -p octos-cli --lib
 
       - name: Test octos-cli (integration, excluding serve process e2e)
-        run: cargo test -p octos-cli --tests -- --skip serve_broken_pipe
+        run: cargo test -p octos-cli --tests -- --skip serve_broken_pipe --skip serve_sigterm
 
-      # Real-octos-process BrokenPipe/SIGINT e2e (serve_broken_pipe). These
-      # spawn real `octos serve` children that contend for model catalog /
-      # profile store resources with other test binaries; under the default
+      # Real-octos-process BrokenPipe/SIGINT e2e (serve_broken_pipe) and
+      # SIGTERM gateway-reap e2e (serve_sigterm). These spawn real
+      # `octos serve` children that contend for model catalog / profile
+      # store resources with other test binaries; under the default
       # thread-per-core parallelism the children starve and the 15s port
       # barrier times out. Dedicated serial step per the #21 outer-loop
       # ruling (2026-08-27): layer at the CI command level, never #[ignore].
-      - name: serve BrokenPipe process e2e (serial)
-        run: cargo test -p octos-cli --features api --test serve_broken_pipe -- --test-threads=1
+      - name: serve BrokenPipe/SIGTERM process e2e (serial)
+        run: cargo test -p octos-cli --features api --test serve_broken_pipe --test serve_sigterm -- --test-threads=1
 
       # Co-located here so they reuse this job's warm octos-cli build cache.
       - name: API feature tests
@@ -621,8 +622,13 @@ jobs:
         run: cargo test -p octos-agent --tests
       - name: Test octos-cli (lib)
         run: cargo test -p octos-cli --lib
-      - name: Test octos-cli (integration)
-        run: cargo test -p octos-cli --tests
+      # Same serve-process exclusions as the test-octos-cli PR job: the
+      # serve_broken_pipe / serve_sigterm e2e suites spawn real `octos serve`
+      # children that starve under default thread-per-core parallelism. They
+      # get their dedicated serial run in test-octos-cli; skipping here keeps
+      # this manual lane's coverage semantics identical to the PR lane's.
+      - name: Test octos-cli (integration, excluding serve process e2e)
+        run: cargo test -p octos-cli --tests -- --skip serve_broken_pipe --skip serve_sigterm
       - name: Test harness starter skills
         run: |
           cargo test -p harness-starter-generic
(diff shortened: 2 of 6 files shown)


anantheparty:
Adversarial pre-merge review verdict — **code: nit-level, no blocker; the architecture and evidence chain are the right fix for #2086.** The merge gate is OPERATIONAL: the required `test-octos-cli` integration step is red from the `http_trace_uses_route_templates_and_omits_credentials` flake — which is **unrelated to this PR (diff touches no router.rs) and is now ROOT-CAUSED AND FIXED in #2312** (the callsite interest cache staleness; `rebuild_interest_cache()` in the victim's with_default closure).

Verified clean: the SIGTERM handler registered synchronously before auto-start (killing the N×2s default-disposition window — the #2086 root), the drain clock starting at signal (the `timeout(grace, serve)` anti-pattern warning is accurate), SSE hard-cut at 10s still reaching stop_all, no zombie risk (serve exits → init reaps; monitor wait + tokio SIGCHLD reaper), zero new fork-exec windows vs #2270 (pure sigaction registration), the non-SIGTERM exit path now also funnels into stop_all (fixing the old `.await?` harvest-skip hole), and the three SIGTERM e2e tests use REAL processes + REAL libc::kill with pre-fix-red assertions and the timing discipline per #2288/#2252.

Sequence proposal: **land #2312 (the flake root-cause fix) → this PR's CI re-runs green → merge.** The red flake should not block this PR once its own root-cause fix lands.

Nits (follow-up ledger, priority on Nit-1): (1) the harvest is fire-and-forget `/bin/kill` with no wait/ladder/verification (recommending libc::kill + bounded wait + SIGKILL escalation + result verification per the #2272 fail-closed discipline — also removes the /bin/exists dependency); (2) the profile watcher/monitor/auto-restart don't know the signal latch (µs-level escape window, mem::take catches most); (3) the http_listener expect/? early-exit paths bypass the harvest (pre-existing, theoretical); (4) SIGINT lazy-registration asymmetry (ms-level window, worth a comment); SIGHUP uncovered (acceptable).

anantheparty (approved):
Re-approving the updated head - the SIGTERM reaping review stands (the synchronous pre-auto-start registration, the bounded drain clocked from signal, the stop_all funnel on non-signal exits, and the three real-process SIGTERM e2e tests). CI green. Approving - merging left to you per convention.

PR #2317, opened 2026-09-12
fix: bind native peer reports to their runtime turn identity

ZhangHanDong:
A completed peer turn can lose its best-effort `result-N.md` / `turns.txt` writes. The next persisted file then has a different ordinal from the runtime ledger, so review tooling cannot safely use the file number to identify the model that produced it.

This change writes the existing runtime `TurnId` into both native result frontmatters. Both completed and errored terminal call sites pass the actual ID. File numbering, the turns index and the existing safe-write behavior stay unchanged. The verifier consumer is octos-org/octoscode#637.

The native writer roundtrip test now checks distinct completed/errored IDs and forces real version/index write failures using test-owned directories. After recovery, `result-1.md` contains the later actual ID rather than the lost turn's ID. A behavior contract accompanies the change.

Validation: fmt, API/all-targets Clippy with `-D warnings`, the native writer roundtrip, and production binary build pass. Complete API/all-targets run: 3,851 passed, 1 failed, 13 ignored. The sole failure is the pre-existing bridge-port test assuming 3101 is free; an existing host Node service occupies it, so allocation correctly moves to 3103. That service was preserved. All integration targets pass. CI is still running.

Two distinct native GLM peers completed independent initial and cross reviews on the exact source hashes; both found no blocker. Four actual GLM model / native turn-ID anchors were re-verified from the archived ledgers after shutdown. The review comment includes the IDs, receipt digest, actual report-ordinal / ledger-ordinal mismatch, and remaining nonblocking notes. The default product GLM/K3 gate is unchanged. The native goal completion verifier separately rejected completion for incomplete ledger coverage / operator-approval wording; it remains rejected/active and was not counted as completed.

Old native reports are not migrated or backfilled. The consumer requires a new review context and new reviewers on the updated runtime when `turn_id` is missing.


ZhangHanDong (commented):
Final review of `63ef8643`: both terminal writer paths now emit their real runtime TurnId. The real writer/gather roundtrip verifies completed/error IDs and actual failed version/index writes followed by recovery; the recovered file keeps the later turn's ID. Consumer: octos-org/octoscode#637.

Validation: fmt, API/all-targets Clippy with `-D warnings`, targeted native writer test, and production binary build pass. Complete API/all-targets run: **3,851 passed / 1 failed / 13 ignored**. The sole failure is the pre-existing `process_manager::tests::should_allocate_bridge_ports_as_consecutive_pair`, which assumes port 3101 is free; existing host Node PID 67790 occupies it, so allocation correctly chooses 3103. The unrelated service was preserved. All integration targets passed, including the seven API feature gates. CI is still running at this comment time.

Two distinct native GLM peers completed independent initial reviews and then cross-reviewed the frozen reports on the exact source snapshots for octoscode `0278373b` and octos `63ef8643`. Both cross verdicts are **no blocker**. This is the operator-authorized GLM×2 review method; the product's default GLM/K3 acceptance gate is unchanged and is not claimed to pass from these same-model receipts.

| Peer | Phase | Actual native / ledger turn ID | Result ordinal / ledger ordinal | Actual model |
|---|---|---|---|---|
| glmfix-a | initial | `4cf3db03-0d99-4f7d-b30a-b4a45272397c` | 1 / 2 | `glm-5.3` |
| glmfix-a | cross | `63650e3b-6169-47d5-887a-b5184f02713d` | 2 / 3 | `glm-5.3` |
| glmfix-b | initial | `aed64a82-132b-4057-a59e-2bf26d1a1229` | 1 / 2 | `glm-5.3` |
| glmfix-b | cross | `0d001546-7e00-4c64-8e96-8157e3c5a5a8` | 2 / 3 | `glm-5.3` |

All four report IDs were checked against the complete raw session ledgers, terminal completion and actual `token_cost_update` model events. Initial hashes were frozen before cross. After clean shutdown of only the dedicated review runtime, the archived ledgers and native reports were re-verified with the same verifier. Aggregate receipt SHA256: `bf61eedb1dce020dff33fcf6302002f7db59b8a7958296685e9ad10299035855`.

The first independent turns encountered a temporary host profile setup error and produced no native report. After fixing only that temporary config and restarting its runtime, each peer independently confirmed unchanged sources and issued a fresh native initial. No old IDs were backfilled and no ledger prefix was discarded or renumbered. Thus the admitted result-1 / ledger-turn-2 mismatch is real, not a synthetic receipt.

Remaining review notes are nonblocking: missing-cross / corrupt-leaf messages could be more specific; frozen absolute paths do not survive relocation; legacy reports need a new review context; host/state integrity remains the trust boundary. `turns.txt` still gates terminal authority. Ordering uses ledger start order, not UUID comparison.

Process limitation: the native goal completion verifier rejected completion because it recognized only part of the ledger coverage and demanded operator approval of the conclusion. Its rejected/active state was preserved; it was not counted as goal completion or as a product acceptance-gate pass. This does not substitute for or alter the four directly re-verified raw model/identity receipts above.

anantheparty (approved):
Adversarial pre-merge review passed - both terminal writer paths emit the real runtime TurnId into the native reports (result-N.md carries the turn_id field, turns.txt the versioned entries), making the file ordinal safely derivable from the runtime ledger - the foundation #637's review tooling consumes. The failure-injection tests use the dirs-as-files trick (no root/platform assumptions) and pin the recovery. Approving - merging left to you per convention.

PR #2327, opened 2026-09-14
fix(serve): initialize profile inbox before steer sweep

alan0x:
## Summary

- create each profile inbox during the canonical ProfileRuntime bootstrap
- prevent a never-used inbox from producing a WARN every five seconds
- preserve the existing loud sweep failure if the directory disappears after startup or becomes unreadable
- add a regression test for nested profile data directories

## Verification

- cargo test -p octos-cli should_create_inbox_when_ensuring_profile_runtime_directories
- cargo test -p octos-cli olp_ctrl_steer
- cargo fmt --all -- --check
- cargo clippy --workspace
- cargo test --workspace: 2896 octos-agent tests passed; 5 unrelated environment-sensitive tests failed because this host does not provide the expected Docker sandbox and resolves the nonexistent-domain fixtures to a private IP


BH3GEI (approved):
Thanks for putting this together — clean diagnosis, and the fix lands at exactly the right funnel point.

What I verified:

- **Live A/B against main.** Built both `origin/main` (c5bae543) and this head (3009ee2), ran `octos serve` ~15s each against a scratch instance dir with one enabled deepseek profile. On main: `WARN steer inbox sweep: failed to read inbox dir ... kind: NotFound` on the 5s drain cadence (GLOBAL_MASTER_CONTINUATION_DRAIN_INTERVAL_SECS = 5, ui_protocol_transport.rs:23350) and no `profiles/<id>/data/inbox` ever created. On this head: zero such warnings, a successful `steer inbox sweep tick ... sidecars=0`, and the inbox dir exists after bootstrap.
- **Coverage of the sweep's path.** `ensure_profile_runtime_directories` sits at the top of `bootstrap_resolved`, which every production `ProfileRuntime` construction funnels through (serve.rs:1027, gateway_runtime.rs:433, local_oup.rs, ui_protocol_transport.rs); the sweep reads `profile_runtime.data_dir`, set from that same parameter, so the created dir is exactly the one read_dir sees. Producers still create the inbox defensively before writing, and the sweep's WARN for post-startup disappearance is untouched — matching the doc comment.
- **Tests.** Full `cargo test -p octos-cli --lib`: 3695 passed / 0 failed, twice (exactly +1 over main = the new regression test); `olp_ctrl_steer` 12/12, `steer_integration` 1/1, profile module 23/23; `cargo fmt --all -- --check` clean. `cargo clippy --workspace --all-targets -- -D warnings` hits only api/handlers.rs:2024 (`result_large_err`) — that file is byte-identical on main, so it's pre-existing under local clippy 1.98.1, not from this change. One full-suite run tripped a pre-existing flake in `stdio_connection_stops_dispatch_after_writer_failure` (asserts on the process-global `STDIO_DISPATCH_COUNT_FOR_TEST` atomic under parallel execution; that test never runs profile bootstrap) — passes solo and in two subsequent full runs. CI on this head is green across the matrix, including test-octos-cli and check-windows.

One deliberate semantic edge worth knowing (fine as-is): if `<data_dir>/inbox` exists as a non-directory, bootstrap now fails fast instead of serving with the side-channels broken — consistent with the new # Errors contract. Non-blocking suggestion for a follow-up: adding `inbox` to the seed list in `ProfileStore::save` (profiles.rs:1798) would keep the two directory-seeding sites from drifting further apart.

PR #2332, opened 2026-09-14
test(windows): unbreak main's check-windows lane — 23 deterministic failures

BH3GEI:
## Problem

**Build break on main.** The `check-windows` job (`cargo test -p octos-cli --lib` on windows-latest) has failed on every recent main CI run (e.g. runs 34788267908, 34760927364, 34744870100), taking `CI required` red with it. The failure set is stable and deterministic — the same 26 tests across 8 modules, not flakes. They slipped in because PR lanes skip check-windows via path filters, so each Unix-only assumption broke main only after merge.

## Root causes (verified against the failing job logs)

1. **Lock contention misclassified (6 tests)** — `build_cache/pool.rs`: `try_lock_slot` only recognized `WouldBlock` as contention. Windows `LockFileEx` fails an already-locked region with `ERROR_LOCK_VIOLATION` (os error 33), which std never maps to `WouldBlock`, so ordinary slot contention surfaced as a hard Io error (`failed to access any candidate slot` / `failed to lock …\.lock`) instead of `PoolExhausted`/`Locked`.
2. **Space gate measured the wrong filesystem (2 tests)** — fs2's Windows `available_space` resolves the volume root from the path *string* only; for a deleted pool root it happily measured C:\ (30GB free < 50GB → `FreeSpaceLow`). Unix `statvfs` reports ENOENT, which the gate maps to fail-closed `FreeSpaceUnknown`.
3. **git rejects `\\?\`-prefixed paths (2 tests)** — `raw_peer_prepare` used `std::fs::canonicalize`, which returns `\\?\C:\…` on Windows; `stage_peer` fed it to `git clone`, which parses it as a URL (`hostname contains invalid characters`). Peer prepare with a worktree was therefore *always* broken on Windows — a real product bug, not a test artifact.
4. **Wire paths leaked `\` (2 tests)** — pane workspace/artifact entries used `to_string_lossy()` verbatim, producing `src\main.rs` on Windows while the wire convention is `/`.
5. **Unix-only test assumptions (11 tests)** — admin shell tests need `sh -c` (the handler hardcodes it); durable peer writes intentionally fail closed off Unix (`File::open` on a directory is `ERROR_ACCESS_DENIED` on Windows); dead-holder GC depends on `kill(pid,0)`, whose non-Unix stub conservatively reports every pid alive; the ledger corruption test's Windows fallback (directory named like a log file) is exactly the case `list_log_files` silently skips, so nothing latched.
6. **Keychain relocation test (1 test)** — its test-store injection is `cfg(linux)`; on Windows the relocation fails before the profile save, so the asserted env slot never exists.

## Fix

- **pool.rs**: new `is_lock_contention` (`WouldBlock` or fs2's platform `lock_contended_error()`, the same helper `commands/serve.rs` and `memory_refresh/service.rs` already use); contention leaving `try_lock_slot` is normalized to `WouldBlock` so all three call sites classify identically on every platform. `space_gate` fails closed when the pool root does not exist (`acquire` creates it before gating, so production behavior is unchanged; TOCTOU still lands in `FreeSpaceUnknown`).
- **peers/mod.rs + ui_protocol_transport.rs**: `dunce::canonicalize`/`dunce::simplified` (same crate and usage as octos-agent) on every path handed to git — clone src/dst, `-C` for checkout/config, identity carry-over read, and `fence_collision_reasons`' `symbolic-ref`.
- **ui_protocol_transport.rs**: `wire_relative_path` helper — `/` separators on Windows, native elsewhere (backslash is a legal filename character on Unix).
- **Tests**: Unix-only-by-design tests gated to `cfg(unix)` with reasons in comments; the cache detached-slot test split into a cross-platform live-holder half and a `cfg(unix)` dead-holder half; the ledger test now poisons by swapping the session dir for a same-named regular file *after* boot indexing — `read_dir` fails with `NotADirectory` (not the tolerated `NotFound`) on every platform, so the failure-latch path is exercised on Windows too instead of gated away. `my_profile_rejects_…_off_macos` gated to linux; Windows reject-path coverage remains via the `admin.rs` twin test.

## Verification

- Local (macOS): `cargo test -p octos-cli --lib` **3696/3696 pass** (incl. new `lock_contention_recognizes_platform_error_shapes` unit test), `cargo clippy -p octos-cli --all-targets` no new warnings (one pre-existing `result_large_err` in untouched handlers.rs), `cargo fmt --check` clean.
- Every previously-failing test that can run on Unix passes locally (pool/cache/peers/orchestrator/admin/ledger/transport families run and green).
- Windows-only behavior can't execute on this host (no Windows toolchain for the ring C dependency), so the authoritative end-to-end check is this PR's own `check-windows` lane.
- Independent review pass (picky-maintainer subagent, given only the failure evidence + diff): no Blockers; its Major (read-side git calls in `stage_peer`/`fence_collision_reasons` still using the prefixed path) and all Minors are incorporated.

## Expected check-windows result

This PR fixes **23 of 26** failures. The remaining 3 (`api::auth_handlers` voice-readiness fixture, `invalid escape` from interpolating Windows paths into JSON via `format!`) are owned by open **#2261**, which rewrites that exact fixture with `serde_json::json!` — verified: #2261's own check-windows run fails *only* on the 4th test, which this PR gates. Once both land, check-windows is fully green. `CI required` will stay red on this PR until then solely due to those 3.


anantheparty (approved):
Verified against the failing nightly logs (octos#2267 Night-6): the deterministic set this fixes is real, the root causes hold (lock-contention misclassification in build_cache/pool; the env-dependent class), and the check-windows lane is green on this PR with `CI required` restored.

One follow-up worth a tracking note: `matrix_channel::tests::test_handle_room_query_requires_token` (the Night-6 single failure — reqwest to a test-local port got ConnectionRefused, winsock 10061) did **not** recur on this PR's run, consistent with a startup/port race flake rather than the deterministic class. It will resurface under load; a wait-for-ready on the test-local listener (or a fixed-to-available port pattern) would close it. Happy to take that as a follow-up if you'd rather keep this PR scoped.

PR #2343, opened 2026-09-15
Default Linux manual builds to main

chrislearn:
## Summary

- change the manual Linux build workflow's default ref from the stale feature branch to main
- retain the ref input so callers can still build another branch or tag explicitly

## Validation

- git diff --check


BH3GEI (approved):
Thanks for putting this together.

Verified before approving:
- `fix/pipeline-stream-errors` no longer exists on the remote (`git ls-remote origin` has no matching branch or tag), so the old default made a no-input manual dispatch fail at the checkout step; `main` is present and is the repo default branch, so the new default resolves correctly.
- The `ref` input is untouched (still optional, same description) and its only consumer is the `actions/checkout@v6` step, so callers passing an explicit branch or tag are unaffected.
- `actionlint` passes on the modified workflow and the YAML parses with the expected structure; the change is a single line in a single file, matching the stated scope.
- This matches the existing convention in `ci-self-hosted.yml`, whose equivalent dispatch input already defaults to `main`.
- Nothing else in the repo references `fix/pipeline-stream-errors`, and since the branch is long gone, no caller could have been relying on the old default without already failing.

PR #2347, opened 2026-09-15
fix(dashboard): route WeChat QR login to the profile being edited

### LINKED ISSUE #1440: Bug: WeChatTab Ignores the Current Profile and Starts QR Login on the Wrong Account
**Severity:** High | **Platform:** Dashboard

## Symptom

When an administrator edits a profile other than their own in the Dashboard messaging settings, opening the WeChat tab and starting QR login still calls `/api/my/profile/wechat/...` instead of the admin route for the profile currently being edited.

## Code Location

- `dashboard/src/pages/profile/MessagingPage.tsx:26` already reads `profileId` from `useProfile()`.
- `dashboard/src/pages/profile/MessagingPage.tsx:54` renders `<WeChatTab>` without passing `profileId`.
- `dashboard/src/components/tabs/WeChatTab.tsx:41` and `:59` compute `const id = profileId || "admin"`, but that value is not used in the request URL.
- `dashboard/src/components/tabs/WeChatTab.tsx:42` and `:60` always call `/api/my/profile/wechat/qr-start` and `/api/my/profile/wechat/qr-poll`.
- The backend already exposes admin profile routes in `crates/octos-cli/src/api/router.rs:352-357`: `/api/admin/profiles/{id}/wechat/qr-start` and `/api/admin/profiles/{id}/wechat/qr-poll`.

## Impact

The user is viewing and editing profile A, but the WeChat QR login can be started for the currently authenticated user's `/api/my` profile instead. This can attach the WeChat token, login state, or QR session to the wrong profile. A later gateway start may then use the wrong WeChat identity.

## Reproduction Steps

1. Sign in to the Dashboard as an administrator.
2. Open a profile page for a profile that is not the current user's profile.
3. Go to Messaging / WeChat settings.
4. Start the QR login flow.
5. Inspect the Network request. The actual URL is `/api/my/profile/wechat/qr-start`, not `/api/admin/profiles/{profileId}/wechat/qr-start`.

## Expected Behavior

When the page is editing a specific profile, WeChatTab should start QR login and polling for that profile. It should only use `/api/my/profile/wechat/...` when the current user's own profile is being edited.

## Suggested Fix

1. Pass `profileId` from `MessagingPage.tsx` into `<WeChatTab config={config} onChange={setConfig} profileId={profileId} />`.
2. Build the WeChat request URL in `WeChatTab.tsx` based on `profileId`:
   - When `profileId` exists, use `/api/admin/profiles/${profileId}/wechat/qr-start` and `/qr-poll`.
   - When `profileId` is absent, use `/api/my/profile/wechat/qr-start` and `/qr-poll`.
3. Remove the unused `id` variable to avoid misleading future maintainers.
4. Add a frontend regression test for the admin-editing-profile path.

## Verification

- Trigger WeChat QR start from an admin profile page and assert the request URL contains `/api/admin/profiles/{id}/wechat/qr-start`.
- Poll for QR status and assert the request URL uses the same `{id}`.
- Trigger QR login from the current user's own profile and assert it still uses `/api/my/profile/wechat/...`.

BH3GEI:
## Problem

#1440: on the Dashboard messaging page for a profile other than the signed-in admin's own, the WeChat tab's QR login flow called `/api/my/profile/wechat/qr-start` and `/qr-poll` — so scanning the QR attached the WeChat bot token and login state to the admin's own profile instead of the one being edited.

## Root cause

Two stacked defects:

1. **Frontend**: `MessagingPage.tsx` read `profileId` from `useProfile()` but rendered `<WeChatTab>` without it, and `WeChatTab.tsx` hardcoded the `/api/my/profile/wechat/*` URLs (its `const id = profileId || "admin"` was computed and never used).
2. **Backend**: the admin `POST /api/admin/profiles/{id}/wechat/qr-poll` handler named its path parameter `_id` and never used it — on confirm it only wrote the token to a shared `/tmp/octos-wechat-token` file, so the token never landed on the named profile even if the frontend had called the right route.

## Fix

- `MessagingPage` passes `profileId={isOwn ? undefined : profileId}` to `WeChatTab`: the own-profile page keeps the user-scoped `/api/my` route (a non-admin cannot call `/api/admin/*`); an admin editing another profile gets `/api/admin/profiles/{id}/wechat/*` (URL-encoded).
- Admin `wechat_qr_start`/`wechat_qr_poll` validate the path id via a shared `require_admin_profile` (404 `profile '{id}' not found`, same convention as the other admin handlers) instead of reporting "confirmed" while the token lands nowhere.
- A shared `persist_wechat_bot_token` helper now lands the confirmed token on the named profile for **both** routes (ensure WeChat channel, set `WECHAT_BOT_TOKEN`, save), with persistence failures logged instead of silently swallowed. The `/tmp/octos-wechat-token` write is removed: a repo-wide search confirms nothing ever read that file, and it was a symlink-following write of a live credential into a world-writable directory.
- `WeChatTab` stops the polling interval when the edited profile changes (the component stays mounted across `/profile/A` → `/profile/B`), and on confirm mirrors the server-pushed WeChat channel into local config state — otherwise a later click on Save would PUT the stale `channels` array and delete the channel the QR flow just created.

## Test evidence

- **Red → green, page level**: new `MessagingPage.test.tsx` fails on the base commit exactly as the issue describes (admin editing `prof-a` → request goes to `/api/my/profile/wechat/qr-start`) and passes after (goes to `/api/admin/profiles/prof-a/wechat/qr-start`; own page still uses `/api/my`).
- **Component level**: new `WeChatTab.test.tsx` pins the own/admin/URL-encoded routes, the poll carrying the session key to the same admin route, and the confirm-mirror `onChange`; 3 of 4 failed on base.
- **Backend**: 3 new `persist_wechat_bot_token` unit tests over a real on-disk `ProfileStore` (targets the named profile only, repeat confirm doesn't stack channels, empty token/unknown profile no-op) + 2 handler-level tests asserting 404 for unknown profiles. `cargo test -p octos-cli --lib`: **3711 passed, 0 failed**, first run; CI's named filter steps `api::auth_handlers` (93) and `api::admin` (86) green.
- **Dashboard**: `vitest run` **61/61**, `tsc -b` clean, `vite build` green. (Local Node 26 needs `NODE_OPTIONS=--localstorage-file=...` for any dashboard test — Node's built-in `localStorage` shadows jsdom's; CI pins Node 22 and is unaffected. Under that workaround the files must run with `--no-file-parallelism`, or parallel workers race on the shared storage file — a harness artifact, not the code.)
- `cargo fmt --check` clean; `cargo clippy -p octos-cli --all-targets` shows only the pre-existing `result_large_err` at `api/handlers.rs:2024` (untouched by this diff).

**End-to-end note**: a full HTTP e2e of the confirm path is not automatable — both handlers hardcode the upstream `https://ilinkai.weixin.qq.com` and confirmation requires a human scanning the QR with WeChat. Substitute evidence: the frontend tests drive the real component through the real `fetch` call sites and assert the exact request URLs/bodies before and after; the backend tests exercise the real persist path against a real on-disk store; the route registration and auth gating (`admin_auth_middleware`) are unchanged by this diff.

## Review

One independent adversarial review pass: REQUEST_CHANGES with 4 majors, all addressed — unknown-profile 404 (both admin handlers), persistence failures logged, Save-clobbers-channel fixed via the confirm-mirror, and the dead `/tmp` credential write removed (verified zero readers). Pre-existing items intentionally left out of scope: empty-`bot_token` confirm semantics, admin qr-start not using the bridge path, the 200-wrapped timeout, and disconnect not clearing the stored token. All verification re-run green after those changes.

Closes #1440


BH3GEI:
Rebased onto latest main to resolve the conflict with #2349 (both PRs had appended tests at the end of the same module in `crates/octos-cli/src/api/admin.rs` — kept both sides). Re-verified after the rebase: the 6 affected lib tests pass (wechat_qr + platform_service_logs families), `cargo fmt --check` clean, dashboard 61/61 + `tsc -b` green.

PR #2363, opened 2026-09-16
ci: fail api-feature test steps whose name filter selects zero tests (#2029)

BH3GEI:
## Problem

Tests behind the `api` feature only run in CI when their name matches a hand-written substring filter in the workflow. A filter that matches nothing does not error: `cargo test <filter>` prints `running 0 tests / test result: ok` and exits 0, so a renamed test silently drops out of CI while its step reports green. #2029 documents three instances of this; the `usage` and `commands::serve::tests` clusters were two more, caught by hand.

## Root cause

`api` is not a default feature, so coverage of that module comes entirely from named filter steps in `ci.yml` (and the same pattern in `ci-self-hosted.yml`). Nothing validates that a filter actually selects anything — a dead filter is indistinguishable from a passing step.

## Fix

Add `scripts/ci-api-filter-test.sh` and route all 13 api-feature filter steps (10 in `ci.yml`, 3 in `ci-self-hosted.yml`) through it. For each filter the script:

1. asserts `cargo test -p octos-cli --features api -- <filter> --list` selects at least one test, else fails the step with an `::error::` naming the filter and pointing at the fix;
2. runs the union of the step's filters exactly as the unguarded command did (same package/features, libtest args like `--nocapture` passed through and mirrored into the guard's selection);
3. asserts at least one test binary reports a non-zero `passed` count, so a selection surviving only as `#[ignore]`d entries cannot masquerade as coverage either (asserted on the `passed` count rather than the `running N` line, whose format for ignored-only selections varies across toolchains).

This covers the "zero-match guard" acceptance item the issue marks "(independent, do now)". It does not attempt the full-suite replacement — that track stays blocked on the parallel-interference isolation discussed in the issue. Two adjacent unguarded shapes remain and are out of scope here: the three `octos-llm` QoS named-test steps, and `--test <target>` steps whose binary could compile to zero tests.

One deliberate semantic note: the run uses the libtest positional form (`cargo test ... -- <filter>`), which — unlike cargo's positional TESTNAME — would also select matching doctests. octos-cli currently has zero runnable doctests, so selection is identical today (verified: `usage` selects 20 tests under both forms); any future matching doctest is additive coverage.

## Test evidence

- Baseline reproduction on origin/main: `cargo test -p octos-cli --features api renamed_away_filter_xyz` → every binary `running 0 tests ... ok`, exit 0 (the silent-green failure mode).
- Same filter through the guard → exit 1, `::error::api-feature CI filter 'renamed_away_filter_xyz' matches zero tests ...`.
- Ignored-only selection (real `#[ignore]`d test `ledger_soak_10k_events_10_sessions`): `--list` guard passes (count 1), run-layer assertion fires → exit 1.
- Mixed live+dead filter pair (`usage renamed_away_filter_xyz`) → exit 1 naming the dead one (per-filter attribution; the old union form let one of two filters die silently).
- No args → usage text, exit 2. Compile failure in the guard propagates cargo's exit and diagnostics, never misreports as zero-match.
- End-to-end: all 13 modified workflow steps replayed locally exactly as CI invokes them (extracted from the YAML, bash word-splitting as Actions does) — every guard passed and every step executed its real tests: 93 / 1 / 7 / 3 / 20 / 259 / 24 / 6 / 3 / 4 in ci.yml, and the three ci-self-hosted steps. All green in a single pass.
- `shellcheck` clean; `actionlint` on both workflows reports only the repo's pre-existing custom-runner-label warnings (identical on origin/main).

## Review

- Round 1 self-review: diff limited to the two workflows + new script; behavioral equivalence of each replaced step checked line by line.
- Round 2 independent review (fresh agent, issue text + diff only): REQUEST_CHANGES with 1 Major — the `--list`-only guard counted `#[ignore]`d tests, so an ignored-only selection would still report green without executing anything (the exact "passes without meaning" property the issue warns about) — plus 4 Minor / 1 Nit. All adopted: run-layer passed-count assertion added and verified against a real ignored test; guard switched to the libtest-positional selection form to mirror the run; stdout-only counting; libtest args mirrored into the guard; error message no longer names a specific workflow file. One Minor (unguarded octos-llm / `--test` shapes) rejected as out of the issue's revised acceptance scope and disclosed above instead.
- Round 3: after those changes, all negatives re-verified and the full 13-step replay rerun — all green.

Refs #2029


BH3GEI:
The `test-octos-cli` failure on the previous run (`build_cache::pool::tests::transient_flock::inherited_flock_live_holder_and_wrong_token_remain_excluded`, "live unrelated ownership must not wait through the retry window") is unrelated to this PR's diff, which only touches `scripts/ci-api-filter-test.sh` and the two CI workflow files — no Rust code.

Investigation notes:
- The failing assertion is a hard 180 ms wall-clock budget over a window spanning `acquire` + `release_detached` + `reclaim_stale` (filesystem + flock + process-liveness syscalls), while the retry window it guards against is 200 ms — only a 20 ms margin on a shared runner executing the full 3784-test suite.
- `try_lock_slot` enters the retry window only when holder metadata is missing, the recorded pid is verifiably dead (`kill(pid, 0)` → ESRCH), or the release token matches; a live unrelated holder (this test's setup) takes the fail-fast path, so the observed overrun is consistent with runner load, not with an unintended retry.
- The same test passed on the latest main CI run (d7612a31) at the identical code, in both feature configurations.

Branch updated onto latest main to re-run CI.

anantheparty (approved):
Verification review: **approve** — the double-layer guard is sound (per-filter `--list` count with a union-passed assertion catching the all-`#[ignore]` case — exactly #2029's silent-green root), no false-positive surface today (all 12 filter steps run on linux x64 where no filter's selection is wholly cfg'd out), and the migration is complete (10 ci.yml steps + 3 self-hosted; the two remaining target-level runs are out of scope by design). Final gate: `test-octos-cli` pending at review time — worst case is a loud red, not a new silent green.

Issue #2367, opened 2026-09-17
monitor/resume on an expired monitor re-arms with a stale deadline and silently re-expires


BH3GEI:
## Summary

`monitor/resume` (MonitorControlKind::Resume in `crates/octos-cli/src/autonomy/agent_orchestrator.rs`) re-arms a monitor in ANY non-deleted status — including `expired` — without touching `expires_at_ms`. A resumed-past-due monitor therefore goes `active` with its deadline still in the past, and the next `monitor_reconcile_pass` (global drain tick, ~5s) re-expires it. The client sees `ok: true` from resume, then the monitor dies again seconds later.

## Evidence

- The Resume arm (agent_orchestrator.rs, `control_monitor`) sets `status = "active"`, clears `pause_reason` and resets the flood window, but never recomputes `expires_at_ms`; only `deleted` is refused.
- Reconcile sweep expires any `active`, non-persistent monitor with `expires_at_ms <= now`.
- Pinned by test `resumed_expired_monitor_re_expires_and_emits_a_second_frame` (added alongside #2080's fix): expire via sweep → resume (succeeds) → next sweep re-expires and emits a second `monitor/expired` frame. The durable note path has always had the same double-fire shape, so this is pre-existing semantics, not a regression from wiring the wire frame.

## Ask

Decide the intended semantics:

1. Refuse `monitor/resume` on `expired` (treat it as terminal like `deleted`) — client must re-create the watch; or
2. Refresh `expires_at_ms` on resume (now + timeout_secs) so resume actually re-arms.

Until then, `resumed_expired_monitor_re_expires_and_emits_a_second_frame` pins the current behavior; whichever route is chosen must update that test.

anantheparty:
Confirmed still present at `main@d7612a3` (2026-09-17 audit), mechanism pinned at code level:

- `agent_orchestrator.rs:11801` — resume only rejects `deleted`; `expires_at_ms` is not validated against now.
- `:11817-11824` — resume sets `active`, clears pause, resets the flood window; `expires_at_ms` untouched.
- `monitor_reconcile_pass` (`:12667-12672`) then flips `active && expires_at_ms <= now` → `expired` on the next drain tick (~5s), so a resumed expired monitor silently re-expires exactly as described.

Two scope notes for whoever picks this up:

1. The pinning test mentioned in the issue body only exists on the open PR branch #2368 — it is not on main.
[...]

Issue #2019, opened 2026-08-14
Monitor events: let long-running work narrate progress into the conversation


ymote:
## Summary

Long-running background work in octos can report **status** (running / completed) and a **final result**, but it cannot narrate *while it runs*. There is no way for a loop to say "round 3 scored 7.2, critique: …" and have that land in the conversation as it happens.

Claude Code has this primitive — a background monitor that emits named observations into the transcript:

```
⏺ Monitor event: "phone UX self-critic loop: round scores + critiques"
```

octos has no equivalent surface. The closest primitives all fall short:

| existing | why it isn't this |
|---|---|
| `spawn_only` task notifications | terminal only — one event at the end, no mid-run content |
| task status (`task_updated`) | lifecycle state, not author-chosen content |
| goal chip | aggregate counters, one line, not a stream |
| peer blackboard / findings ledger | durable artifacts read on demand, not pushed into the conversation |

The gap matters most for the workloads octos is actually being pushed toward: self-critic loops, multi-round evaluation, long fleets. Today a 40-minute loop is a black box that emits one summary — which is exactly the "looks dead" complaint that produced #2003 and octos-tui#533, in a different costume.

## Proposal

A first-class **monitor event**: a named, timestamped, user-visible observation emitted by running work into the session that owns it.

1. **Emit API** — a tool (`monitor_emit`) plus an internal helper so native runners, `spawn_only` skills, peers, and pipeline nodes can all emit. Payload: `label` (stable, groups a stream), `message` (the observation), optional `data` for structured rounds/scores.

2. **Transport** — a new `UiNotification::MonitorEvent` sent through `send_notification_durable`, mirroring `PeerStaged`, so a briefly-disconnected client still replays the stream rather than losing the middle of a run.

3. **Rendering (octos-tui)** — a distinct transcript line attributed to the emitting task/peer, visually separable from assistant prose. Collapsible by `label` so a 50-round loop is one foldable group, not 50 loose lines.

## Design constraints, learned the hard way

- **Every event MUST carry `session_id`.** Activity emitted without one renders in whichever session happens to be focused — a defect this project has shipped at least three times (octos-tui#461, #466, #483). Non-negotiable, and worth a test that asserts routing rather than emission.

- **Attribution must name the emitter.** A peer's or sub-agent's observation rendered unattributed reads as the master speaking. Carry `task_id` / peer slug and render it.

[...]

PR #2368, opened 2026-09-17
fix(ui-protocol): emit monitor/expired with a profile-scope filter arm; drop the producer-less loop/completed variant

### LINKED ISSUE #2080: ui-protocol: LoopCompleted / MonitorExpired have no producer — wire them with a profile-scope filter arm, or delete them
## Summary

`LoopCompleted` and `MonitorExpired` are declared wire notifications with a `profile_id: Option<String>` field, but nothing in `crates/` ever constructs them. They are the only two profile-carrying `UiNotification` variants left out of the `ledger_event_matches_profile_scope` filter after #2075, and the reason is that there is no producer to test a filter arm against.

## Detail

- `LoopCompletedEvent` — `crates/octos-core/src/ui_protocol.rs:6086`
- `MonitorExpiredEvent` — `crates/octos-core/src/ui_protocol.rs:6176`

Both appear only in the enum declaration, the method-name mapping, and exhaustive match arms. `book/src/advanced.md:512` documents them as not currently emitted. `monitor_reconcile_pass` does expire monitors durably but emits no `monitor/expired` frame.

## Why file it rather than pre-emptively filter

Adding a filter arm now would be an untestable code path — there is no real delivery path to drive an effect test through, and this repo's convention is that a coverage claim needs a test that goes through the production path. But leaving them unfiltered means that whenever a producer *is* added, the cross-tenant leak fixed in #2067 silently reopens for that variant.

## Ask

Either:

1. Wire the producers and add the scope-filter arms together, with an effect test per variant (preferred — the events exist because something was meant to emit them); or
2. Delete both variants from the wire protocol if they are genuinely abandoned, so the filter's variant list stays exhaustive by construction.

Whichever route, the filter arm and the producer must land in the same change.

## Provenance

Audit during #2067 / PR #2075; codex review classified these two as tracking-only (agreed) while classifying the other deferred variants as must-fix.

BH3GEI:
## Problem

#2080: `LoopCompleted` and `MonitorExpired` are declared wire notifications carrying `profile_id`, but nothing in `crates/` ever constructs them. They are the last two profile-carrying `UiNotification` variants without an arm in `ledger_event_matches_profile_scope` — meaning whenever a producer *is* added, the cross-tenant leak fixed in #2067 silently reopens for that variant. The issue's constraint: filter arm and producer must land in the same change.

## Root cause

`monitor_reconcile_pass` and `expire_monitor` both transition a monitor to its terminal `expired` status durably (record + note), but sit deep in the runtime with no handle to the transport, so no `monitor/expired` frame ever existed. `loop/completed` has no producer and no natural emission site at all — the book already tells clients not to wait for it.

## Fix

- **`monitor/expired` wired (issue route 1):** both expiry transition sites build the frame from the post-transition record (nested `monitor` snapshot reuses `autonomy_monitor_json`, the exact wire shape `monitor/list`/`monitor/updated` publish) and hand it — after the state lock is dropped — to a process-global best-effort sink, the same discipline as the #2019 `background/activity` human sink: bounded-channel `try_send`, drop-count metric, installed once at serve boot (`spawn_monitor_expired_sink`, covering both `serve --stdio` and HTTP), drained onto the owning session's durable ledger via `send_notification_durable` over a detached connection. Connected clients receive it live; disconnected clients replay it by cursor. Emission is exactly once per active→expired transition (both sites no-op on an already-terminal status).
- **Profile-scope filter arm in the same change:** top-level `profile_id` stamp first, nested `monitor.profile_id` as fallback, mirroring every other monitor arm; fail-closed via `optional_profile_scope_matches` (absent stamp ⇒ main profile only). The existing `coding.monitor_runtime.v1` capability gate already covers the variant; test extended to assert it.
- **`loop/completed` deleted (issue route 2):** variant, method constant, inventory entry, decode arm and golden-test entries removed from octos-core; spec, wire inventory and both books updated; errata note added to the historical UPCR so future audits don't re-flag it. No client can break: the variant was never emitted, and a stray inbound `loop/completed` now decodes as an unknown method. The filter's variant list is exhaustive by construction again.

## Test evidence

- **Baseline reproduction on main (d7612a31):** a scratch test (uncommitted, removed after) drove `create_monitor` → `expire_monitor` with serve's sink wiring installed: the durable transition happened (`status = expired`) while the owning session's durable stream held **0 entries** — the issue symptom through the real ledger path. `git grep` on main confirms no production code constructs either variant.
- **New tests (all asserting real behavior):**
  - `monitor_reconcile_pass_emits_monitor_expired_exactly_once` / `expire_monitor_emits_monitor_expired_with_reason_exactly_once` — both emit sites, exactly-once, already-terminal and unknown-id silence; recorders filter to their own session to survive process-global cross-talk.
  - `should_deliver_monitor_expired_through_the_production_sink_to_the_owning_session` — production wiring end-to-end: real `spawn_monitor_expired_sink` → drain → durable ledger → cursor replay; asserts frame contents, profile filter (alpha passes / beta refused), and that a sibling session's stream stays empty.
  - `should_deliver_monitor_expired_from_the_reconcile_sweep_through_the_production_sink` — the same end-to-end through the OTHER emit site (TTL lapse → `monitor_reconcile_pass`).
  - `monitor_expired_frames_are_visible_only_to_their_profile` — filter arm unit test incl. the nested-fallback shape.
  - `resumed_expired_monitor_re_expires_and_emits_a_second_frame` — pins the pre-existing `monitor/resume` semantics this tap inherits (see below).
  - Capability-gate test extended with the `monitor/expired` arm.
- **Mutation red-checks:** neutering either emit site turns its tests red (acceptance test hits its 5s deadline deterministically); restored afterwards.
- **Full suites:** `cargo test -p octos-cli --lib` 3801/3801 green (99.6s), `cargo test -p octos-core` 381/381, `cargo fmt --check` clean, workspace `clippy --all-targets` zero new lints (only the pre-existing local-clippy-1.98 `result_large_err` at api/handlers.rs, untouched by this diff).

## Review

Independent adversarial review returned REQUEST_CHANGES with one Major, since resolved: the sink docs claimed "exactly once per monitor per lifetime", but `monitor/resume` can re-arm an expired monitor (only `deleted` is refused) without refreshing `expires_at_ms`, so the next sweep re-expires it with a second frame. This is pre-existing resume semantics (the durable note path has always double-fired the same way), not a regression — so the fix here is corrected docs ("exactly once per active→expired transition") plus a pinning test, and the underlying resume-semantics decision is filed as #2367 for the maintainers. Review Minors also adopted: sink-install ordering assumption documented at the serve call site, UPCR errata, sweep-site production-path e2e added. Rejected with reasons: renaming the shared `BACKGROUND_ACTIVITY_QUEUE_CAPACITY` (widens the diff into #2019 code for no semantic gain) and poisoned-mutex/drop-metric coverage (mirrors the #2019 sink, which is likewise uncovered). Post-fix round re-ran all suites green.

## Disclosure

- Scope interpretation: the issue offers "wire both" or "delete both"; this change wires `MonitorExpired` (it has a natural, durable emission site) and deletes `LoopCompleted` (it has neither producer nor emission site), satisfying the invariant per-variant — no producible frame is left unfiltered. Happy to adjust if maintainers prefer symmetric treatment.
- One full-suite run on this machine SIGSEGV'd inside `MTLCreateSystemDefaultDevice` (llama.cpp Metal init in `llm_select_promotes_fallback`, a native path this diff doesn't touch); the test passes solo and the rerun was fully green — local-environment flake, not observable on Linux CI.

Closes #2080


BH3GEI:
Merged current `main` into this branch ahead of review. The merge required flipping this branch's own pin `resumed_expired_monitor_re_expires_and_emits_a_second_frame`, which documented itself as pinning the pre-`#2367` resume semantics: `monitor/resume` now refuses a record whose deadline has already passed, so the pin became `resumed_expired_monitor_is_refused_and_emits_no_second_frame` — same sink discipline and guard, asserting the typed `monitor_invalid_state` refusal, that the record stays `expired`, and that the sweep emits no second frame after the refusal. The sweep/watcher expiry-frame pins are untouched; monitor family 44/44 and the full `agent_orchestrator` module 412/412 ran green locally on the merged tree.

anantheparty (approved):
Adversarial review (zero-context, head `e0341724`): **approve** — this closes the #2482-item-4 gap (clients can't learn about monitor expiry proactively) with the right mechanics.

Verified: the profile-scope filter arm aligns verbatim with its sibling arm (top-level `profile_id` first, nested fallback, pinned-profile connections filtered, stdio None short-circuits — no cross-profile leak), emission is exactly-once per active→expired transition (both sites share the state lock, guard-then-emit after release), the frame is a normal durable ledger row (replays on reconnect per standard at-least-once semantics), the producer-less loop/completed deletion is verified workspace-wide with spec/inventory/golden/books/UPCR-errata all updated in-PR and a §6-superset consistency test reading the spec markdown directly, and the payload carries the full `UiMonitorRecord` snapshot — the one-click-recreate flow from #2482-item-4 works (only `cwd` is lost, defaulted at create; `timeout_secs` derivable). The acceptance tests run through the real production sink + ledger replay, not mocks, and #2367's interaction (resumed-expired is refused, no second frame) is pinned by test.

Notes:
1. **[Medium, doc — should fix in-PR or immediately after]** The tap-invariant docblock at `agent_orchestrator.rs:12988-12991` still claims a resumed-past-due monitor "re-expires on the next sweep … that re-expiry is a NEW transition with its own frame" — the exact wire path #2367 removed (and which this PR's own test `resumed_expired_monitor_is_refused_and_emits_no_second_frame` pins as absent). An invariant doc promising a nonexistent frame is the kind of rot the doc-reference lint (#2492) now catches in ARCHITECTURE.md — this one is in a rustdoc comment, so it needs the human pass.
2. **[Low]** `agent_orchestrator.rs:20063`'s `.ok()` silently downgrades the payload on snapshot-conversion failure (structurally unreachable today, but debug-log it anyway).
3. **[Low, documented tradeoff]** Delivery is best-effort: the 512-slot shared channel `try_send` drops permanently under pressure (counter `ws.monitor_expired.drop`, untested path) and retained-ring eviction means a late reconnecting client may have been squeezed out — #2482-item-4 technically reopens under load. Consistent with the #2019 human-sink precedent and the durable record stays queryable via `monitor/list`; just noting the boundary honestly.
4. **[Low, pre-existing]** The spec has no params section for `monitor/expired` (the whole monitor family is "backfilled from code constants") — a spec-side gap worth its own pass, not this PR's job.

Scope: 12 files all on-proposition, CI zero-red (two heavy jobs still running). Approving; auto-merge covers the tail.

PR #2369, opened 2026-09-17
test(cli): filter background-activity recorders to their own sessions (#2364)

### LINKED ISSUE #2364: test flake: background-activity recorder tests race non-guarded handle_monitor_batch emitters (process-global sink cross-talk)
## Summary

`autonomy::agent_orchestrator::tests::should_emit_attributed_human_activity_on_the_owning_session_when_a_monitor_fires` can fail when run in the same test process as `monitor_wake_idle_gates_and_paused_monitor_wake_is_unschedulable` (or any other test that calls `handle_monitor_batch` without holding `background_activity_test_guard`).

## Evidence

Observed on 2026-09-16 on `main` (9a140495), `cargo test -p octos-cli --lib monitor_`:

```
---- autonomy::agent_orchestrator::tests::should_emit_attributed_human_activity_on_the_owning_session_when_a_monitor_fires stdout ----
assertion `left == right` failed: one event per observed line: [BackgroundActivityEvent { session_id: SessionKey("api:mon-idle"), ... text: "c", ... }, ...]
  left: 5
```

The recorder holds `background_activity_test_guard` and installs a process-global sink that records **every** `BackgroundActivityEvent`. The contaminating `api:mon-idle` / `text: "c"` event comes from `monitor_wake_idle_gates_and_paused_monitor_wake_is_unschedulable`, which calls `handle_monitor_batch("monitor_01", &["c"], 1)` — a `Fired` outcome emits a human-activity event when a sink is installed — **without** taking the guard. The guard only serializes cooperative (sink-installing) tests; it cannot stop a non-cooperative emitter from writing into the currently-installed recorder.

## Root cause

Same class as the `monitor/expired` recorder cross-talk fixed alongside #2080: a process-global sink plus count-everything recorders is only deterministic if either

1. every test that can emit through the sink holds the guard (untenable — dozens of batch call sites), or
2. each recorder filters to the events its own test produces (e.g. by `session_id`), so foreign emissions are ignored by construction.

Option 2 is the local, robust fix: filter the recording closure in `should_emit_attributed_human_activity_on_the_owning_session_when_a_monitor_fires` (sessions `api:mon-human-a` / `api:mon-human-b`) and audit the other `background_activity_test_guard` consumers for the same pattern.

## Reproduction

Timing-dependent; more likely under a crowded test filter:

```sh
for i in $(seq 1 20); do
  cargo test -p octos-cli --lib monitor_ || break
done
```

## Provenance

Found while stress-running the new `monitor/expired` sink tests for #2080; verified independent of that change (the diff does not touch `handle_monitor_batch` or the human-activity path).

BH3GEI:
## Problem

`should_emit_attributed_human_activity_on_the_owning_session_when_a_monitor_fires` flakes when it shares a test process with any test that calls `handle_monitor_batch` without holding `background_activity_test_guard` — e.g. `monitor_wake_idle_gates_and_paused_monitor_wake_is_unschedulable`. Observed on main (9a140495): `left: 5` for the 4-event routing assertion, with the extra event carrying `session_id: api:mon-idle`, `text: "c"`.

## Root cause

The human-activity sink is process-global. The test guard serializes sink *installers*, but it cannot stop a **non-guarded emitter**: a parallel test whose `handle_monitor_batch` reaches a `Fired` outcome emits through whatever sink happens to be installed. The failing test's recorder counted *every* event, so one foreign emission inflated the count.

## Fix

Per the issue's option 2 (forcing every emitter under the guard is untenable — dozens of batch call sites), each recorder now filters to the sessions its own test produces, so foreign emissions are ignored by construction:

- `agent_orchestrator.rs`: the routing test's recorder accepts only `api:mon-human-a` / `api:mon-human-b`; a deterministic foreign emission (session `api:mon-idle`, text `"c"` — mirroring the real contaminator exactly) is emitted mid-test to pin the immunity.
- `human_events.rs`: `should_refuse_to_emit_when_the_session_key_is_empty`'s recorder filters to `dev:local:tui`, with the same foreign-emission pin.
- `background_activity_test_guard`'s doc now states the real contract: the guard serializes installers only; recorders must filter.

Audit for the same pattern (repo-wide): the only sink installers are these two tests and production `spawn_background_activity_sink` (boot-only, unreachable from lib tests); the two no-sink tests install no recorder and `emit_background_activity` early-outs before touching shared state with no sink installed, so they cannot be poisoned. Cap keys are per-(session, origin), and the tests' own keys are test-unique, so foreign emissions cannot suppress their own events either.

## Test evidence

- **Baseline red (deterministic)**: a scratch test installing the old count-everything recorder plus one foreign `api:mon-idle` emission fails on main with `left: 1, right: 0` — the flake channel, reproduced without timing luck.
- **Mutation pins**: stripping either filter turns its test red — the orchestrator test fails with `left: 5, right: 4`, byte-for-byte the CI flake signature from the issue.
- **Post-fix green**: the issue's repro invocation `cargo test -p octos-cli --lib monitor_` ran 20x + 10x with zero failures; an independent review pass ran it another 12x at `--test-threads=16`, all green.
- Full `cargo test -p octos-cli --lib`: 3795 passed / 0 failed (twice: pre-review and post-review). `cargo fmt --check` clean; `cargo clippy -p octos-cli --lib --tests` shows only the pre-existing local-toolchain `result_large_err` at `api/handlers.rs:2140`, untouched by this diff.

## End-to-end note

The change is test-only (`#[cfg(test)]` modules; no production code touched), so the real path being verified is the test suite itself: the deterministic baseline repro, the mutation pins reproducing the exact observed signature, and 42 consecutive green runs of the issue's own stress loop are the end-to-end evidence.

## Review

Independent adversarial review: SHIP, zero Blocker/Major. Both Minor findings (guard doc overstating its guarantee; pin liveness depends on cap admission) and one Nit (filter style consistency) are addressed in this commit.

Closes #2364


anantheparty (approved):
Verification review: **approve** — the fix matches #2364's crosstalk exactly (the same polluting event is injected as a regression pin, and the assertions pin both the exclusion and the undisturbed count/order of owned events — not a vacuous pin given the budget caps clear on guard acquisition). Test-only: zero production paths touched.

Issue #2181, opened 2026-08-30
goal transitions (#2178) follow-ups: offline TOCTOU, instance-data-dir blind spot, docs lag, archived guard


anantheparty:
Non-blocking notes from the #2178 pre-merge adversarial review, bundled for tracking. None are regressions from #2178.

1. **Offline fallback TOCTOU** (`goal.rs:250-263,558-570`): `serve_liveness` is try_lock→unlock→Offline, then `cmd_transition_located` appends **without** the lock. If a serve finishes starting + initial restore in that ms-scale window, the stale-live-cache overwrite can still fire. Robust shape: hold the lock across the offline append (a starting serve blocks on the lock and re-restores after, seeing the append).

2. **instance-data-dir blind spot** (pre-existing): `GoalCommand` has no `--instance-data-dir`, so liveness always probes the state_home lock; for per-instance-dir serves it can misdetect Offline and write the wrong store. Fix: add the flag (or document the limitation).

3. **Docs lag** (3 spots): `api/OCTOS_UI_PROTOCOL_WIRE_INVENTORY_2026-05-24.md` goal method table, `book/src/advanced.md:505`, `book-zh/src/advanced.md:505` don't list the new transition methods yet.

4. **Reconciliation fail-noisy** (`goal.rs:195-219`): when the profile is unresolved or the ledger has no rows, the transition commits but returns ok:false "committed but reconciliation failed" — an operator retry then hits "already archived (terminal)". Consider a distinct non-error outcome.

5. **Pre-existing: `set_goal` has no archived guard** (`agent_orchestrator.rs:9876-10151`): a user/backend path goal_set can flip an archived goal back to active, contradicting the "TERMINAL irreversible" doc at :2224 (the model path is already blocked by the unfinished-goal check).

Context: #2178 (merged `305eb3ae`) fixed the production-reported live-cache overwrite of offline archive/reopen.

PR #2374, opened 2026-09-17
fix(serve): wait out transient data-dir lock holders before emitting OCTOS_DATA_DIR_LOCKED (#2357)

### LINKED ISSUE #2357: serve: retry the data-dir single-writer lock briefly before emitting OCTOS_DATA_DIR_LOCKED
## Summary

`acquire_serve_data_dir_lock` (crates/octos-cli/src/commands/serve.rs) try_locks `<data_dir>/.octos-serve.lock` exactly once and, on contention, exits with `OCTOS_DATA_DIR_LOCKED`. octoscode greps that marker on child-exit and **stops relaunching** the serve.

The lock is no longer held only by a long-lived serve: the goal operator CLI (`octos goal archive/reopen`, #2181) now holds it across its offline append — an ms-scale window. A serve (re)spawned inside that window gets the marker, and the octoscode respawn loop stops permanently, requiring manual intervention even though the conflict was transient.

## Proposal

Retry the try_lock on a short budget (e.g. 50 ms steps up to ~2 s) before emitting the marker. A genuinely running serve holds the lock for its whole lifetime, so the refusal contract (and the greppable marker octoscode depends on) is unchanged — only delayed by the budget in the true-conflict case. Transient holders (goal CLI append, future short critical sections) no longer kill a serve spawn.

## Notes / risks

- The deliberate double-serve refusal becomes slower by the retry budget; the existing `second_serve_on_same_data_dir_is_refused_with_a_greppable_marker` test still passes but its wall time grows.
- Fixed-duration waits are a known check-windows flake family (#2053, #2082); the retry budget must follow the repo's duration-scaling convention on Windows runners.

Follow-up to #2181 item 1.

BH3GEI:
## Problem

`acquire_serve_data_dir_lock` try_locks `<data_dir>/.octos-serve.lock` exactly once and, on contention, exits with `OCTOS_DATA_DIR_LOCKED`. octoscode greps the serve child's stderr for that marker on child-exit and **stops relaunching**.

Since #2181, the lock is also held by the goal operator CLI (`octos goal archive/reopen`) across its ms-scale offline append. A serve (re)spawned inside that window was refused with the marker on the first try_lock, so a transient conflict permanently killed the session until manual intervention.

## Root cause

One-shot refusal was correct when only a long-lived serve could hold the lock. With a short-lived transient holder in the mix, first-contention no longer implies a live serve.

## Fix

Retry `try_lock_exclusive` on a bounded budget (50ms steps, 2s total, monotonic-deadline, first try before any sleep) before emitting the marker. A genuinely running serve holds the lock for its whole lifetime, so the refusal contract — and the byte-stable greppable marker — is unchanged, only delayed by at most the budget in the true-conflict case. Non-contention errors still return immediately.

## Test evidence

- `serve_lock_acquisition_waits_out_a_transient_holder` (new): holder thread releases mid-budget behind a channel handshake (no scheduling-sensitive window); acquisition must succeed, and only after the release. Verified **red** against a simulated one-shot baseline (`Err` at the first contention), **green** with the fix.
- `serve_lock_acquisition_still_refuses_a_live_holder_after_the_budget` (new): a live holder is still refused with the same marker, and only after the full budget (red on the same baseline, green now).
- Pre-existing `second_serve_on_same_data_dir_is_refused_with_a_greppable_marker` and the duplicate-descriptor release test still pass unchanged.
- `cargo test -p octos-cli --lib`: 3797/3797, single full run.
- serve process e2e (real `octos serve` children): `serve_broken_pipe` 4/4 + `serve_sigterm` 3/3 serial, green.
- `cargo fmt --all -- --check` clean; `cargo clippy -p octos-cli` minimal + full-feature variants: zero new warnings (only the known local rust-1.98 `result_large_err` in api/handlers.rs, untouched by this diff; CI's 1.97 does not emit it).

## End-to-end evidence (real binaries, before/after)

Built `octos` from origin/main (baseline) and from this branch, drove both against a real flock-held `.octos-serve.lock`:

| scenario | baseline | this PR |
|---|---|---|
| transient holder releases after 1s | exits 0.11s with `OCTOS_DATA_DIR_LOCKED` (the bug) | waits, acquires, **starts listening**, no marker |
| persistent live holder | marker refusal | same marker refusal, delayed by the budget |

## Review

Two rounds: self-review of the full diff, then an independent maintainer-perspective review (SHIP, no Blocker/Major). Its findings are all addressed or disclosed:

- transient-holder test reworked to a channel handshake so a descheduled main thread can never release early (false pass/fail removed);
- comments updated where they still described the one-shot behavior (goal.rs ×2, the pre-existing \"refused FAST\" test comment);
- disclosed: the 2s budget is a heuristic bound — the goal CLI's offline hold includes an append that can fsync a snapshot compaction, so a pathologically slow/large append could still outlast the budget (strictly narrower than before: previously *any* overlap refused);
- disclosed: octoscode lives in a separate repo, so its spawn-side timing can't be checked here — but it greps the marker on child-exit (timing-insensitive), and its pre-marker behavior tolerated a ~5s crash loop, so a ≤2s later marker is safe.

Closes #2357


anantheparty (approved):
Adversarial review (zero-context, code read at head `3a303ed`, base `d7612a3` = main tip): **APPROVE**.

- Semantics match #2357 exactly: 50ms steps / 2000ms budget before the marker; marker contract and error message byte-identical, only delayed in the contended case.
- Concurrency verified line-level: deadline anchored once, no underflow/busy-spin, same-fd repeated try_lock valid for both flock and LockFileEx; SIGTERM/ctrl-C during the retry window take the default kill path (shutdown watcher spawns later, `serve.rs:1743` vs acquisition at `:770`) so exit stays prompt.
- Goal CLI unaffected: `serve_liveness` (`goal.rs:509`) is an independent one-shot try_lock; the new probe race is correctly absorbed by "no store opens before the lock".
- Tests are fix-sensitive both ways (transient success + persistent refusal), no new flake family — not using `waiting_budget` is correct per #2082 (scaling applies to upper-bound deadlines only).
- Scope clean: goal.rs hunks are comment-only (programmatically filtered); nothing smuggled in.
- Cross-checked the octoscode side (transport.rs:316-339, 1910-1927): the marker is consumed by grepping captured stderr on child exit — there is no spawn handshake timeout, so the +2s worst case only delays terminal detection, it cannot be misread as a spawn failure.

Nits for the record (all P3, none blocking):
1. The 2s budget is a heuristic — an fsync-heavy goal append on a slow disk could exceed it and reintroduce the original stop-respawning mode on such machines (author-documented at serve.rs:616-621). Worth watching real p99 lock-hold times; consider telemetry if it creeps.
2. `std::thread::sleep` in async context is safe today (call site precedes any spawn, multi-thread runtime) but would become a real bug if the call site moves — a comment or `spawn_blocking` would armor it.
3. The two new tests add ~4.3s wall time (both budget-burning, doc-commented).

CI: 13/13 green (incl. check-windows, the relevant one for the timing tests). Merging as squash.

Issue #2380, opened 2026-09-17
security: deploy.ps1 NSSM service registration keeps the dashboard bearer token in service argv


BH3GEI:
Found while fixing #2371 (verified at main@6b794fa6).

`scripts/deploy.ps1:291` registers the Windows service via NSSM with the bearer token in argv:

```powershell
& $nssmExe install $serviceName $octosExe "serve" "--host" "0.0.0.0" "--port" "$servePort" "--data-dir" $dataDir "--auth-token" $authToken
```

Windows service command lines are readable by any local user (`sc qc`, `Get-CimInstance Win32_Process`), so this is the same failure class as #2371. The sibling generators (`install.sh` launchd/systemd, `install.ps1` scheduled-task wrapper, `local-tenant-deploy.sh`) all carried the token via the `OCTOS_AUTH_TOKEN` env var already and were scrubbed of the argv form under #2371; deploy.ps1 is the remaining persistent-argv site.

Suggested fix: pass the token via NSSM's `AppEnvironmentExtra` (`OCTOS_AUTH_TOKEN=...`) and drop `--auth-token` from the argv list. Note that `scripts/test-windows-deploy.sh:56` currently asserts the `--auth-token` argv form and must be updated in the same PR.

Related transient (install-time-only) argv sites, lower severity: `scripts/cloud-host-deploy.sh:599`, `scripts/frp/bootstrap-tenant.sh:147`, and the admin-generated setup script at `crates/octos-cli/src/api/admin.rs:4592` (overlaps #2344's setup-script surface).

PR #2381, opened 2026-09-17
fix(serve): warn when --auth-token arrives via argv; keep installed services' tokens out of argv (#2371)

### LINKED ISSUE #2371: security: octos serve --auth-token keeps the bearer token in argv (readable via ps)
**Found in the 2026-09-17 zero-context security audit** (evidence at `main@d7612a31d4`).

`octos serve --auth-token <value>` keeps the dashboard bearer token in `argv` (`crates/octos-cli/src/commands/serve.rs:905-915` — CLI arg takes priority over env/config), where any local process can read it via `ps`. The env (`OCTOS_AUTH_TOKEN`) and config-file sources already exist; the CLI flag is the only path that leaks.

**Suggested fix**
- Print a one-time warning when the token arrives via argv.
- Docs should steer operators to env/config (0600) as the recommended sources.

Low severity (local-only, requires another local process), but it undercuts the otherwise-correct default where a non-loopback bind without a token auto-generates a random one (`serve.rs:918-927` — nice touch, worth keeping).

BH3GEI:
## Problem

`octos serve --auth-token <value>` keeps the dashboard bearer token in the process argv, where any local process can read it via `ps`. The env (`OCTOS_AUTH_TOKEN`) and config-file sources already exist; the CLI flag was the only leak path — and the repo's own service generators (launchd plist, systemd unit, Windows scheduled task) all put the token in the installed service's argv permanently, even though each *also* delivered it via `OCTOS_AUTH_TOKEN`.

## Root cause

The token-resolution chain in `commands/serve.rs` (`--auth-token` > `OCTOS_AUTH_TOKEN` > config) gave operators no signal that the argv source is world-readable, and the installer service templates passed `--auth-token` redundantly alongside the env var.

## Fix

- **serve.rs**: extracted the precedence chain into `resolve_auth_token`, which also reports the token's source; an argv-sourced token now earns a one-time `tracing::warn!` (same channel as the existing "Binding without --auth-token is dangerous" warning) steering operators to env/config. Behavior is otherwise identical — argv (even empty) still wins, empty config token still counts as absent.
- **install.sh / install.ps1 / local-tenant-deploy.sh**: dropped `--auth-token` from the launchd `ProgramArguments`, systemd `ExecStart`, and the scheduled-task wrapper. Each template already sets `OCTOS_AUTH_TOKEN` with the same value, and env is next in precedence, so the effective token is unchanged.
- **Docs**: `--auth-token` clap help and the `book/src/cli-reference.md` table row now state the ps visibility and recommend env/config.
- **Tripwire test**: a CI-covered unit test asserts the repo's service templates never carry `--auth-token` in argv while still delivering `OCTOS_AUTH_TOKEN` (mutation-checked: it fails against every one of the three pre-fix templates).

`deploy.ps1`'s NSSM registration remains an argv site (NSSM needs `AppEnvironmentExtra` for env injection, and `test-windows-deploy.sh` pins the current argv form) — filed as #2380, deliberately out of scope here.

## Test evidence

- New unit tests pin precedence + source tagging and the empty-config fallback; tripwire pins the templates. `cargo test -p octos-cli --lib`: **3800/3800 green** in one run; `cargo fmt --all -- --check` clean; clippy has only the pre-existing local-1.98 `result_large_err` lint (untouched by this diff); `bash -n` + shellcheck on both shell scripts match the main baseline exactly (20 pre-existing warnings, zero new).
- **End-to-end (real binary, before/after)**: built `octos` from this branch and ran `octos serve --auth-token E2ESECRET2371` — `ps` shows the token in argv (the issue's symptom, confirmed live) and the new warning appears in the serve log ("--auth-token exposes the bearer token in the process list (ps); prefer the OCTOS_AUTH_TOKEN env var or the config file"). Rerunning with `OCTOS_AUTH_TOKEN` instead: token absent from `ps`, no warning. On main there is no such warning (the code path simply doesn't exist).
- Independent adversarial review: **SHIP**, zero Blocker/Major. Findings addressed: stale call-site comment updated; sibling generator `local-tenant-deploy.sh` brought into the fix; installer tripwire test added. Rejected with reason: aligning empty-env-var semantics (pre-existing behavior, deliberately preserved); deploy.ps1 scrub (needs NSSM env design + its pinning test updated — #2380).

Closes #2371


anantheparty (approved):
Adversarial review (zero-context, code read at head `8aa1be3`, single commit on main tip): **APPROVE**.

- #2371 implemented exactly: one-time `tracing::warn!` on argv-sourced token at startup (`serve.rs:974-987`), clap help + book guidance to env/config. Warning lands on stderr + rolling file; `--stdio` protocol stdout stays clean.
- Priority chain (argv > env > config) verified branch-equivalent to main, including the empty-string edges; the config_layer backfill cannot cause false Argv warnings (`auth_token` is secret and non-layerable, pinned by test).
- All three in-repo service generators verified: launchd×2 and systemd move the token to env delivery (`EnvironmentVariables` / `Environment=`), the ps1 scheduled task routes through a wrapper that `set`s the env before exec — removing argv breaks nothing.
- Tripwire independently simulated: 0 hits on HEAD, 2/1/2 hits on the three pre-fix templates — the mutation claim holds. New tests are fix-sensitive.
- Scope clean (5 files, all serving #2371); the deploy.ps1/NSSM exclusion is an honestly-disclosed trim with #2380 tracking it.

Nits for the record (none blocking):
1. The tripwire is shape-matching (three forms only) — `--auth-token=$T` inline form, multi-line `ExecStart`, or env delivery moved out of the service block would evade it; the `OCTOS_AUTH_TOKEN` assertion is file-level `contains`. Worth tightening to the service-block region + the inline form later.
2. Pre-existing, out of scope but adjacent: installed plist/unit files are 0644 while containing the token in their env blocks (`install.sh:684/:727`), and the install.ps1 wrapper .cmd holds it in plaintext — env/file-plane follow-up candidate, same family as #2380.
3. #2371 asked for a "(0600)" hint next to the config-file guidance — currently just "prefer the env var or the config file"; a docs follow-up can add it.
4. Cosmetic: second same-name `resolve_auth_token` in admin.rs shadows nothing but invites confusion.

anantheparty:
CI follow-up on this merge: `test-octos-cli` on main went red at exactly this commit (`e89e266f`) with `GGML_ASSERT(!key.empty()) failed` (llama-cpp-sys-2, gguf.cpp:132) — green at the parent (`6b794fa6`). The changed files (serve.rs, install scripts, docs) don't touch llama, so this looks like an environment/data flake that happened to land here, but attribution is undetermined and main's required CI is red regardless. Anyone rebasing/landing next will inherit it — worth a look before the next merge.

PR #2365, opened 2026-09-16
ADR-001 personal memory: three tiers by trust, one index (design + phases 2–3)

ymote:
## Summary

ADR-001 (`docs/adr/personal-memory-tiers.md`) plus its kernel implementation: app data (mail, calendar, contacts…) becomes part of octos's searchable personal memory without a second memory system.

**Design** — three tiers by lifecycle and trust (Working = prompt; Recall = episodes + app records, untrusted, on demand; Knowledge = curated markdown bank, injected + searchable) over ONE hybrid index; record-level vectors, MRL-truncated to 256-d, int8 at rest, persisted HNSW, heat aging, hot-window residency, BM25 always on. Budget for 10k mails + 5k events ≈ 14 MB disk, < 15 MB RAM.

**Implemented here (phases 2–3)**
- `octos-memory`: `Record`/`RecordKind`/`Trust`, `quant.rs`, `RecallStore` (`recall.redb` + `recall-index/` persisted graph with manifest, residency, `touch`/`age`/`rebuild`/`nominate`, degraded in-memory fallback), `HybridIndex` dump/load/attach, relevance-ranked bank summary.
- `octos-agent`: `memory_search`, `memory_load`, `recall_memory {query}`; `memory_index` (bank sync, vector backfill, page ranking, promotion into staging with provenance); episode mirroring; per-turn ranked memory segment.
- `octos-cli`: profile/gateway/session wiring, background upkeep, `memory.recall_dimension`, `octos memory search|ingest|promote`; UI protocol `memory/search|load|ingest` (+ spec §6 and per-method sections, wire inventory).
- `octos-ffi`/`octos-uniffi`: `octos_memory_upsert|search|load|stats`, `data_dir`/`recall_dimension`, regenerated header and Python bindings.

Phase 1 (the `personal-data` plugin skill with an app-side index) is OctoSense-org/Octoscript-AppCard#98. Phase 4 (cross-device) stays optional.

## Test plan

- [x] `cargo test -p octos-memory` (153), `-p octos-agent --lib` (2916), `-p octos-cli --features api --lib` (3787 incl. `spec_section6_catalog_lists_every_advertised_method`), `-p octos-core ui_protocol` (143), `-p octos-ffi` (45), `-p octos-uniffi` (12)
- [x] `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`
[...]

PR #2393, opened 2026-09-18
fix(cli): unbreak main CI — model-fetch race, Windows voice fixture, stdio stack

ymote:
`main` CI is red on two jobs. It comes down to three unrelated bugs, one per commit so each can be reviewed or reverted on its own.

## 1. `test-octos-cli`: intermittent llama.cpp abort (a real bug, not only a test issue) — `ec274821`

Since #2365 made `embed-llama` a default feature, `test-octos-cli` sometimes dies with `gguf.cpp:132: GGML_ASSERT(!key.empty()) failed`.

With no `embedding` config, `create_embedder` falls back to llama.cpp, and `ensure_default_model` fetches the 300 MB EmbeddingGemma GGUF into the shared data dir. Every fetch writes `<model>.gguf.part.<pid>`. The pid keeps separate **processes** apart, but parallel tests are **threads in one process**, so they share one part file, and `File::create` truncates it underneath the other writer. Each fetch still passes its SHA-256 check, because it hashes the bytes *it* streamed rather than the file on disk. So the interleaved garbage gets renamed into place, and the next load aborts.

**This can crash production too**: any process that builds two embedders at once on first run (a gateway with several profiles, say) can corrupt the model and take itself down. CI hits it on every fresh runner, because nothing there sets `OCTOS_NO_MODEL_DOWNLOAD`.

**Fix:** fetches are serialized behind a process-wide lock, and the status is re-checked once the lock is held, so a waiting caller finds the finished model instead of fetching it again.

**Evidence:** the fetch is injected so the test runs without the network. Before the lock, 4 racing callers overlapped **4** fetches; after it, **1**. Separately, the full `octos-cli` lib suite on the real default build, in a fresh `OCTOS_HOME` that downloaded the real 333,590,944-byte model: **3801 passed, 0 failed**, with no leftover `.part` file. The crash was intermittent, so that green run shows no regression rather than proving the race is closed. The unit test is the proof.

## 2. `check-windows`: voice-readiness fixture — `1f8c683e`

Three `api::auth_handlers` voice tests failed on every Windows run. Their fixture built JSON with `format!`, pasting paths into string literals. `C:\Users\runneradmin\AppData\Local\Temp\…` contains `\U`, `\A`, `\L` and `\T`, which are invalid JSON escapes. It now uses `json!`. Checked: a Windows temp path through the old construction gives `Invalid \escape`; through `json!` it round-trips exactly.

## 3. `check-windows`: stack overflow that aborted the whole binary — `363b8461`

`stdio_connection_stops_dispatch_after_writer_failure` overflowed its stack on Windows. A stack overflow can't unwind, so it aborted the entire test binary. That also hid #2's panic messages, which print in an end-of-run summary the abort never reached.

Reproduced on macOS by running the **built test binary** under `RUST_MIN_STACK`: it passes at 2 MiB and overflows at 1 MiB and below. (Setting that variable for `cargo test` also shrinks rustc's stack and crashes the compiler.) Boxing the future **doesn't** help, because the cost is the depth of the poll chain, not the size of the stored state. The test now runs on an 8 MiB thread, matching the `thread_stack_size(8 * 1024 * 1024)` that the chat, ACP, gateway and MCP runtimes already use for deep futures. It passes at every stack size down to 256 KiB.

## Verification
- `cargo fmt --all -- --check` clean
- `cargo clippy -p octos-cli --all-targets` on the default build (with llama.cpp): no warnings
- full `octos-cli` lib suite on the default build: 3801 passed, 0 failed
- **Windows:** check-windows only compiles on PRs, so its tests don't run here. A manual `workflow_dispatch` on this branch runs them: https://github.com/octos-org/octos/actions/runs/35314434304. **Please don't merge until that's green**; it's the only real Windows evidence.

## Two things this surfaced, not fixed here
- **Windows test failures can only appear after merge.** `check-windows` compiles on PRs and runs tests only on `main`, which is how this red streak built up (including #2362, which fixed Windows breakage in tests from #2360). Running the Windows tests on PRs, or on a label, would close that gap.
- **`serve` has no 8 MiB stack.** `serve` builds its runtime without the `thread_stack_size(8 * 1024 * 1024)` that chat, ACP, gateway and MCP all set, and `serve --stdio` is octoscode's default backend. That's a production change, so it's left for its own PR.


anantheparty (approved):
Adversarial review (zero-context, per-commit read at head `363b8461`): **APPROVE** — and thank you for the three-commit split with independent revertability; that's exactly how an unbreak-CI PR should land.

- **Commit 1 (model-fetch race)**: attribution chain fully verified — pid-only part filename (`embed_model.rs:160`) + `File::create` truncation + stream-only hashing + rename; parallel tests in one process share the pid, matching the main-branch failure logs line-for-line (`run 35259835472`, job 105332491114, `gguf.cpp:132 GGML_ASSERT(!key.empty())` + SIGABRT). The fix is a genuine root-cause closure: process-level mutex + in-lock re-check (double-checked pattern, single fetch funnel — the only `.part.` writer in the repo), no re-entrancy, no deadlock, callers audited.
- **Commit 2 (voice fixture)**: real fixture repair (`format!` → `json!` so Windows backslashes stop producing invalid JSON escapes), not a skip; all three voice tests still assert real behavior.
- **Commit 3 (stdio stack)**: dedicated 8 MiB-stack thread is semantically equivalent for `#[tokio::test]` and aligns with the four existing production `thread_stack_size(8MiB)` sites; serve's own gap honestly left for later instead of being smuggled in.

Independent-revert claim verified mechanically (files disjoint, `git revert --no-commit` clean on all three).

Nits for the record (P3, none blocking):
1. The race test asserts `max_in_flight == 1` but not total fetch count — a future removal of the in-lock re-check would still pass it. Consider asserting the injected fetch runs exactly once (with the fixture writing the full byte length so the re-check path is actually exercised).
2. `model_status` completeness is byte-length-only; on-disk SHA-256 verification after rename would close the "bypassing writer reproduces the abort" depth (you already note the hash is stream-only).
3. The fetch mutex is process-global, serializing unrelated data_dirs (throughput only, acknowledged in-code).
4. The manual run's `check-arm` failure looks like runner loss (no step-level failure, PR doesn't touch octos-agent) — worth a one-line note when you close the loop on it.

Windows evidence now closed: `check-windows` green at head (12m9s full run). Merging as squash.

BH3GEI (approved):
Thanks for putting this together — the one-commit-per-fix split and the write-up made this pleasant to verify.

I reproduced all three failures and verified each fix independently:

**1. Model-fetch race (ec274821)**
- Confirmed the mechanism against `download_and_verify` (embed_model.rs:160-211): the part file is keyed by pid only, so threads in one process share it; `File::create` truncates under the other writer, and the SHA-256 is computed over the bytes each caller streamed (never re-read from disk), so both pass and interleaved garbage lands at the destination — matching the `gguf.cpp:132` abort I see in main's test-octos-cli log (run 35259835472).
- The new race test is load-bearing: 4 barrier-synchronized callers report max 1 in-flight fetch; with the lock temporarily removed (mutation check in an isolated worktree) the test fails immediately. 10/10 repeat runs stable.
- All fetch paths route through the lock (checked every `ensure_default_model` caller in octos-ffi, commands/memory.rs, commands/chat.rs; `download_default_model` is private and its spawned thread never re-enters, so no deadlock).
- Full `octos-cli` lib suite in a fresh `OCTOS_HOME` that really downloaded the 333,590,944-byte model during the run: **3801 passed, 0 failed**, no `.part` leftover.

**2. Voice fixture (1f8c683e)**
- A/B through the real `VoicesRegistry::parse` with a Windows-style path (`C:\Users\runneradmin\AppData\Local\Temp\...`): the old `format!` construction fails to parse, the `json!` construction parses and round-trips. On macOS the values are identical to before (so it's a no-op on Unix), and the auth_handlers suite is 93/93.

**3. stdio stack (363b8461)**
- Reproduced on macOS: the old `#[tokio::test]` version of `stdio_connection_stops_dispatch_after_writer_failure`, run from the built test binary, passes at 2 MiB and aborts with a stack overflow at `RUST_MIN_STACK=1048576` — the same abort that killed the whole binary on Windows main (exit code 0xc00000fd in the check-windows log). The new 8 MiB-thread version passes at every size I tried down to 256 KiB.
- The manual current_thread + enable_all runtime preserves `#[tokio::test]` semantics for this body (spawn/timeout/Notify all fine), and matches the 8 MiB convention already in chat/acp/gateway/mcp_serve.

**CI**
- The workflow_dispatch run you flagged is green where it matters: check-windows ran the full per-crate test suite (not just the PR compile) including Test octos-cli (lib) — success, so both Windows fixes are proven on Windows. test-octos-cli also passed there and on the PR itself.
- `cargo fmt --all -- --check` clean; clippy on the default build reports only the pre-existing `result_large_err` at handlers.rs:2140, in a file this PR doesn't touch.

Two non-blocking observations for possible follow-ups (neither blocks this PR):
- `model_status` treats any file of exactly `DEFAULT_MODEL_BYTES` as complete, so a file already corrupted by the pre-fix race stays "complete" — the pinned hash is only checked against the stream, never against disk. A one-time hash probe would close that loop for already-affected installs.
- The manual-only check-arm lane on the dispatch run failed, but its logs are no longer retained so it can't be attributed from here; the ARM64 nightly lane on main was green on 09-17 and this diff has no arch-specific code, so it doesn't look related — just flagging so it isn't a surprise.

Issue #2400, opened 2026-09-18
docs: the reference layer is hand-copied — §15.1 misses ≥10 live config fields, says latency_threshold_ms=30000 vs code 10000 (book appendix C says 10000); 'octos docs' already generates TOOLS.md — route it


anantheparty:
**From the 2026-09-18 documentation audit** — the reference layer is hand-copied from code, and the copy has already diverged in both directions:

1. **`latency_threshold_ms`: the user-guide is wrong and contradicts the book.** guide says 30000 (en L535/544/2117, zh L455/464/2022); code default is 10_000 (`crates/octos-cli/src/config.rs`, `default_latency_threshold_ms`); **octos-book appendix C says 10000** (`chapters/appendix-c-config-reference.md:253`). Two doc surfaces, two different values, and the one wired into the deploy pipeline (user-guide) is the wrong one.
2. **§15.1 "Full Config Structure" is missing ≥10 live fields** (`context_window`, `model_temperature`, `model_top_p`, `model_reasoning_effort`, `env_vars`, `model_hints`, `auth_token`, `mcp_servers`, `sandbox`, `snapshots` — all present in `config.rs`).
3. **Route census drift**: guide claims "~140 REST endpoints"; `router.rs` has 184 `.route(` sites.
4. **Slash-command table lists 3 commands** (`/new /config /exit`); the chat/gateway dispatchers register a much larger family (the rest is left to §11.2 prose).
5. **The generation machine already exists and isn't routed**: `octos docs` generates TOOLS.md live from the ToolRegistry — the codebase knows how to generate reference docs; it's just not connected to the config/API surfaces.

**Suggested direction**
- Generate, don't copy: derive the config reference from serde/schemars annotations (or wire `octos docs` output into the docs build), and the API reference from the router table; let CI fail on drift.
- Until generation lands, fix the 30000→10000 contradiction now (it's user-facing and the book already disagrees with the guide).
- Add the 184-route census as a generated artifact so the "~140" claim can't silently rot.

PR #2402, opened 2026-09-18
docs(user-guide): fix latency_threshold_ms default 30000 -> 10000 (en+zh)

anantheparty:
Part of #2400 (reference-layer drift — the concrete user-facing piece, fixable now).

`docs/user-guide.md` and `docs/user-guide-zh.md` documented `adaptive_routing.latency_threshold_ms` as `30000` in their sample configs, while the code default is **10_000** (`crates/octos-cli/src/config.rs`, `default_latency_threshold_ms`) and octos-book appendix C already says 10000. Two doc surfaces, two values — and the guide wired into the deploy flow was the wrong one.

Changes: 2 occurrences in the English guide, 2 in the Chinese guide (`30000` → `10000`). No prose changes; the surrounding adaptive-routing examples stay as-is.

Remaining follow-ups from #2400 (§15.1 missing fields, route census, routing `octos docs` output) left deliberately out of scope for this small fix.


BH3GEI (commented):
Thanks for putting this together, and for the follow-up commit covering the book surfaces too.

I verified the core claim against the code: the default really is 10_000 — `crates/octos-cli/src/config.rs` defines `fn default_latency_threshold_ms() -> u64 { 10_000 }` (wired in via `#[serde(default = ...)]` and the `Default` impl), and the `Default` for the routing config in `crates/octos-llm/src/adaptive.rs` agrees. A repo-wide sweep found no other stale `30000`/`30s` mentions of this field, and the neighboring adaptive-routing defaults (`error_rate_threshold` 0.3, `probe_probability` 0.1, `probe_interval_secs` 60, `failure_threshold` 3, weights) all match the code across all five doc surfaces. I also built both books with mdbook 0.5.4 (the version `deploy-docs.yml` pins) — both build cleanly; the only warning is a pre-existing unclosed `<f32>` in `architecture.md`, untouched by this PR.

One concern before this fully achieves its goal: each guide states the default a third time in the prose bullet list, and those two spots still say 30s — now contradicting the JSON sample 9 lines above them in the very same section:

- `docs/user-guide.md:544` — "`latency_threshold_ms` — Providers with average latency above this are penalized (default: 30s)"
- `docs/user-guide-zh.md:464` — "平均延迟超过此值的提供商被降权（默认：30 秒）"

Since the PR titles itself as fixing the documented default, could we update those two bullets to 10s here as well (or call them out as an explicit follow-up)? As it stands, the guides go from uniformly wrong to internally contradictory, which is arguably more confusing for readers.

anantheparty:
Thank you for the sweep — the repo-wide "no other stale mentions" check and the neighboring-defaults cross-check are exactly the assurance this small PR needed. Both books building clean on the pinned mdbook is also good to know (that was my main worry with touching book surfaces).

The PR is armed with auto-merge; a formal approve (if you're satisfied) will let it land — no other changes planned on this branch.

anantheparty:
Gentle nudge — this has been green and MERGEABLE for ~2 days with your verification comment on record; it's waiting only on a formal approve before the armed auto-merge lands it. (Main has moved since, so the branch may need a one-click update first — happy to do that if you prefer.)

anantheparty:
Good catch — both prose bullets updated to 10s / 10 秒 (`docs/user-guide.md:544`, `docs/user-guide-zh.md:464`) in `50a7198`. The remaining "30s" hits in those files are a different field's timeout comment (handshake), legitimately untouched. The guides are now internally consistent: JSON sample = prose bullet = code default = book appendix C.

BH3GEI (approved):
Thanks for the follow-up — closing the loop at the new head.

Both prose bullets are fixed exactly as discussed (`docs/user-guide.md:544` now "default: 10s", `docs/user-guide-zh.md:464` now "默认：10 秒"), and commit `50a71986` touches just those two lines. Re-verified at `abf4ad3a` (including the two main merges):

- **Code default triple-checked**: `crates/octos-cli/src/config.rs:1277` (`fn default_latency_threshold_ms() -> u64 { 10_000 }`, wired via the serde default and the `Default` impl, with a matching doc comment) and `crates/octos-llm/src/adaptive.rs:65` both say 10_000, unchanged by the main merges.
- **Coverage complete**: a repo-wide sweep finds 13 `latency_threshold_ms` mentions across the doc surfaces — all now say 10000 / 10s / 10 秒 (`book-zh/src/providers.md` was already correct on main). The remaining "30s" hits all belong to other settings (the 429 retry-after fallback per `crates/octos-llm/src/retry.rs:310`, MCP handshake timeout, heartbeat intervals) and are correctly left alone.
- **Direction confirmed via history**: `affb458f` deliberately tuned this threshold 30s → 10s, so the old docs were lagging an intentional change — not the reverse.
- **Scope exact**: 5 files, 10 insertions / 10 deletions, every hunk a `latency_threshold_ms` 30000→10000 or 30s→10s edit; JSON sample, prose bullet, and table now agree within each section.
- **Build**: both books build cleanly with mdbook 0.5.4 (the version `deploy-docs.yml` pins); the only warning is the pre-existing unclosed `<f32>` in `architecture.md`, untouched by this PR.

The guides are now internally consistent with the code. Approving — thanks for the patience through the extra round.

Issue #2163, opened 2026-08-28
Store provider API keys in a server-side write-only credential store, not profile JSON


anantheparty:
## Summary

The AppUI wire contract is already write-only: profile reads expose `has_api_key` and do not return a secret. The server-side persistence contract should match that boundary. Ordinary provider API keys currently remain as plaintext values in the profile JSON `env_vars` map; only selected structured credentials such as Vertex service-account JSON are relocated to the OS keychain.

Move provider secrets behind one server-side credential-store abstraction and persist only an opaque reference or logical credential slot in profile configuration.

## Current behavior and code evidence

At [`main@e188664c`](https://github.com/octos-org/octos/tree/e188664ca252712d2ef18c3991b87a068b0367d3):

- The read projection is correctly sanitized: [`configured_provider_json`](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-cli/src/api/ui_protocol_transport.rs#L8991-L9015) returns `has_api_key`, not the value.
- [`profile/llm/upsert`](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-cli/src/api/ui_protocol_transport.rs#L11828-L11915) inserts a supplied key directly into `profile.config.env_vars` and then relocates only keychain-backed credential shapes before saving the profile.
- The adjacent sub-provider path documents the shared at-rest model explicitly: [ordinary API keys remain in the mode-0600 profile JSON while structured credentials are relocated](https://github.com/octos-org/octos/blob/e188664ca252712d2ef18c3991b87a068b0367d3/crates/octos-cli/src/api/ui_protocol_transport.rs#L15615-L15620).

File permissions are useful defense in depth, but they do not provide secret/config separation. The profile file, copies, backups, diagnostics, or future synchronization paths can contain reusable provider credentials even though AppUI never returns them.

## Minimal reproduction

Use a disposable profile and a non-secret marker value:

1. Call `profile/llm/upsert` with a route that has `api_key_env` and a marker as `api_key`.
2. Read the persisted profile registry JSON.
3. Observe the marker under `config.env_vars[api_key_env]`.
4. Call `profile/llm/list` and observe that the wire response is redacted to `has_api_key: true`.

This demonstrates that the network contract is write-only while the persistence contract is not a credential store.

## Proposed system contract

- Introduce one profile-scoped credential-store interface used by primary, fallback, and sub-provider model configuration.
[...]

Issue #1510, opened 2026-06-29
octos init: confusing API key prompt causes users to paste key value into env var name field


LeonRust:
## Problem

When running `octos init`, the wizard prompts for an API key like this:

```
Environment variable containing the API Key [OPENAI_API_KEY]:
```

Many users paste their actual API key here (e.g. `sk-proj-abc123...`) instead of an environment variable name. This results in:

```json
{ "api_key_env": "sk-proj-abc123..." }
```

At runtime, `octos chat` tries to look up `$sk-proj-abc123...` as an environment variable, which doesn't exist, producing:

```
Error: sk-proj-abc123... not set. Run `octos auth login -p deepseek` or set the env var
```

The user is confused because they *did* enter their key — the wizard just didn't make it clear that it was asking for the env var name, not the key value itself.

## Proposed solution

1. Add an `api_key` field to config that stores the key value directly, as a fallback when no environment variable is set.
2. Redesign the init wizard key step to default to direct key input, matching user expectations.
3. Update the error message to mention all three configuration methods (auth login, env var, direct key).

## Acceptance criteria

- [ ] `octos init` wizard defaults to prompting for the key value directly
- [ ] Environment variable name mode is still available as option 2
- [ ] Runtime prefers env var over direct key when both are present
- [ ] `octos chat` works with a key stored directly in config.json
- [ ] Error message when no key is found lists all configuration options

PR #2403, opened 2026-09-18
fix(init): disambiguate API-key env prompt and validate the env name (#1510)

anantheparty:
## Problem

`octos init` prompted `Environment variable containing the API Key [OPENAI_API_KEY]:` and users answered with the key **value** itself, which was stored as `api_key_env` and failed the runtime lookup (`Error: sk-proj-abc123... not set`).

## Why not the issue's proposal 1

The issue proposes adding an `api_key` config field that stores the key value directly. That is intentionally **not** adopted here: it conflicts with the #2163 direction of keeping credentials out of the profile JSON. This fix instead addresses the root cause — an ambiguous prompt with no input validation.

## Changes

- `crates/octos-cli/src/commands/init.rs` only:
  - Both key-env prompts (custom-provider flow, previously L360, and the provider wizard flow, previously L1019–1022) now read `Environment variable NAME that holds the API key (not the key value itself) [<default>]:` and share one new `prompt_api_key_env(default)` helper.
  - New pure validator `is_valid_env_name(&str) -> bool` (ASCII identifier: first char letter/underscore, rest alphanumeric/underscore) and a `looks_like_api_key` heuristic (`sk-` prefix or len > 40).
  - Validation loop: empty input → keep the provider default; invalid name → explain and re-ask; key-shaped input → extra hint to `export <NAME>=<key>` or use `octos auth login`; bounded at 3 invalid attempts, after which the input is accepted with a warning so piped stdin cannot hang the wizard.
  - No new dependencies; `--defaults` and non-interactive paths untouched.

## Tests

- 4 new unit tests in `mod tests`: valid names accepted; empty / non-identifier answers rejected; pasted `sk-`-prefixed key values rejected as env names; key-shaped input flagged for the extra hint.

## Notes

- **Not compiled locally; CI is the gate** (fmt / clippy / tests run in CI).


BH3GEI (approved):
Thanks for putting this together — the shared `prompt_api_key_env` helper replacing the two duplicated prompt blocks, with the pure validators split out for unit testing, is a clean way to fix this.

Verified locally on the head SHA:

- **Build/tests**: `cargo fmt --check` clean; `cargo clippy -p octos-cli --all-targets -- -D warnings` reports only the pre-existing `result_large_err` at `api/handlers.rs:2140` (identical on main, untouched by this PR). `cargo test -p octos-cli --lib`: 3805 passed / 0 failed (10 ignored); `--test cli_tests`: 19 passed / 0 failed.
- **Behavior, against the real binary with piped stdin** (`octos init --cwd <tmp>`, both call sites):
  - Valid name (`MY_SECRET_KEY`) is accepted and written to config.json; empty input keeps the provider default.
  - Invalid answer (`1BAD-NAME`) is rejected with the escaped `{:?}` echo and re-asked; a valid retry is then stored.
  - A pasted key value (`sk-proj-…`) is echoed only as `«pasted value masked»` with the `export`/`octos auth login` hint — I confirmed the raw key appears nowhere in the output or the written config.
  - After 3 invalid answers the wizard warns and falls back to the default name (the fail-closed path from the head commit) instead of accepting garbage.
  - EOF before any answer returns the default and exits cleanly — the bounded loop plus the empty-input path mean piped stdin can never spin.
  - Custom-provider flow (option 10) behaves the same around the `CUSTOM_API_KEY` default and stores the accepted name with `"provider": "custom"`.
- Scope: the PR's net diff vs main is `commands/init.rs` only; `--defaults` and flag-driven paths are untouched, and the existing `init` integration tests (all `--defaults`-based) still pass.

Two non-blocking follow-ups, no change requested here:
1. The PR description still says the 3-attempt path "accepts the input with a warning", but the head commit intentionally returns the default instead — worth updating the description so future readers review the code that actually shipped (the fail-closed behavior is the better of the two).
2. The prompt loop itself isn't unit-tested since it reads global stdin; the file's own `capture_api_key_from_reader` shows the injected-`BufRead` pattern that would make it testable. Likewise `--custom-api-key-env` still accepts unvalidated values under the old help wording — fine to leave for a follow-up given this PR's declared scope.

anantheparty:
Thank you for the thorough local verification — especially testing both call sites against the real binary with piped stdin; that's the exact coverage this couldn't get from CI alone.

One piece of context: the fail-closed fallback (3 invalid answers → default name, no longer accepting the input) landed as a second commit (`526475f`) *after* the initial PR, from an internal security-audit pass that flagged the bounded-accept path as leaving the #1510 failure mode reachable via piped stdin. Your piped-stdin test validating that fallback is exactly the confirmation it needed.

Great to see #1510 (open since July, user-reported confusion) close the loop: report → fix → adversarial review → independent local verification, all inside a day.

PR #2404, opened 2026-09-18
ci: give check-windows room to finish — raise its timeout from 90 to 150 minutes

ymote:
Follow-up to #2393.

**check-windows on `main` is no longer failing — it's being killed.** The re-run on `751913d` ran 16:39 → 18:10 and was cancelled at exactly the 90-minute cap. GitHub reports a timed-out job as *cancelled*, and no newer commit landed on `main`, so it wasn't cancel-in-progress.

**Why now:** the full Windows suite takes ~82 min when it passes (octos-cli integration ~19, lib ~13, builds ~20, doc tests ~6). Earlier failing runs ended in 54–81 min partly because the stack overflow #2393 fixed aborted the octos-cli lib step early. With that gone the whole suite runs, and runner variance tips it past 90.

| check-windows run | duration | result |
|---|---|---|
| #2393 verification (dispatch) | 82 min | success |
| main re-run on 751913d | 90 min | cancelled (timeout) |

**This PR** raises it to 150: headroom, without letting a real hang run for long.

**Caveats:**
- check-windows only compiles on PRs, so **this PR's own CI doesn't exercise the change**; `main` will.
- It stops the kill; it doesn't make Windows faster. Splitting octos-cli's tests into their own Windows job would cut the wall-clock and is the better long-term fix. Since #2365 the default build also compiles llama.cpp and the tests fetch a 333 MB model, which adds to it.


anantheparty (approved):
Approve. The evidence chain is complete: post-#2393 the suite no longer aborts early, the passing-run duration (~82 min) sits inside the old cap, and the 751913d re-run was killed at exactly 90:00 (reported as cancelled). 150 gives ~1.8× headroom over the passing baseline without being so loose that a true hang burns a whole evening; no other hunks in the diff.

PR #2406, opened 2026-09-18
fix(serve): tell users how to resolve a port-in-use bind failure (#2385)

### LINKED ISSUE #2385: ux(serve): port-in-use error is a bare io wrap — give it the same remediation quality as OCTOS_DATA_DIR_LOCKED (--port / how to find the holder)
**From the 2026-09-17 developer-experience audit.**

Within the same command, two startup failures get wildly different treatment:

- Data-dir lock (serve.rs:710-715): greppable `OCTOS_DATA_DIR_LOCKED` marker + two remediation paths + (since #2374) a transient-retry window. Gold standard.
- Port-in-use (serve.rs:623-625): `failed to bind octos API server to {host}:{port}` wrapping the raw io error. No `AddrInUse` recognition, no `--port` hint, no way to find the holder.

**Suggested fix**: match the bar the lock error already set — recognize `AddrInUse` and emit something like: port {port} is in use; find the holder with `lsof -i :{port}` (or your platform equivalent), or pass `--port` to choose another.

BH3GEI:
## Problem

Within one `octos serve` invocation, two startup failures get wildly different treatment:

- Data-dir lock: greppable `OCTOS_DATA_DIR_LOCKED` marker + two remediation paths (+ the #2374 transient-retry window) — the gold standard.
- Port-in-use: `failed to bind octos API server to {host}:{port}` wrapping the raw io error. No `AddrInUse` recognition, no `--port` hint, no way to find the holder.

Closes #2385.

## Root cause

`bind_http_listener` wraps every bind failure in the same bare message; nothing branches on `io::ErrorKind::AddrInUse`.

## Fix

Split the wrap into `bind_listener_error`. On `AddrInUse` the top-level error now reads e.g.:

```
failed to bind octos API server to 127.0.0.1:64435: the port is already in use. Find the holder with `lsof -i :64435`, or pass `--port` to choose another.
Caused by: Address already in use (os error 48)
```

Holder hint per platform (the issue's "or your platform equivalent"):

- **Linux**: `ss -ltnp 'sport = :PORT'` — iproute2 `ss` ships with the base system (minimal container images included); `lsof` frequently doesn't.
- **macOS/BSD**: `lsof -i :PORT`
- **Windows**: `netstat -ano | findstr :PORT`

Every other bind failure keeps the byte-identical bare wrap, and the io error stays as the `Caused by:` source in both arms.

Deliberately **no** greppable marker (unlike `OCTOS_DATA_DIR_LOCKED`): that marker exists because octoscode greps it verbatim; this HTTP-bind failure has no machine consumer today, and freezing the wording into a contract now would foreclose future rewording. Note the old message remains a prefix of the new one, so exact-line (`grep -x`) consumers would notice — none exist in-repo.

## Test evidence

- New tests (all in `commands::serve::tests`):
  - `bind_error_on_addr_in_use_carries_remediation` — message content per-platform + io error preserved in the source chain.
  - `bind_error_other_kinds_keep_the_bare_wrap` — non-`AddrInUse` output is byte-identical to the old message.
  - `bind_http_listener_reports_remediation_for_occupied_port` — holds a real socket and drives the production `bind_http_listener` through it (pins the wiring, not just the helper). Mutation check: reverting the call site to the old bare wrap turns this test red with exactly the old message.
- `cargo test -p octos-cli --lib`: **3804 passed / 0 failed** in a single run (plus 10 pre-existing ignored).
- `cargo test -p octos-cli --features api --test serve_broken_pipe --test serve_sigterm -- --test-threads=1`: 4/4 + 3/3, matching the CI serve-process lane.
- `cargo fmt --all -- --check` clean; `cargo clippy -p octos-cli --all-targets` shows only the pre-existing `handlers.rs` `result_large_err` warning (local toolchain 1.98, untouched by this diff).

### End-to-end (real binary)

Baseline (main build), port occupied by a socket:

```
Error: failed to bind octos API server to 127.0.0.1:63217
Caused by: Address already in use (os error 48)
```

With this patch:

```
Error: failed to bind octos API server to 127.0.0.1:64435: the port is already in use. Find the holder with `lsof -i :64435`, or pass `--port` to choose another.
Caused by: Address already in use (os error 48)
```

Both claims in the message were verified truthful: `lsof -i :PORT` returns the holder process on macOS, and `octos serve --port <free>` boots to the listening line.

- Linux hint runtime-verified on a separate Ubuntu host: `ss -ltnp 'sport = :PORT'` returns the exact LISTEN row with holder process/pid, and `lsof` is absent there — confirming the availability-based choice.
- Windows hint is the standard `netstat -ano | findstr` form; not runtime-verifiable from this machine — the `netstat` branch is compile-verified by the `check-windows` CI lane. The message assertions in the new unit test are platform-`cfg`-matched, so the ubuntu test lane exercises the Linux string.

## Review

Round-2 independent review verdict: APPROVE WITH NITS (reviewer independently built, ran the suite, and re-did the e2e). Findings: 1 Minor adopted (Linux hint switched from `lsof` to `ss`, with the test assertion re-gated); 1 Nit adopted (source-chain assertion added to the e2e test); 1 Nit accepted as-is (`Location:` line moved from the call site into the helper — same as the existing lock-error pattern); marker deliberation documented above. All verification re-run green after the changes.


anantheparty (approved):
Adversarial review (zero-context, code read at head `c0777d2`): **APPROVE**.

- **AddrInUse recognition is correct cross-platform**: `error.kind() == ErrorKind::AddrInUse` (`serve.rs:639`) — no raw-errno hardcoding; Unix EADDRINUSE and Windows WSAEADDRINUSE both map to this kind in std. One known blind spot worth a line in the docs eventually: Windows excluded port ranges surface as WSAEACCES → PermissionDenied, so those won't get the remediation (std limitation, not introduced here).
- **Remediation quality matches the bar**: port number, platform-appropriate holder command (`ss` on Linux / `lsof` elsewhere-Unix / `netstat findstr` on Windows, `serve.rs:652-667`), and the `--port` hint (flag verified at `serve.rs:324-325`). Construction follows the same `Report::new().wrap_err()` pattern as the non-contended lock branch (`serve.rs:756-759`), with the io error preserved as `Caused by:` (test asserts the chain).
- **Behavior boundary clean**: non-AddrInUse failures keep the old message byte-for-byte (pinned by an exact `assert_eq`), so existing greps don't break; all +108 lines accounted for as one replacement + two helpers + three tests.
- **Tests**: three of them, platform-independent (ephemeral ports, and tokio doesn't set SO_REUSEADDR by default so the second bind reliably fails on all three platforms). Honest CI note: PR runs of check-windows compile but don't execute tests, so the netstat/lsof assertion branches never run in CI — fine for now, worth knowing.

One open question for the maintainers (not blocking): the data-dir-lock error's greppable `OCTOS_DATA_DIR_LOCKED` marker exists because octoscode consumes it; port-in-use has no machine consumer today (octoscode's backoff-retry is the correct response there), so I don't think this error needs a marker — flagging it only because #2385's wording said "match the bar". If the maintainers want symmetry, a tiny follow-up can add one.

Issue #2388, opened 2026-09-18
security: installed systemd/launchd units ship the bearer token in world-readable (0644) env blocks — #2371's local-read threat model is not closed


anantheparty:
**Found in the 2026-09-17/18 follow-up audit of #2381's merge (commit `e89e266f`)** — this closes the gap my #2381 review noted as nit and elevates it with the full threat model.

#2381 correctly moved the token out of argv for installed services, but the installed units still carry the plaintext token in a **world-readable file**:

- `scripts/install.sh:727` — systemd unit installed with `sudo chmod 644`: `/etc/systemd/system/octos-serve.service` contains `Environment=OCTOS_AUTH_TOKEN=<plaintext>`; any local user can read it via `systemctl cat` or by just reading the file.
- `scripts/install.sh:684` — launchd plist installed with `sudo chmod 644`: `/Library/LaunchDaemons/io.octos.serve.plist` contains the token in `EnvironmentVariables`; same exposure.
- `scripts/install.ps1:171` (context) — the wrapper `.cmd` embeds `set "OCTOS_AUTH_TOKEN=$AuthToken"` in plaintext on disk.

Net: the #2371 threat model ("a local low-privilege user can read the dashboard bearer token") is **not closed by #2381** — the read moved from `ps` to a static 0644 file.

The correct pattern already exists in the same script: `install.sh:490` chmods `/etc/frp/frpc.toml` to **600**.

**Suggested fix**
- systemd: `EnvironmentFile=/etc/octos/serve.env` with that file at 0600 (root-owned), unit stays 644 — or `LoadCredential=`.
- launchd: 0600 on the plist (LaunchDaemons runs as root anyway), or move the token to a 0600 env file the plist sources.
- install.ps1: keep the token out of the wrapper (e.g. read from a 0600 file at start).
- Consider a follow-up tripwire asserting no `OCTOS_AUTH_TOKEN=` line lands in any 0644-installed file.

Issue #2326, opened 2026-09-13
governance: six decision-record conventions coexist with no charter, no routing, no lint — UPCR numbering already colliding


anantheparty:
Architecture-decisions lens deep-round finding (the last of the six-lens fresh-eyes rotation; zero overlap with #2325).

**Current state**: SIX decision-record conventions coexist — octoscode-web ADR 0001-0019 (the only one with CI metadata lint + supersession chains), octos's M-ADR flat namespace + docs/adr/ (4 files, unnumbered, unindexed) + UPCR×30, octoscode knowledge/decisions/ (schema + lint but ZERO decision entries), octos-arc ARC_BASELINE+lock (the best evidence binding), octos-learn dual doc dirs plan-prose, octos-web/arcbench-runtime/octos-hub ZERO records. **The meta-decision — which record type for which decision class — has never been written.**

**Immediate hazard (P1)**: UPCR-2026-022/026/027 numbering collisions (022 is TWO files with Accepted vs proposed coexisting) — the OUP protocol decisions' foreign keys are ambiguous, no uniqueness lint, and the #2320 conformance suite will consume these identifiers.

**Fix direction**:
1. The org **decision charter** in .github profile: which record class (ADR/UPCR/spec/M-note) for which decision type, with the routing table.
2. A **10-line uniqueness lint** (CI or pre-push) + renumber the colliding UPCR + Superseded-by pointers.
3. **The AGENTS.md hard-routing**: the decision records are in NO agent-required path (CLAUDE.md/AGENTS.md mention ADR zero times) — agent traffic bypasses them faster as agent count grows; route decision-record touches through the charter.
4. The ADR template's mandatory **Evidence section** (the decision-evidence binding best practice scattered in edge repos — octos-arc's lock format, octoscode-web's supersession chains — consolidated).
5. Restore the octos-learn issue channel or designate the alternative ledger (its closure conflicting with LEP-003's issue-ledger paradigm in the same week).

Refs #2320 #2324 #2273

anantheparty:
Enrichment from the full architecture-decisions deep report (supplementing the filed issue):

[...]

Issue #2319, opened 2026-09-13
octos serve hardening: AppUI token first-frame auth (mid-term) and output redaction layer


anantheparty:
### Summary

Two server-side hardening items surfaced by a security audit of the octoscode-web browser client against `octos serve` (files/lines reference the client repo where they document the client side of the contract).

**1. AppUI token first-frame authentication (mid-term)**

The UI Protocol WebSocket currently authenticates via `?token=` in the query string (`packages/client/src/url.ts:22` in octos-org/octoscode-web; documented as a known compromise in its `docs/deployment.md` because browsers cannot attach an `Authorization` header to WebSocket handshakes). Consequences:

- any intermediary that logs full request lines (transparent proxies, APM gateways) retains a long-lived credential;
- the reference reverse-proxy config must deliberately scrub query strings from logs (`log_format` with `$uri` instead of `$request`), and any deployment that forgets this leaks tokens to logs;
- there is no TTL or revocation story for the token, and a 401 handshake failure is indistinguishable from an expired token.

Suggested direction: accept the token in the first protocol frame after the socket opens (or via `Sec-WebSocket-Protocol`), keep query-token as a fallback during a transition; add TTL + revocation and distinguish "expired" from "invalid" in the 401 semantics so clients can prompt re-auth instead of retrying.

**2. Output redaction layer for terminal/task output**

The browser client renders task output, protocol event params and workspace paths verbatim into the DOM (octos-org/octoscode-web `EventInspector.tsx:60`; the client intentionally owns no data semantics, so redaction responsibility falls to the server). In screen-share scenarios, environment values, API-key echoes and absolute workspace paths are visible to anyone looking at the screen. The onboarding error path already demonstrates the pattern client-side (`redactSecret()` in octoscode-web `use-onboarding.ts:260,322-323`).

Suggested direction: an opt-in secret-detection/redaction pass on `task` output and read channels in `octos serve`, plus a capability flag so clients can offer a "redacted view" toggle.

Happy to coordinate on the client side of both (octoscode-web is prepared to consume a first-frame auth handshake and a redaction capability flag).

PR #2407, opened 2026-09-18
feat(api): let a UI client stop a local `octos serve --solo` (server/shutdown)

ymote:
## Why
A web client connected to `octos serve` could not stop it; the only way was Ctrl+C in the terminal that launched it.

## What
- New raw AppUI method `server/shutdown`. It flips the same `watch` switch SIGINT/SIGTERM use, so it takes the same graceful path (drain, stop gateways, exit). The switch flips 250 ms after the reply so the client gets its answer first.
- **Advertised and callable only** when `local_solo_danger_allowed` (solo login + local deployment) **and** this is an HTTP serve with a stop switch (`AppState.serve_shutdown`; `None` for stdio and in tests). Otherwise it returns `invalid_request` with `kind: server_shutdown_unavailable` and stops nothing.
- Added to the session-ingress deny list: session-scoped connections can never call it.

## Tests
`should_advertise_server_shutdown_only_on_a_local_solo_http_serve`, `should_flip_the_serve_stop_switch_when_server_shutdown_is_called`, `should_refuse_server_shutdown_and_stop_nothing_when_unavailable`, `should_bar_session_scoped_connections_from_server_shutdown`; fmt + clippy clean.

Also checked end to end with the web client (octos-org/octoscode-web, feat/stop-server): Settings → Stop server → confirm → the server logs the graceful shutdown and exits in about 0.5 s, and the tab returns to the connect screen.


anantheparty (changes requested):
Adversarial review (zero-context, code read at head `8fce6ee1`, three-dot diff against `7f3cb5ce`): **design verified sound — requesting changes on one P1.**

**The security design holds.** Verified end-to-end:
- The gate chain (`--solo` opt-in ∧ Local deployment ∧ HTTP serve holds the stop switch) is evaluated per-request from `AppState` fields that are immutable after publication (both assignment sites precede `Arc::new`) — no TOCTOU, not forgeable without controlling process launch.
- The session-ingress raw-method deny is enforced at dispatch *before* `handle_raw_appui_rpc`, backed by the single-source-of-truth surface list, with `unreachable!` guarding dispatcher drift; ingress scope is fixed at upgrade; there is no mid-connection escalation path and no method-name normalization to exploit.
- The only reachable route is the AppUI WS, which enforces the Origin gate at upgrade — a DNS-rebinding page gets 403, so a malicious webpage cannot stop a user's server. No non-WS path reaches `serve_shutdown` (router table swept).
- Shutdown flips the same watch sender SIGINT/SIGTERM use: identical drain → gateway stop → exit path, idempotent on repeat.

**P1 — CI red: spec §6 catalog drift (must fix before merge).** `server/shutdown` was added to `APPUI_EXTRA_METHODS` (`ui_protocol_transport.rs:412`) without a §6 entry in `api/OCTOS_UI_PROTOCOL_V1_SPEC_2026-04-24.md`; `spec_section6_catalog_lists_every_advertised_method` fails (`ui_protocol_tests.rs:322`, `missing: ["server/shutdown"]`, run 35398984754: test-octos-cli 3813/1) with the usual `CI required` cascade. This is the drift-guard doing exactly its job — and since fixing it means editing the spec, the repo's convention also calls for a **UPCR document** for the new protocol method (per #2326/#2334; note the current `check-ui-protocol-upcr.sh` only enforces pairing when the spec is edited, so this PR bypassed the weaker gate — adding the UPCR now is the chance to do it right).

**P3s (none blocking, worth taking with the fix):**
1. `ui_protocol_transport.rs:9509-9515` — the "250 ms after the reply" wording is inaccurate: the flip is scheduled 250 ms from *request handling*, while the reply is written by the WS loop *after* the handler returns; if that connection's outbound queue is backpressured >250 ms the client never reads the ack before drain. Harmless (idempotent re-call), but the comment should say "250 ms from handling" or schedule after `send_rpc_result`.
2. `ui_protocol_tests.rs:20381-20405` — the raw-surface enumeration pin table is missing `APPUI_METHOD_SERVER_SHUTDOWN` (single-source-of-truth hygiene).
3. Record in the UPCR's risk statement: on a tokenless loopback solo serve, **one RPC stops the whole server** (any local process; any page behind the dev-allowlist origins). That's the #2388 trust model, not introduced here — but "one call = global stop" deserves to be written down, and is a natural input to #2319/#2147.

Also: base has moved (#2403) — a rebase/update will re-run the suite on the merged tree.

Everything else I checked came back clean: no TOCTOU in the gate, ingress can't reach the method, no non-WS exposure, dispatch-table completeness with drift guard, graceful-path identity with SIGINT, and all four new tests green in CI. Fix the catalog + UPCR and this is an easy approve.

ymote:
Thanks, all points addressed in `296289c7` (rebased on main incl. #2403):

- **P1:** the §6 catalog entry (`13df886f`) now cites the new **UPCR-2026-032** (`docs/OCTOS_UI_PROTOCOL_CHANGE_REQUEST_UPCR_2026_032_SERVER_SHUTDOWN.md`); `check-ui-protocol-upcr.sh` passes against it.
- **P3.1:** the comment now says the flip is scheduled 250 ms from *handling* and that under outbound backpressure the ack can be lost (the stop still happens; repeat calls are idempotent).
- **P3.2:** `raw_method_is_dispatched_covers_full_raw_surface` pins `APPUI_METHOD_SERVER_SHUTDOWN`.
- **P3.3:** the UPCR's Risk section records that on a tokenless loopback solo serve one call stops the whole server, under the #2388 trust model, with pointers to #2319/#2147.

Full octos-cli lib suite: 3812 passed, clippy clean.

anantheparty (approved):
Re-review at head `6289b774` — all points from my request-changes are addressed:

- **P1 resolved**: `server/shutdown` is in the §6 method catalog, and the new `UPCR_2026_032_SERVER_SHUTDOWN.md` carries the change — numbered sequentially, no collision with the historical double-files.
- **P3.1 adopted**: the 250ms comment now states accurately that the flip is scheduled from request *handling* and that a backed-up outbound queue means the client may never read the ack.
- **P3.2 adopted**: the raw-surface enumeration pin table includes `APPUI_METHOD_SERVER_SHUTDOWN`.
- **P3.3 adopted**: the UPCR's Risk section states "one call stops the whole server" on a tokenless loopback solo serve, anchored to the #2388 trust model as pre-existing rather than new.

The branch also carries the rebase across today's merges cleanly. The design verification from my earlier review (gate chain, ingress deny, Origin coverage, SIGINT-identical shutdown path) is unchanged by these doc/table/test additions. Approving — the pending test-octos-cli run is the final gate (its §6 drift-guard case is the one that matters for this fix).

PR #2408, opened 2026-09-18
fix(api): never blank a reply when shrinking an oversized frame

ymote:
## Why
Reopening a long session in the web client showed `[oversized field omitted]` where assistant replies should be, so finished turns looked like they had no summary.

The `session/hydrate` reply for a 2.9 MB session was squeezed under the 1 MiB frame cap largest-first. Each field got whatever budget the *rest* of the frame left, and while the rest was still over target that budget was 0. So the biggest fields were blanked wholesale: 408 stubs, 27 of them replies.

## What
`preview_oversized_frame` now cuts by **tier** — reasoning traces → tool arguments/output → conversation text — and within a tier caps every field at **one shared length** (water-filling, binary search):
- fields under the cap stay whole; fields over it become head+tail previews with the existing `[N bytes truncated]` marker;
- per-tier floors (reasoning/tool I/O ~384 B, conversation text ~1 KiB), so **no field is blanked** and the frame fits without the structural fallback dropping messages;
- replies are only shortened when reasoning and tool output can't absorb the excess, and even then they keep their head and tail.

Still a single serialize-to-verify, and the array-shrink structural fallback is unchanged.

## Tests
- `should_keep_every_assistant_reply_whole_when_reasoning_and_tool_output_can_absorb_the_cut`: ~3 MiB hydrate, all 40 replies whole, no stub, no message dropped.
- `should_preview_rather_than_blank_replies_when_the_replies_alone_are_over_the_cap`: each reply keeps head + tail + marker.
- Both fail on `main` (stub present). Existing oversized/truncation/preview/hydrate/marker tests pass (188), full octos-cli lib suite 3807 passed; fmt + clippy clean.

- `should_keep_every_message_and_reply_when_a_long_session_has_many_medium_fields`: 300 turns, ~5 MiB, all 1200 messages and every reply whole. (A first cut with a flat 2 KiB floor failed this: 1200 → 300 messages via the array-shrink fallback.)

## Checked on a real session
Hydrating the 1400-message session that showed the bug. Before: 408 stubs, 27 of them replies. After: 0 stubs, 1453/1453 messages, 407/407 replies whole, 276 tool outputs previewed, 830 KB frame.

Note: this PR and #2407 both append tests to the end of `ui_protocol_tests.rs`, so whichever merges second needs a trivial rebase (keep both blocks).

Longer term, hydrate should page history rather than send one frame; this makes the single frame degrade gracefully meanwhile.


anantheparty (approved):
Adversarial review (zero-context; the algorithm was extracted verbatim into an independent crate and re-verified against the author's scenario plus adversarial boundaries — 17/17 passing): **APPROVE**.

The water-filling math is correct: binary search on a monotonically non-increasing savings function converges without off-by-one, the frame budget is met exactly when the tier absorbs the excess (running_len measured against real serde_json escaping, 1x/2x cases land in [TARGET−300, TARGET]), and the structural fallback has a bounded 64-round termination. Tier semantics hold under the worst case I could construct (200×20K reasoning tokens dominating a 500B reply): the reasoning tier's cap absorbs everything alone and the reply survives byte-for-byte — the strongest evidence that conversation text really is cut last.

UTF-8 safety verified: head/tail cut points land on char boundaries by construction, and the marker's N equals the dropped raw bytes exactly (head.raw + tail.raw + N == original_raw). The old "blank the field" stub is unreachable in the new code (cap ≥ floor ⇒ preview always builds), and the new regression test asserting `!out.contains(UNPREVIEWABLE_STUB)` is red on main — fix-sensitive as it should be.

Non-blocking notes:
1. The key-name heuristic (any `output`/`stdout`/`stderr`/`arguments` string classifies as ToolIo, cut before session text) can mis-order non-tool user data in RPC results — affects cut order/retention only, never blanking. Fine as-is; tightening to require a tool_calls container path would be the follow-up if it ever matters.
2. The ~3000-turn × ~300B shape still falls into the structural fallback (fields under floor, string layer can't meet target) — pre-existing behavior, correctly disclosed, and the real fix is pagination. The second commit's floor drop (2 KiB → 384) is what keeps realistic shapes out of it.
3. base has moved (#2407) — expects a trivial rebase; test-octos-cli on the final head is the remaining gate.

PR #2409, opened 2026-09-18
feat(ui-protocol): say when a turn is certainly not running (UPCR-2026-031)

ymote:
## Why
If a client sends `turn/start` as the server dies, the restarted process has neither a registry entry nor a ledger record for that turn. `turn/state/get` then answers `unknown` every time. A client that won't risk running a prompt twice holds sending indefinitely: the web client sat on "Sending paused" until the tab was reloaded.

Turns the ledger *did* record are already settled at boot (`orphaned_by_restart`). This covers only the unrecorded case.

## What (UPCR-2026-031, additive)
- `TurnStateGetResult.running: Option<bool>` is sent as **`false`** with `state: "unknown"` only when the server is **certain** it isn't executing the turn: the session is known, there's no registry entry, no ledger lifecycle state, and **no `turn/start` for that `(session_id, turn_id)` mid-admission**.
- An RAII `TurnAdmission` marker is held across `handle_turn_start_with_accept`, so the window before the registry insert is never reported as not running.
- `state` keeps its UPCR-2026-011 meaning. Absent whenever not certain. Old clients ignore the field.
- Spec §`turn/state/get` updated; new `docs/…UPCR_2026_031_TURN_STATE_NOT_RUNNING.md`.

## Tests
- `turn_state_get_returns_unknown_for_missing` now also pins `running: false`
- `should_not_claim_a_turn_is_stopped_while_its_start_is_still_being_admitted`
- `should_say_a_turn_is_not_running_once_its_admission_has_ended_without_a_record`
- Full octos-cli lib suite 3807 passed, octos-core 381 passed; clippy clean.

Client side: octos-org/octoscode-web, which settles a turn automatically on `unknown` + `running: false` and keeps the manual "Continue without it" button for older servers.


anantheparty:
Blocking correctness issue at [`ui_protocol_transport.rs:26685-26689`](https://github.com/octos-org/octos/blob/bcf469577e43b59f9a52fa40c8da09556a52678c/crates/octos-cli/src/api/ui_protocol_transport.rs#L26685): the certainty check combines a **pre-admission registry/ledger snapshot** with a **post-admission marker**. I reproduced `state: "unknown", running: false` while the same turn's live registry state was still `Active`, using the real handler at this exact head.

The reachable ordering is:

1. A start has entered `TurnAdmission`, but has not inserted its registry entry.
2. The lookup captures `registry_state = None` and an empty projection at lines 26602-26622.
3. The lookup awaits the SessionManager/context work at lines 26628 onward. Admission can now insert the active turn and drop its RAII marker.
4. The final `in_progress()` check sees false, combines it with the stale snapshots, and incorrectly reports `running: false`.

The existing admission test holds its marker for the entire lookup, so it cannot expose this transition. I added this deterministic reproduction beside those tests in an isolated worktree; no production hook or sleep is needed:

```rust
#[tokio::test(flavor = "current_thread")]
async fn not_running_must_not_mix_pre_admission_registry_with_post_admission_marker() {
    let session_id = SessionKey("local:not-running-admission-race".into());
    let state = prg_state_with_session(&session_id, |_| {});
    let active_turns: SharedActiveTurns = Arc::new(TokioMutex::new(HashMap::new()));
    let turn_id = TurnId::new();
    let admitting = TurnAdmission::enter(&session_id, &turn_id);

    // Pause before the registry read, then at the existing metadata await.
    let registry_guard = active_turns.lock().await;
    let lookup = turn_state_frame(&state, &session_id, &active_turns, turn_id.clone());
    futures::pin_mut!(lookup);
    assert!(futures::poll!(lookup.as_mut()).is_pending());
    let sessions_guard = state.sessions.as_ref().unwrap().lock().await;
    drop(registry_guard);
    assert!(futures::poll!(lookup.as_mut()).is_pending());

    // The normal admission transition: register the live turn, drop marker.
    let task = tokio::spawn(std::future::pending::<()>());
    active_turns.lock().await.insert(
        session_id.clone(),
        test_active_turn(turn_id, task.abort_handle()),
    );
    drop(admitting);
    drop(sessions_guard);

    let frame = lookup.await;
    let registry = active_turns.lock().await;
    assert!(matches!(
        *registry.get(&session_id).unwrap().state.lock().await,
        TurnState::Active
    ));
    task.abort();
    assert_ne!(frame["result"]["running"], false, "{frame}");
}
```

`cargo test -p octos-cli --lib --no-default-features --features api not_running_must_not_mix_pre_admission_registry_with_post_admission_marker -- --nocapture` fails on that final assertion (1 run / 1 failure).

Please compute the negative certainty from a coherent registry/admission observation, or revalidate both together after the intervening awaits so a completed admission cannot disappear between the two observations. The automatic browser release in octoscode-web#138 should wait for this fix; adding browser retry/defensive layers would not establish the missing server guarantee. Separately, `not running` still must not be rendered as proof of completion or interruption—the client patch is being adjusted to preserve that distinction.

anantheparty (changes requested):
Adversarial review (zero-context, code read at head `bcf4695`): **request changes** — the design direction is right and the RAII placement *inside* `handle_turn_start_with_accept` is airtight, but the `state/get`-side check sequence has a cross-connection race, and the guarantee the UPCR promises to clients is currently stronger than what the code enforces.

**F1 (Medium) — cross-connection TOCTOU in the state/get path.** The registry read (`ui_protocol_transport.rs:26602-26614`) and the `TurnAdmission::in_progress` check (`:26688`) are separated by several awaits. Interleaving: conn-B `state/get` reads registry (miss) → conn-A `turn/start` inserts into the registry *and* drops its admission marker (`:23094`) → conn-B then evaluates `in_progress == false` → answers `unknown` + `running:false` → the turn runs immediately after. Same-connection dispatch is serialized (`select!`, :6714/:7063) so it's safe there; the race needs two connections (e.g. the web polling connection vs the user's sending connection) — exactly the shape the web client's "Sending paused" bug lives in. Fix: evaluate `in_progress` inside the same registry lock acquisition (or double-check after acquiring).

**F2 (Medium) — the wiring has no test pinning it.** Both new tests construct `TurnAdmission::enter` by hand (`ui_protocol_tests.rs:26805, :26820`); deleting the `let _admission = ...` wiring at `:22833` leaves the full 3807-test suite green. Please add: (a) an end-to-end test where a `turn/start` is in-flight (pre-registry-insert) while `state/get` runs, asserting `running` is absent; (b) ideally a concurrent variant.

**F3 (Low-Med) — topic-folding key mismatch.** The marker is keyed on the raw `session_id` (`:22833`) while the registry uses the folded id (`:22944` → `:23094`); `state/get` takes no topic parameter (`ui_protocol.rs:3115-3118`), so a topic client necessarily queries by the folded id — during the admission window the keys don't match and `running:false` is emitted for a turn that is mid-admission.

**F4 (Low-Med) — two sister startup paths bypass the marker**: `handle_review_start` (`:22279`, insert `:22410`, accepts a client-supplied turn_id) and `maybe_spawn_appui_master_continuation_runner` (`:23456`, insert `:23839`). The UPCR's client-facing "proof"条款 (`:44-46`) is unconditional; these pre-insert windows can produce wrong certainty too.

**F5 (Low)** — a ledger read failure (`:26619-26622`, Err → None) is treated as "no lifecycle state" and asserts `false` without having verified the four conditions. Limited blast radius (a running turn should have a registry entry), but the certainty claim should degrade to *absent* on read errors.

**F6 (Info)** — spec catalog (`:423`) and wire inventory (`:40`) lack the UPCR-2026-031 annotation (repo convention, cf. inventory `:42-45`); the static comment at `:23150` is a truncated sentence; and the docs should state explicitly that the certainty is **per-process** (a restart loses the registry — that's the bug being fixed, but clients should know the certainty scope).

Everything else verified clean: all 8 early-return paths inside the marker scope, insert strictly precedes marker drop, refusal/accept-fail handled, non-Clone type + poison-lock recovery, no lock-ordering inversion, the `running:false ⟹ state:"unknown"` matrix proven (`:26663-26689`), additive serde compat both directions (golden fixtures + wire presence/absence assertions), no UPCR number collision with 022/026/027 (those double-files predate this branch), single commit, purely additive, no smuggled changes. CI 26/26 green at review time; branch needs a rebase (base moved to #2403).

ymote:
Thanks, all six addressed in the latest commit (rebased on main):

- **F1:** `TurnAdmission::in_progress` is now read inside the same `active_turns` lock acquisition as the registry lookup. Admission inserts under that lock before its marker drops, so a cross-connection start is seen as admitting or registered, never as neither.
- **F2:** a `#[cfg(test)]` seam (`turn_admission_test_pause`) holds a **real** `turn/start` inside its admission window. `should_withhold_not_running_while_a_real_turn_start_is_mid_admission` and `…_for_a_topic_turn_asked_by_its_folded_id` each include a control that asserts `running: false` for an idle turn on the same state, so the check can fail. Mutation-checked: deleting the `let _admission`/folded wiring fails both.
- **F3:** a topic turn is also marked under the folded session id.
- **F4:** `handle_review_start` (when the client supplies `turn_id`) and `maybe_spawn_appui_master_continuation_runner` now hold markers across their pre-insert windows.
- **F5:** a ledger read error now withholds `running` (`projection.is_some_and(state.is_none())`).
- **F6:** the spec catalog and wire inventory cite UPCR-2026-031, the static doc comment is rewritten, and the UPCR states the certainty is per process and lists every start path covered.

Full octos-cli lib suite 3812 passed, clippy clean, `check-ui-protocol-upcr.sh` passes.

anantheparty (dismissed):
Re-review at head `0554ea98`:

- **F1 (TOCTOU) resolved**: the admission check now reads inside the same `active_turns.lock()` critical section as the registry lookup (`let (registry_state, admitting) = { let registry = active_turns.lock().await; … }`), with a comment citing UPCR-2026-031 and "never as neither". The cross-connection interleaving from my review can no longer produce `running:false` for an admitted turn.
- **F3 (topic-folding key mismatch) resolved, elegantly**: raw-key entry at `turn/start` (as the client sent it), then a conditional second entry with the folded key after the UPCR-2026-015 fold (`let _folded_admission = (params.session_id.0 != raw_session_id.0).then(…)`), so both the raw and the folded keys report in-progress during admission and the refcount handles the double-enter.
- **F4 resolved**: both sister startup paths (`review/start`, master continuation runner) now hold the marker.
- **F5 (Low) open**: a ledger `snapshot_with_cursor` failure still degrades to `projection = None` and asserts `running:false`; the docs' own "certain" standard would prefer falling back to *absent* on read errors. Blast radius is limited (an in-process running turn always has a registry entry).
- **F6 (Info) resolved**: spec §catalog and wire inventory now carry the UPCR-2026-031 annotation.

**One follow-up I'd like tracked (not blocking this merge)**: F2's end-to-end test. The enabling hook is in (`turn_admission_test_pause`, cfg(test), right after the folded-key entry), but no test drives a real `turn/start` to that pause while `state/get` runs — the synthetic-marker tests still don't pin the wiring. I'll file a standalone issue so it stays visible; deleting the marker wiring line should ideally turn that test red.

The rebase across today's merges is clean and CI was green on the pre-rebase head. Approving; auto-merge will land it when the updated head's checks clear.

anantheparty (approved):
Re-review at head `f12eb636` (the 06:45Z force-push) — **APPROVE**.

All request-changes points verified in the incremental diff (`0554ea98…f12eb636`):

- **F1 (TOCTOU) closed**: `turn/state/get` now reads the registry and evaluates `TurnAdmission::in_progress` inside the same `active_turns.lock().await` critical section (`:26700-26701`) — the cross-connection interleaving from my review can no longer emit `running:false` for an admitted turn.
- **F4 closed**: all three sister startup paths now hold the marker — `review/start` (`:22364` via the turn-id map), the folded-key conditional entry (`:22921`), and the master continuation runner (`:23684`).
- **F5 closed**: ledger read failures now withhold the certainty (fail-closed) instead of asserting `running:false` on an unverified absence.
- **F6 closed**: spec §catalog + wire inventory carry the UPCR-2026-031 annotation; the UPCR-2026-031 document (+65 lines) is in the branch.

An independent correctness delta audit also read this exact force-push diff and found no new holes. CI on this head was green (including the SPA reducer fixture replay). Approving — the armed auto-merge can land it.

anantheparty (approved):
Re-review at head `f12eb636` (the 06:45Z force-push) — **APPROVE**.

All request-changes points verified in the incremental diff (`0554ea98…f12eb636`):

- **F1 (TOCTOU) closed**: `turn/state/get` now reads the registry and evaluates `TurnAdmission::in_progress` inside the same `active_turns.lock().await` critical section (`:26700-26701`) — the cross-connection interleaving from my review can no longer emit `running:false` for an admitted turn.
- **F4 closed**: all three sister startup paths now hold the marker — `review/start` (`:22364` via the turn-id map), the folded-key conditional entry (`:22921`), and the master continuation runner (`:23684`).
- **F5 closed**: ledger read failures now withhold the certainty (fail-closed) instead of asserting `running:false` on an unverified absence.
- **F6 closed**: spec §catalog + wire inventory carry the UPCR-2026-031 annotation; the UPCR-2026-031 document (+65 lines) is in the branch.

An independent correctness delta audit also read this exact force-push diff and found no new holes. CI on this head was green (including the SPA reducer fixture replay). Approving — the armed auto-merge can land it.

anantheparty:
Follow-up defect found in a post-merge delta audit — the f12eb636 folded-marker fix is **asymmetric between the two start paths**:

- `handle_turn_start_with_accept` correctly holds **both** markers (raw `:22891` + folded `:22921` after the UPCR-2026-015 fold).
- `handle_review_start` holds **only the raw marker** (`:22364`, `params.session_id` as sent) — then `:22564` folds the topic, `:22624`+ runs async voice-admission resolution (await points), and the registry insert lands under the **folded** id.

Window: a `review/start` carrying a `topic` with a client-supplied `turn_id`, while a second connection queries `turn/state/get` by the folded id during admission → registry miss + empty ledger + `admitting=false` → `state:"unknown", running:false` — the exact false-certainty UPCR-2026-031 set out to eliminate, now on the review path.

Suggested fix: the symmetric `_folded_admission` after the fold (mirroring `:22921`), same shape as the turn/start fix. ~2 lines.

(Found in an independent correctness delta audit; filing here since the defect lives in this merged change.)

Issue #2344, opened 2026-09-15
security: avoid tenant auth tokens in setup-script URLs


chrislearn:
## Summary

The cloud self-service registration flow embeds a tenant's long-lived dashboard authentication token directly in a public setup-script URL:

```text
GET /api/register/setup-script/{id}/{auth_token}
```

The generated Unix command consequently has the form:

```text
curl -fsSL "https://<domain>/api/register/setup-script/<tenant>/<auth_token>" | bash
```

Relevant code:

- Public route registration: `crates/octos-cli/src/api/router.rs:919`
- Public handler and token comparison: `crates/octos-cli/src/api/admin.rs:4715-4737`
- URL generation: `crates/octos-cli/src/api/admin.rs:4746-4753`
- The value is the tenant dashboard auth token and has no expiry field: `crates/octos-services/src/tenant.rs:39-41`

## Impact

Bearer credentials in URLs can be retained by reverse proxies, load balancers, WAF/APM tooling, browser history, terminal history, and other request-line telemetry. Although the application's matched-route tracing records the route template rather than the concrete token, external infrastructure is not guaranteed to do so.

This is not a narrowly scoped download token: the URL contains the tenant's persistent dashboard auth credential. A leaked setup URL can therefore remain useful beyond the setup operation.

## Existing safer path

An authenticated owner-scoped endpoint already exists:

```text
GET /api/register/setup-script
Authorization: Bearer <user session>
```

It resolves the tenant from `AuthIdentity`, so the public long-lived-token route is avoidable.

## Suggested direction

1. Stop generating URLs containing `TenantConfig.auth_token`.
2. Remove the public `/{id}/{auth_token}` route without redirecting the credential-bearing URL.
3. Use the existing authenticated endpoint for dashboard downloads.
4. Make registration email link to the authenticated dashboard rather than embedding the persistent credential.
5. If unattended one-command installation is required, mint a separate cryptographically random setup grant that is short-lived, single-use, owner-scoped, rate-limited, and atomically consumed. It must not reuse `tenant.auth_token`.

## Acceptance criteria

- No generated setup URL, email, or request path contains `TenantConfig.auth_token`.
- The legacy public credential-bearing route no longer serves setup scripts.
- An authenticated tenant owner can still retrieve the setup script.
- If temporary setup grants are retained, expiry, replay, cross-tenant access, and concurrent double-consumption are tested.
- Request-tracing tests assert that persistent credentials are absent from emitted URLs/logs.

## Related

[...]

PR #2410, opened 2026-09-18
security(session-ingress): carry the work secret in the Authorization header, drop query-token aliases (#2370)

### LINKED ISSUE #2370: security: session-ingress WS accepts the work-secret via query string (?token=/?_token=/?session_ingress_token=) — request-line retention risk
**Found in the 2026-09-17 zero-context security audit** (evidence at `main@d7612a31d4`).

`/v1/session_ingress/ws/{session_id}` (`crates/octos-cli/src/api/session_ingress.rs:66-80`) accepts its work-secret not only from the `Authorization` header but also from **three query parameters**: `?token=`, `?_token=`, `?session_ingress_token=`. This endpoint is the new external CLI-agent session entry shipped alongside the browser client, so the pattern is live on a fresh attack surface.

Credit where due: this token already has TTL + revocation and distinguishes invalid/expired/revoked (`session_ingress.rs:31-48`) — better than the main AppUI token (#2319). But placing a bearer credential in the request line exposes it to reverse-proxy/APM request-line telemetry and intermediary logs — the same failure class as #2319 (AppUI WS `?token=`) and #2344 (setup-script URLs).

**Suggested fix**
- Header-first extraction; keep `?token=` as a documented-deprecated fallback; drop `_token`/`session_ingress_token` aliases.
- Or carry the credential in `Sec-WebSocket-Protocol` / first-frame auth (the #2319 mid-term design).
- Add a request-tracing test asserting credentials never appear in any emitted URL/log line.

Related: octoscode-web#2319-class work, #2344.

BH3GEI:
## Problem

`/v1/session_ingress/ws/{session_id}` accepted its work secret not only from the `Authorization` header but also from three query parameters (`?token=`, `?_token=`, `?session_ingress_token=`). A bearer credential in the request line is exposed to reverse-proxy and APM request-line telemetry and intermediary logs — the failure class flagged in #2370 (same class as #2319 and #2344).

## Root cause

`extract_session_ingress_token` treated the query string as a co-equal credential carrier and accepted three spellings of it. On the pre-fix binary, a real `octos serve` upgraded the WebSocket to 101 with `?session_ingress_token=<valid secret>` — the token lands verbatim in any request-line retention path between client and server.

## Fix

- Header-first extraction, unchanged precedence: `Authorization: Bearer <token>` wins; a non-empty query `token=` remains as the **documented-deprecated fallback** for WebSocket clients that cannot set headers, and its use now logs one warning per authenticated connect so operators can migrate clients off it.
- The `_token=` / `session_ingress_token=` aliases are removed. A request that carries its credential only under a removed alias now gets a diagnostic 401 (`credential query parameter removed; send Authorization: Bearer <token> or ?token=`) instead of a misleading bare "missing" message.
- The deprecation warning fires only after the grant validates, so unauthenticated scanners fuzzing `?token=` cannot flood the log with it.
- The serve-log redaction layer (`SERVE_LOG_QUERY_TOKEN_RE`) learns the removed alias spellings, so retained URLs from stale clients are masked like the canonical ones; `my_token=`-style keys are untouched.
- The doc example and the work-secret guide now show the header carrier, note the deprecated fallback, and the `websockets >= 14` requirement for `additional_headers=`.

The warning message is a static string — no credential material is ever interpolated.

## Test evidence

Baseline red / fixed green (issue reproduction):

- `work_secret_ws_rejects_removed_query_token_aliases` (new): on pristine `origin/main` this fails — the server upgrades to 101 for `?session_ingress_token=<valid secret>`; with the fix it asserts 401 **and** the diagnostic body naming the removal.
- `work_secret_ws_accepts_authorization_bearer_header` (new): pins that the header carrier performs a full JSON-RPC round trip with no query parameter; green on main before the fix (documents the pre-existing path) and after.
- `extracts_percent_decoded_query_token` / `extracts_bearer_token_before_query_token` updated to the carrier-aware return; new unit tests pin alias rejection, the removed-alias diagnostic detector (including `not_token=` / `xsession_ingress_token=` non-matches), and missing/empty-token 401s.
- `http_trace_uses_route_templates_and_omits_credentials` extended with the ingress route: asserts the emitted span is `/v1/session_ingress/ws/{session_id}` and that neither the token marker nor the session id appears in any log line (issue's tracing requirement).
- `serve_log_redaction_masks_removed_ingress_token_aliases` (new) pins the redaction regex against both removed spellings and the `my_token=` false-positive guard.
- `cargo test -p octos-cli --lib`: **3809 passed; 0 failed; 10 ignored** in a single run.
- `cargo test -p octos-cli --test session_ingress_work_secret`: 5/5.
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets`, `cargo clippy -p octos-cli --all-targets --no-default-features --features api,telegram,discord,dingtalk,slack,whatsapp,feishu,email,matrix,audio_mp3`, and the minimal-production-lib variant all clean under a `-D warnings` gate (local clippy 1.98 flags one pre-existing `result_large_err` at `handlers.rs:2140` that CI's pinned 1.97.1 does not; the function is untouched by this PR).

## End-to-end evidence

Real-binary before/after with `octos serve` + `octos auth issue-work-secret` on loopback (raw-socket RFC6455 client):

| carrier | main (before) | this branch (after) |
|---|---|---|
| `?session_ingress_token=<secret>` | **101 upgrade succeeds** | 401 + diagnostic body |
| `?_token=<secret>` | **101 upgrade succeeds** | 401 + diagnostic body |
| `?token=<secret>` | 101 + JSON-RPC answered | 101 + JSON-RPC answered, exactly one deprecation WARN in the rolling log |
| `Authorization: Bearer` | 101 + JSON-RPC answered | 101 + JSON-RPC answered, no warning |

After the fix, the token appears in neither the serve process output nor the rolling log (0 occurrences), and the session id never appears in a log line.

One transparency note on local test runs: during verification, one full `--lib` run reported 3 failures while a `cargo build` I had launched concurrently was still compiling against the same target dir. The output of that run was truncated (tail-only), so the failing test names were unrecoverable; I re-ran the identical sources twice under controlled compile load (once with a concurrent octos-cli build, once with a full `--workspace --all-targets` build) — both 3809/0, with every test touched by this diff green in all runs. The identical-sources rerun rules out any deterministic effect of this change; the transient failures are consistent with the suite's documented load-sensitive tests. The 3-failure run was my orchestration error (concurrent builds), disclosed here for completeness.

## Review

An independent adversarial review (maintainer perspective, zero context) returned APPROVE WITH NITS with no Blocker/Major findings. Four findings were adopted in the follow-up commit (post-validation warning placement, diagnostic 401 for removed aliases, serve-log redaction coverage, websockets version note); two were declined with reasons: the throwaway local e2e probe script is not part of this PR, and a log-capture assertion inside the integration binary would reintroduce the process-global-subscriber cross-talk class that #2364 just fixed — the warning is a compile-time static string and the trace-span leak path is already pinned with a positive control.

Closes #2370


anantheparty (approved):
Verification review (the PR body already carries an unusually complete self-evidence trail — before/after real-binary table, 3809/0, baseline-red regression, disclosed orchestration flake — so this was a spot-check of the load-bearing claims, all of which held):

1. **Removed aliases → diagnostic 401**: confirmed in `session_ingress.rs` — `?_token=` / `?session_ingress_token=` no longer authenticate and return the diagnostic body naming the removal.
2. **Redaction regex**: the updated `SERVE_LOG_QUERY_TOKEN_RE` alternation puts the longest alternative first (`session_ingress_token` before `_token`/`token`), and the `\b`-prefix semantics naturally protect `my_token=` (no word boundary inside `my_token`) — with a new test pinning exactly that false-positive guard.
3. **Header precedence + deprecated fallback**: `Authorization: Bearer` wins; non-empty `?token=` remains as the documented fallback with a single post-validation warning (unauthenticated fuzzing can't flood it).
4. **Diff audit**: 5 files, deletions are the alias arms + their old tests; no behavioral stowaways. The tracing test extension (route template asserted, token marker + session id absent from log lines) covers the issue's tracing requirement.

Approving. CI tail (check / check-windows / test-octos-cli) pending at review time — auto-merge will land it when they clear.

One follow-up candidate for a later PR (not this one): the same header-first + deprecation pattern is exactly what #2319's AppUI WS token path needs when it lands.

PR #2411, opened 2026-09-19
fix(deploy): keep the dashboard bearer token out of the NSSM service argv (#2380)

### LINKED ISSUE #2380: security: deploy.ps1 NSSM service registration keeps the dashboard bearer token in service argv

BH3GEI:
## Problem

`scripts/deploy.ps1` registered the Windows service with the dashboard bearer token on the `nssm install` command line:

```powershell
& $nssmExe install $serviceName $octosExe "serve" ... "--data-dir" $dataDir "--auth-token" $authToken
```

NSSM stores those arguments as `AppParameters` and re-spawns `octos.exe` with them, so the token sits in a service command line readable by any local user via `Get-CimInstance Win32_Process` / `sc qc` — the same failure class #2371 closed for the four sibling generators (`install.sh` launchd/systemd, `install.ps1`, `local-tenant-deploy.sh`). deploy.ps1 was the remaining persistent-argv site.

## Root cause

The script carries the token twice: once in argv and once in `AppEnvironmentExtra` (`OCTOS_AUTH_TOKEN=...`). `resolve_auth_token` ranks `--auth-token` > `OCTOS_AUTH_TOKEN` > config, so the argv copy always won while being the only copy with process-list exposure.

## Fix

- Drop `--auth-token` from the `nssm install` argv. The effective token is unchanged: `AppEnvironmentExtra` already delivers the identical `OCTOS_AUTH_TOKEN` value, deploy.ps1 guarantees a non-empty token (auto-generates one if empty) before registration, and re-runs go through a full `stop`/`remove`/`install` cycle so no stale value survives. The env channel is the exact mechanism #2381 left in place for this script.
- Add a `$LASTEXITCODE` guard on the `AppEnvironmentExtra` set — now that it is the only token channel, a silent failure must not quietly fall back to a (possibly rotated-away) `config.json` token.
- Extend the `#2371` tripwire `service_templates_never_pass_auth_token_via_argv` (octos-cli `serve.rs` tests) to cover `deploy.ps1`; its doc comment previously named this path as the known remaining exception.
- `scripts/test-windows-deploy.sh`: the argv-form assertion is replaced with a negative guard plus env-channel assertions, statically and in the pwsh dry-run branch.
- CI `windows-deploy-script` (real windows-latest pwsh execution of the script): add a `--auth-token` negative assertion and an `AppEnvironmentExtra` positive assertion.

## Test evidence

- Red→green: with the tests applied and deploy.ps1 still at main state, the bash test fails (`deploy.ps1 must not pass the bearer token via service argv`) and the Rust tripwire panics naming the exact issue line: `deploy.ps1 puts --auth-token in the service argv: & $nssmExe install ... "--auth-token" $authToken`. After the fix both pass.
- End-to-end before/after (PowerShell 7.4.10, `deploy.ps1 -DryRun` — the same execution shape the CI job runs on windows-latest):
  - before: `& $nssmExe install ... "--data-dir" $dataDir "--auth-token" $authToken`
  - after: `& $nssmExe install $serviceName $octosExe "serve" "--host" "0.0.0.0" "--port" "$servePort" "--data-dir" $dataDir` with `AppEnvironmentExtra ... "OCTOS_AUTH_TOKEN=$authToken"` intact.
- `cargo test -p octos-cli --lib`: 3808 passed / 0 failed / 10 ignored, single run. `cargo fmt` clean. `cargo clippy -p octos-cli --all-targets`: only the pre-existing local-rustc `handlers.rs:2140 result_large_err` warning (function untouched by this diff). `shellcheck`/`actionlint`: finding counts identical to the main baseline (6 SC2016 / 1 runner-label).
- Real NSSM service registration on a live Windows host is not runnable from this machine; the authoritative execution gate is this PR's `windows-deploy-script` job, which actually invokes the script under pwsh on windows-latest and now fails if the argv form returns.

## Review

One independent adversarial review round (with its own compile/test/dry-run verification): no Blocker or Major findings. Adopted: the `AppEnvironmentExtra` `$LASTEXITCODE` guard, the tripwire's `AppParameters` arm (a `nssm set AppParameters --auth-token ...` reintroduction shape), and the comment anchor moved from #2371 to #2380. Rejected with evidence: merging the `OCTOS_AUTH_TOKEN=` and token-value assertions into one grep — the dry run prints the template source, where the env line is the literal `"OCTOS_AUTH_TOKEN=$authToken"` (the value interpolates only at remote runtime), so the merged assertion can never match; the separate assertions pin the channel and the `$authToken` assignment that feeds it. All validations above were re-run to green after the review-driven changes.

Pre-existing exposures left out of scope (static-store class, tracked by #2388 / #2344 surfaces): the token also lands in the remote `config.json`, NSSM's `Parameters` registry key is default-readable by local users, the deploy-window `EncodedCommand` base64 line transiently carries the script, and the install-time-only argv sites (`cloud-host-deploy.sh`, `frp/bootstrap-tenant.sh`, the admin setup script).

Closes #2380


anantheparty (approved):
Adversarial review (zero-context, code read at head `abeb9e0e`): **APPROVE**.

- The token now travels via `nssm set <svc> AppEnvironmentExtra "OCTOS_AUTH_TOKEN=…"` (`deploy.ps1:303`) — the documented NSSM environment mechanism — with a `$LASTEXITCODE` guard that fails closed (service stays stopped rather than silently falling back to a possibly-rotated config token). `AppParameters` residue: zero (`--auth-token` count in the head script = 0), and the reinstall path (stop/remove/install) clears the old Parameters key.
- The tripwire now covers deploy.ps1: two new arms (`"$nssmExe install"` and `AppParameters` — the latter catching the "re-add via `nssm set AppParameters`" regression form), the stale "known remaining exception" doc note removed, and the red→green claim verified: the same predicate trips at exactly base:291 and passes on head. The bash negative grep in test-windows-deploy.sh provides the second layer.
- Token semantics unchanged: argv removal hands off to the env channel at the same value (priority env>config verified at `serve.rs:483-498`), and the non-empty guarantee precedes registration.

Minor notes for the record (none blocking):
1. "Out of argv" ≠ "unreadable by local users": the token still lives in the NSSM registry value (`HKLM\…\AppEnvironmentExtra`, REG_MULTI_SZ, default ACL readable by local users) and in the remote config.json — both explicitly out of scope here and tracked by #2388/#2344. The remaining hardening step for this family is a credential store or 0600-style env file.
2. The Rust tripwire is line-shape matching: a multiline splat (`$svcArgs = @("serve","--auth-token",$t)` + splatting call) evades both arms (the bash full-file grep still catches it) — inherent limit of string tripwires, fine for now.
3. Pre-existing PS 5.1 quoting edge (user-supplied token containing both `"` and a space) — same risk existed in the argv form; auto-generated tokens are hex and safe.

windows-deploy-script (the authoritative gate for this change) is green at head. Merging as squash.

Issue #2390, opened 2026-09-18
perf(memory): RecallStore::open is O(records × manifest-layout) + full-table reparse — ADR-001 cold-start acceptance has no executable gate


anantheparty:
**From the 2026-09-17/18 performance audit** (octos `e89e266f`).

`RecallStore::open` has three compounding costs, and ADR-001's cold-start acceptance has no executable gate anywhere in the repo:

1. `octos-memory/src/recall.rs:379-383` — on manifest hit, every record not in the layout does `m.layout.iter().any(...)`: **O(records × layout)**. At ADR-scale (15k records × 10k layout) that's ~1.5×10⁸ String comparisons — directly threatening the ADR's "cold start < 1 s" acceptance (personal-memory-tiers.md:65). The fix is a HashSet.
2. `recall.rs:254-267` — full-table load: per-record `serde_json::from_str` + int8 vector decode into a HashMap **on every process start**.
3. `recall.rs:371` — even when the graph reloads from manifest, the BM25 corpus is rebuilt for all records (re-tokenizing everything in the `attach_hnsw` closure); and `recall.rs:389-415` on manifest mismatch does a full HNSW re-insert plus a **synchronous** `persist_index()` during open.

Meanwhile ADR-001's three acceptance criteria (recall p95 < 50ms on a OnePlus 6; cold start < 1s; 500-page ingest ≥90%) have **no harness, no benches, no CI gate** — `.github/workflows` contains zero perf jobs and neither octos-memory nor octos-agent has benches. #2323 covers the org's four measurement assets; this is the kernel memory subsystem itself flying unmeasured.

**Suggested actions**
1. HashSet for the layout membership check (a one-line-class fix).
2. A `criterion` micro-bench for `RecallStore::open` on a synthetic 15k-record set, run in nightly.
3. Write the three ADR acceptance criteria as executable assertions wired to that bench.

Related: #2377 (memory/ingest governance), #2323 (measurement consolidation).

Issue #2330, opened 2026-09-14
security: sandbox posture is per-host auto-negotiation with fail-open default — no guaranteed-isolation tier


anantheparty:
Fresh-eyes security-lens deep round finding.

**Current state**: SandboxConfig::default { mode: Auto, fail_closed: false } — hosts without bwrap/sandbox-exec/AppContainer/docker helpers (minimal containers, most CI runners) → NoSandbox full pass with ONE tracing warn (warn_once, anti-spam by design); doctor surfaces it only on demand. The workspace path boundaries only constrain the FILE tools — the SHELL is the primary execution face, so the boundary system's foundation is empty on most backend-less hosts.

**Fix direction**: a confinement-required profile (fail_closed=true preset) for security-sensitive deployments; a CI assertion tier asserting 'no-backend must refuse to start'; the doctor sandbox verdict surfaced into the transcript.

**Overlap**: #2195 is the Windows platform special case; #2321-② is the install-time special case — this is the unified sandbox-posture parent issue referencing both.

Refs #2195 #2321

Issue #2147, opened 2026-08-27
Make effective runtime policy a durable session resource; decouple dangerous authority from solo login


anantheparty:
## Summary

Permission choice is durable session runtime state, not ephemeral Web/TUI state. Persist the effective runtime policy against an authoritative session identity and expose an exact server-owned preset/transition catalog.

Separately, decouple dangerous execution authority from passwordless local-solo login. An operator should be able to keep solo login disabled while explicitly allowing an authenticated administrator to select Full Access. This remains default-off and cannot create a client-side bypass.

## Current contract and failure mode

At `main@10022387`:

- explicit choices live in a process-global `OnceLock<Mutex<HashMap<SessionKey, ...>>>`; they are neither durable nor scoped by profile + Workspace. [`ui_protocol_transport.rs`](https://github.com/octos-org/octos/blob/10022387da44949cdacb07e2e2fcf38a85155455/crates/octos-cli/src/api/ui_protocol_transport.rs#L1537-L1593)
- SPA sessions may use raw `web-*` keys; source comments already say `SessionKey` alone is not a Workspace isolation boundary. [`ui_protocol_transport.rs`](https://github.com/octos-org/octos/blob/10022387da44949cdacb07e2e2fcf38a85155455/crates/octos-cli/src/api/ui_protocol_transport.rs#L1525-L1535)
- the selectable catalog normally returns Read/Deny and WorkspaceWrite/Deny, plus Full/Allow only behind the solo gate. [`ui_protocol_transport.rs`](https://github.com/octos-org/octos/blob/10022387da44949cdacb07e2e2fcf38a85155455/crates/octos-cli/src/api/ui_protocol_transport.rs#L9003-L9042)
- a fresh runtime can effectively be WorkspaceWrite/NetworkAllow even when that combination is absent from the selectable catalog, so `current` is not closed over the advertised presets.
- Full Access authority is keyed directly to `solo_login_enabled`. [`ui_protocol_transport.rs`](https://github.com/octos-org/octos/blob/10022387da44949cdacb07e2e2fcf38a85155455/crates/octos-cli/src/api/ui_protocol_transport.rs#L8173-L8191)
- `permission/profile/set` only updates the in-memory map. [`ui_protocol_transport.rs`](https://github.com/octos-org/octos/blob/10022387da44949cdacb07e2e2fcf38a85155455/crates/octos-cli/src/api/ui_protocol_transport.rs#L9359-L9401)

[...]

PR #2430, opened 2026-09-19
Add confined host-managed ACP workers

kevinaboos:
> Superseded by **[PR #2443](https://github.com/octos-org/octos/pull/2443)**, which ports the confined worker directly onto upstream main and preserves ordinary OUP behavior. This branch remains available for Robrix’s existing dependency pin. The original implementation details and base-specific validation are retained below.

Protected host applications need to run Octos as a separate process without giving the worker direct access to provider credentials, private files, network connections, native tools, or persistent history. This adds `octos acp --host-managed`: the existing agent loop obtains model completions and tool results through its parent over a versioned ACP extension.

## Draft status and upstream integration

This implementation is built and tested on `ce2add2677b5fc8e03fd4129e1784a071d557d7a`, the `gterzian/octos` ACP branch revision previously used by Robrix. It is already integrated into Robrix, but is **not yet adapted to current upstream main**.

Upstream subsequently moved ACP execution to the OUP dispatcher in #2265 (`d51601df`). The host-managed worker currently uses the earlier ACP session/turn helpers. Comparing with upstream `main` at `878bae8a` finds conflicts in `crates/octos-cli/src/commands/acp.rs`, `crates/octos-agent/src/agent/execution.rs`, and `crates/octos-llm/src/registry/mod.rs`. Before this is ready to merge, the host-managed entry point and its session/notification behavior need to be integrated with that architecture, preserving upstream's existing ordinary ACP behavior, then revalidated. The validation below applies to this PR's head on its original base, not to a resolved upstream merge.

The upstream diff also includes three prerequisite commits from the previously pinned ACP branch: client-provided MCP servers/long-running tool calls (`9a6f92cb`), host notifications and a hosted zero-tool profile (`c9f408f3`), and per-URL web approval (`ce2add26`). The four subsequent commits implement and validate host-managed workers. These prerequisites can be separated or replaced by upstream equivalents during integration.

## Host-managed worker changes

- Add typed, bounded model and tool broker requests, capability negotiation, one memory-only session per worker, dynamic host tool discovery, and cancellation of outstanding broker waits. Model and tool identity and authorization remain the parent's responsibility; no provider endpoint or credentials enter the worker.
- Add `octos_sandbox::host_managed_command` to apply confinement before executable entry, followed by worker initialization checks before creating threads or reading input. Use default-deny Seatbelt on macOS and bubblewrap namespaces plus seccomp and Landlock on Linux. Unsupported or unavailable isolation fails closed.
- Keep episode storage in RAM, disable configuration/plugins/native tools/remote MCP/session loading in this mode, bound the worker runtime, and validate payload sizes without a second serialized allocation.
- Document the protocol, parent security responsibilities, supported platforms, and limitations. Add native Linux/macOS CI with actual escape probes and a confined Octos model → tool → model turn.

## Validation on the current head

- Octos ACP tests: 44 passed; host protocol tests: 2 passed; in-memory episode store test: 1 passed.
- [Native confinement CI on this exact head](https://github.com/project-robius/octos/actions/runs/35434148088): Linux and macOS passed, including private file contents/metadata, inherited descriptors, network, subprocess, parent-memory and platform-specific escape probes, plus actual worker startup and model/tool execution.
- Downstream [Robrix integration](https://github.com/a2app/robrix_a2app/commit/e876d27acc9dabbe3011d5231a42ff32d3e699c7): 99 default-agent tests, 100 embedded-agent tests, 61 application tests, embedded application check, and a real confined worker model/tool/cancellation test passed.
- Live Matrix/cloud-provider UI flows were not exercised. Confinement does not claim protection against kernel compromise or timing/resource side channels. Version 1 returns complete model responses rather than streaming tokens.

Robrix currently depends on `project-robius/octos` branch `host-managed-ifc`, locked to `8e4d841dee3ffa6df31a688fdb4aa5c158497d06`. Linux requires bubblewrap 0.8+, enabled namespaces/seccomp, and Landlock ABI 3 (Linux 6.2+); the macOS launcher requires system Seatbelt support.


anantheparty:
Thank you for this — the confinement model is exactly the direction the org wants for embedded/hosted workers (it's the affirmative answer to the posture questions in #2330 and the isolation goals in #2147), and having Robrix already running it proves a real consumer. Also good to see a returning contributor (#1914) on board.

To set expectations for review: the PR is currently based on the pre-#2265 ACP branch (`ce2add26`, gterzian's `acp-embedding-seam`), while main has since moved ACP execution **into the OUP dispatcher** (#2265, 133 files) — hence the conflicts. Before a line-level review makes sense, two scoping questions:

1. **Is host-managed a distinct mode from, or a replacement for, the OUP-dispatcher path?** If they're complementary (e.g. host-managed = confined worker behind the parent, dispatcher path = direct execution), the PR should describe the boundary — which turn operations go through the parent extension vs the dispatcher.
2. **Rebase strategy**: rebasing 4.6k lines across 34 files onto the post-#2265 tree is the expensive part. If the confinement boundary (the parent-mediated channel for completions/tools, with no direct credentials/files/network/history access) is a small set of seams, extracting those seams first and rebasing the rest mechanically may be cheaper than a full re-rebase.

For the security review (which this will need, given the promise is "no direct access to provider credentials, private files, network connections, native tools, or persistent history"):
- that exclusion list maps to the standing issues #2147 (durable runtime policy), #2163 (write-only credential store), #2330 (sandbox posture — currently fail-open by default, which matters a lot for what "confined" guarantees), and #2390-adjacent store access;
- the review will want the enforcement mechanism spelled out: what enforces "no network/native tools" — capability config, seccomp/namespace confinement, or convention?

Happy to do a full adversarial pass once the branch sits on a current base — and if it helps, I can summarize the delta between `ce2add26`'s ACP seam and the current OUP dispatcher path as a starting point for the rebase.

anantheparty:
One addition to the scoping questions above, from an independent security delta pass that reviewed this branch's design yesterday: the approach — **zero-credential workers (`env_clear`) with completions/tool-results proxied through a versioned parent extension, partial-confine failures terminating rather than degrading, and CI escape probes** — matches where the industry consensus is heading ("sandbox confines the process, not the token" — [NAV's writeup](https://ki-utvikling.nav.no/en/news/sandbox-confines-the-process-not-the-token) makes the exact point that leaving credentials in the worker environment defeats the confinement; see also [GitLab's agent sandbox](https://about.gitlab.com/ja-jp/blog/ai-agent-sandbox/)).

That independent assessment flagged the real risk as **merge, not design**: the conflict resolution against #2265's OUP-dispatcher move (`acp.rs`/`execution.rs`/registry, three sites) is where the confinement invariants are most likely to get broken accidentally. When you do that rebase, it may be worth calling out in the PR description which hunks touch the confinement boundary, so the reviewer knows where to look hardest.

Also noted and appreciated: the standalone CI workflow with real escape probes on a read-only-token branch — that's the part most attempts skip.

kevinaboos:
thanks, this is still a work in progress, but I'll take a look at the points above and see if it's possible for us to incorporate them and mesh them into our octos fork's needs

kevinaboos:
Thanks for the scoping and merge-boundary review. Host-managed is a complementary mode: model completions and tool discovery/calls go through the parent extension; ordinary ACP stays on the OUP dispatcher. The parent owns credentials and authorization. The confined worker shares the canonical `Agent` loop, with private RAM-only session state, rather than bootstrapping OUP's workspace tools and disk-backed stores.

I extracted the upstream-facing implementation into **#2443**, based directly on current main. It excludes the older branch's unrelated prerequisite changes and leaves ordinary `acp/oup.rs`, agent execution, and provider registration unchanged. This keeps Robrix's already-tested consumer branch intact while making the upstream change smaller and independently reviewable.

The new PR identifies the security-sensitive startup, broker/session, and confinement paths explicitly. Network/native-tool exclusion is enforced by parent-installed Seatbelt or bubblewrap/seccomp plus worker Landlock/seccomp, alongside broker-only provider/tool construction; it does not depend on capability names or the ordinary sandbox's fallback policy. Confinement failure remains fatal. [Detailed mode and review boundaries](https://github.com/project-robius/octos/blob/host-managed-confined-worker/docs/HOST_MANAGED_ACP.md).

The replacement remains draft while its upstream-base integration and Linux/macOS confinement checks complete. Those results will be reported there before it is marked ready; the old-base checks are not being treated as proof of the new integration.

kevinaboos:
Superseded by #2443, which carries the confined-worker implementation on the current upstream architecture and addresses both review comments. Closing this older-base PR to keep review in one place; the `host-managed-ifc` branch remains intact for Robrix’s existing dependency pin.

PR #2438, opened 2026-09-20
feat(api): keep a Session a client holds open out of the idle sweep

ymote:
## Why
A user kept hitting "Session recovery required" in the web client and could only escape by reloading the tab. Measuring their live server explains the window:

| | cold (after the idle sweep) | warm |
|---|---|---|
| `session/open` (2258-message session) | **18.7 s** | 1.0 s |
| `session/hydrate` | 0.6 s (761 KB) | 0.7 s |

`octos serve` builds `SessionRuntimeCache::new(64, 1800s)`. A client that sits **reading** touches nothing for 30 minutes, so its Session ages out underneath it and the next `session/open` pays the rebuild — against the web client's 30 s request timeout. The session only grows.

## What
- `SessionRuntimeCache::keep_session_alive(&SessionKey) -> usize` refreshes `last_used` for resident entries of that Session. It never builds a runtime, so it cannot resurrect an evicted Session or pin memory for a closed one.
- Each UI Protocol connection (WS and stdio) renews every Session in its live-forwarder set from the existing 2 s tick, at `idle_ttl / 4` clamped to 5 s..5 min — derived from the cache's own TTL so changing the TTL cannot invalidate the cadence.
- `OCTOS_SESSION_CACHE_IDLE_TTL_SECS` overrides the TTL (min 1 s; malformed/zero keeps the 30 min default).

## Tests
- `should_keep_an_open_session_cached_past_its_idle_ttl`: a renewed Session survives repeated sweeps at a 100 ms TTL while an abandoned one is evicted; renewing an evicted Session reports 0.
- `should_pace_session_keepalive_inside_the_cache_idle_window`: 30 min → 5 min, 2 min → 30 s, both clamps.
- `should_wait_a_full_interval_before_the_first_session_keepalive` and `should_keep_renewing_open_sessions_on_every_interval` (two simulated hours, 24 renewals, no drift).
- Full octos-cli lib suite 3827 passed; fmt + clippy clean.

**Honest limit:** the renewal itself is proven by unit tests and by reading the two call sites; I could not observe it on a live server because this `serve` prints no `debug!`/`warn!` to stdout (verified: the existing `server/shutdown` warning does not appear either). A live check with `OCTOS_SESSION_CACHE_IDLE_TTL_SECS=20` could not distinguish warm from cold either, because a freshly created test session rebuilds in single-digit milliseconds.

Client side (a way out when recovery does fail): octos-org/octoscode-web#146.


anantheparty (approved):
Adversarial review (zero-context, head `9d904f8a`, all 4 new tests executed locally via the --no-default-features --features api bypass): **approve**.

The fix is a **sliding-window renewal, not a pin** — `keep_session_alive` only refreshes `last_used` on already-resident entries, so the 64-slot soft cap, the 1800s sweep, and `invalidate` semantics all survive. No permanent pin, no unbounded-memory path:

- "Holding" = a live forwarder on the connection (scope-validated `session/open` only); disconnect cleanup rides the existing WS/stdio paths, and after cleanup the entry returns to the sweep within ≤1860s.
- Renewal period derives from the cache's own `idle_ttl()` (idle/4, clamped [5s, 300s]), so staleness stays ≤ ~302s — well under the 1800s TTL; TTL is *not* exempted.
- Multi-client semantics are at-least-one; no corruptible reference count.
- The measured incident (18.7s rebuild against a 30s web timeout) justifies the change on its own.

Non-blocking notes:
1. **The wiring is untested** (the two transport integration points have no test — deleting them leaves the suite green; you've flagged this yourself). The renewal logic is pure and tested, but the #2438-class regression protection needs at least one test through the connection lifecycle — same shape as the octos#2412 ask on #2409.
2. Pre-existing, adjacent: when continuations are disabled, `continue` still runs before the keepalive (`:7717-7719`) — same-family quirk, unchanged here.
3. `keep_session_alive` renews across profiles (cache is global) — bounded by both caps, no leak, but worth a comment.
4. `api/mod.rs:506` hardcodes 1800 instead of reading the env knob.

CI in progress at review time; auto-merge will land it when the suite clears.

anantheparty:
An independent security delta pass over this branch (separate from my approval review) converged on the same conclusion — **LOW risk, safe to merge** — with two hardening suggestions worth recording as follow-ups:

1. **Single-connection pinning cap**: a client with many concurrent connections can pin the entire 64-slot cache via `keep_session_alive` (each connection renews its own open sessions), which converts other sessions' rebuilds into churn. A per-connection pinning cap (or a global pinned-count ceiling) would bound that.
2. **The new env knob has no upper bound**: `OCTOS_SESSION_CACHE_IDLE_TTL_SECS` is a raw u64 — a very large value effectively disables idle eviction (the capacity cap still bounds memory). Suggest an upper clamp or a doc line stating the effective maximum.

Neither blocks this merge — the renewal mechanism is sound, TTL is not exempted (staleness ≤ ~302s ≪ 1800s), and disconnect cleanup rides existing paths.

ymote:
Correction to this PR's "honest limit" note, after verifying properly.

I wrote that the renewal could not be observed on a live server because `serve` prints no `debug!`/`warn!` to stdout. **That was wrong**: the server logs to `<OCTOS_HOME>/logs/serve.<date>.log`, I was reading the wrong sink. Re-ran the check there, with the renewal temporarily raised to `info` and `OCTOS_SESSION_CACHE_IDLE_TTL_SECS=20`:

- **Works in a realistic setup** (profile runtime + a real 2 465-message session held open on one connection): 6 renewal rounds, each reporting `sessions=1`, with the requested key matching the cached key exactly (`dev|dev:api:web-…`).
- **No-op without a profile runtime**: in a bare test home the same probe reported `sessions=0 open=1` on every round — `session/open` there never populates `SessionRuntimeCache` at all (`cached=[]`). So the renewal is correctly scoped, but its blast radius is narrower than this PR's description implies: it only matters where the session-runtime cache is actually used.

Also worth recording against the motivation in the description: the 18.7 s cold `session/open` I cited was measured on a **debug** binary. On release the same open is ~0.9 s, and the dominant remaining cost was the context-ledger verification, fixed separately in #2442 (0.9 s → 0.09 s). The idle-eviction window this PR closes is real, but it was not the main cause of the stall that motivated it.

Issue #2412, opened 2026-09-19
test: pin the turn-admission wiring end-to-end via turn_admission_test_pause (follow-up to #2409 review) + ledger-Err should keep running absent


anantheparty:
Follow-up from the #2409 review (F2, Medium, partially addressed in the merged fix).

The turn-admission wiring in `turn/start` now has a test hook — `turn_admission_test_pause(&params.turn_id)` (`ui_protocol_transport.rs`, cfg(test), holding admission open right after the folded-key marker entry, with a per-turn registry at `:23172-23183`) — but **no test drives it**: both tests added in #2409 construct `TurnAdmission::enter` synthetically, so deleting the marker wiring line leaves the whole suite green.

**Ask**: an end-to-end test through the real handler —
1. Enable the pause for a `turn_id`, send a real `turn/start` over the WS test harness (it parks in admission);
2. Send `turn/state/get` from a second connection; assert `running` is **absent** (not `false`);
3. Release the pause; assert the turn proceeds and a subsequent `state/get` reports the known state.
A concurrent variant (two `state/get` connections racing the insert) would close the F1-class blind spot permanently.

Related low-severity leftover from the same review: a ledger `snapshot_with_cursor` failure currently degrades to `projection = None` and still asserts `running:false` — the "certain" standard (UPCR-2026-031 §Contract) would prefer keeping `running` absent on read errors.

PR #2442, opened 2026-09-20
perf(api): stop replaying the whole history to verify a context ledger

ymote:
## Why
`session/open` verifies that the cached context ledger still matches durable history by **rebuilding the entire history into a throwaway ledger** and comparing (`context_ledger_covers_history`). It then rewrites the 11 MB snapshot even when nothing changed. Measured on a copy of a real **2 465-message** session (release build):

| | before |
|---|---|
| `session/open` | **0.79–0.97 s**, every time |
| — of which the replay | 0.65 s |
| — of which the snapshot rewrite | 0.065 s |
| `session/hydrate` | 0.016 s |

It grows with the session: 0.13 s at 721 messages, 0.47 s at 1 321, 0.9 s at 2 465. The answer is almost always "yes, it matches".

(Context: a user hit repeated "Session recovery required" in the web client. The dominant cause there was a debug build — release is ~20× faster on this path — but this per-open cost is real and growing.)

## What
**1. Fingerprint instead of replay.** A ledger records which messages it consumed: `SourceHistoryChain { messages, digest }`, persisted in the snapshot (`serde default`, so old snapshots simply have none).

- Maintained by the two durable consumers only: `record_persisted_message` and `record_persisted_message_merging_prompt_equivalent`. The second is the one production actually uses — it builds a probe manager and splices rows, bypassing `record_message_with_source_ref`; an earlier attempt hooked only the message recorder and therefore did nothing.
- **Dropped** by every other way a source row can arrive: hand-built items (`record_item_with_source_ref`), the ordered splice, twin stamping, `adopt_source_items_after`, forked children. Dropping costs the replay; trusting a stale chain would load a ledger that does not match history.
- **Kept** across changes that touch the projection but not the consumed set: compaction, `mark_source_event_kind`, synthetic aborts, checkpoints, semantic regrouping.
- The digest hashes exactly what the recorder reads (role, content, media, tool calls, `tool_call_id`, reasoning, `thread_id`) and excludes what it doesn't (`timestamp`, `client_message_id`, `ToolCall::metadata`) — hashing those would rebuild for nothing. It counts **messages**, so a message recording no rows (System, empty assistant, tool without a call id) can't skew it.
- A ledger that passes the old check **adopts** the chain at that moment, so sessions written before this change become fast on their second open instead of never.
- `OCTOS_CONTEXT_FINGERPRINT=0` forces the replay.

**2. Don't rewrite an unchanged snapshot** on open (loaded clean + no compaction + nothing adopted).

## Result

| | before | after |
|---|---|---|
| `session/open`, steady state | 0.79–0.97 s | **0.09–0.11 s** (~9×) |
| first open of a pre-existing snapshot | 0.79–0.97 s | ~1.0 s (one replay, then fast) |

## Safety
The property that matters is that the fast path is **never more permissive** than the replay. `should_never_claim_coverage_the_replay_would_deny` runs ten history shapes (zero-row, multi-row reasoning + tool calls, answered and unanswered batches, threaded) through **both** recorders and, for each, the unchanged / appended / truncated / edited variants, asserting:
- the fast path never claims coverage the replay denies, and
- an unchanged history always takes the fast path (or the optimisation is dead).

It is deliberately allowed to be **stricter**: a history whose trailing message records no rows (a System message) is indistinguishable from a shorter one to the replay, and the fingerprint tells them apart. That only costs a rebuild. This asymmetry is what the test encodes, and it found a real disagreement during development.

Plus: chain-drop tests for each bypass path, projection-only mutations keeping it, snapshot round-trip, ignored-field equality, and the kill switch.

Full octos-cli lib suite **3834 passed**; fmt + clippy clean. Benchmark harness and measurements are reproducible against a copy of the session (never the live one).


anantheparty (changes requested):
Adversarial review (zero-context, head `21e4be8`, injection-tested with an independent probe): **request changes — two Medium fixes, both cheap, then this is an easy approve.**

The semantic-equivalence work is real (digest read-surface aligned with the recorder field-by-field; load self-check intact; chain maintenance discipline verified across both durable consumers; migration/degradation paths probe-tested). The performance claim holds for today's code. What's missing is closing the two verification-strength regressions the fast path introduces:

**1 [Medium] — coherent bad writes are now silently accepted.** The old replay compared against the *actual source rows* (session_id/thread_id/seq/kind payload + watermark, `source_head_hash_for_items`); the new digest trusts the `source_history` field the writer maintained. Injection test: delete the seq=2 source row, recompute `source_head_hash`, leave the chain intact → `load_or_rebuild` returns **Loaded** with the truncated items, where the old code returned Stale and rebuilt. Worst case: a restored session silently missing tail rows/ledger items — no log, and (unlike the old path) never self-heals on the next open. Trigger is a future writer bug or tampering, not today's code — but the fix is an O(1) cross-check: assert row-count/max(source_seq)+1 equals `chain.messages` at load. That keeps the fast path and closes the class.

**2 [Medium] — the digest folds `thread_id`, a field the recorder never reads** (`extend_source_history_digest:970-992` vs recorder reading `source_ref.thread_id` from the manager at `:2077`). Safe today only because `synthesize_thread_ids` is deterministic (`synth_{seq}`); the moment a non-deterministic gap-fill or migration rewrite appears (a dead-code path with `uuid::now_v7` already exists), every `session/open` sees a changed digest → permanent replay+11MB rebase per open — the exact cost this PR exists to kill, with no error. Suggest excluding `thread_id` from the digest (like `timestamp`/`client_message_id`) or documenting the invariant in the digest.

**Non-blocking notes:**
- The perf property isn't fix-sensitive: deleting the entire fast path leaves the new tests green (the "unchanged must take fast path" assertion holds via the replay fallback too). Your legacy-migration probe (strip chain → Loaded → adopt → chain persisted) is worth upstreaming — it pins the adopt path.
- `context_manager.rs:908` references `should_answer_coverage_exactly_as_the_replay_does`; the actual test is `should_never_claim_coverage_the_replay_would_deny` (:4574).
- The :4574 test reads an env var (`OCTOS_CONTEXT_FINGERPRINT=0` would false-red it) — calling `context_ledger_covers_history_with(…, true)` directly is sturdier.
- `unwrap_or_default()` on digest serialization (≈:984) — silent failure degrades to an empty body where different messages collide into false-covered; `expect`/debug_assert fits the file's style.
- Loaded fast path computes two sha256 passes over full history (check + adopt) — still ≪ the 0.65s replay; pass-through of the digest would shave it.
- The skip condition embeds a cross-module contract ("every open-path change emits a lifecycle notification") — true today (all three compaction branches notify) but untested; a pin would future-proof it.

Scope/CI: strictly 2 files, +459/−5 matches the claim, no smuggles; everything except test-octos-cli green at review time.

ymote:
Both Mediums addressed in `22587cbd` (rebased on main).

**1 — coherent bad writes.** Fixed as suggested, with the check on the ledger rather than on the seq arithmetic (a trailing zero-row message, e.g. System, makes `max(seq)+1 == chain.messages` false for a perfectly good ledger). The chain now carries a witness of the ledger it was written against — count of source-stamped rows plus their high watermark — and the fast path is only taken when the ledger still matches it. Your injection is now a test: delete the seq=1 row, recompute `source_head_hash`, keep the chain → falls back to the replay and rebuilds (`should_replay_when_the_ledger_lost_rows_behind_the_chain_s_back`).

**2 — `thread_id`.** Kept in the digest, and I think excluding it would be wrong: it reaches the ledger through every row's `source_ref.thread_id`, so the replay *does* distinguish two histories that differ only there. Dropping it would make the fast path **more permissive** than the replay — the one property this design must never break — and the differential test would fail (it now covers that case: `should_notice_a_history_whose_thread_ids_changed`). Your underlying concern is real but it is a cost, not a correctness risk: a non-deterministic thread-id fill would make every open miss. That is now `debug!`-logged on each miss instead of being a silent slowdown, so the condition is observable if it ever appears.

**Non-blocking notes, all taken:**
- The perf property is fix-sensitive now: a thread-local replay counter asserts an unchanged history is answered **by the chain**, not by a replay that agrees. Deleting the fast path fails the suite.
- Legacy-migration probe upstreamed (`should_adopt_a_chain_for_a_ledger_persisted_before_chains_existed`): chainless snapshot → Loaded → chain adopted → persisted → second load answers from the chain.
- Stale test name in the `:908` doc comment fixed.
- Tests call `context_ledger_covers_history_with(…, true)`, so `OCTOS_CONTEXT_FINGERPRINT=0` can no longer false-red them.
- Digest serialization `expect`s instead of `unwrap_or_default()`.
- Two sha256 passes on the adopt path: left as is (it happens once per ledger, then never again); say the word if you want it threaded through.
- The skip-persist contract ("every open-path change emits a lifecycle notification") is still untested — noted as a follow-up rather than done here.

Full octos-cli lib suite **3847 passed**, clippy clean. Re-benchmarked against the same session copy, release, identical data: **origin/main 0.99–1.03 s** steady-state `session/open` vs **0.12–0.13 s** here.

anantheparty (approved):
Re-review at head `9d0afa79` — **approve, with two items moved to tracked follow-ups.**

What landed since my review (the +620 in context_manager.rs is substantial hardening):

- **Rebase-over-appended-history with source_high_watermark cross-checks** (`rebase_context_manager_over_appended_history`): appended-only snapshots rebase without replay; any edit to the covered prefix rejects the rebase and falls back to a full rebuild. This is a real addition to the verification story.
- The compacted/adopt path now filters chains by `chain.messages == source_seq` (:1845) — a partial density check on one path.

Still open from my review (both fine as tracked follow-ups, not merge blockers at this point):

1. **F2 — `thread_id` is still folded into `extend_source_history_digest`** (:1018-1050 at HEAD). Today it's safe (deterministic `synth_{seq}` ids), but any future non-deterministic gap-fill or migration rewrite flips every `session/open` into permanent full replay + 11MB rewrite. Excluding it (like `timestamp`/`client_message_id`) is a two-line change plus a comment.
2. **F1's load-side density check** — the coherent-truncation-at-load hole (snapshot rows tampered with a recomputed hash) is now only covered by the digest-vs-durable comparison, which passes if the *writer* was coherently wrong. The `chain.messages == source_seq` filter at :1845 covers one path; the load path itself still trusts the declaration.

Both are now tracked in #2412 (extending that issue's scope to cover them). The performance claim stands (the rebase path's watermark check is a genuine addition), CI on the head was green at review time, and the rebase across today's merges is clean. Approving; the armed auto-merge lands it when the tail clears.

Issue #2418, opened 2026-09-19
perf(ci): root-cause check-windows — mirror line-tables-only on Windows, parallelize build-windows, split octos-cli Windows lane (122→~50min wall)


anantheparty:
**From the second performance audit** (Actions API run data, `ci.yml@7f3cb5ce`). #2404 stopped the kill; this issue is the root cause.

Measured on main (run 35421711063): check-windows **84.9min** (recent: 73.2 → 93.8-cancelled-at-90 → 84.9), build-windows **37min**, and `build-windows needs: check-windows` (`ci.yml:1041`) — so a main push pays **~122min of serial Windows wall-clock** for what is mostly "it compiles".

Root causes, each with code evidence:
1. **check-windows lacks the compile mitigation Linux got**: the three heavy Linux jobs set `CARGO_PROFILE_{TEST,DEV}_DEBUG=line-tables-only` (`ci.yml:245-247,380-382,425-428`); the Windows job (`ci.yml:768-786`) has **no env block at all** → full DWARF debuginfo linked into 27+ test binaries. Windows integration alone is 26.4min.
2. **octos-cli's 40min (lib 13.3 + integration 26.4) runs serially in one job** — the workflow's own comment (`ci.yml:773-774`) says splitting it into its own Windows job would cut wall-clock; not done.
3. **build-windows only needs compilation success**, yet it waits behind the entire test job and uses a separate cache key (`windows-release`, `ci.yml:1050`) — cold cache every time.
4. No sccache anywhere; cache miss = full rebuild.

**Estimated wins**: ① mirror line-tables-only on Windows (−15~25min across build+link); ② split octos-cli Windows lane into its own job (85→~50min); ③ build-windows in parallel with tests (−35min wall); ④ pilot `rust-lld` linker on Windows MSVC (openai/codex does this: [commit](https://github.com/openai/codex/commit/24aee3eb5ee90dd211ef183d65fec774dc7049c9)). Combined: **~122min → ~50min** for the main-push Windows path.

Related: #2404 (the cap raise — symptom), #2323 (cost gates), #2334 (verification convention).

PR #2443, opened 2026-09-20
Add confined ACP workers alongside the OUP runtime

kevinaboos:
Host applications need an Octos worker that cannot bypass their authorization by using its own provider credentials, files, network connections, or native tools. This adds `octos acp --host-managed`: a confined process whose model completions and tools are supplied by its parent through a versioned ACP extension.

This replaces #2430 on upstream `main` at `baf4e519`. It excludes the older ACP branch's unrelated MCP, hosted-profile, native-network-approval, and scenario-provider prerequisites. Robrix's existing consumer branch and dependency pin remain intact while this upstream implementation is validated.

## Execution boundary

Host-managed mode complements ordinary ACP/OUP execution. Ordinary ACP still uses the unchanged OUP adapter and configured runtime. The confined adapter shares the current `octos_agent::Agent` loop, while owning only private, memory-only ACP session bookkeeping. It does not bootstrap the ordinary OUP runtime, which currently constructs workspace tools and disk-backed stores.

- Every model call, including compaction and retries, uses `_octos/host/model`. Tool discovery and invocation use `_octos/host/tools/list` and `_octos/host/tools/call`. There is no direct-provider or local-dispatcher fallback.
- The parent retains credentials, compartment identity, information-flow policy, recipient consent, and current tool authorization. No worker-provided identity or taint labels are trusted.
- Each worker accepts one negotiated session, refuses persistent history loading and external MCP servers, and uses a functional RAM-only episode store. Notifications, bounded fold/drop queues, and cancellation generations remain supported.
- Host-managed turns retain the existing finite 20-iteration default; positive overrides remain available. Ordinary ACP retains its upstream unlimited-zero convention. Truncated upstream model responses are reported as errors rather than successful completion.

## Enforcement and review focus

These restrictions have both an OS boundary and a broker-only construction path; they do not depend on prompt instructions or ordinary tool-sandbox policy.

1. **Startup:** `main.rs` selects host-managed execution before configuration or logging workers; `AcpCommand::execute` confines before Tokio or ACP input. The parent launcher clears environment/cwd/unrelated descriptors and applies restrictions before executable entry. Missing or partial confinement terminates startup.
2. **Kernel isolation:** `octos-sandbox` uses default-deny Seatbelt on macOS and bubblewrap namespaces plus seccomp/Landlock on Linux. The worker cannot use private files, direct network, native subprocess tools, or persistent writes. Exact runtime-file exceptions and unsupported platforms are documented; the protocol's `confined` flag is not attestation.
3. **Agent/session wiring:** `acp/host_managed.rs` directly constructs the host provider, broker tools, and RAM store. Review notification/cancellation and model/tool feedback here. Ordinary `acp/oup.rs`, agent execution, and provider registry are unchanged.

[Protocol and integration boundaries](https://github.com/project-robius/octos/blob/host-managed-confined-worker/docs/HOST_MANAGED_ACP.md) · [Platform restrictions and escape probes](https://github.com/project-robius/octos/blob/host-managed-confined-worker/crates/octos-sandbox/HOST_MANAGED.md)

## Validation

Local validation on the upstream port passed:

- 18 minimal-worker ACP tests; 24 ACP tests with the ordinary OUP runtime enabled.
- All 7 ordinary ACP integration tests, including persistence, reload, history sanitization, and protocol output.
- 2 host-protocol tests and the RAM-store isolation test.
- Strict Clippy across all targets in the CLI, sandbox, model, and memory crates with the API feature enabled.
- Native macOS escape probes and actual confined model/tool execution, using both API-enabled and minimal worker builds.
- Robrix's real parent-to-worker model/tool/cancellation test against the new worker, while retaining Robrix's existing dependency pin.

The [Linux/macOS confinement workflow](https://github.com/octos-org/octos/actions/runs/35501745898) passed on final head `5c9e0b53`, including actual kernel escape probes, ordinary ACP integration, and confined model/tool execution. The [broader upstream CI](https://github.com/octos-org/octos/actions/runs/35501745764) has passed formatting, workspace Clippy, agent tests, macOS checks, and the general Linux sandbox tests; its remaining repository-wide checks are still running.

Live cloud-provider and Matrix UI flows are outside these fixture tests. OS confinement does not promise protection against kernel compromise or timing/resource side channels. Protocol v1 returns complete model responses rather than incremental token streaming.


kevinaboos:
Marking this ready for review. The implementation is now based directly on upstream main, with ordinary ACP/OUP behavior preserved and the parent authority and OS confinement boundaries documented. Both Linux and macOS native confinement workflows pass, including actual model/tool execution, and Robrix’s real parent-to-worker model/tool/cancellation test passes against this worker. Missing isolation still fails closed. The two scoping/security comments on #2430 are addressed here; this replaces that older-base draft.

anantheparty (commented):
Adversarial review (zero-context; the reviewer **actually ran the Linux confinement** — bwrap 0.9.0 + userns, real escape probes ×3 modes all passing, a real `octos acp --host-managed` binary completing a full confined turn (initialize→session/new→prompt→tools/call→end_turn) with `confined:true`, 18 acp + 2 host-protocol + 1 RAM-store isolation tests green locally). Verdict: **the Linux confinement boundary is real and the design is sound — two items block ready-for-review.**

**Verified working (Linux, locally executed)**
- Confinement is **OS-enforced, not config-declared**, and every failure path is fail-closed with no fallback code path anywhere: `env_clear` three layers (launcher/seccomp/bwrap `--clearenv`), zero credential/PATH/HOME whitelist, seccomp allowlist with default-ENOSYS, Landlock V3 HardRequirement, clone restricted to thread flags, `openat` entry-phase-only with all write flags denied, memfd+BPF+F_SEAL, bwrap `--unshare-all --die-with-parent --cap-drop ALL --remount-ro /` with root-owned/immutable-dependency validation including ancestors.
- The five exclusions each have a real enforcement mechanism: credentials/private files = OS+code (early-exit skips config load, clap conflicts), network = OS (no socket/connect in the allowlist, net unshared), native tools = OS (thread-only clone flags, execve entry-phase-only, Seatbelt fork deny), persistent history = in-memory store + `load_session` always refuses, external MCP = protocol (non-empty `mcpServers` rejected, tested).
- Escape surface: the worker→parent channel is pure request-model with `deny_unknown_fields` (no identity injection), version negotiation is exact-equality with no downgrade path, no notification-frame trigger, no batch injection, no mid-connection scope escalation. The complementary boundary vs the OUP dispatcher is clean (early-exit before config_layer/tracing; `ensure!(!host_managed)` on the dispatcher path; CI validates the no-default-features build).
- Scope: 17/17 files traceable to the declared feature, zero smuggles.

**Blocking for ready-for-review**

1. **[P1] macOS `sandbox_check` polarity is unverifiable, and the macOS CI leg was cancelled — zero verification record.** `macos.rs:106` uses the documented NULL-op form (`>0 = sandboxed`, confirmed by Chromium's seatbelt usage), but `:108-115` requires `sandbox_check(pid, "network-outbound", 0) > 0` etc. while the error messages say "inherited sandbox *allows*…"/"*permits*…" — i.e. the code assumes **>0 = denied**. Apple does not document per-operation return semantics (checked SDK headers, Chromium, sb_validator). If the standard "1 = allowed" semantics hold, the polarity is inverted: a correct profile fails every check → `confine()` errs → every macOS worker launched via the parent dies at startup (fail-closed, not an escape — but the macOS platform is dead and the validation proves nothing). The author reports local macOS probes passing (implying reading (b)), but this head's `confinement (macos-14)` run was **cancelled** — there is no verifiable pass record. Before ready: get the macOS leg actually green on this head, cite the semantics source in a comment, or add a positive-control assertion (probe a profile-*allowed* path to prove the two polarities are distinguishable).

2. **[P2] cargo fmt failure + a predicted new red in the main CI.** `check` fails at "Check formatting" (reproduced locally on octos-sandbox — cfg-attribute ordering); the "CI required" cascade followed. And the main `ci.yml` `Test octos-sandbox` step will newly execute the escape probes on GitHub runners, where bwrap is absent and `apparmor_restrict_unprivileged_userns=1` — reproduced locally (`bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted`), exactly the condition the PR's own workflow sysctl-fixes (`host-managed-sandbox.yml:32-42`) but the existing ci.yml jobs don't. Fixes: run `cargo fmt`, and add the same bubblewrap/sysctl prep to the existing ci.yml ubuntu jobs (or decouple the probe tests from the plain test step). Also note check-arm/check-riscv run this test and were skipped on this head — seccompiler arch support and bwrap availability there are unverified.

**Non-blocking notes**
- [L] The worker executable itself lacks the root-owned/immutable validation the ELF dependencies get (`linux.rs:203-212`) — a validate→spawn TOCTOU window if the host places it somewhere user-writable. Fine under the stated "host picks a trusted binary" boundary; document the placement requirement or reuse the immutability check.
- [L] `(import "dyld-support.sb")` depends on Apple-controlled system profile content that can silently weaken the worker profile in future macOS versions — inlining the needed authorizations (they're already enumerated) would remove the dependency.
- [L] No RLIMIT_AS/rss on the worker (the old Windows AppContainer path had a 512MB job limit) — defense-in-depth for trusted code.
- [nit] `--unshare-user` after `--unshare-all` is redundant; the Landlock grants on /dev/null|urandom are unreachable post-seccomp (harmless depth); the body's docs link points at project-robius/octos rather than this repo.
- Scope note: the `session/notify` extension (Robrix auto-respond) is in this diff and declared — reviewers should know it's a beyond-confinement feature riding along.

Fix the fmt, close the macOS verification gap, and prep the shared CI lanes — then this is an easy approve. The Linux side is genuinely the best-confinement PR this repo has seen.

kevinaboos:
Thanks for the independent execution and boundary review. Both blocking items now have passing evidence on current head `5c9e0b53ec725d0b3a247c6335e714c1af65e66c`:

1. **macOS:** the [macOS confinement job](https://github.com/octos-org/octos/actions/runs/35501745898/job/106054606275) completed successfully, including parent/worker/standalone native escape probes and a real confined Octos model/tool turn. This exercises successful inherited-sandbox initialization together with actual denied file/network/process operations and working pipe/thread I/O. It is current-head native execution evidence, rather than relying on the protocol flag or a cancelled earlier run. The [Linux job](https://github.com/octos-org/octos/actions/runs/35501745898/job/106054606395) passed too.
2. **Formatting and generic CI:** the [general check job](https://github.com/octos-org/octos/actions/runs/35501745764/job/106054657227) is green, including formatting, workspace Clippy, and `Test octos-sandbox`. `5c9e0b53` shares the existing Linux prerequisite setup across the dedicated workflow, general job, manual ARM/RISC-V lanes, and nightly ARM workspace tests. Persistent self-hosted runners retain administrator-controlled policy; their required prerequisites are documented. Source review confirmed the locked seccompiler supports ARM64 and RISC-V, but those manual/native lanes are not being represented as tested by this PR run.

The remaining low-priority points are separate hardening considerations, not additional confinement guarantees in this change. The system-runtime profile dependency and resource/side-channel limits are documented. The parent still chooses the worker executable, and `session/notify` remains explicitly declared as the host-driven notification support required by the consumer.

anantheparty (approved):
Re-review at head `5c9e0b53` (ready-for-review conversion) — **approve**.

Blocking items from my comment review, resolved:
1. **P1 (macOS polarity)** resolved empirically: `confinement (macos-14)` ran to completion and **passed** (11m47s) on this head — the `sandbox_check` loops execute on real macOS and the profile+check combination works end-to-end. The remaining ask from my review stands as documentation only: cite the semantics source in a comment near `macos.rs:106-115` (or keep the positive-control idea) so the next reader doesn't have to re-derive it.
2. **P2 (fmt)** resolved — `check` passes (17m29s).
3. **P2 (shared CI lanes)** — `check` is green on this head including the octos-sandbox tests; whatever decoupling/gating makes that work, the empirical result is what matters. The residual question (what happens to `Test octos-sandbox` in the existing ci.yml jobs at merge time, and arm/riscv seccompiler support) is now a merge-time observation rather than a review blocker — it lands in #2418's CI-economics scope if it bites.

Non-blocking notes from the earlier review stand (worker-executable immutability check, dyld-support.sb dependency, memory rlimit, docs link). The Linux confinement remains locally-probe-verified, scope discipline holds, and this is the strongest confinement PR the repo has seen. Approving; auto-merge will land it when the tail suites clear.

Issue #740, opened 2026-05-01
P1: M8.10 fast-burst sticky-map drift in foreground SSE (rapid-fire scenarios mis-route under wave-4)


ymote:
## Summary

Wave-4 soak (today, against `0.1.1+b78703bb` which already merged #739) deterministically reproduces an M8.10 thread-binding regression on `live-overflow-stress.spec.ts` scenarios that contain ONLY fast prompts (no `spawn_only`/`run_pipeline`/`deep_research` involvement). Late-arriving assistant tokens cluster under the LAST user thread rather than each turn's originating user.

This is the SAME symptom as the #649/#664/#673/#680 regression chain, but on a different code path: PR #739 (the most recent fix) only covers `spawn_only` background delivery via `originating_client_message_id`. The fast-burst foreground SSE turn does not flow through that fix path, so it remains broken.

## Hard evidence — `rapid-fire-five-fast` scenario, mini3 wave-4

Five 1+1=…/5+5=… arithmetic prompts sent 800ms apart. Live URL: `https://dspfac.octos.ominix.io`. Build: `0.1.1+b78703bb`. DOM dump captured at test failure (full log: `/tmp/wave4-soak-mini3.log:424-495`):

```text
User 1+1 → assistant ".\n.\n.\nThinking (iteration 1)\n2026-05-01 11:28:50"   ← placeholder, never finalised
User 2+2 → assistant ".\n.\n.\nThinking (iteration 1)\n2026-05-01 11:28:51"   ← placeholder
User 3+3 → assistant "1 + 1 = 2\n\n2026-05-01 11:28:52"                        ← Q1's answer, mis-bound to Q3's thread
            assistant ".\n.\n.\nThinking (iteration 1)\n2026-05-01 11:28:51"   ← stray placeholder
User 4+4 → assistant ".\n.\n.\nThinking (iteration 1)\n2026-05-01 11:28:52"   ← placeholder
User 5+5 → assistant "3 + 3 = 6\n\n2026-05-01 11:28:57"                        ← Q3's answer, mis-bound to Q5
            assistant "2 + 2 = 4\n\n2026-05-01 11:28:58"                        ← Q2's answer, mis-bound to Q5
            assistant "5 + 5 = 10\n\n2026-05-01 11:29:00"                       ← Q5's answer (correct binding)
            assistant "2026-05-01 11:28:53"                                     ← stray timestamp-only bubble
```

Observations:
[...]

PR #2452, opened 2026-09-20
view_image shows the model the image it asked for

ymote:
## What

A tool result can carry images for the model: `ToolResult::model_media: Vec<PathBuf>`. The dispatcher puts them on the tool message's `media`; `execute_tools` appends **one user-role message** carrying them after the whole batch, so every provider (OpenAI-compatible, Anthropic, Gemini, Responses) renders them the way it already renders a user upload — no provider code changes.

`view_image` is the first caller. It returned `{format, mime_type, byte_length}` and nothing else (the bytes were for the host UI's artifact channel), so a vision-capable model could learn that its own screenshot was a 2.1 MB PNG and nothing more. Now a raster file the vision APIs accept (png/jpeg/gif/webp, ≤ 5 MB, recognised extension) comes back as `model_media` with `shown_to_model: true` in the output; anything else stays metadata with `not_shown_because`, so the model always knows whether it is looking at the image or at a description of it.

## Why

In the OctoSense image-to-app evaluation, DeepSeek, GLM and Kimi — all vision-capable — found every visual defect in their own renders only when a person attached the picture, because `view_image` could not show it to them. Codex's `view_image` puts the image in the model's context; this brings octos to parity.

## Design notes

- The follow-up goes **after all of the batch's tool results**, never between them: Anthropic requires the `tool_result` blocks to sit together immediately after the assistant's `tool_use`, and `build_anthropic_messages` would otherwise fall back to plain text for the split-off result.
- It names the tool call it answers (`[image returned by tool call <id>: <path>]`), so the model can tell a returned screenshot from a person's attachment.
- The tool message keeps the path in `media` for the UI; providers deliberately inline images only on user rows (see the comment in `openai.rs::build_openai_content`), so that is text-only on the wire.
- Persisted and re-sent like any user upload; compaction handles it as it handles uploads.
- `files_to_send` is untouched: it goes to the human over the channel, never to the model.

## Tests

- `should_follow_the_batch_with_one_user_image_message_when_a_tool_returns_model_media` (ordering with a second, plain tool in the batch), `should_add_no_follow_up_when_no_tool_returns_model_media`
- `view_image_hands_a_raster_image_to_the_model`, `view_image_keeps_an_svg_to_metadata_and_says_why`
- `cargo test -p octos-agent --lib`: 2920 passed; `-p octos-pipeline`: 359 passed; dependents type-check; fmt and clippy clean.


anantheparty (changes requested):
Adversarial review (zero-context, code read at head `f901b84`, provider wire shapes cross-checked against each API's documented constraints): **request changes**. The tool-side work (`model_media`, view_image degradation semantics, in-batch ordering) is good; the problem is concentrated in one transport-layer decision — **carrying the follow-up as an independent user-role message** — which breaks two provider families and has persistence/budget side effects.

**F0 · Blocker (Anthropic) — consecutive user messages, request fails with 400.** `messages.extend(model_media_followups)` (`execution.rs:2729`) appends a User line after the tool_result user line; `build_anthropic_messages`' `_` arm pushes it as a new AnthropicMessage without merging (`anthropic.rs:937-948`). Wire: `assistant(tool_use) → user(tool_result) → user(images)` — Anthropic requires alternating roles and 400s ("roles must alternate"; the repo's own test comment at `anthropic.rs:1463-1467` says "separate user messages 400"). Classified as failover-worthy (`retry.rs:97-151`) → hard turn failure on single-Anthropic configs, unexplained provider flapping on multi. And the bad shape **permanently enters the transcript** (`repair_message_order` only moves Tool lines) until compaction. The Anthropic-correct fix is image blocks inside the `tool_result` content array ([vercel/ai#2047](https://github.com/vercel/ai/issues/2047)) — which requires provider code, in tension with the PR's headline.

**F1 · High (Gemini) — adjacent same-role contents with incompatible part types.** The tool line merges into `push_or_merge("user", [FunctionResponse])` (`gemini.rs:681-696`) while the follow-up is `push_or_merge("user", text+inlineData)` (`:629-632`); `parts_compatible` (`:703-740`) explicitly refuses to merge FunctionResponse with text/inlineData → two adjacent `"user"` contents, which Gemini rejects (the repo's own comment at `:587` says so), with no 400-retry on the Gemini path.

**F2 · High (persistence) — the follow-up is an orphan User line that roots a new thread.** The follow-up carries `thread_id: None` (`execution.rs:3040-3060`); the turn-end persistence loop pre-marks only `Assistant | Tool` (`session_actor.rs:8346-8360`), so it lands in `derive_thread_id_for_new_write`'s User arm → **a synthetic UUIDv7 thread** per view_image turn — re-introducing the M8.10 thread-binding bug class (#649→#740) for this synthetic line, while the assistant line stays pinned to the real cmid.

**F3 · Medium (budget) — the token estimator is blind to images.** `estimate_message_tokens` (`context.rs:208-218`) counts text+tool_calls only; a 5MB screenshot (~6.8MB base64, thousands of image tokens) counts as ~20 tokens of caption text. Route-fit and compaction triggers systematically under-count, and there's no per-turn aggregate cap on `model_media` (N parallel view_image = N×5MB into one request).

**F4 · Medium (TOCTOU at render)** — the tool-time defenses (O_NOFOLLOW, ancestor symlink walk, magic/size checks, `coding_tools.rs:2857-2931`) are all bypassed at render time: `vision.rs:10-28` `encode_image` is a bare `std::fs::read`. A parallel sibling tool or background writer can swap the validated PNG for a symlink (any readable file leaves the process labeled `image/png`) between validation and the next request. Deletion race is benign-ish (all four builders silently drop the image while still telling the model "[image returned by tool call…]"). Suggest: read the bytes at dispatch (not PathBuf), or re-apply no-follow+size at encode; mark render failures as not-shown.

**F5 · Medium (privacy)** — screenshots (terminal contents: keys, files) go to the provider with no approval gate (the human sees only a path in the tool event; the #2410 text-redaction layer is blind to pixels), and because the follow-up is a **persistent user row**, every request until compaction re-reads, re-base64s, re-sends, and re-bills the image (the "~1 MB per slide per call" waste pattern, now as a user line). Suggest: strip media from the row after first successful render (send-once), and give `model_media` visibility/approval semantics analogous to `files_to_send`.

**F6 · Low** — `VISION_FORMATS` includes gif (`coding_tools.rs:2982`) but Gemini inline_data doesn't accept gif → 400 on Gemini after the gate passes. On lacks_vision models the follow-up degrades to a `[user-uploaded files: …]` note that mislabels a tool screenshot as a user upload.

Everything else checked clean: in-batch follow-up ordering, no double-render/billing of the Tool-row media, media serde persistence, episode 500-char summaries keep bytes out of episodes, compaction replaces the user line with `[media omitted]`, memory_refresh extraction path is pixel-safe, SVG/BMP dual-track handling, scope discipline (5 files, all traceable).

**Suggested direction**: promote "zero provider changes" from headline to to-do — Anthropic wants the images in `tool_result` content, Gemini wants same-content merging or the Files API, and the follow-up line needs a non-User identity (or pre-marked thread_id). F0/F1/F2 then collapse together. Happy to re-review the provider-side revision.

anantheparty:
The new video commit (`view_video`, `video_url` parts, refuse-and-retry) **extends the same broken follow-up pattern** rather than resolving the review: `messages.extend(model_media_followups)` is unchanged (`execution.rs:2729` at `f1a07704`), so video parts now ride the independent user-role message too — meaning on Anthropic the consecutive-user 400 (F0) now fires with videos in the payload, and on Gemini the adjacent same-role contents (F1) gain inlineData/video parts on top.

The REQUEST_CHANGES therefore stands, and the video feature makes the provider-side fix *more* urgent, not less: once `tool_result` carries the media (Anthropic) / compatible merging or Files API (Gemini) lands, video and image parts get the same treatment for free.

One additional video-specific note for that fix: `refuse-and-retry` semantics — if a provider rejects a video-bearing message, make sure the retry path strips the media rather than retrying the identical failing payload (the OpenAI path's text-only retry pattern is the in-repo precedent).

Happy to re-review as soon as the provider-side revision lands — the review's detection matrix carries over as-is.

ymote:
Thanks — F0/F1/F2 were real, and they all came from the one transport decision, so the new commit replaces it rather than patching around it.

**What changed**
- No follow-up user row at all. The transcript carries only the media on the tool row (`ToolResult::model_media` → `Message.media`). Nothing synthetic is persisted, so F2 (orphan user line rooting a thread) is gone by construction.
- A new `octos_llm::tool_media` decides for every builder what renders: **only the current tool batch** (the rows after the last assistant tool-call row); an older row's media becomes a note naming it, so nothing is re-encoded per request (F5's re-send cost). Media the model can't take (`lacks_vision` / `lacks_video`) becomes a note that says so — never a silent drop or a "user-uploaded files" mislabel (F6 second half).
- Provider shapes, per protocol: **Anthropic** — image blocks inside the `tool_result` content array; roles keep alternating (test asserts `user, assistant, user` and no adjacent duplicates); video is named since no block exists. **Gemini** — `functionResponse.parts[].inlineData` (v1beta multimodal function response), so the user content stays a pure functionResponse turn; not live-tested, flagged as such. **Chat completions / Responses** — one user turn built at wire time after the batch's outputs, naming the calls (`[media returned by tool call call_1: grab.png]`); both protocols accept consecutive user turns. The image/video modality retries now count current tool-row media too.
- F4: `vision::read_media_no_follow` re-applies `O_NOFOLLOW` + a symlink-metadata refusal + a 20 MB cap at encode time; a file that can't be read then becomes a `[… could not be read when this request was sent; not shown]` note, not a silent drop (test: symlink and oversize refused).
- F3: `estimate_message_tokens` charges ~1.6k per image and ~8k per video on any row carrying media (over-counts stale tool rows, which is the safe side).
- F6: GIF is metadata-only in `view_image` (png/jpeg/webp remain).
- Also fixed the CI failure (two `ModelHints` literals in `octos-cli` tests) and, found while live-testing this: the CLI's context manager rebuilds the prompt from typed items and its `ToolOutput` envelope had no media field, so the tool row's image was dropped on every rebuild — it now carries `media` under the same `supports_media` policy as user media.

**Not done here, agreed as follow-ups:** an approval/visibility gate for `model_media` analogous to `files_to_send` (F5 privacy), and a per-turn aggregate cap on parallel `view_image` calls.

**Verified live after the rework:** DeepSeek V4 Flash (chat completions), GLM 5.3 Flash (Anthropic-protocol endpoint), Kimi K3 and kimi-for-coding all read the test image through `view_image`; K3 watched the video; DeepSeek hit the 422, retried without it and reported it could not watch. Tests added for each protocol's shape and the send-once rule.

ZhangHanDong (commented):
lgtm

ZhangHanDong (approved):

anantheparty (approved):
Re-review at head `f351c58a` — **approve**. The restructure resolves the blocking findings properly:

- **F0 (Anthropic 400) fixed the right way**: the media now rides the Tool row itself, and the Anthropic builder renders it as image blocks **inside the tool_result content** (`tool_media::for_tool_row` + the merge machinery for consecutive Tool rows at `:862-870`) — roles alternate, no orphan user line. There's a named test for exactly this (`should_put_tool_media_inside_the_tool_result_and_keep_roles_alternating`).
- **F2 (thread pollution) resolved by construction**: with the follow-up line gone, media rides Tool rows, which the persistence loop already pre-marks with thread_id — no synthetic thread.
- **F1 (Gemini)**: the same `tool_media` shaping handles the Gemini side; the FunctionResponse/text incompatibility no longer produces adjacent same-role contents for the media path.

Remaining follow-ups (Medium, non-blocking — worth a tracked hardening pass, happy to file):
1. **Token estimation is still blind to media** (`context.rs:208-218` counts text+tool_calls only) — a 5MB screenshot rides as ~20 caption tokens, so route-fit/compaction triggers under-count exactly when images inflate the request. Suggest estimating image tokens by size/mime.
2. **Render-time TOCTOU**: `encode_image` is a bare `std::fs::read` — the tool-time defenses (no-follow, size, magic) don't re-apply at render. Suggest reading bytes at dispatch or re-validating at encode; and mark render failures as not-shown (currently a deleted file silently drops the image while the caption remains).
3. **Screenshot approval gate + send-once semantics**: pixels go to the provider ungated and re-send every turn until compaction — the files_to_send visibility/approval analogy is the natural shape.

The video extension in this head inherits all of the above correctly (same path), so the follow-up covers images and videos uniformly.

